PR #5, opened 2015-09-09
bump to dor-services 4.22 and dor-workflow-service to 1.7.7

drh-stanford:
Note that you have to install oracle for this to install. the dev box already has it installed (see config/deploy.rb). I added a .travis.yml, but it doesn't work because travis does not support Oracle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,0 @@
-sudo: false
-notifications:
-  email: false
-
-rvm:
-  - 1.9.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+sudo: false
+notifications:
+  email: false
+
+rvm:
+  - 1.9.3
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
 gem "grape", "0.2.1"
-gem "dor-services", "~> 4.21"
+gem "dor-services", "~> 4.22"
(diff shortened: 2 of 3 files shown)


tingulfsen:
Thanks for doing this, just one question - why the version specification for pry-debugger?

atz:
Debuggers shouldn't need to be included as dependencies at all, let alone pinned to a specific version... only for one version of ruby.

tingulfsen:
Joe, can you be a bit more specific? What is the problem with including debuggers in the Gemfile?

drh-stanford:
I got the pin from here https://github.com/sul-dlss/dor-services/blob/develop/Gemfile#L5

drh-stanford:
@atz @tingulfsen should I take out the debugger pin? and if so, should we also take it our of dor-services? thanks.

tingulfsen:
I have to admit that I'm not quite sure what @atz is referring to, so I'd like to wait for his response if you don't mind.

drh-stanford:
moved travis config to https://github.com/sul-dlss/dor-services-app/pull/7

tingulfsen:
@drh-stanford, I've spoken to @atz, and the rationale for not referencing debuggers in the dependencies is that there shouldn't be any code that needs the debugger to run and that we want to minimize the number of dependencies. Could you remove the debugger dependency, but leave a comment stating what the preferred debugger version is? I think you could just comment out that one line in the Gemfile. Thanks!

drh-stanford:
ok i snipped that out and updated the `Gemfile.lock`

PR #8, opened 2015-09-14
switch to pry-byebug and use helpers to DRY up code and enforce a tem…

drh-stanford:
…porary workspace


tingulfsen:
@drh-stanford I like your changes. I think we agreed yesterday to leave debuggers out of the application, so if you remove the references to any debuggers and update Gemfile.lock I will merge. Thanks for the work you've put in on this.

PR #15, opened 2016-01-15
dor-services bump and dependency rework

### LINKED ISSUE #12: Update to dor-services 5.x
This will be required as part of the major dor index conversion.

atz:
Notice how much could be removed.

Closes #12


tingulfsen:
Thanks for this, @atz - bonus for removing unnecessary stuff.

PR #18, opened 2016-02-18
Add SDR service gateway endpoints

cbeer:
This allows the dor-services-app to mediate requests to the SDR services, instead of requiring all dor-services consumers to get special access to SDR. There will be a corresponding change in dor-services to consume these endpoints instead of accessing SDR directly.

PR #20, opened 2016-02-26
Backport SDR endpoints into dor-services 4.x branch

cbeer:
This backports #18 to the dor-services-4.x branch, allowing us to deploy a new version of `dor-services-app` to production that supports https://github.com/sul-dlss/dor-services/pull/164.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ group :test do
   gem "rspec"
 end
 
-group :development do
+group :deployment do
   gem "capistrano", '~> 3.0'
   gem 'capistrano-bundler', '~> 1.1'
   gem "lyberteam-capistrano-devel"
--- a/Gemfile
+++ b/Gemfile
@@ -7,12 +7,14 @@ gem "workflow-archiver", "~> 1.3"
 gem "rubydora", "1.6.5"
 gem "rack-test", :require => "rack/test"
 gem "addressable", '2.3.5'
+gem "rest-client"
 
 group :test do
   gem "awesome_print"
   gem "equivalent-xml"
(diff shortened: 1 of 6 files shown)


cbeer:
Note that travis doesn't work with `dor-services-4.x`, but the tests do pass locally.

darrenleeweber:
Generally seems OK.  Could be deployed and tested against sdr-dev or sdr-test.

Gemfile
```diff
@@ -1,21 +1,23 @@
...
 gem "rack-test", :require => "rack/test"
 gem "addressable", '2.3.5'
+gem "rest-client"
 
 group :test do
   gem "awesome_print"
   gem "equivalent-xml"
   gem "simplecov"
   gem "rspec"
+  gem "fakeweb"
```
drh-stanford: we're already using Faraday. Can we use it for the testing rather than bringing in FakeWeb?

drh-stanford:
Re: travis. Travis for `dor-services-app` does not work at all due to our requirement for Oracle oci8 support. We probably shouldn't even be running it.

Gemfile
cbeer: Are we using Faraday in dor-services 4.x?

cbeer:
That's true for the 4.x branch, but we fixed it up to work in `master`.

Gemfile
drh-stanford: It's in the `Gemfile.lock` for dor-services, but we don't use it explicitly. Since this is a backport I think it's ok -- 4.x uses FakeWeb too :-/.

PR #25, opened 2016-03-30
add sdr endpoint for retrieving preserved content

jmartin-sul:
* dor_services_app.rb: add `get '/objects/:druid/content/:filename'` under the sdr resource
* dor_services_app_spec: add tests for the following common scenarios:
 * passing through error codes (404 and 500)
 * retrieving files with names that require URI encoding
 * retrieving text files
 * retrieving binary files


spec/dor_services_app_spec.rb
```diff
@@ -53,6 +53,69 @@ def app
...
+        FakeWeb.register_uri(:get, "#{Dor::Config.sdr.url}/objects/#{item.pid}/content/unexpected_error?version=2", status: 500)
+
+        get "/v1/sdr/objects/#{item.pid}/content/no_such_file?version=2"
+        expect(last_response.status).to eq 404
+
+        get "/v1/sdr/objects/#{item.pid}/content/unexpected_error?version=2"
+        expect(last_response.status).to eq 500
+      end
+
+      context 'URI encoding' do
```
ndushay: FYI: when it's just a single test, it doesn't need to be wrapped in context or describe block.  But no harm, no foul.

ndushay:
:+1:   from me.

jmartin-sul:
this was tested in stage, and it appears to work.  i'd like to add tests for the `get_preserved_file` methods in argo and dor-services, but i think this is ready for review.  i'm a bit worried about the fact that we're loading up the entire file before passing it along to argo, which will then do the same thing.  but it didn't break argo before, and i can't see a clean way to stream the file from sdr to argo in the current setup, so i guess i'm happy enough with the naive implementation for now if others are.

to @ndushay's comment, i was mostly just trying to parallel the current test structure, but definitely a fair point.

opinions from @cbeer since he was the original implementer of the proxy functionality?

tingulfsen:
:+1:

PR #31, opened 2016-09-21
Add new update marc record web service call

peetucket:
Implements a new method called "update_marc_record" which will write out a marc record for syncing up with symphony.  The logic used to be in the item-release robot suite, but it is being extracted here so that it can be called directly without having to invoke a workflow.  The robot will simply call the web service now.  See https://github.com/sul-dlss/item-release/pull/64

fixes sul-dlss/discovery-indexing#1

### DIFF
--- a/bin/write_marc_record_test
+++ b/bin/write_marc_record_test
@@ -1,3 +1,3 @@
-#Do kerbose authentication here if it is required by the server
+#Do kerberos authentication here if it is required by the server
 echo "${1}" >> $2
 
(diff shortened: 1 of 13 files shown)


bin/write_marc_record_test
```diff
@@ -0,0 +1,3 @@
+#Do kerbose authentication here if it is required by the server
```
jmartin-sul: typo? (`kerbose` -> `kerberos` ?)

peetucket: fixed

lmcglohon:
@jmartin-sul @peetucket John are you still reviewing this pull request?  If so, what is the status?

jmartin-sul:
sorry for the confusion, i was all done, my one contribution was the comment i left (and that's been fixed).

PR #33, opened 2016-09-30
switch to new .thumb method from dor-services

### LINKED ISSUE #32: Update file_id method to use new thumb logic in dor-services gem
use .thumb method in dor-services and escape / in output to %2F

peetucket:
fixes #32

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -300,7 +300,7 @@ DEPENDENCIES
   capistrano-bundler
   capistrano-passenger
   dlss-capistrano
-  dor-services (~> 5.11)
+  dor-services (~> 5.12)
   equivalent-xml
   fakeweb
   grape (~> 0.14)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -288,7 +288,7 @@ DEPENDENCIES
   capistrano-bundler
   capistrano-passenger
   dlss-capistrano
-  dor-services (~> 5.10, >= 5.11.0)
+  dor-services (~> 5.11)
   equivalent-xml
   fakeweb
   grape (~> 0.14)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,10 +11,10 @@ GEM
       rdf-rdfxml (= 1.0.1)
       rsolr
       rubydora (~> 1.6, >= 1.6.5)
-    activemodel (4.2.7)
-      activesupport (= 4.2.7)
+    activemodel (4.2.7.1)
+      activesupport (= 4.2.7.1)
       builder (~> 3.1)
-    activesupport (4.2.7)
+    activesupport (4.2.7.1)
       i18n (~> 0.7)
       json (~> 1.7, >= 1.7.7)
       minitest (~> 5.1)
@@ -63,11 +63,11 @@ GEM
       capistrano-releaseboard
     docile (1.1.5)
     docopt (0.5.0)
-    domain_name (0.5.20160615)
+    domain_name (0.5.20160826)
       unf (>= 0.0.5, < 1.0.0)
     dor-rights-auth (1.2.0)
       nokogiri
-    dor-services (5.10.1)
+    dor-services (5.11.0)
       active-fedora (~> 6.0)
       activesupport (>= 3.2.18)
       confstruct (~> 0.2.7)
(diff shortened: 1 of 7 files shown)


Gemfile.lock
```diff
@@ -63,11 +63,11 @@ GEM
       capistrano-releaseboard
     docile (1.1.5)
     docopt (0.5.0)
-    domain_name (0.5.20160615)
+    domain_name (0.5.20160826)
       unf (>= 0.0.5, < 1.0.0)
     dor-rights-auth (1.2.0)
       nokogiri
-    dor-services (5.10.1)
+    dor-services (5.11.0)
```
drh-stanford: there's a `5.12.0` now

spec/update_marc_spec.rb
```diff
@@ -189,236 +187,172 @@
...
-      expect(d).to receive(:id).and_return('bb111bb2222')
-      expect(d).to receive(:datastreams).exactly(4).times.and_return({'contentMetadata' => content_metadata_ds})
-      expect(content_metadata_ds).to receive(:ng_xml).twice.and_return(content_metadata_ng_xml)
+      expect(d).to receive(:id).twice.and_return('bb111bb2222')
+      expect(d).to receive(:datastreams).exactly(3).times.and_return({'contentMetadata' => content_metadata_ds})
+      expect(content_metadata_ds).to receive(:ng_xml).and_return(content_metadata_ng_xml)
 
       updater = Dor::UpdateMarcRecordService.new(d)
-      expect(updater.file_id).to eq('|xfile:bb111bb2222%2Fwt183gy6220_00_0001.jp2')
+      expect(updater.thumb).to eq('|xfile:bb111bb2222%2Fwt183gy6220_00_0001.jp2')
```
drh-stanford: what is this `|xfile:` business? I couldn't find any documentation on it in the codebase

drh-stanford:
this PR looks good, just a couple minor comments...

spec/update_marc_spec.rb
peetucket: the |xfile: syntax is something specific to the line in the file that is generated for updating the marc record... i am not sure of the specifics either - i presume it is a MARC specification

Gemfile.lock
peetucket: well, i just tried updating to 5.12.0 and now it is force downgrading activesupport to a lower version of 4, which bundle-audit is complaining about...need some investigation

drh-stanford: yes, I can confirm that `dor-services v5.12.0` breaks `bundle-audit` when you do a `bundle update dor-services`:

```
Name: activesupport
Version: 4.0.13
Advisory: CVE-2015-3227
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/bahr2JLnxvk
Title: Possible Denial of Service attack in Active Support
Solution: upgrade to >= 4.2.2, ~> 4.1.11, ~> 3.2.22
```

so I would suggest:

```
-gem 'dor-services', '~> 5.11'
+gem 'dor-services', '~> 5.11.1' # 5.12.0 break with a bundle-audit error for activesupport
```

cbeer: Hm. I don't know why it is downgrading `activesupport` on you. Pinning it in the app seems to be ok, though.

drh-stanford: hmmm. it looks like the problem is doing a `bundle update dor-services`... if you just change:

```
-gem 'dor-services', '~> 5.11'
+gem 'dor-services', '~> 5.12'
```

and then do a `bundle install` (not update), `bundle-audit` will pass...

peetucket: I did some sleuthing in the Gemfile.lock after upgrading and couldn't find which update was causing activesupport to downgrade.

cbeer: e.g. https://github.com/sul-dlss/dor-services-app/pull/34

peetucket: what do you think of @drh-stanford comment right above yours @cbeer ?  Should we still pin activesupport just to be sure?

peetucket:
ok, pinned activesupport via #34 and updated to 5.12, bundle-audit passes, bundle install works as expected

drh-stanford:
sorry, i didn't see #34 before my last comment. i looked over the resulting `Gemfile.lock` and it's correctly upgrading to ActiveFedora 6.8.0 and rubydora 1.9.0 (which were both released recently).

PR #35, opened 2016-11-01
Goobi integration

peetucket:
Add new web service endpoint to make calls to Goobi Digitization Service to indicate when an object is registered.   see https://docs.google.com/document/d/127Qbx9LFL6Ver9qWj5M_ZmRBERW9s1umpMVR38pwvMI and https://github.com/sul-dlss/sul-goobi/issues/7


spec/dor_services_app_spec.rb
```diff
@@ -222,6 +222,16 @@ def app
     end
   end
 
+  describe '/publish' do
```
drh-stanford: curious this didn't have a test as it's probably the most used route we have :)

peetucket:
Thanks, just took the simplistic approach of moving all of those methods for fixture data into their own file and requiring it separately to keep the spec_helper itself more clean.

PR #36, opened 2016-11-21
Convert to a Rails application

cbeer:


app/controllers/workflows_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  def archive
+    version = params.fetch(:ver_num) { @item.current_version }
+    archiver.archive_one_datastream 'dor', params[:object_id], params[:id], version
+    
+    render plain: "#{params[:id]} version #{version} archived"
+  end
+  
+  private
+
+  def load_item
```
cbeer: Sure, I can move the definition.

app/models/dor/service_item.rb
```diff
@@ -44,7 +44,8 @@ def barcode
     # the @id attribute of resource/file elements including extension
     # @return [String] thumbnail filename (nil if none found)
     def thumb
-      @druid_obj.encoded_thumb unless @druid_obj.datastreams.nil?
+      return if @druid_obj.datastreams.nil? || !@druid_obj.is_a?(Dor::Contentable)
```
drh-stanford: where did this extra logic come from? `!@druid_obj.is_a?(Dor::Contentable)` I see the method defined in Publishable here: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/concerns/publishable.rb#L45 

assuming you're checking to see if the method exists, why not use a `respond_to?`

cbeer: Fallout from the ActiveFedora update.

I think there's a bug in `Dor::Publishable#thumb` that doesn't work if non-`Dor::Contentable` classes (like `Dor::Collection`), and I have a question in to @lmcglohon to confirm and figure out the expected upstream behavior.

cbeer: Upstream PR: https://github.com/sul-dlss/dor-services/pull/280. If that looks good, I'll remove this commit.

drh-stanford: merged.

config/routes.rb
```diff
@@ -0,0 +1,44 @@
...
+        collection do
+          get 'current'
+          post 'current/close', action: 'close_current'
+        end
+      end
+      
+      resources :workflows, only: [] do
+        member do
+          post 'archive'
+          post 'archive/:ver_num', action: :archive
```
drh-stanford: :( ok. btw i don't seem to have permissions to read the AFS archive logs

app/models/dor/service_item.rb
cbeer: Ok, released and fixed up that commit.

PR #37, opened 2016-11-22
Add catalog service

cbeer:


config/initializers/okcomputer.rb
```diff
@@ -20,4 +20,13 @@ def version_from_version_file
...
+class SymphonyCheck < OkComputer::HttpCheck
+  def perform_request
+    Timeout.timeout(request_timeout) do
+      SymphonyReader.client.get(url.to_s)
+    end
+  end
+end
+
 OkComputer::Registry.register 'version', CustomAppVersionCheck.new
+OkComputer::Registry.register 'symphony', SymphonyCheck.new(Settings.CATALOG.SYMPHONY.JSON_URL % { catkey: 12345 })
```
drh-stanford: if symphony is an external service that's directly monitored elsewhere, then this check should be named `external-symphony`

drh-stanford:
There's several XSLTs in the old code -- are they not used at all? Only `MARC21slim2MODS3-6.xsl`? See lyberservices in `apps/xslt/home/xslt_files`

config/routes.rb
```diff
@@ -16,6 +16,12 @@
       get 'initial', to: 'workflows#initial'
     end
     
+    scope :catalog do
+      get 'marcxml', to: 'marcxml#marcxml'
+      get 'mods', to: 'marcxml#mods'
+      get 'catkey', to: 'marcxml#catkey'
+    end
+
```
drh-stanford: did you look at the log history to see whether all these routes are being used? it'd be nice to deprecate or remove them if not.

spec/controllers/marcxml_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    login
+  end
+
+  let(:resource) { MarcxmlResource.new(catkey: '12345') }
+
+  describe 'GET catkey' do
+    it 'returns the provided catkey' do
+      get :catkey, params: { catkey: '12345' }
+      expect(response.body).to eq '12345'
+    end
```
drh-stanford: is this really all it does is an echo? or is it suppose to return a 500 if the catkey is invalid or something?

spec/controllers/marcxml_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  describe 'GET mods' do
+    it 'transforms the MARCXML into MODS' do
+      FakeWeb.register_uri(:get, Settings.CATALOG.SYMPHONY.JSON_URL % { catkey: resource.catkey }, body: '{}')
+
+      get :mods, params: { catkey: '12345' }
+      expect(response.body).to match(/mods/)
+    end
+  end
+end
```
drh-stanford: can we add a MARC and its converted MODS fixture to the test suite? it'd be nice to have that input/output data as reference for the API since it's not documented (I'm assuming).

cbeer:
The other XSLTs are not used at all by the catalog service, no.

config/initializers/okcomputer.rb
cbeer: 👌

spec/controllers/marcxml_controller_spec.rb
cbeer: Isn't that simply testing the LoC XSLT?

PR #42, opened 2016-12-02
Add ContentController for listing and retrieving content from the dor workspace

cbeer:
Addressing https://github.com/sul-dlss/dor-services/issues/205 by removing the need for SSH keys in common scenarios


peetucket:
updated documentation here with new endpoints (not sure if anyone is using that documentation, but was easy enough to add to): https://consul.stanford.edu/display/chimera/REST+mappings+for+dor-services+gem

PR #46, opened 2016-12-12
Update  MARC21slim2MODS3-4 using actual deployed version of the file

cbeer:


cbeer:
I can't explain the discrepancy, so I'm also going to disable the jenkins build for the xslt-service, knowing that this PR should allow us to replace the service entirely.

PR #51, opened 2017-01-04
#body_params should fallback to an empty hash

cbeer:


ndushay:
"the problem is rack doesn’t define what `request.body` returns, and passenger chose something more restrictive than whatever the rails test harness provides."

PR #68, opened 2017-02-15
set second indicator depending on the objects APO

### LINKED ISSUE #67: Set indicator2=0 for ETDs, EEMs in SDR managed Purls
These are two cases where we know the Symphony record was created to point to a born-digital resource. In this case the second indicator of the 856 should be a 0 instead of a 1.

Set Ind2=0 for any object whose APO/whose isGovernedBy relationship is

ETD - druid:bx911tp9024
EEMs - druid:jj305hm5259
else Ind2=1

moved from https://github.com/sul-dlss/discovery-indexing/issues/54

peetucket:
fixes #67

### DIFF
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -2,6 +2,9 @@ require 'open3'
 
 module Dor
   class UpdateMarcRecordService < ServiceItem
+    # objects goverened by these APOs (ETD and EEMs) will get indicator 2 = 0, else 1
+    BORN_DIGITAL_APOS = %w(druid:bx911tp9024 druid:jj305hm5259).freeze
+
     def update
       push_symphony_record if ckey.present?
     end
@@ -13,7 +16,7 @@ module Dor
 
     # catkey: the catalog key that associates a DOR object with a specific Symphony record.
     # druid: the druid
-    # .856. 41
+    # .856. 41 or 40 (depending on APO)
     # Subfield u (required): the full Purl URL
     # Subfield x #1 (required): The string SDR-PURL as a marker to identify 856 entries managed through DOR
     # Subfield x #2 (required): Object type (<identityMetadata><objectType>) - item, collection,
@@ -73,9 +76,9 @@ module Dor
       '4'
     end
 
-    # It returns Second Indicator for Version of resource (1)
+    # It returns Second Indicator for Version of resource
     def get_2nd_indicator
-      '1'
+      born_digital? ? '0' : '1'
     end
 
     # It's a plceholder for the uri label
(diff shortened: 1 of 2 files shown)


spec/update_marc_spec.rb
```diff
@@ -199,11 +199,30 @@
   end
 
   describe '.get_2nd_indicator' do
-    it 'should return 1' do
+    it 'should return 1 for a non dorn digital APO' do
```
cbeer: typo: dorn -> born

app/models/dor/update_marc_record_service.rb
```diff
@@ -128,9 +135,12 @@ def released_to_searchworks
...
       return [] unless @druid_obj.relationships(:is_constituent_of)
       @druid_obj.relationships(:is_constituent_of).map do |cons|
-        cons_druid = cons.sub('info:fedora/', '')
-        Dor::Item.find(cons_druid)
+        Dor::Item.find(extract_druid(cons))
       end
     end
+
+    def extract_druid(druid)
+      druid.match(/druid(:[a-z]{2})(\d{3})([a-z]{2})(\d{4})$/i).to_s
```
cbeer: Surely we already have logic to do this.. somewhere.. right? Either in dor-services, druid-tools, or up in active-fedora.

Or, maybe it's time to finally get https://github.com/sul-dlss/dor-services/pull/263 in, so we aren't re-inventing our own tools for managing relationships?

spec/update_marc_spec.rb
peetucket: make it so

worftng

app/models/dor/update_marc_record_service.rb
```diff
@@ -118,6 +121,10 @@ def get_x2_constituent_info
       end.join('')
     end
 
+    def born_digital?
+      @druid_obj.relationships(:is_governed_by).map { |apo| BORN_DIGITAL_APOS.include? extract_druid(apo) }.include? true
```
cbeer: `#relationships` is a pretty low-level method. Can we use e.g. `#admin_policy_object` (or maybe `#admin_policy_object_id` here?

app/models/dor/update_marc_record_service.rb
peetucket: yes, probably, just extracted since it was better than what was happening before and was going to use it twice...but if use admin_policy_object_id than it is not necessary to change this method and i can go back

app/models/dor/update_marc_record_service.rb
peetucket: yup will use admin_policy_object_id

peetucket:
Now using admin_policy_object_id to check for apo to avoid using relationships.  Leaving all else alone.

PR #77, opened 2017-05-25
fix writing out of marc records that contain quotes in them

### LINKED ISSUE #76: Allow single quotes in Collection titles and elsewhere in update_marc_service
Per https://app.honeybadger.io/projects/50568/faults/33388400#notice-summary , the update_marc step of the releaseWF returns a 500 when collection titles contain a single quote or apostrophe.

Per @jkeck "Looks like the the command sent to `bin/write_marc_record` is an unescaped string wrapped in single quotes (which will break if single-quotes are in the data anywhere)."

I'm not sure if this belongs in the dor-services-app where the update-marc-service sits, or somewhere else in either dor-services or the release robots. Please feel free to let me know and I'll move it.

peetucket:
fixes #76

### DIFF
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -52,7 +52,7 @@ module Dor
       return if symphony_records.blank?
       symphony_file_name = "#{Dor::Config.release.symphony_path}/sdr-purl-856s"
       symphony_records.each do |symphony_record|
-        command = "#{Dor::Config.release.write_marc_script} \'#{symphony_record}\' #{symphony_file_name}"
+        command = "#{Dor::Config.release.write_marc_script} #{Shellwords.escape(symphony_record)} #{Shellwords.escape(symphony_file_name)}"
         run_write_script(command)
       end
     end
--- a/spec/update_marc_spec.rb
+++ b/spec/update_marc_spec.rb
@@ -294,33 +294,62 @@ RSpec.describe Dor::UpdateMarcRecordService do
       @updater = Dor::UpdateMarcRecordService.new(@dor_item)
       expect(File.exist?(@output_file)).to be_falsey
     end
-
-    it 'should write the symphony record to the symphony directory for a single record' do
-      marc_records = ['abcdef']
-      @updater.write_symphony_records marc_records
-      expect(File.exist?(@output_file)).to be_truthy
-      expect(File.read(@output_file)).to eq "#{marc_records.first}\n"
-    end
-
-    it 'should write the symphony record to the symphony directory for multiple records' do
-      marc_records = %w(abcdef 12345)
-      @updater.write_symphony_records marc_records
-      expect(File.exist?(@output_file)).to be_truthy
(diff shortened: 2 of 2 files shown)


app/models/dor/update_marc_record_service.rb
```diff
@@ -52,7 +52,7 @@ def write_symphony_records(symphony_records)
       return if symphony_records.blank?
       symphony_file_name = "#{Dor::Config.release.symphony_path}/sdr-purl-856s"
       symphony_records.each do |symphony_record|
-        command = "#{Dor::Config.release.write_marc_script} \'#{symphony_record}\' #{symphony_file_name}"
+        command = "#{Dor::Config.release.write_marc_script} \"#{symphony_record.gsub('"', %q(\\\"))}\" #{symphony_file_name}" # escape any double quotes in the record itself
```
jcoyne: Can we use `Shellwords.escape` here rather than roll our own?

spec/update_marc_spec.rb
```diff
@@ -302,11 +302,25 @@
       expect(File.read(@output_file)).to eq "#{marc_records.first}\n"
     end
 
-    it 'should write the symphony record to the symphony directory for multiple records' do
-      marc_records = %w(abcdef 12345)
+    it 'should write a symphony record containing single quotes' do
```
jcoyne: Style suggestion for rspec. Use context blocks like:

```ruby
describe "#write_symphony_records" do
  subject { @updater.write_symphony_records marc_records }

  content "for a record with single quotes" do
    let(:marc_records) {  ["this is | a record | that has 'single quotes' in it | and it should work"] }
    it "writes the record" do
      expect(File.exist?(@output_file)).to be_truthy
      expect(File.read(@output_file)).to eq "#{marc_records.first}\n"
    end
  end
  content "for a record with double and single quotes" do
    let(:marc_records) {  ['record with "double quotes" in it | and it should work'] }
    it "writes the record" do
      expect(File.exist?(@output_file)).to be_truthy
      expect(File.read(@output_file)).to eq "#{marc_records.first}\n"
    end
  end
end
```
There's a great guide here: http://www.betterspecs.org/#contexts

peetucket: Ok, I'll rewrite all of the write_symphony_test to match that style.

app/models/dor/update_marc_record_service.rb
peetucket: Nice. Didn't know about this method.  Changing.

peetucket:
Rebased and pushed.

spec/update_marc_spec.rb
```diff
@@ -294,33 +295,62 @@
       @updater = Dor::UpdateMarcRecordService.new(@dor_item)
       expect(File.exist?(@output_file)).to be_falsey
     end
+    after :each do
+      FileUtils.rm_f(@output_file)
+    end
+
+    subject { @updater.write_symphony_records marc_records }
```
jcoyne: It looks like rubocop would like us to name this like `subject(:writer)` and then `expect(writer).to_not be_nil` down below.

peetucket:
fixed rubocop stuff and rebased/squashed

PR #85, opened 2018-02-21
Update dependencies

cbeer:


jmartin-sul:
should this get rebased in light of #86?

cbeer:
I don't think it'll be a problem, but sure, done.

Issue #70, opened 2017-02-27
Restore MODS parm on new /catalog/mods service call 


LynnMcRae:
The lyberservices-based catalog service was set up to have old, current and new XSLT versions installed and addressable via a query parm:

http://lyberservices-prod.stanford.edu/catalog/mods?catkey=3195754&mods=3.3
http://lyberservices-prod.stanford.edu/catalog/mods?catkey=3195754&mods=3.4

That way we could test a new stylesheet before cutting over the default. Just as important, we could continue to compare current with old behaviors when questions arose any time after the cutover. The new dor-services-app version does not implement this option.

https://dor-services-prod.stanford.edu/v1/catalog/mods?catkey=319575
https://dor-services-prod.stanford.edu/v1/catalog/mods?catkey=319575&mods=3.3
https://dor-services-prod.stanford.edu/v1/catalog/mods?catkey=319575&mods=3.6

all deliver 3.4. The ability to invoke specific versions needs to be retained, both for testing now and for later use if questions arise about particular forms of data.

cmh2166:
@cbeer have you scoped this work already?

cbeer:
No; it seems straightforward enough, but we should make sure it's something we actually need and will use. In my log analysis way back when, it was clear it had never been used (at least as far back as we had logs).

cmh2166:
kk, thanks.

PR #101, opened 2018-05-08
LC MARC2MODS 3.6 v.1.118 with local changes to 3.4 merged in

arcadiafalcone:
LC stylesheet updated with previous local changes (see https://consul.stanford.edu/display/chimera/MODS+schema+and+MARC+transformation).

More intensive bugfixes/revisions in progress.

### DIFF
--- a/app/xslt/MARC21slim2MODS3-6_SDR.xsl
+++ b/app/xslt/MARC21slim2MODS3-6_SDR.xsl
@@ -5,7 +5,7 @@
 
 	<!-- Maintenance note: For each revision, change the content of <recordInfo><recordOrigin> to reflect the new revision number.
 	MARC21slim2MODS3-6 (Revision 1.118) 20180131
-	
+
 	MODS 3.6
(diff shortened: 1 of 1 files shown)


cmh2166:
@arcadiafalcone please confirm: you *only* want conversion to MODS 3.6 using this XSLT for all resources being transformed for load from Symphony MARC to DOR MODS/XML from this point forward?

cmh2166:
No, there are open tickets related to this / on this repo [1] + one on catalog-service [2] that speaks not just to adding MODS 3.6, but supporting a MODS version flag. Checking in on is that is requirement for you, or you're seeing this addition as the only XSLT we expect to use.

1. https://github.com/sul-dlss/dor-services-app/issues/70
2. https://github.com/sul-dlss/catalog-service/issues/1#issue-208311679 (see the last block in that comment; additionally, you state in this ticket that this work as needed has already been done here, hence my just confirming here)

arcadiafalcone:
@cmh2166 Got it, thanks. I do not see any need for a versioning flag.

PR #104, opened 2018-05-23
Include only the catkey in the 001 field

### LINKED ISSUE #91: [marc maint] Process Arcadia's feedback re: correct ID in 001 when retrieving data from Symphony web services
See the feedback here: 

> The marc2xml stored process removes any existing 001 fields from the MARC record, and then adds a single 001 field containing the catkey with the “a” prefix. That process’s output also includes an 003 field with value “SIRSI”. The MARC2MODS XSLT expects a single instance of each of these fields, which it uses to generate the recordIdentifier value for the metadata source in the MODS output. The web services MARCXML output may include zero or multiple 001s containing non-catkey identifiers, and lacks the 003 when the catkey 001 is not already present in the MARC, so this needs to be normalized before the data goes to the XSLT for transformation.

what should be in the recordIdentifier in the generated MODs, does that require the XSLT to work from the 001 field, and if so, then what should be (but isn't) in the 001 field.

larger question: do we care about having recordIdentifier in the generated MODS, or can we use that identifier / catkey field from the Identification metadata? If needed, add data injection on the symphony reader (with @arcadiafalcone reviewing that output).

aaron-collier:
Exclude any provided 001 and explicitly set the catkey as the 001 and SIRSI as 003

Fixes #91

### DIFF
--- a/app/models/symphony_reader.rb
+++ b/app/models/symphony_reader.rb
@@ -16,9 +16,15 @@ class SymphonyReader
     record.leader = leader if leader
 
     fields.uniq.each do |field|
-      record << marc_field(field)
+      record << marc_field(field) unless field['tag'] == '001' # explicitly remove all 001 fields from the record
     end
 
+    # explicitly inject the catkey into the 001 field
+    record << marc_field('tag' => '001',
+                         'subfields' => [
+                           { 'code' => '_',
+                             'data' => "a#{catkey}" }
+                         ])
     record
   end
 
--- a/app/models/symphony_reader.rb
+++ b/app/models/symphony_reader.rb
@@ -16,15 +16,15 @@ class SymphonyReader
     record.leader = leader if leader
 
     fields.uniq.each do |field|
-      record << marc_field(field) unless field['tag'] == '001' # explicitly remove all 001 fields from the record
+      record << marc_field(field) unless %w[001 003].include? field['tag'] # explicitly remove all 001 and 003 fields from the record
     end
 
     # explicitly inject the catkey into the 001 field
-    record << marc_field('tag' => '001',
-                         'subfields' => [
-                           { 'code' => '_',
-                             'data' => "a#{catkey}" }
-                         ])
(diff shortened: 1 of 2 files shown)


cmh2166:
Does this add the catkey based off what was known in the ruby code, or just the first one in the retrieved xml?

aaron-collier:
We aren't doing any explicit tag logic in the ruby code, so I put this in the XSLT, but I'm not opposed to adding it to the ruby code if that we feel that is more appropriate.

cmh2166:
I'm worried that this doesn't actually solve the issue as outlined in the ticket. See https://github.com/sul-dlss/dor-services-app/issues/91#issuecomment-385507251

arcadiafalcone:
@cmh2166 @aaron-collier I would prefer doing this in the Ruby code over the XSLT as it is normalization of the MARC data input, and is more closely related to the system providing that input than the transformation of the provided MARC into MODS. Other systems might have different normalization requirements.

cmh2166:
@arcadiafalcone yep, and that's explicitly what is in the ticket.

cbeer:
(and, 👍 to doing it in ruby; the less we have to customize the upstream XSLT the better)

aaron-collier:
Just pushed up a fix.

app/models/symphony_reader.rb
```diff
@@ -16,9 +16,15 @@ def to_marc
     record.leader = leader if leader
 
     fields.uniq.each do |field|
-      record << marc_field(field)
+      record << marc_field(field) unless %w[001 003].include? field['tag'] # explicitly remove all 001 and 003 fields from the record
```
jcoyne: 👏 for documentation

app/models/symphony_reader.rb
```diff
@@ -16,9 +16,15 @@ def to_marc
...
     end
 
+    # explicitly inject the catkey into the 001 field
+    record << marc_field('tag' => '001', 'subfields' => [{ 'code' => '_', 'data' => "a#{catkey}" }])
+
+    # explicitly inject the SIRSI into the 003 field
```
jcoyne: "the" is probably unnecessary

PR #139, opened 2018-08-24
add <tags> to goobi notify call with object tags in specified format

### LINKED ISSUE #138: notify_goobi call needs to add all tags to XML request sent to goobi
Update the XML format generated in  “app/models/dor/goobi.rb” to include all object tags, with the name of the tag being the first value of the tag before the first colon, and the value of the tag being anything after the first colon.

E.g. If the obj.tags array is  ["Process : Content Type : Image", "Project : DIGREQ One-off Requests", "JIRA : SPECPAT-1976", "DPG : Patron Request : Noakes", "DPG : Patron Request : Noakes", "Registered By : astrids", "DPG : Workflow : Muybridge_WideTek_Workflow"] 

we should add this to the XML generated in the goobi call:

```
<tags>
   <tag name="Process" value="Content Type : Image"/>
   <tag name="Project" value="DIGREQ One-off Requests"/>
   <tag name="JIRA" value="SPECPAT-1976"/>
   <tag name="DPG" value="Patron Request : Noakes"/>
   <tag name="DPG" value="Patron Request : Noakes"/>
   <tag name="Registered By" value="astrids"/>
   <tag name="DPG" value="Workflow : Muybridge_WideTek_Workflow"/>
</tags>
```

The goobi system will then pick up these tags on their end and add to the object.  Timing of this development relative to the work done on the Goobi side is unimportant (either can occur first), but integration testing with goobi needs to occur after both sides have been developed and deployed.

peetucket:
fixes #138

### DIFF
--- a/app/models/dor/goobi.rb
+++ b/app/models/dor/goobi.rb
@@ -15,6 +15,12 @@ module Dor
       # rubocop:enable Metrics/LineLength
     end
 
+    def goobi_xml_tags
+      goobi_tag_list.map do |tag_name, tag_value|
+        "<tag name=\"#{tag_name}\" value=\"#{tag_value}\"></tag>"
+      end.join
+    end
+
     def xml_request
       <<-END
         <stanfordCreationRequest>
@@ -31,9 +37,7 @@ module Dor
             <sdrWorkflow>#{Dor::Config.goobi.dpg_workflow_name}</sdrWorkflow>
             <goobiWorkflow>#{goobi_workflow_name}</goobiWorkflow>
             <ocr>#{goobi_ocr_tag_present?}</ocr>
-            <tags>
-              #{goobi_tag_list.map { |tag_name, tag_value| "<tag name=\"#{tag_name}\" value=\"#{tag_value}\"></tag>" }.join("\n")}
-            </tags>
+            <tags>#{goobi_xml_tags}</tags>
         </stanfordCreationRequest>
       END
     end
--- a/app/models/dor/goobi.rb
+++ b/app/models/dor/goobi.rb
@@ -31,6 +31,9 @@ module Dor
             <sdrWorkflow>#{Dor::Config.goobi.dpg_workflow_name}</sdrWorkflow>
             <goobiWorkflow>#{goobi_workflow_name}</goobiWorkflow>
             <ocr>#{goobi_ocr_tag_present?}</ocr>
+            <tags>
+              #{goobi_tag_list.map { |tag_name, tag_value| "<tag name=\"#{tag_name}\" value=\"#{tag_value}\"></tag>" }.join("\n")}
+            </tags>
         </stanfordCreationRequest>
       END
     end
--- a/app/models/dor/service_item.rb
+++ b/app/models/dor/service_item.rb
@@ -112,5 +112,15 @@ module Dor
         @druid_obj.tags.any? { |tag| tag.casecmp(dpg_goobi_ocr_tag).zero? } # case insensitive compare
       end
     end
+
+    # returns an array of arrays, each element contains an array of [name, value] of DOR object tags in the format expected to pass to Goobi
+    # the name of the tag is the first namespace part of the tag (before first colon), value of the tag is everything after this
+    # @return [Array]
+    def goobi_tag_list
+      @druid_obj.tags.map do |tag|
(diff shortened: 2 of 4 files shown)


app/models/dor/goobi.rb
```diff
@@ -31,6 +31,9 @@ def xml_request
             <sdrWorkflow>#{Dor::Config.goobi.dpg_workflow_name}</sdrWorkflow>
             <goobiWorkflow>#{goobi_workflow_name}</goobiWorkflow>
             <ocr>#{goobi_ocr_tag_present?}</ocr>
+            <tags>
```
jcoyne: maybe extract this into a method that just generates the `<tags>` element? That would make this method less complex and easier to test the behavior.

app/models/dor/service_item.rb
```diff
@@ -112,5 +112,15 @@ def goobi_ocr_tag_present?
         @druid_obj.tags.any? { |tag| tag.casecmp(dpg_goobi_ocr_tag).zero? } # case insensitive compare
       end
     end
+
+    # returns an array of arrays, each element contains an array of [name, value] of DOR object tags in the format expected to pass to Goobi
+    # the name of the tag is the first namespace part of the tag (before first colon), value of the tag is everything after this
+    # @return [Array]
+    def goobi_tag_list
+      @druid_obj.tags.map do |tag|
+        tag_split = tag.split(':', 2).map(&:strip) # only split on the first colon
```
jcoyne: How about instantiating a `Tag` object that `responds` to name and `value`?

app/models/dor/service_item.rb
```diff
@@ -112,5 +112,15 @@ def goobi_ocr_tag_present?
         @druid_obj.tags.any? { |tag| tag.casecmp(dpg_goobi_ocr_tag).zero? } # case insensitive compare
       end
     end
+
+    # returns an array of arrays, each element contains an array of [name, value] of DOR object tags in the format expected to pass to Goobi
+    # the name of the tag is the first namespace part of the tag (before first colon), value of the tag is everything after this
+    # @return [Array]
```
jcoyne: Please specify what the `Array` contains.

spec/goobi_spec.rb
```diff
@@ -59,6 +64,32 @@
...
+            <tag name="DPG" value="Workflow : book_workflow"></tag>
+            <tag name="DPG" value="OCR : TRUE"></tag>
+        </tags>
+      </stanfordCreationRequest>
+    END
+  end
+
+  it 'creates the correct xml request with no tags present' do
+    allow(item).to receive(:tags).and_return([])
+    expect(@goobi.xml_request).to be_equivalent_to <<-END
```
jcoyne: this test might be easier to read if we do 
`...to include('<tags></tags>')`

spec/service_item_spec.rb
```diff
@@ -42,6 +42,27 @@
     end
   end
 
+  describe '.goobi_tag_list' do
```
jcoyne: typically instance methods would be described as `#goobi_tag_list`

app/models/dor/service_item.rb
peetucket: So create a new tag object and "Dor" namespace and use it here?

spec/service_item_spec.rb
peetucket: ok, will update the rest of the describe blocks in this spec too

app/models/dor/service_item.rb
peetucket: Described current behavior, but if we move to a new Tag object, this would change the return structure to an array of `Tag`s...do you think its worth creating a new Tag class for this use (agree the array of array is not ideal, but the name of a tag can be repeated as we define it, so we can't use a hash)

app/models/dor/service_item.rb
jcoyne: Yes, It could even be a nested class in ServiceItem, if we aren't going to use it externally.

jcoyne:
Thanks @peetucket. This looks good. It's still `[WIP]` though, right?

peetucket:
Yeah, still WIP -- waiting for some additional stakeholder verification/clarification.

peetucket:
No longer WIP - we can move forward with this

PR #140, opened 2018-08-28
break out tags to its own class

peetucket:
experiment in breaking out the goobi tag into a small class instead of using array of arrays to represent

### DIFF
--- a/app/models/dor/goobi_tag.rb
+++ b/app/models/dor/goobi_tag.rb
@@ -1,0 +1,10 @@
+module Dor
+  class GoobiTag
+    attr_accessor :name, :value
+
+    def initialize(p_hash)
+      @name = p_hash[:name]
+      @value = p_hash[:value]
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/models/dor/goobi_tag.rb
```diff
@@ -0,0 +1,10 @@
+module Dor
+  class GoobiTag
+    attr_accessor :name, :value
+
+    def initialize(p_hash)
+      @name = p_hash[:name]
+      @value = p_hash[:value]
+    end
```
jcoyne: You could even add a `to_xml` method here which would return:
```ruby
        "<tag name=\"#{name}\" value=\"#{value}\"></tag>"
```

jcoyne:
👏 I like this very much. Thank you.

app/models/dor/goobi_tag.rb
peetucket: good idea, i'm going to push to this branch and then submit a new PR against master (since the goobi-tag work is now merged)

peetucket:
closed in favor of #141

PR #150, opened 2018-10-03
encode tag name and values when generating goobi request xml

### LINKED ISSUE #149: Tags generated in goobi xml passed to the goobi server need to be XML encoded
When tags include an ampersand (and potentially other special characters), invalid XML is generated if we don't encode properly, which causes a 400 issue when posting to the goobi API endpoint.

peetucket:
fixes #149

### DIFF
--- a/app/models/dor/goobi_tag.rb
+++ b/app/models/dor/goobi_tag.rb
@@ -13,7 +13,10 @@ module Dor
     end
 
     def to_xml
-      "<tag name=\"#{name.encode(xml: :text)}\" value=\"#{value.encode(xml: :text)}\"></tag>"
+      Nokogiri::XML::Node.new('tag', Nokogiri::XML::Document.new).tap do |node|
+        node['name'] = name
+        node['value'] = value
+      end.to_xml
     end
   end
 end
--- a/app/models/dor/goobi_tag.rb
+++ b/app/models/dor/goobi_tag.rb
@@ -13,7 +13,7 @@ module Dor
     end
 
     def to_xml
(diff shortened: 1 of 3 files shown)


app/models/dor/goobi_tag.rb
```diff
@@ -13,7 +13,7 @@ def initialize(p_hash)
     end
 
     def to_xml
-      "<tag name=\"#{name}\" value=\"#{value}\"></tag>"
+      "<tag name=\"#{name.encode(xml: :text)}\" value=\"#{value.encode(xml: :text)}\"></tag>"
```
jcoyne: Would you consider this:
```ruby
doc = Nokogiri::XML::Document.new
Nokogiri::XML::Node.new('tag', doc).tap do |node|
  node['name'] = name
  node['value'] = value
end.to_xml
```

peetucket: i am ambivalent - the rest of the XML generation uses inline strings instead of Nokogiri, but this if fine too and takes care of encoding

peetucket: pushed up a new commit (had to alter the tests slightly since nokogiri closes tags differently)

jcoyne: Thanks Peter, I'm just thinking that if we try to use Nokogiri everwhere, then we'll never get caught on a bug like this one again.

peetucket: yup, makes sense - we can convert the other bits to nokogiri in the future

PR #152, opened 2018-10-08
remove the constant that loads an .xsl from loc so that the app can c…

peetucket:
…ome up even if loc.gov is down

### DIFF
--- a/app/models/marcxml_resource.rb
+++ b/app/models/marcxml_resource.rb
@@ -1,6 +1,5 @@
 # MARC resource model for retrieving and transforming MARC records
 class MarcxmlResource
-
   def self.find_by(catkey: nil, barcode: nil)
     if catkey
       new(catkey: catkey)
@@ -30,6 +29,7 @@ class MarcxmlResource
   end
 
   private
+
   def marc_to_mods_xslt
     @marc_to_mods_xslt ||= Nokogiri::XSLT(File.read(File.join(Rails.root, 'app', 'xslt', 'MARC21slim2MODS3-6_SDR_v1.xsl')))
   end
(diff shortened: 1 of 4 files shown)


spec/controllers/marcxml_controller_spec.rb
```diff
@@ -30,7 +30,9 @@
   end
 
   describe 'GET mods' do
-    it 'transforms the MARCXML into MODS' do
+    pending 'transforms the MARCXML into MODS' do
```
ndushay: don't we want a test that shows what happens when LoC file isn't available?  and a test that shows what happens when it IS available?  Maybe this should be a separate ticket ...  See #153

ndushay:
I won't merge b/c travis build failed.

app/models/marcxml_resource.rb
```diff
@@ -32,6 +30,10 @@ def marcxml
 
   private
 
+  def marc_to_mods_xslt
+    @marc_to_mods_xslt ||= Nokogiri::XSLT(File.open(File.join(Rails.root, 'app', 'xslt', 'MARC21slim2MODS3-6_SDR_v1.xsl')))
```
jcoyne: This is now increasing the number of calls from one time in the whole app to one call per instance.  This may increase our error rate.  Can this be a class method instead?

jcoyne: Oops, I didn't recognize this is a local file now.  This is probably not a peformance impact.

spec/controllers/marcxml_controller_spec.rb
```diff
@@ -30,7 +30,9 @@
...
+      # 10/8/2018 -- this test is currently failing due to an .xsl file at loc.gov being unavailable
```
jcoyne: But does it still depend on something at loc.gov? I thought it was switched to be local?

peetucket: indeed, test was updated to removing pending status and back to the way it was before

PR #156, opened 2018-10-11
Add part label and sort keys to the 856 export

cbeer:
This is part of https://github.com/sul-dlss/SearchWorks/issues/2037.

The desired behavior is to display the part name and number as given in the MODS record (instead of the artificial "part 1" "part 2" etc), and order them in the Searchworks sidebar according to a sortable key given in the MODS.

A future enhancement, per the emerging spec, will be to only do this behavior for items that are explicitly parts (pages, issues, etc) of the whole MARC record, but this is pending some additional infrastructure work. In this meantime, the SearchWorks app will ignore part data when it isn't useful.  

A couple example searchworks records:

http://searchworks.stanford.edu/view/8923346
https://searchworks.stanford.edu/view/8810055


jkeck:
👍 from the meeting the spec perspective.  This appears to parse the correct data and present it as will be expected by the SW app.

PR #158, opened 2018-10-16
Use namespaces for our MODS, just like our real data has

cbeer:


cbeer:
I checked with @blalbrit; it's our expectation that all MODS descMetadata will declare that MODS namespace. If it doesn't, that's a bug and should be fixed at the source.

ndushay:
works for me.  I do wonder about what will happen if anyone decides to rebuild the index truly from scratch, with all our mods data coming in afresh, but I guess that's not my concern.

PR #179, opened 2018-12-17
Containerize dor-services-app

mjgiarlo:
Include a working docker-compose configuration (w/ Solr configs) as an example to show dependencies, including config values. Also include dummy certs to override defaults that cause SSL errors when spinning up in test/dev environments.


mjgiarlo:
@jcoyne I used `vendor/` to ship the Oracle dependencies necessary for `bundle install` to build the `ruby-oci8` gem. Is there a better way to take care of this?

jcoyne:
@mjgiarlo do they grant a license to redistribute those binaries?

jcoyne:
How about `bundle --without production`?  Can we get away with that?

mjgiarlo:
@jcoyne I think so? I'll give that a shot once I have the Argo build passing.

PR #199, opened 2019-02-08
Remove dependency on Assembleable

jcoyne:
This doesn't provide anything that we use.


justinlittman:
@jcoyne Approved, but since PR has been sitting around leaving merge decision to you.

PR #206, opened 2019-03-13
Moved RegistrationService from dor-services.

justinlittman:
references https://github.com/sul-dlss/dor-services/issues/500


.rubocop_todo.yml
```diff
@@ -1,26 +1,53 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2019-02-08 14:54:24 -0600 using RuboCop version 0.57.2.
+# on 2019-03-13 08:55:20 -0400 using RuboCop version 0.65.0.
```
jcoyne: The version of rubocop in Gemfile.lock is 0.57.2.  Did you somehow use a different version?

app/services/registration_service.rb
```diff
@@ -0,0 +1,200 @@
+# frozen_string_literal: true
+
+require 'uuidtools'
```
jcoyne: Add this to the `Gemfile`

PR #208, opened 2019-03-13
Added 422 error for ArgumentErrors.

### LINKED ISSUE #193: [dor-services-app/stage] ArgumentError: Source ID must follow the format 'namespace:value', not 'foo'
## Backtrace

    line 19 of [PROJECT_ROOT]/app/controllers/objects_controller.rb: create

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/44150027)

This should be a 422 error, not a 500

justinlittman:
closes #193

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -16,7 +16,11 @@ class ObjectsController < ApplicationController
   # Register new objects in DOR
   def create
     Rails.logger.info(params.inspect)
-    reg_response = RegistrationService.create_from_request(create_params)
+    begin
+      reg_response = RegistrationService.create_from_request(create_params)
+    rescue ArgumentError => e
+      return render status: 422, plain: e.message
+    end
     Rails.logger.info(reg_response)
     pid = reg_response[:pid]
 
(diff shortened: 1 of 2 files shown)


app/controllers/objects_controller.rb
```diff
@@ -13,6 +13,10 @@ class ObjectsController < ApplicationController
     render status: 409, plain: e.message
   end
 
+  rescue_from(ArgumentError) do |e|
```
jcoyne: I think it's easier to understand this code, if you put this just on the `create` method.   That's the only place we know will throw this error, right?

app/controllers/objects_controller.rb
```diff
@@ -24,6 +24,8 @@ def create
       format.all { render status: 201, location: object_location(pid), plain: Dor::RegistrationResponse.new(reg_response).to_txt }
       format.json { render status: 201, location: object_location(pid), json: Dor::RegistrationResponse.new(reg_response) }
     end
+  rescue ArgumentError => e
```
jcoyne: Could this specifically wrap the `RegistrationService.create_from_request`? is that what raises this error?

PR #225, opened 2019-04-23
Move PublishMetadataService in from common-accessioning

jcoyne:
This was previously moved out of dor-services, but it was still being used here.
Once this is here we can refactor the common-accessioning robots to call these
services in dor-services-app.


justinlittman:
Approve pending fixed conflict.

PR #226, opened 2019-04-23
Update the ruby version for rubocop

jcoyne:


bin/repub
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env ruby
```
justinlittman: Where did the changes in this file come from?

jcoyne: This file wasn't meant for this PR.  I'll remove.

PR #233, opened 2019-04-26
Create a mechanism to refresh the metadata from Symphony

jcoyne:
Both Argo and Common-accessioning can use this


app/services/refresh_metadata_action.rb
```diff
@@ -0,0 +1,32 @@
...
+    datastream.ng_xml.normalize_text!
+    datastream.content = datastream.ng_xml.to_xml
+  end
+
+  private
+
+  def fetch_datastream
+    candidates = @object.identityMetadata.otherId.collect(&:to_s)
+    metadata_id = Dor::MetadataService.resolvable(candidates).first
+    metadata_id.nil? ? nil : Dor::MetadataService.fetch(metadata_id.to_s)
```
justinlittman: Should other candidates be tried if first doesn't succeed?

jcoyne: This is just the code that's already in place in dor-services.  I don't want to change that here.

PR #242, opened 2019-04-30
Expose the descriptive metadata as an endpoint

### LINKED ISSUE #230: Dor::PublicDescMetadataService should be merged in
Then it must be exposed as an endpoint so Argo can consume it here:
https://github.com/sul-dlss/argo/blob/fe0a3bf0ec511b36fdd6c5981e633c9a3c73f825/app/controllers/items_controller.rb#L58

Then it can be removed from dor-services

jcoyne:
This will be consumed by common-accessioning and argo.
We can then move that class to dor-services from dor-services-app

Fixes #230


justinlittman:
Approve pending resolved conflict.

PR #256, opened 2019-05-10
Workflows are no longer required as seeds.

jcoyne:
They moved to workflow-server-rails


jcoyne:
Turns out they are required for creating APOs in Argo.

PR #260, opened 2019-05-10
Use the workflow service client for starting workflows

### LINKED ISSUE #259: Avoid deprecated Dor::CreateWorkflowService
Use the workflow-service client instead

jcoyne:
Avoid deprecated Dor::CreateWorkflowService

Fixes #259

### DIFF
--- a/spec/controllers/versions_controller_spec.rb
+++ b/spec/controllers/versions_controller_spec.rb
@@ -41,7 +41,6 @@ RSpec.describe VersionsController do
     let(:open_params) do
       {
         assume_accessioned: false,
-        create_workflows_ds: false,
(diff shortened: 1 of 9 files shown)


spec/controllers/versions_controller_spec.rb
```diff
@@ -41,7 +41,7 @@
     let(:open_params) do
       {
         assume_accessioned: false,
-        create_workflows_ds: false,
+        #  create_workflows_ds: false,
```
justinlittman: Cruft?

jcoyne:
Got it removed.  Depends on https://github.com/sul-dlss/hydrus/pull/306

PR #269, opened 2019-05-16
Return a conflict when goobi returns a conflict

jcoyne:


config/settings.yml
```diff
@@ -87,7 +87,7 @@ RELEASE:
   PURL_BASE_URI : 'http://purl.stanford.edu'
 
 GOOBI:
-  URL: 'http://localhost:9292'
+  URL: 'http://goobi-env.stanford.edu:9292'
```
justinlittman: Did you intend this to change?

jcoyne: yes.  Webmock was set to allow connections to localhost. Changing the url is the easiest way for it to say "stub this connection"

PR #283, opened 2019-05-22
Adds new route for checking if a version can be opened for an object.

### LINKED ISSUE #282: Add route for checking if new version can be opened
Hydrus currently has it's own logic for checking if a version can be opened. This creates a situation where Hydrus determines that it is OK to open a new version, but the Version Service's own check throw an error.

To avoid this, dor-services-app should expose a route for checking if a version can be opened for an object.

justinlittman:
closes #282

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -21,6 +21,10 @@ class VersionsController < ApplicationController
     render build_error('Unable to close version', e)
   end
 
+  def openeable
+    render plain: VersionService.can_open?(@item, open_params).to_s
+  end
+
   private
 
   # JSON-API error response
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -48,6 +48,7 @@ Rails.application.routes.draw do
 
       resources :versions, only: [:create] do
         collection do
+          get 'openeable', to: 'openeable'
           get 'current'
           post 'current/close', action: 'close_current'
         end
--- a/spec/services/version_service_spec.rb
+++ b/spec/services/version_service_spec.rb
@@ -70,27 +70,16 @@ RSpec.describe VersionService do
       end
     end
 
-    context 'when the object has not been accessioned' do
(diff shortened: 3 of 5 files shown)


app/controllers/versions_controller.rb
```diff
@@ -21,6 +21,10 @@ def close_current
     render build_error('Unable to close version', e)
   end
 
+  def can_open?
```
jcoyne: I wouldn't use a question mark on a controller action method since the `?` doesn't translate into a URL.

config/routes.rb
```diff
@@ -48,6 +48,7 @@
 
       resources :versions, only: [:create] do
         collection do
+          get 'can_open', to: 'can_open?'
```
jcoyne: how about `openeable` ?

.rubocop.yml
```diff
@@ -31,3 +31,7 @@ RSpec/MultipleExpectations:
 RSpec/ExpectActual:
   Exclude:
     - 'spec/routing/**'
+
+RSpec/AnyInstance:
+  Exclude:
+    - 'spec/**/*'
```
jcoyne: Boo!

spec/services/version_service_spec.rb
```diff
@@ -70,37 +70,79 @@
       end
     end
 
-    context 'when the object has not been accessioned' do
+    context 'when a new version cannot be opened' do
+      before do
+        allow_any_instance_of(described_class).to receive(:raise_for_open).and_raise(Dor::Exception, 'Object net yet accessioned')
```
jcoyne: Can you expect that a version is initialized?
```
let(:instance) { instance_double(described_class) }
before do
  allow(instance).to receive(:raise_for_open).and_raise(Dor::Exception, 'Object net yet accessioned')
  allow(described_class).to receive(:new).and_return(instance)
end
```

PR #288, opened 2019-05-23
Create an endpoint for exposing the data as JSON

jcoyne:
This is the cocina data model


app/models/cocina/dro.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Cocina
+  # A digital repository object.  See https://github.com/sul-dlss-labs/taco/blob/master/maps/DRO.json
+  class DRO < Dry::Struct
```
mjgiarlo: I'm cool with DRO since that's consistent with prior work, FWIW: https://sul-dlss.github.io/taco-truck/COCINA.html#cocina-data-models--shapes

Would you consider `s/Cocina/COCINA/g` for moar consistency, or do you think in this case it is preferable to adhere to Ruby naming standards over consistency?

jcoyne: 🤷‍♂

mjgiarlo: I don't care that much about it, and it might be better not to resist the code conventions of that language we're writing in.

PR #298, opened 2019-05-29
Move the DeleteService from dor-services

jcoyne:


app/services/delete_service.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def cleanup_stacks
+    DruidTools::StacksDruid.new(druid, Dor::Config.stacks.local_stacks_root).prune!
+  end
+
+  def cleanup_purl_doc_cache
+    DruidTools::PurlDruid.new(druid, Dor::Config.stacks.local_document_cache_root).prune!
+  end
+
+  def remove_active_workflows
```
justinlittman: Mismatch between _active_ and _all_ is confusing.

jcoyne: This is just copied from dor-services without changes. I'd prefer not to make any improvements to the names  in this PR.

PR #300, opened 2019-05-29
Add a route to clear the workspace

jcoyne:
depends on #299


spec/controllers/workspaces_controller_spec.rb
```diff
@@ -64,4 +61,19 @@
...
+
+  describe '#destroy' do
+    before do
+      login
+      allow(Dor::CleanupService).to receive(:cleanup_by_druid)
+    end
+
+    let(:druid) { 'druid:aa222cc3333' }
+
+    it 'is successful' do
```
justinlittman: Need test for Dor::CleanupService.cleanup_by_druid branch.

app/controllers/workspaces_controller.rb
```diff
@@ -12,7 +10,28 @@ class WorkspacesController < ApplicationController
...
+      CleanupResetService.cleanup_by_reset_druid druid
+    else
+      Dor::CleanupService.cleanup_by_druid druid
+    end
+    head :no_content
+  end
+
+  private
+
+  # determines if the druid is the regular druid tree or the truncated one
```
justinlittman: Can you describe what a regular druid treee / truncated druid tree is?

jcoyne: I don't know, this is just moving the code from https://github.com/sul-dlss/common-accessioning/blob/09b3aa6f4740daf7cc53d61213793d5ce7148fff/lib/robots/dor_repo/dissemination/cleanup.rb

PR #321, opened 2019-08-14
Add a way to do virtual merge through an API

jcoyne:
Ref https://github.com/sul-dlss/argo/issues/1458

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,14 @@
 class ApplicationController < ActionController::API
   include ActionController::MimeResponds
 
+  rescue_from ActionController::ParameterMissing do |exception|
+    render json: {
+      errors: [
+        { title: 'bad request', detail: exception.message }
+      ]
+    }, status: :bad_request
+  end
+
   before_action :check_auth_token
 
   # Since Basic auth was already using the Authorization header, we used something
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -28,6 +28,20 @@ class ObjectsController < ApplicationController
     end
   end
 
+  # Handles updates to the record.
+  # Presently this only needs to handle the merge object use case.
+  # Do this by providing: constituent_ids => ['druid:123', 'druid:345']
+  def update
+    # validate that the constituent_ids parameter is an present, raises ActionController::ParameterMissing
+    params.require(:constituent_ids)
+    filtered_params = params.permit(constituent_ids: [])
+    raise ActionController::ParameterMissing, 'constituent_ids must be an array' unless filtered_params[:constituent_ids]
+
+    # Update the constituent relationship
+    ConstituentService.new(parent_druid: params[:id]).add(child_druids: filtered_params[:constituent_ids])
+    head :no_content
+  end
+
   def publish
     PublishMetadataService.publish(@item)
     head :created
--- a/app/services/constituent_service.rb
+++ b/app/services/constituent_service.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+# Adds a constituent relationship between a parent work and child works
+# by taking the followin actions:
+#  1. altering the contentMD of the parent
+#  2. add isConstituentOf assertions to the RELS-EXT of the children
+#  3. saving the parent and the children
+class ConstituentService
+  # @param [String] parent_druid the identifier of the parent object
+  def initialize(parent_druid:)
+    @parent_druid = parent_druid
+  end
+
+  # This resets the contentMetadataDS of the parent and then adds the child resources.
+  # Typically this is only called one time (with a list of all the pids) because
+  # subsequent calls will erase the previous changes.
+  # @param [Array<String>] child_druids the identifiers of the child objects
+  def add(child_druids:)
+    ResetContentMetadataService.new(druid: parent_druid).reset
+
+    child_druids.each do |child_druid|
+      add_constituent(child_druid: child_druid)
+    end
+    parent.save!
+  end
+
+  private
+
+  attr_reader :parent_druid
+
+  def add_constituent(child_druid:)
+    child = ItemQueryService.find_modifiable_work(child_druid)
+    child.contentMetadata.ng_xml.search('//resource').each do |resource|
+      parent.contentMetadata.add_virtual_resource(child.id, resource)
+    end
+    child.add_relationship :is_constituent_of, parent
+    child.save!
+  end
+
+  def parent
+    @parent ||= ItemQueryService.find_modifiable_work(parent_druid)
+  end
+end
--- a/app/services/reset_content_metadata_service.rb
+++ b/app/services/reset_content_metadata_service.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Clears the contentMetadata datastream to the default, wiping out any members.
+class ResetContentMetadataService
+  def initialize(druid:, type: 'image')
+    @druid = druid
+    @type = type
+  end
+
+  def reset
+    work = ItemQueryService.find_modifiable_work(@druid)
+    work.contentMetadata.content = "<contentMetadata objectId='#{work.id}' type='#{@type}'/>"
+    work.save!
+  end
+end
--- a/spec/requests/virtual_merge_spec.rb
+++ b/spec/requests/virtual_merge_spec.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Virtual merge of objects' do
+  let(:payload) { { sub: 'argo' } }
+  let(:jwt) { JWT.encode(payload, Settings.dor.hmac_secret, 'HS256') }
+  let(:parent_id) { 'druid:mk420bs7601' }
+  let(:child1_id) { 'druid:child1' }
+  let(:child2_id) { 'druid:child2' }
+
+  let(:object) { Dor::Item.new(pid: parent_id) }
+  let(:service) { instance_double(ConstituentService, add: true) }
+
+  before do
+    allow(Dor).to receive(:find).and_return(object)
+    allow(ConstituentService).to receive(:new).with(parent_druid: parent_id).and_return(service)
+  end
+
+  context 'when constituent_ids is provided' do
+    it 'merges the objects' do
+      put "/v1/objects/#{parent_id}",
+          params: { constituent_ids: [child1_id, child2_id] },
+          headers: { 'X-Auth' => "Bearer #{jwt}" }
+      expect(service).to have_received(:add).with(child_druids: [child1_id, child2_id])
+      expect(response).to be_successful
+    end
+  end
+
+  context 'when constituent_ids is not provided' do
+    it 'renders an error' do
+      put "/v1/objects/#{parent_id}",
+          params: { title: 'New name' },
+          headers: { 'X-Auth' => "Bearer #{jwt}" }
+      expect(service).not_to have_received(:add)
+      expect(response).to be_bad_request
+      json = JSON.parse(response.body)
+      expect(json['errors'][0]['detail']).to eq 'param is missing or the value is empty: constituent_ids'
+    end
+  end
+
+  context 'when constituent_ids is not an array' do
+    it 'renders an error' do
+      put "/v1/objects/#{parent_id}",
+          params: { constituent_ids: child1_id },
+          headers: { 'X-Auth' => "Bearer #{jwt}" }
+      expect(service).not_to have_received(:add)
+      expect(response).to be_bad_request
+      json = JSON.parse(response.body)
+      expect(json['errors'][0]['detail']).to eq 'param is missing or the value is empty: constituent_ids must be an array'
+    end
+  end
+end
--- a/spec/services/constituent_service_spec.rb
+++ b/spec/services/constituent_service_spec.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ConstituentService do
+  let(:parent) do
+    Dor::Item.new.tap do |item|
+      item.contentMetadata.content = <<~XML
+        <contentMetadata objectId="druid:parent1" type="image">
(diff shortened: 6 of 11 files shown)


app/services/constituent_service.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# TODO: this should move to dor-services-app and dor-services-client
+# because we don't want Argo to have to know about the XML backing the model.
+#
+# Adds a constitutent relationship between a parent work and child works
```
mjgiarlo: `s/constitutent/constituent/g`

app/controllers/objects_controller.rb
```diff
@@ -28,6 +28,31 @@ def create
...
+    end
+
+    # Update the constituent relationship
+    begin
+      ConstituentService.new(parent_druid: params[:id]).add(child_druids: params[:constituent_ids])
+      head :no_content
+    rescue StandardError => e
+      return render json: {
+        errors: [
+          { title: 'bad request', detail: e }
```
mjgiarlo: I wonder if this ought to be an internal server error 500 given this catches all sorts of exceptional behavior.

app/services/constituent_service.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# TODO: this should move to dor-services-app and dor-services-client
```
mjgiarlo: I think this TODO can be removed.

app/services/constituent_service.rb
```diff
@@ -0,0 +1,41 @@
...
+# because we don't want Argo to have to know about the XML backing the model.
+#
+# Adds a constituent relationship between a parent work and child works
+# by taking the followin actions:
+#  1. altering the contentMD of the parent
+#  2. add isConstituentOf assertions to the RELS-EXT of the children
+#  3. saving the parent and the child
```
mjgiarlo: `s/child/child(ren)/` ?

app/controllers/objects_controller.rb
```diff
@@ -28,6 +28,31 @@ def create
...
+      return render json: {
+        errors: [
+          { title: 'bad request', detail: 'Currently only constituent_ids are supported for update' }
+        ]
+      }, status: :bad_request
+    end
+
+    # Update the constituent relationship
+    begin
+      ConstituentService.new(parent_druid: params[:id]).add(child_druids: params[:constituent_ids])
```
mjgiarlo: If only one value is in `constituent_ids`, will it still come across as an array?

app/services/constituent_service.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+    child.add_relationship :is_constituent_of, parent
+    child.save!
+  end
+
+  def parent
+    @parent ||= WorkQueryService.find_modifiable_work(parent_druid)
+  end
+
+  attr_reader :parent_druid
```
mjgiarlo: I have a mild preference for seeing this immediately below the `private` keyword but nbd.

app/services/reset_content_metadata_service.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Clears the contentMetadata datastream to the default, wiping out any members.
+class ResetContentMetadataService
+  def initialize(druid:, type: 'image')
+    @druid = druid
+    @type = type
+  end
+
+  def reset
```
mjgiarlo: Appreciate the rename from `purge` to `reset`!

app/controllers/objects_controller.rb
ndushay: I further wonder if the ConstituentService would throw a specific error, or set of them to be caught here?

app/services/constituent_service.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+    parent.save!
+  end
+
+  private
+
+  def add_constituent(child_druid:)
+    child = WorkQueryService.find_modifiable_work(child_druid)
+    child.contentMetadata.ng_xml.search('//resource').each do |resource|
+      parent.contentMetadata.add_virtual_resource(child.id, resource)
```
ndushay: I'm assuming there's good reason to leave the `add_virtual_resource` method in dor-services rather than port it here?  See dor-services/issues/652.   (Like -- we'd have to pull over all kinds of things to get the same functionality natively in this app?  or ... this app is just gating the access to dor-services gem?)

app/services/reset_content_metadata_service.rb
ndushay: agree - nice.

app/services/constituent_service.rb
ndushay: I think we *might* want to keep the logging message for debugging?  https://github.com/sul-dlss/argo/blob/master/bin/virtual-merge#L33

I'm ok with removing it if @andrewjbtw is ok with it.  I understand it's on a diff VM than the one running the merge script ... just making sure we don't need the breadcrumb.

app/services/work_query_service.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# Responsible for retrieving information based on the given work (Dor::Item).
+class WorkQueryService
```
ndushay: ooo - work is an overloaded term.   so is "object" ...  `DorItemQueryService` ?   `ItemQueryService`?  `ObjectQueryService` ?  

I'm concerned the existing naming will be confusing, esp. over time.

app/services/work_query_service.rb
```diff
@@ -0,0 +1,36 @@
...
+  # @param [#exists?, #find] work_relation - How we will query some of the related information
+  def initialize(id:, work_relation: default_work_relation)
+    @id = id
+    @work_relation = work_relation
+  end
+
+  delegate :allows_modification?, to: :work
+
+  # @raises [RuntimeError] if the item is not modifiable
+  def self.find_modifiable_work(druid)
```
ndushay: is "find" the right concept here?   It's more like `assert_modifiable` or even `.modifiable?` or eve `.allows_modification?`, no?   

Same concern about naming --  this is really a method to assert that the item allows_modification;  I'm concern that the terms "find" and "work" will create confusion.

spec/requests/virtual_merge_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    allow(ConstituentService).to receive(:new).with(parent_druid: parent_id).and_return(service)
+  end
+
+  it 'merges the objects' do
+    put "/v1/objects/#{parent_id}",
+        params: { constituent_ids: [child1_id, child2_id] },
+        headers: { 'X-Auth' => "Bearer #{jwt}" }
+    expect(service).to have_received(:add).with(child_druids: [child1_id, child2_id])
+    expect(response).to be_successful
+  end
```
ndushay: Add a test showing the ResetContentMetadataService.reset is called on the parent, as well?

spec/services/constituent_service_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    context 'when the parent is open for modification' do
+      before do
+        add
+      end
+
+      it 'merges objects' do
+        expect(parent.contentMetadata.content).to be_equivalent_to <<~XML
+          <contentMetadata objectId="druid:parent1" type="image">
+            <resource sequence="1" id="#{namespaceless}_1" type="image">
```
ndushay: I believe this is incorrect.  See:

https://github.com/sul-dlss-deprecated/dor-utils/issues/17#issuecomment-518770583 -- we are missing externalFile part.

or 

https://argo.stanford.edu/view/druid:bb258cw1178   contentMetadata

spec/services/constituent_service_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+              <relationship type="alsoAvailableAs" objectId="#{child1.id}"/>
+            </resource>
+            <resource sequence="2" id="#{namespaceless}_2" type="image">
+              <relationship type="alsoAvailableAs" objectId="#{child2.id}"/>
+            </resource>
+          </contentMetadata>
+        XML
+        expect(child1.object_relations[:is_constituent_of]).to eq [parent]
+        expect(child2.object_relations[:is_constituent_of]).to eq [parent]
```
ndushay: it would be good to show this relationship doesn't exist before running the command.

app/services/constituent_service.rb
```diff
@@ -0,0 +1,41 @@
...
+# by taking the followin actions:
+#  1. altering the contentMD of the parent
+#  2. add isConstituentOf assertions to the RELS-EXT of the children
+#  3. saving the parent and the child
+class ConstituentService
+  def initialize(parent_druid:)
+    @parent_druid = parent_druid
+  end
+
+  def add(child_druids:)
```
ndushay: would this be clearer as "add_children"  or "add_constituents"?

spec/services/constituent_service_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      it 'merges nothing' do
+        expect { add }.to raise_error RuntimeError
+        expect(parent.contentMetadata.content).to be_equivalent_to <<~XML
+          <contentMetadata />
+        XML
+        expect(child1.object_relations[:is_constituent_of]).to be_empty
+      end
+    end
+
+    context 'when the child is closed for modification' do
```
ndushay: when _any_ child is close for modification?  Should test show second child closed for modification, and should it show what happens to parent contentMd in that case??

app/controllers/objects_controller.rb
jcoyne: I'm just going to drop the rescue block and let the Rails middleware + Honeybadger deal with any errors.  I don't see any benefit to specially handling the errors at this point.

app/services/work_query_service.rb
jcoyne: I don't like Item as it seems to be too generic (it is used in every domain).  "Work" is what Samvera has choosen and in the interest of aligning our practices with the greater community, that's what I'm inclined to choose.

spec/requests/virtual_merge_spec.rb
jcoyne: ResetContentMetadataService is called by ConstituentService, which we are checking for.

spec/services/constituent_service_spec.rb
jcoyne: The method is moving over the whole node, but the xml we're using as a fixture just doesn't have any children.

spec/services/constituent_service_spec.rb
jcoyne: We are showing that by creating the child fixture objects without any relationships at all.

app/services/constituent_service.rb
jcoyne: My thinking is that the ConstituentService class provides that context.

spec/services/constituent_service_spec.rb
jcoyne: I'm not sure I understand what you are asking for here.   Why does it matter which child is closed?

app/services/constituent_service.rb
jcoyne: Im trying to keep the changes to a manageable size. Moving add_virtual_resource is not required to gain the functionality added in this PR. 652 is still valid and can be done later.

app/services/work_query_service.rb
jcoyne: It is finding and ensuring that it’s modifable. I’m not sure I understand why this is a problem.

app/services/work_query_service.rb
mjgiarlo: Alignment with Samvera is generally fine by me and I'm accustomed to the "work" verbiage at this point. That said, the term we use most in other SDR codebases is "Object," in my reckoning, so I'd be inclined to switch to that if we're switching to anything. (I see a Dor::Item as a subclass of SDR "objects.")

jcoyne: In the TACO work, we called it a DRO, so maybe we should go that direction? https://github.com/sul-dlss-labs/sdr3-models/blob/master/taco-maps/DRO.json

mjgiarlo: @jcoyne I figured `Object` was defensible because DSA is Fedora-facing, and `Object` is standard Fedora nomenclature. I don't care *too* much; work, item, object, and DRO all work for me.

app/services/constituent_service.rb
ndushay: totally makes sense.  thanks for the explanation.

spec/services/constituent_service_spec.rb
ndushay: I was thinking it would be appropriate to show that the process ends even if a child in the middle is the one that can't be modified.  That is, showing a partially "merged" situation could be a more informative use case for future spelunking.

app/services/constituent_service.rb
ndushay: k.  certainly not something I'd go to the mat over.

spec/services/constituent_service_spec.rb
ndushay: It would be more useful to show the full "what we want it to look like when it is successful" here ... imagine trying to figure out if it's working right when you don't have a good example of what is supposed to happen in the tests.

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,14 @@
 class ApplicationController < ActionController::API
   include ActionController::MimeResponds
 
+  rescue_from ActionController::ParameterMissing do |exception|
+    render json: {
+      errors: [
+        { title: 'bad request', detail: exception.message }
+      ]
+    }, status: :bad_request
+  end
```
ndushay: I'm glad to see trapping for this problem and a better error, but ... should it be a separate commit, in a perfect world?  (no need to rework, just asking)

spec/services/constituent_service_spec.rb
jcoyne: That depends on the starting value of the publish attribute: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/datastreams/content_metadata_ds.rb#L188  That varying behavior is tested in the dor-services gem.

spec/services/constituent_service_spec.rb
jcoyne: done

PR #328, opened 2019-08-15
Move DruidTools::Druid#prune! and #prune_ancestors from druid-tools

### LINKED ISSUE #327: Move DruidTools::Druid#prune! and #prune_ancestors to dor-services-app
This repo is the only place they are used.

jcoyne:
This will allow the methods in druid-tools to be deprecated because dor-services-app is the only place they are used

Fixes #327

### DIFF
--- a/app/services/cleanup_reset_service.rb
+++ b/app/services/cleanup_reset_service.rb
@@ -36,7 +36,7 @@ class CleanupResetService
     reset_directories.each do |path|
       FileUtils.rm_rf(path)
     end
-    base_druid.prune_ancestors(base_druid.pathname.parent)
+    PruneService.new(druid: base_druid).prune_ancestors(base_druid.pathname.parent)
   end
 
   # @param [String] base_druid_tree The base directory to delete from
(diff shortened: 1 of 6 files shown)


jcoyne:
@mjgiarlo thanks for catching that.  Looks like I forgot a `git add`

app/services/cleanup_reset_service.rb
```diff
@@ -36,7 +36,7 @@ def self.cleanup_reset_workspace_content(druid, last_version, base)
     reset_directories.each do |path|
       FileUtils.rm_rf(path)
     end
-    base_druid.prune_ancestors(base_druid.pathname.parent)
+    PruneService.new(druid: base_druid).prune_ancestors(base_druid.pathname.parent)
```
mjgiarlo: Ideally the CleanupResetService specs would change to do less testing of effects caused by other components and test, in this case, that PruneService was instantiated and `#prune_ancestors` was called. That said, I don't think these changes are necessary for this PR.

app/services/cleanup_service.rb
```diff
@@ -14,7 +14,7 @@ def self.cleanup_by_druid(druid)
   # @param [String] base The base directory to delete from
   # @return [void] remove the object's data files from the workspace area
   def self.cleanup_workspace_content(druid, base)
-    DruidTools::Druid.new(druid, base).prune!
+    PruneService.new(druid: DruidTools::Druid.new(druid, base)).prune!
```
mjgiarlo: Same comment as on CleanupResetService (in which I convinced myself that changing the specs for these two services does not *need* to change in this PR, though we should revisit that at some point, lest these unit tests be a tad on the integration-y side).

app/services/delete_service.rb
```diff
@@ -33,11 +33,13 @@ def destroy
   attr_reader :druid
 
   def cleanup_stacks
-    DruidTools::StacksDruid.new(druid, Dor::Config.stacks.local_stacks_root).prune!
+    stacks_druid = DruidTools::StacksDruid.new(druid, Dor::Config.stacks.local_stacks_root)
```
mjgiarlo: Same as above.

PR #329, opened 2019-08-15
Move mkdir and mkdir_with_final_link from DruidTools

jcoyne:
Also moved rmdir, but it doesn't appear to be used

### DIFF
--- a/app/controllers/workspaces_controller.rb
+++ b/app/controllers/workspaces_controller.rb
@@ -2,7 +2,7 @@
 
 # Handles API routes for managing the DOR workspace
 class WorkspacesController < ApplicationController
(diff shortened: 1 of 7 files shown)


mjgiarlo:
@jcoyne 	:speech_balloon: 
> Also moved rmdir, but it doesn't appear to be used

Here, I guess? https://github.com/sul-dlss/dor-services-app/blob/master/app/services/prune_service.rb#L21

jcoyne:
@mjgiarlo nope, that's a `Pathname`

jcoyne:
Closing, because I'm seeing mkdir is used elsewhere in druidTools.

PR #333, opened 2019-08-19
Remove all traces of `openeable` and prepare for 3.0.0 release

### LINKED ISSUE #322: Change all references to `openeable` to `openable`
Note that this typo was already corrected and deprecated in dor-services-client as of version 2.0.0.

mjgiarlo:
Fixes #322

Includes:

* Remove all traces of `openeable` (which is no longer used by any code in the `sul-dlss` organization now that dor-services-client 2.1.0 was rolled out to all affected codebases) 
* Remove deprecation dependency now that it is no longer being used
* Remove deprecated workspace initialization route (which is no longer used by any code in the `sul-dlss` organization)
* Bump version to 3.0.0 in advance of release

**NOTE**: Without the spec changes in `spec/controllers/workspaces_controller_spec.rb` related to the object identifier param, four tests fail with:
```
   Failure/Error: post :create, params: { id: item.pid }                                                                                            
                                                                                                                                                      
     ActionController::UrlGenerationError:                                                                                                            
       No route matches {:action=>"create", :controller=>"workspaces", :id=>"druid:aa123bb4567"}
```


spec/controllers/workspaces_controller_spec.rb
```diff
@@ -26,12 +26,12 @@
     end
 
     it 'creates a druid tree in the dor workspace for the passed in druid' do
-      post :create, params: { id: item.pid }
+      post :create, params: { object_id: item.pid }
```
jmartin-sul: i see how this lines up with the other test in the file, but not sure i understand why the old version would have worked?  rails routing magic?  mostly curious for my own understanding.

mjgiarlo: @jcoyne I have the same question @jmartin-sul has :point_up:. Maybe you can shed light on this?

mjgiarlo: @jmartin-sul I chatted with @jcoyne about this and here's my understanding:

Prior to [this change removing the member route for initializing the workspace](https://github.com/sul-dlss/dor-services-app/pull/333/files#diff-21497849d8f00507c9c8dcaf6288b136L25), `params[:id]` was available when hitting `WorkspacesController#create`. To see evidence of this, check out the member route for publishing objects which used to be the line below:                

```
publish_object POST   /v1/objects/:id/publish(.:format)                       objects#publish
```

You can see the `:id` there. Why does `WorkspacesController#create` now use `:object_id` instead? Because it is a nested route, which changes Rails' parameter naming behavior. See related post on SO here: https://stackoverflow.com/questions/7666513/naming-params-of-nested-routes

Does that make sense?

mjgiarlo: Merging in the meantime. Thanks for reviewing and flagging this!

jmartin-sul: ah yeah, makes sense, and there was indeed a lack of understanding of rails routing on my part.  thanks for the explanation!

mjgiarlo: (and mine as well!)

PR #335, opened 2019-08-22
Return HTTP 400 with an informative message when metadata to be published

### LINKED ISSUE #331: dor-services-app errors raised need to surface more info to caller
(related to sul-dlss/common-accessioning/issues/188)

For example, this is what the dor-services-client gets from a raised error in the app for the publish action:

"Dor::Services::Client::UnexpectedResponse: Internal Server Error: 500 ()"  (see https://app.honeybadger.io/projects/52894/faults/50003137, which was thrown for `bq454cj1549` going through common-accessioning)

the corresponding dor-services-app  error: https://app.honeybadger.io/projects/50568/faults/49013058, which is the same as:

When I take the mods from `bq454cj1549`, which is 

```xml
      <mods xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.loc.gov/mods/v3" version="3.6" xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-6.xsd">
        <typeOfResource/>
        <originInfo>
          <issuance/>
        </originInfo>
        <recordInfo>
          <recordIdentifier source="SIRSI">a13132918</recordIdentifier>
          <recordOrigin>Converted from MARCXML to MODS version 3.6 using MARC21slim2MODS3-6_SDR.xsl (SUL version 1 2018/06/13; LC Revision 1.118 2018/01/31)</recordOrigin>
        </recordInfo>
      </mods>
```

and I run that mods xml through the DublinCoreService `.ng_xml` method, there is more detail:

```
#<DublinCoreService::CrosswalkError: DublinCoreService#ng_xml produced incorrect xml (no children):
       <...ion="http://www.openarchives.org/OAI/2.0/oai_dc/ http://www.openarchives.org/OAI/2.0/oai_dc.xsd"/>
       > with backtrace:
         # ./app/services/dublin_core_service.rb:21:in `ng_xml'
         # ./spec/services/dublin_core_service_spec.rb:11:in `block (3 levels) in <top (required)>'
         # ./spec/services/dublin_core_service_spec.rb:49:in `block (4 levels) in <top (required)>'
         # ./spec/services/dublin_core_service_spec.rb:49:in `block (3 levels) in <top (required)>'
```

I would be great to get the specific error and its message included in the 500 error consumed by the caller. 

This is probably a rails config setting or settings that can be changed  (e.g. in `config/application.rb`):

https://guides.rubyonrails.org/v5.2/configuring.html#rails-general-configuration

> config.consider_all_requests_local is a flag. If true then any error will cause detailed debugging information to be dumped in the HTTP response, and the Rails::Info controller will show the application runtime context in /rails/info/properties. true by default in development and test environments, and false in production mode. For finer-grained control, set this to false and implement local_request? in controllers to specify which requests should provide debugging information on errors.

`local_request?`  on the object controller at a minimum, I'd say.

https://apidock.com/rails/Rails/ApplicationController/local_request%3F

mjgiarlo:
Fixes #331

@ndushay This is a very specific fix to #331. I'm not sure if you were hoping for a more general (across-the-board) fix. I'm probably inclined to fix the narrowest case and add more error handling as we find other exceptions that don't bubble up useful information, but I am open to pushback on this if you think it's too short-sighted.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -11,6 +11,10 @@ class ObjectsController < ApplicationController
     render status: 409, plain: e.message, location: object_location(e.pid)
   end
 
+  rescue_from(DublinCoreService::CrosswalkError) do |e|
+    render status: 400, plain: e.message
+  end
+
   # Register new objects in DOR
   def create
     Rails.logger.info(params.inspect)
--- a/spec/controllers/objects_controller_spec.rb
+++ b/spec/controllers/objects_controller_spec.rb
@@ -53,10 +53,21 @@ RSpec.describe ObjectsController do
   end
 
   describe '/publish' do
-    it 'calls publish metadata' do
+    it 'calls PublishMetadataService and returns 201' do
       expect(PublishMetadataService).to receive(:publish).with(item)
       post :publish, params: { id: item.pid }
       expect(response.status).to eq(201)
+    end
+
+    context 'with bad metadata' do
+      let(:error_message) { "DublinCoreService#ng_xml produced incorrect xml (no children):\n<xml/>" }
+
+      it 'returns a 400 error' do
+        allow(PublishMetadataService).to receive(:publish).and_raise(DublinCoreService::CrosswalkError, error_message)
+        post :publish, params: { id: item.pid }
+        expect(response.status).to eq(400)
+        expect(response.body).to eq(error_message)
+      end
     end
   end
 


ndushay:
I was thinking more broadly, as, as stated in the ticket (which this PR closed, ahem), I believe we can possibly change a rails setting and get more info from the specific error.

Should I re-open the ticket?

ndushay:
And, by the way, it was a **500** error for the publish step, so not sure this even fixes that.

mjgiarlo:
@ndushay It does fix the 500 error, which was caused by a lack of more specific exception handling in the app.

app/controllers/objects_controller.rb
```diff
@@ -11,6 +11,10 @@ class ObjectsController < ApplicationController
     render status: 409, plain: e.message, location: object_location(e.pid)
   end
 
+  rescue_from(DublinCoreService::CrosswalkError) do |e|
+    render status: 400, plain: e.message
```
ndushay: 400 is misleading -- there is nothing wrong with the REQUEST -- it is the dor object empty descMetadata that is the problem.

PR #345, opened 2019-08-27
Combinable items

### LINKED ISSUE #334: ConstituentService should also check objects for (not) dark or citation_only
~~requires sul-dlss/dor-rights-auth#40 (new release of dor-rights-auth with `dark?` and `citation_only?` methods)~~

both the parent and the child objects should be
- open for modification
- not dark 
- not citation_only (ditto)

ndushay:
Fixes #334 - ensures ConstituentService does not combine dark or citation_only objects.


ndushay:
@jcoyne I changed ResetContentMetadataService to take the object, as we discussed this morning.

PR #356, opened 2019-08-29
Create an endpoint for modifying an object's embargo date

### LINKED ISSUE #353: Create an endpoint that allows an embargo date to be modified.
It can be called from Argo's ItemsController#embargo_update: https://github.com/sul-dlss/argo/blob/master/app/controllers/items_controller.rb#L132-L133 

It should also create an event in the events datastream.

mjgiarlo:
Fixes #353

And create a new event on the object associated with the modification.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -56,6 +56,16 @@ class ObjectsController < ApplicationController
     head :created
   end
 
+  def update_embargo
+    # validate that the embargo_date and requesting_user parameters were provided
+    params.require([:embargo_date, :requesting_user])
+    Dor::EmbargoService.new(@item).update(params[:embargo_date])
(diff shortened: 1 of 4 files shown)


app/controllers/objects_controller.rb
```diff
@@ -56,6 +56,16 @@ def update_marc_record
     head :created
   end
 
+  def update_embargo
```
jcoyne: I wonder if this should be on an embargo controller (e.g. `EmbargosController#update)?  Embargos controller could be nested under objects.

mjgiarlo: Oh, I like that even better. Worth ticketing/revising, or meh?

jcoyne: Yes, let's fix it now before we change the client.

mjgiarlo: @jcoyne 	ok, will do that momentarily.

PR #366, opened 2019-09-05
Validate constituents before creating virtual object

### LINKED ISSUE #365: Validate all children before adding as constituents
Validate all children early before touching any objects and fail fast if any children can't be combined.

Blocks https://github.com/sul-dlss/argo/issues/1463

mjgiarlo:
Fixes #365 

Includes:
* Add `ItemQueryService.validate_combinable_items`. This new class method takes an array of druids and makes sure each is combinable. If any are not, a hash of errors is created and returned.
* Validate all items to be combined before modifying any of them.
* Raise exception from objects controller when constituent service raises.

### DIFF
--- a/app/services/constituent_service.rb
+++ b/app/services/constituent_service.rb
@@ -15,13 +15,21 @@ class ConstituentService
   # Typically this is only called one time (with a list of all the pids) because
   # subsequent calls will erase the previous changes.
   # @param [Array<String>] child_druids the identifiers of the child objects
+  # @raises ActiveFedora::RecordInvalid if AF object validations fail on #save!
+  # @returns [NilClass, Hash] true if successful, hash of errors otherwise (if combinable validation fails)
   def add(child_druids:)
+    errors = ItemQueryService.validate_combinable_items([parent_druid] + child_druids)
+
+    return errors if errors.any?
+
     ResetContentMetadataService.new(item: parent).reset
 
     child_druids.each do |child_druid|
       add_constituent(child_druid: child_druid)
     end
     parent.save!
+
+    nil
   end
 
   private
--- a/app/services/item_query_service.rb
+++ b/app/services/item_query_service.rb
@@ -2,6 +2,8 @@
 
 # Responsible for retrieving information based on the given Dor::Item.
 class ItemQueryService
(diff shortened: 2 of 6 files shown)


app/services/item_query_service.rb
```diff
@@ -11,6 +11,20 @@ def initialize(id:, item_relation: default_item_relation)
 
   delegate :allows_modification?, to: :item
 
+  # @param [Array] druids a list of druids
+  # @raises [RuntimeError] if any items are dark, citation_only, or not modifiable
```
jcoyne: I don’t think this raises

app/services/constituent_service.rb
```diff
@@ -15,7 +17,12 @@ def initialize(parent_druid:)
   # Typically this is only called one time (with a list of all the pids) because
   # subsequent calls will erase the previous changes.
   # @param [Array<String>] child_druids the identifiers of the child objects
+  # @raises NotCombinableError if validation of parent and child objects returns errors
```
jcoyne: Could this return errors rather than raising an exception?

app/services/item_query_service.rb
```diff
@@ -11,6 +11,20 @@ def initialize(id:, item_relation: default_item_relation)
...
+  def self.validate_combinable_items(druids)
+    errors = {}
+
+    druids.each do |druid|
+      find_combinable_item(druid)
+    rescue RuntimeError => e
```
jcoyne: RuntimeError seems a bit broad.  Can we update find_combinable_item to raise a custom `UncombinableItemError`?

app/services/item_query_service.rb
mjgiarlo: It does not. This was from a prior implementation I worked on. Removing.

app/services/constituent_service.rb
mjgiarlo: Done.

app/services/item_query_service.rb
mjgiarlo: Done.

mjgiarlo:
@jcoyne 	this is ready for another look now. Thanks for the review!

Issue #375, opened 2019-09-09
Close versions when adding constituents?


mjgiarlo:
When virtual objects are created in Argo, it first uses dor-services-client to hit dor-services-app's `ObjectsController#update` endpoint to add constituent objects to a virtual objects:

https://github.com/sul-dlss/argo/pull/1552/files#diff-66bc912cfe9ab565ae0f4b8a8c2414a7R34-R35

You can see here that if this operation encounters any errors, the rest of the method is short-circuited and an error is returned without modifying any objects:

https://github.com/sul-dlss/dor-services-app/blob/1f486a6976e62b818d8a6175fa9921921971a3ec/app/controllers/objects_controller.rb#L53

The specific error conditions that are checked are 1) does the object allow modification?, 2) is it dark?, and 3) is it citation-only? See:

https://github.com/sul-dlss/dor-services-app/blob/1f486a6976e62b818d8a6175fa9921921971a3ec/app/services/item_query_service.rb#L33-L35

The `Dor::StateService` is responsible for determining whether an object may be modified:

https://github.com/sul-dlss/dor-services/blob/16135e6b831bae47356c95057d087b5733ce4ba9/lib/dor/services/state_service.rb#L10-L12

If this all goes fine, Argo then loops over the parent object and all child objects to close their versions:

https://github.com/sul-dlss/argo/pull/1552/files#diff-66bc912cfe9ab565ae0f4b8a8c2414a7R36-R40

This is undesirable for three reasons, particularly if/when version closing raises an exception:

1. It makes calculating and reporting overall success/failure numbers challenging. (This is where we lament the lack of transactionality in Fedora.) If the `#add_constituents` call succeeds and a single child object fails on version-close, is that a single failure even though the child object was added as a constituent of the virtual object?
2. It potentially leaves affected objects in an unexpected state (added to a virtual object but still possibly open for modification?).
3. Performance/latency: for each object in the virtual, it adds an additional HTTP call... to the service it already hit. 

Meanwhile, if we don't have a use case for adding constituents *without* closing versions, why **not** move the version closing into the `#add_constituents` operation for these reasons?

The potentially tricky part of doing this work is figuring out whether the current `#allows_modification?` conditions are more or less the same as the conditions checked when closing a version:

https://github.com/sul-dlss/dor-services-app/blob/91a0305a4a1ca65a51221bf80e63c409bd25149a/app/services/version_service.rb#L74-L76

mjgiarlo:
[...]

PR #379, opened 2019-09-10
Allow virtual objects to be created in batches

mjgiarlo:
Connects to #375 
Connects to sul-dlss/argo#1463

Includes:
* Mark `ObjectsController#update` as soon as be removed
* Make clear in `ItemQueryService` which parents are associated with which combination errors
* Add class method for convenience of checking if a work is open for versioning
* Check whether item is open or openable, NOT whether it allows modification, within `ItemQueryService`
* Handle versioning concerns when adding constituents
* Extract `VirtualObjectsController` from `ObjectsController`: it functions identically to the `ObjectsController` implementation except it can create many virtual objects at a time.
* Enhance parameter validation in `VirtualObjectsController` by using dry-schema

### DIFF
--- a/app/controllers/virtual_objects_controller.rb
+++ b/app/controllers/virtual_objects_controller.rb
@@ -2,27 +2,62 @@
 
 # Create virtual objects
 class VirtualObjectsController < ApplicationController
-  # Create one or more virtual objects represented by JSON:
-  # {
-  #   virtual_objects: [
-  #     { parent_id: '', child_ids: [] }
-  #   ]
-  # }
-  def create
-    # validate that the virtual_objects parameter is an present, raises ActionController::ParameterMissing
-    params.require(:virtual_objects)
-    filtered_params = params.permit(virtual_objects: [:parent_id, child_ids: []])
-    raise ActionController::ParameterMissing, 'virtual_objects must be an array' unless filtered_params[:virtual_objects]
+  before_action :validate_params!, only: :create
 
+  # Create one or more virtual objects represented by JSON (see `#schema` below):
+  def create
     errors = []
 
-    filtered_params[:virtual_objects].each do |virtual_object|
+    create_params[:virtual_objects].each do |virtual_object|
+      parent_id, child_ids = virtual_object.values_at(:parent_id, :child_ids)
       # Update the constituent relationship between the parent and child druids
-      errors << ConstituentService.new(parent_druid: virtual_object[:parent_id]).add(child_druids: virtual_object[:child_ids])
+      errors << ConstituentService.new(parent_druid: parent_id).add(child_druids: child_ids)
+    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest => e
+      errors << { parent_id => [e.message] }
     end
 
-    return render json: { errors: errors }, status: :unprocessable_entity if errors.any?
+    return render_error(errors: errors, status: :unprocessable_entity) if errors.any?
 
     head :no_content
+  end
+
+  private
+
+  def create_params
+    params.to_unsafe_h
+  end
+
+  # Use dry-validation to validate params instead of Rails strong parameters.
+  # This gives us finer-grained validation.
+  def validate_params!
+    errors = schema.call(create_params).errors(full: true)
+    return render_error(errors: errors, status: :bad_request) if errors.any?
+  end
+
+  def render_error(errors:, status:)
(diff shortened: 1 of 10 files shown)


spec/services/version_service_spec.rb
```diff
@@ -18,6 +18,20 @@
     allow(obj.inner_object).to receive(:repository).and_return(double('frepo').as_null_object)
   end
 
+  describe '.open?' do
+    let(:instance) { instance_double(described_class) }
```
jcoyne: You can add `open_for_versioning?: true` here and then remove the `allow` on line 29.

app/controllers/virtual_objects_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+class VirtualObjectsController < ApplicationController
+  # Create one or more virtual objects represented by JSON:
+  # {
+  #   virtual_objects: [
+  #     { parent_id: '', child_ids: [] }
+  #   ]
+  # }
+  def create
+    # validate that the virtual_objects parameter is an present, raises ActionController::ParameterMissing
+    params.require(:virtual_objects)
```
mjgiarlo: Extract to private method, `create_params`.

app/controllers/virtual_objects_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    filtered_params = params.permit(virtual_objects: [:parent_id, child_ids: []])
+    raise ActionController::ParameterMissing, 'virtual_objects must be an array' unless filtered_params[:virtual_objects]
```
mjgiarlo: Look into parameter validation.

spec/services/version_service_spec.rb
mjgiarlo: Will do.

app/controllers/virtual_objects_controller.rb
mjgiarlo: Use Json schema?

app/services/constituent_service.rb
```diff
@@ -18,16 +18,21 @@ def initialize(parent_druid:)
   # @raises ActiveFedora::RecordInvalid if AF object validations fail on #save!
   # @returns [NilClass, Hash] true if successful, hash of errors otherwise (if combinable validation fails)
   def add(child_druids:)
-    errors = ItemQueryService.validate_combinable_items([parent_druid] + child_druids)
+    errors = ItemQueryService.validate_combinable_items(parent: parent_druid, children: child_druids)
```
jcoyne: 👏

app/services/constituent_service.rb
```diff
@@ -38,11 +43,16 @@ def add(child_druids:)
 
   def add_constituent(child_druid:)
     child = ItemQueryService.find_combinable_item(child_druid)
+    # Make sure the child is open before making modifications
+    VersionService.open(child) unless VersionService.open?(child)
```
ndushay: isn't this check now superfluous, given your changes to `find_combinable_item` ?

config/routes.rb
```diff
@@ -20,6 +20,9 @@
       get 'catkey', to: 'marcxml#catkey'
     end
 
+    resources :virtual_objects, only: [:create]
+
+    # TODO: Remove :update once Argo, in stage and prod, uses a version of dor-services-client that no longer hits this endpoint
```
ndushay: is the migration / removal of `:update` ticketed?  If not ... make it so??

app/services/constituent_service.rb
mjgiarlo: @ndushay Nope! The changes to `ItemQueryService#find_combinable_item` make sure that every item is either open or openable before proceeding with any operation. This step makes sure every item is *open* before making any modifications. (This prevents users from having to manually open every parent and child object before beginning the virtual object creation.)

config/routes.rb
mjgiarlo: I'm planning to do this immediately after sul-dlss/argo#1463. I wasn't planning to ticket every subtask bc my hope was to push through the PRs rapid-fire style so we can finish this feature, test it, and tidy things before the week is through. I am happy to write 'em up if you think it's worth the time.

app/services/constituent_service.rb
ndushay: ah ha!  Cool!

config/routes.rb
ndushay: your plan makes sense.  :-)

PR #387, opened 2019-09-10
Transition from RestClient to Faraday

jcoyne:
For consistency with other dlss projects


justinlittman:
Recommend giving PSM a heads up before deploying, as this does involve some risk.

jcoyne:
I've notified #dlss-psm

PR #388, opened 2019-09-10
Removed unused catalog/marcxml and catalog/mods endpoints

jcoyne:


jcoyne:
@ndushay Thanks for the feedback.  That is definitely a clearer error message and I've updated to use that wording.

PR #391, opened 2019-09-10
Add druid validation

jcoyne:
## Why was this change made?

Because the druid parameter was poorly specified before.

yes

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -847,7 +847,9 @@
   "components": {
     "schemas": {
       "Druid": {
-        "type": "string"
+        "type": "string",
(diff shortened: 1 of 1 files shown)


jcoyne:
@ndushay I just took this from here: https://github.com/sul-dlss/druid-tools/#druidtools

ndushay:
I guess I'm thinking:

- it's good to validate, even loosely
- would it be better to validate tightly?

So ... is there are reason we wouldn't want to be more restrictive?  I'm not blocking the PR but I think it's worth asking.  It doesn't seem like there would be a down side, and it would catch problems a bit sooner, without making fedora requests ...

jcoyne:
@ndushay this won't validate anything, it's just documentation.

justinlittman:
I'd suggest linking to the more complete description (that is the code) for druid validation.

jcoyne:
Updated the validation

PR #392, opened 2019-09-10
surface Faraday::Timeout errors for Symphony response

### LINKED ISSUE #377: if Symphony request times out, does approp error get surfaced to client?
We want to surface this info to the caller, in the same way we do if the number of bytes received from Symphony doesn't match the HTTP header from Symphony.

- [x] dor-services-app (PR #392)
- [x] dor-services-client (will be surfaced the same as wrong number of bytes or other errors from Symphony)
- [x] Argo
  - [x] registration (will be surfaced the same as wrong number of bytes or other errors from Symphony)
  - [x] refresh metadata action (will be surfaced the same as wrong number of bytes or other errors from Symphony)

ndushay:
## Why was this change made?

A timeout for a Symphony request should get a specific error message to the end user (e.g. in Argo registration or in Argo refresh_metadata).

Fixes #377

- no change to API (we aren't currently documenting error codes for requests)

### DIFF
--- a/app/models/symphony_reader.rb
+++ b/app/models/symphony_reader.rb
@@ -47,12 +47,18 @@ class SymphonyReader
       validate_response(resp)
       return resp
     elsif resp.status == 404
+      # 404 received here is for the catkey, but this app cares about the druid
+      #   for the DOR object, hence raising 404 from here could be misleading
       errmsg = "Record not found in Symphony: #{@catkey}"
     else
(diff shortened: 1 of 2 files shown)


app/models/symphony_reader.rb
```diff
@@ -47,12 +47,18 @@ def symphony_response
       validate_response(resp)
       return resp
     elsif resp.status == 404
+      # 404 received here is for the catkey, but this app cares about the druid
+      #   for the DOR object, hence raising 404 from here could be misleading
```
ndushay: this comment from a follow on discussion to PR #380.

PR #394, opened 2019-09-11
Skip validating symphony response length when chunked.

### LINKED ISSUE #393: cope with "chunked" Symphony response
https://argo-stage.stanford.edu/view/druid:mx445pq9399  has ckey 11958694;  an attempt to refresh metadata gives this error:

image

Using the rails console for dor-services-app, I was able to determine the HTTP headers in the response from Symphony had no Content-Length:

```ruby
{"date"=>"Tue, 10 Sep 2019 22:29:48 GMT", "server"=>"Apache/2.4.6 ()", "strict-transport-security"=>"max-age=31536000; includeSubDomains", "access-control-allow-origin"=>"*", "content-type"=>"application/vnd.sirsidynix.roa.resource.v1+json", "via"=>"1.1 symphony-webservices-prod.stanford.edu", "connection"=>"close", "transfer-encoding"=>"chunked"} 
```

Note especially *"transfer-encoding"=>"chunked"*

So we need a way to cope with multipart responses from Symphony.

Speaking to libsys folks, we think Symphony responses may be chunked when the record is longer than X chars.

justinlittman:
closes #393

### DIFF
--- a/app/models/symphony_reader.rb
+++ b/app/models/symphony_reader.rb
@@ -64,6 +64,9 @@ class SymphonyReader
 
   # expects resp.status to be 200;  does not check response code
   def validate_response(resp)
+    # The length for a chunked response is 0, so checking it isn't meaningful.
+    return resp if resp.headers['Transfer-Encoding'] == 'chunked'
+
     exp_content_length = resp.headers['Content-Length'].to_i
     actual_content_length = resp.body.length
     return resp if actual_content_length == exp_content_length
--- a/spec/models/symphony_reader_spec.rb
+++ b/spec/models/symphony_reader_spec.rb
@@ -61,6 +61,14 @@ RSpec.describe SymphonyReader do
       expect(field.subfields.first.value).to eq 'some data'
     end
 
+    context 'when response is chunked' do
+      let(:headers) { { 'Content-Length': 0, 'Transfer-Encoding': 'chunked' } }
+
+      it 'does not validate content length' do
+        expect(reader.to_marc).to be_a_kind_of MARC::Record
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


app/models/symphony_reader.rb
```diff
@@ -64,6 +64,9 @@ def symphony_response
 
   # expects resp.status to be 200;  does not check response code
   def validate_response(resp)
+    # The length for a chunked response is 0, so checking it isn't meaningful.
```
jcoyne: Do we have any sense that the Transfer-Encoding is ever not "chunked"?

mjgiarlo: @ndushay 	:point_up:

ndushay:
I could have done THIS.  The problem is that we need to get the full response -- chunked or not.  When chunked, the response here is only the FIRST chunk.

Most of the times the response is NOT chunked.

Sorry for my poor explanation.

app/models/symphony_reader.rb
ndushay: see comment below -- this is the wrong approach.  We need to get ALL of the chunked responses, and re-created the json as if it wasn't chunked.   This change is a step backwards -- we still fail on chunked records, but we get less information.

jcoyne: @ndushay do we still need to validate responses for length?  I don't know of a case where Faraday wouldn't raise an error if the request body wasn't of the correct length.

justinlittman: Haven't looked at the code, but I'd be surprised if Faraday did raise an error. Content-Length is notoriously wrong (in the web at large).

PR #398, opened 2019-09-11
[HOLD] new api call for checking if an item is currently embargoed

### LINKED ISSUE #400: Implement a new endpoint for checking the embargo status of an item
An upcoming integration point with library systems code and Yewno for ETD metadata enhancement will require libsys being able to check if a particular item is embargoed.

peetucket:
HOLDING, #426 may end up serving this need instead.

fixes #400 

## Why was this change made?

So that we can tell if an object is currently embargoed via an API call.  Likely will be needed for a new workflow being developed to enhance ETD metadata via Yewno.


app/controllers/objects_controller.rb
```diff
@@ -56,6 +56,14 @@ def update_marc_record
     head :created
   end
```
jcoyne: Can you please move this to the embargoes controller?

spec/controllers/objects_controller_spec.rb
```diff
@@ -52,6 +55,20 @@
     end
   end
 
+  describe '/is_embargoed' do
```
jcoyne: Please use a request spec

app/controllers/objects_controller.rb
peetucket: I think this would be inconsistent with the rest of the tests in this project - is that ok?

spec/controllers/objects_controller_spec.rb
peetucket: I think this would be inconsistent with the rest of the tests in this project - is that ok?

app/controllers/objects_controller.rb
jcoyne: Use this one: https://github.com/sul-dlss/dor-services-app/blob/master/app/controllers/embargoes_controller.rb

spec/controllers/objects_controller_spec.rb
jcoyne: There are lots of request specs: https://github.com/sul-dlss/dor-services-app/tree/master/spec/requests

jcoyne: In fact, if you were on the head of master, this controller spec file no longer exists.

app/controllers/embargoes_controller.rb
```diff
@@ -2,7 +2,12 @@
 
 # Manages the embargo for a repository object
 class EmbargoesController < ApplicationController
-  before_action :load_item, only: :update
+  before_action :load_item, only: [:show, :update]
+
+  def show
+    result = { embargoed: !!@item&.embargoed?, release_date: @item&.embargoMetadata&.release_date }
```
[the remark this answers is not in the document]
mjgiarlo: @jcoyne I'm not sure I follow. The `&.` after `@item` is to protect against `@item` not responding to `embargoMetadata`, not to protect against `@item` being `nil` or not found or something. Would the former case trigger a 404?

mjgiarlo:
Note to reviewers: please ignore the CircleCI error in the checks above; I was in the process of integrating dor-services-app with Circle while @peetucket was submitting this PR.

peetucket:
> I believe openapi.json needs updating too.

updated

openapi.json
```diff
@@ -645,6 +645,42 @@
             }
           }
         ]
+      },
+      "get": {
```
mjgiarlo: @jcoyne 	?

jcoyne: This should have a `tags` so that it's put in the right section of the generated documentation.

peetucket:
> Now that we have Objects#show, can we add the `access.embargoReleaseDate` property there? See [sul-dlss-labs/taco:maps/DRO.json@`839755e`#L88](https://github.com/sul-dlss-labs/taco/blob/839755ef9485ae1c3783338ed3f39cf7b006cea1/maps/DRO.json#L88)
> 
> This avoids having to make a whole new endpoint.

Seems reasonable.  Not sure how to get that property returned in the show method though.  We could also wait for the first consumer (libsys as part of the Yewno integration) to be sure that works for them.  Depending on how much data is returned in the show response and how often they hit it, it could be unnecessarily expensive.

openapi.json
peetucket: done

peetucket:
Note: rebased and force pushed to keep up to date

app/controllers/embargoes_controller.rb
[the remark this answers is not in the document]
jcoyne: No, the safe navigation error only defends against the receiver being null, it does not prevent a NoMethodError:

```
7&.foo
Traceback (most recent call last):
        5: from /Users/jcoyne85/.rbenv/versions/2.6.4/bin/irb:23:in `<main>'
        4: from /Users/jcoyne85/.rbenv/versions/2.6.4/bin/irb:23:in `load'
        3: from /Users/jcoyne85/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):2
        1: from (irb):2:in `rescue in irb_binding'
NoMethodError (undefined method `foo' for 7:Integer)
```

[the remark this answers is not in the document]
mjgiarlo: I'd mistakenly thought `&.` was closer to `#try` than it actually is. Thanks for the heads-up.

jcoyne:
@peetucket could we do this instead? https://github.com/sul-dlss/dor-services-app/pull/426

peetucket:
> @peetucket could we do this instead? #426

That looks like a reasonable solution.  Libsys would need to check for the existence of the date to infer if an embargo exists, but that should be easy.  As they haven't written any code yet, this is a good time to make that decision.  We could leave this work in a branch without deleting for now just in case.

jcoyne:
@peetucket can this be closed?

peetucket:
I think so -- I was hoping libsys would have to chance to weigh in to see if the way you can get dates now works for them, but let's just close and re-visit later if needed.

PR #399, opened 2019-09-12
Accept the Authorization header

### LINKED ISSUE #320: Prefer Authorization header over X-Auth
When token-based authentication was added to dor-services-app, we still supported HTTP Basic (credential-based) authentication, and thus defined an unconventional header (`X-Auth`) to flag token-based connections. Now that dor-services-app and all of its clients have switched to token-based authentication, and dor-services-client no longer supports HTTP Basic authentication, we should consider whether to keep using the `X-Auth` header (advantage: less work; disadvantage: unconventional) or to switch to the `Authorization` header (advantage: conventional/least surprise; disadvantage: have to touch a lot of codebases and coordinate deployments).

jcoyne:
Log a honeybadger notification when we get the old X-Auth header
Fixes #320

## Why was this change made?

We want to be able to accept the standard Authorization header rather than our custom X-Auth header.

Yes!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,7 +15,8 @@ class ApplicationController < ActionController::API
 
   # Since Basic auth was already using the Authorization header, we used something
   # non-standard:
-  TOKEN_HEADER = 'X-Auth'
+  OLD_TOKEN_HEADER = 'X-Auth'
+  TOKEN_HEADER = 'Authorization'
 
   private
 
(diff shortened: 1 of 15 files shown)


app/controllers/application_controller.rb
```diff
@@ -25,6 +26,9 @@ def check_auth_token
     return render json: { error: 'Not Authorized' }, status: :unauthorized unless token
 
     Honeybadger.context(invoked_by: token[:sub])
+    return unless request.headers[OLD_TOKEN_HEADER]
+
+    Honeybadger.notify("Deprecated authentication header '#{OLD_TOKEN_HEADER}' was provided, but '#{TOKEN_HEADER}' is expected")
```
mjgiarlo: s/authentication/authorization/ ?

jcoyne: Thank you! I fixed it.

mjgiarlo:
@jcoyne 🐑 it

mjgiarlo:
Should we look to do lots o' shared configs PRs tomorrow and Friday so Monday morning updates can reduce the number of hb alerts?

Issue #336, opened 2019-08-15
confirm: bulk action for virtual-merge handles > 500 child druids [blocked by #1463]


ndushay:
PO will be checking for this after #1463 (UI hooked up in argo) is deployed to stage for vetting.

- can it handle > 500 druids?
- is there an easily found practical limit of how many druids it can handle?  (note:  order of children matters, so can't parallelized in virtual-merge)

Justin Coyne says:

>I'm concerned that the virtual-merge we put in dor-services-app is not going to meet the 500+ items because it takes place in a request context and it could time out.
>I think we need a ticket to specifically test it on stage.  If it fails, we can look at putting it in a background job, but we'd need to have a callback or some endpoint where one could see "is it done yet?"

It is believed the "500" limit pertains to javascript code for registration specifically ...

ndushay:
@jcoyne would it be easiest to hook this up to UI per sul-dlss/argo/issues/1463 and ask Andrew to try a big batch for testing?

mjgiarlo:
We should consider re-architecting DSA such that it uses background jobs for its long-running processes. I will look into this work next week, alongside the work needed in Argo to accommodate this.

mjgiarlo:
@andrewjbtw can confirm the current implementation allows uploading virtual objects of more than 500 child druids. Largest test was I believe 1,500 children. Closing.

PR #408, opened 2019-09-13
Add Sidekiq to process background jobs

mjgiarlo:
Connects to sul-dlss/dor-services-app#336

~Blocked by sul-dlss/puppet#4463~


mjgiarlo:
I'll fix up the :policeman: on Monday

PR #412, opened 2019-09-17
Fix security syntax in openapi.json

jcoyne:
## Why was this change made?
The openapi specification was incorrectly specifying the authentication

Yes


mjgiarlo:
@jcoyne do you think we should add a CI step that validates `openapi.json`?

PR #413, opened 2019-09-17
upgrade to dor-services 8

jcoyne:
## Why was this change made?

To keep dor-services-app consistent with other Dor projects.

n/a


mjgiarlo:
:policeman:

PR #415, opened 2019-09-17
Move client logic to SdrClient

jcoyne:
## Why was this change made?

Keep logic out of the controller. This makes the client code reusable for shelving (TBD)

n/a

### DIFF
--- a/spec/controllers/sdr_controller_spec.rb
+++ b/spec/controllers/sdr_controller_spec.rb
@@ -63,10 +63,10 @@ RSpec.describe SdrController do
 
     context 'URI encoding' do
       let(:filename_with_spaces) { 'filename with spaces' }
-      let(:uri_encoded_filename) { URI.encode(filename_with_spaces) }
+      let(:cgi_escaped_filename) { CGI.escape(filename_with_spaces) }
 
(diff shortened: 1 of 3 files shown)


spec/controllers/sdr_controller_spec.rb
```diff
@@ -63,10 +63,10 @@
 
     context 'URI encoding' do
       let(:filename_with_spaces) { 'filename with spaces' }
-      let(:uri_encoded_filename) { URI.encode(filename_with_spaces) }
+      let(:cgi_escaped_filename) { CGI.escape(filename_with_spaces) }
```
jcoyne: `URI.encode` is deprecated

mjgiarlo: @jcoyne These produce different values depending on the input, e.g.:

```ruby
2.5.3 :001 > URI.encode('/foo/bar with spaces.txt')
 => "/foo/bar%20with%20spaces.txt" 
2.5.3 :002 > CGI.escape('/foo/bar with spaces.txt')
 => "%2Ffoo%2Fbar+with+spaces.txt" 
```

I guess it doesn't matter, though, given how it's being used in this spec?

PR #417, opened 2019-09-17
Add API endpoint for shelving

jcoyne:
## Why was this change made?

So that common-accessioning doesn't need to be aware of how to shelve something. It can just ensure the process is triggered. 

Yes


app/controllers/shelves_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+        errors: [
+          { title: 'Invalid item type', detail: "A Dor::Item is required but you provided '#{@item.class}'" }
+        ]
+      }, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def druid_tools
```
justinlittman: Is this used?

app/services/digital_stacks_service.rb
```diff
@@ -0,0 +1,113 @@
...
+  # Delete files from stacks that have change type 'deleted', 'copydeleted', or 'modified'
+  # @param [Pathname] stacks_object_pathname the stacks location of the digital object
+  # @param [Moab::FileGroupDifference] content_diff the content file version differences report
+  def self.remove_from_stacks(stacks_object_pathname, content_diff)
+    %i[deleted copydeleted modified].each do |change_type|
+      subset = content_diff.subset(change_type) # {Moab::FileGroupDifferenceSubset}
+      subset.files.each do |moab_file| # {Moab::FileInstanceDifference}
+        moab_signature = moab_file.signatures.first # {Moab::FileSignature}
+        file_pathname = stacks_object_pathname.join(moab_file.basis_path)
+        delete_file(file_pathname, moab_signature)
```
justinlittman: Ignoring when delete_file doesn't succeed (returns false).

app/services/digital_stacks_service.rb
```diff
@@ -0,0 +1,113 @@
...
+  # @param [Moab::FileGroupDifference] content_diff the content file version differences report
+  def self.rename_in_stacks(stacks_object_pathname, content_diff)
+    subset = content_diff.subset(:renamed) # {Moab::FileGroupDifferenceSubset
+
+    # 1st Pass - rename files from original name to checksum-based name
+    subset.files.each do |moab_file| # {Moab::FileInstanceDifference}
+      moab_signature = moab_file.signatures.first # {Moab::FileSignature}
+      original_pathname = stacks_object_pathname.join(moab_file.basis_path)
+      temp_pathname = stacks_object_pathname.join(moab_signature.checksums.values.last)
+      rename_file(original_pathname, temp_pathname, moab_signature)
```
justinlittman: Ignores when rename fails.

app/services/digital_stacks_service.rb
```diff
@@ -0,0 +1,113 @@
...
+    return false if workspace_content_pathname.nil?
+
+    %i[added copyadded modified].each do |change_type|
+      subset = content_diff.subset(change_type) # {Moab::FileGroupDifferenceSubset
+      subset.files.each do |moab_file| # {Moab::FileInstanceDifference}
+        moab_signature = moab_file.signatures.last # {Moab::FileSignature}
+        filename = change_type == :modified ? moab_file.basis_path : moab_file.other_path
+        workspace_pathname = workspace_content_pathname.join(filename)
+        stacks_pathname = stacks_object_pathname.join(filename)
+        copy_file(workspace_pathname, stacks_pathname, moab_signature)
```
justinlittman: Ignores when copy file fails.

app/services/shelving_service.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+# Push file changes for shelve-able files into stacks
+class ShelvingService
+  def self.shelve(work)
```
justinlittman: Why calling "work" instead of "item"?

app/services/digital_stacks_service.rb
jcoyne: I'm just moving these from common-accessioning, so I'd like to take a bye on this.

app/services/digital_stacks_service.rb
jcoyne: This is how it is in common-accessioning.

app/services/digital_stacks_service.rb
jcoyne: This is how it is in common-accessioning.

app/services/shelving_service.rb
jcoyne: This is how it is in common-accessioning.

app/controllers/shelves_controller.rb
jcoyne: nope.

app/services/digital_stacks_service.rb
jcoyne: https://github.com/sul-dlss/common-accessioning/blob/master/lib/digital_stacks_service.rb#L13

app/services/digital_stacks_service.rb
jcoyne: https://github.com/sul-dlss/common-accessioning/blob/master/lib/digital_stacks_service.rb#L45

app/services/digital_stacks_service.rb
jcoyne: https://github.com/sul-dlss/common-accessioning/blob/master/lib/digital_stacks_service.rb#L88

app/services/shelving_service.rb
jcoyne: https://github.com/sul-dlss/common-accessioning/blob/master/lib/shelving_service.rb#L5

PR #418, opened 2019-09-17
Remove current children from virtual object before initializing content metadata

### LINKED ISSUE #409: Before adding children to a parent virtual object, first remove any children attached to the parent
A new requirement from @andrewjbtw

mjgiarlo:
Fixes #409 

This prevents a situation where lots of manual clean-up is required to remove old children from a parent.

Also includes:
* Ignore postgres data directory created by `docker-compose` service
* Correct typo in `RefreshMetadataAction` spec description

## Why was this change made?

To satisfy the new virtual object creation requirement surfaced in #409.

N/A


spec/services/constituent_service_spec.rb
```diff
@@ -64,6 +64,9 @@
       allow(ItemQueryService).to receive(:find_combinable_item).with('druid:parent1').and_return(parent)
       allow(ItemQueryService).to receive(:find_combinable_item).with('druid:child1').and_return(child1)
       allow(ItemQueryService).to receive(:find_combinable_item).with('druid:child2').and_return(child2)
+
+      # Stub this here because it is tested in its own spec
+      allow_any_instance_of(ResetContentMetadataService).to receive(:disown_current_children!)
```
jcoyne: Why can't you stub on the instance itself?  Have you considered stubbing out `ResetContentMetadataService#reset` rather than `disown_current_children!`, which the `ConstituentServices` never references itself?

mjgiarlo: @jcoyne I've changed the test to stop caring about effects from `ResetContentMetadataService`. Take another look?

PR #422, opened 2019-09-18
Handle virtual object creation in a background job

mjgiarlo:
Connects to #336
Deployment blocked by sul-dlss/puppet#4470

* Add ActiveRecord and Postgres as dependencies. This makes it possible to reliably stash information about background jobs in a persistent store, which will be used to run virtual object creation in the background.
* Include `factory_bot_rails` as a dependency which will make it easier to spin
up fixture-like factories when testing with database entities later.
* Create a model and a table for background job results. The model holds any output from the job, an HTTP status code, and a processing status.
* Add factory_bot_rails to the test suite. This allows us to more efficiently test database-backed entities, specifically `BackgroundJobResults`.
* Add new controller and view for background job results.
* Use a background job to create virtual objects. Use the new job from the `CreateVirtualObjectsController`. In the job, record progress and results in the application database

## Why was this change made?

To be able to process large batches of virtual objects outside of the request/response cycle.

Yes.

### DIFF
--- a/app/controllers/virtual_objects_controller.rb
+++ b/app/controllers/virtual_objects_controller.rb
@@ -6,19 +6,10 @@ class VirtualObjectsController < ApplicationController
 
   # Create one or more virtual objects represented by JSON (see `#schema` below):
   def create
-    errors = []
-
-    create_params[:virtual_objects].each do |virtual_object|
-      parent_id, child_ids = virtual_object.values_at(:parent_id, :child_ids)
-      # Update the constituent relationship between the parent and child druids
-      errors << ConstituentService.new(parent_druid: parent_id).add(child_druids: child_ids)
-    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception => e
-      errors << { parent_id => [e.message] }
-    end
-
-    return render_error(errors: errors, status: :unprocessable_entity) if errors.any?
-
-    head :no_content
+    result = BackgroundJobResult.create
+    CreateVirtualObjectsJob.perform_later(virtual_objects: create_params[:virtual_objects],
+                                          background_job_result: result)
+    head :created, location: result
   end
 
   private
@@ -31,11 +22,7 @@ class VirtualObjectsController < ApplicationController
   # This gives us finer-grained validation.
   def validate_params!
     errors = schema.call(create_params).errors(full: true)
-    return render_error(errors: errors, status: :bad_request) if errors.any?
-  end
-
-  def render_error(errors:, status:)
-    render json: { errors: errors }, status: status
+    return render json: { errors: errors }, status: :bad_request if errors.any?
   end
 
   # {
--- a/app/jobs/create_virtual_objects_job.rb
+++ b/app/jobs/create_virtual_objects_job.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# Create virtual objects in the background
+class CreateVirtualObjectsJob < ApplicationJob
+  queue_as :default
+
+  # @param [Array] virtual_objects an array of hashes representing a batch of virtual objects
+  # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
+  def perform(virtual_objects:, background_job_result:)
+    background_job_result.processing!
+
+    errors = []
+
+    virtual_objects.each do |virtual_object|
+      parent_id, child_ids = virtual_object.values_at(:parent_id, :child_ids)
+      # Update the constituent relationship between the parent and child druids
+      errors << ConstituentService.new(parent_druid: parent_id).add(child_druids: child_ids)
+    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception => e
+      errors << { parent_id => [e.message] }
+    end
+
+    if errors.any?
+      background_job_result.output = { errors: errors }.to_json
+      background_job_result.code = 422
+    else
+      background_job_result.code = 200
+    end
+
+    background_job_result.complete!
+  end
+end
--- a/app/models/background_job_result.rb
+++ b/app/models/background_job_result.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Database-backed model to hold results of long-running jobs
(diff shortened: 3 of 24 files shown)


app/controllers/virtual_objects_controller.rb
```diff
@@ -2,27 +2,21 @@
 
 # Create virtual objects
 class VirtualObjectsController < ApplicationController
-  before_action :validate_params!, only: :create
+  before_action :validate_params!, :create_background_job_result, only: :create
```
jcoyne: Is there a reason we need to put this in a callback?  I think it's easier to read if it's just in the method.

app/controllers/virtual_objects_controller.rb
```diff
@@ -2,27 +2,21 @@
...
-      errors << ConstituentService.new(parent_druid: parent_id).add(child_druids: child_ids)
-    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception => e
-      errors << { parent_id => [e.message] }
-    end
-
-    return render_error(errors: errors, status: :unprocessable_entity) if errors.any?
-
-    head :no_content
+    CreateVirtualObjectsJob.perform_later(virtual_objects: create_params[:virtual_objects],
+                                          background_job_result_id: @result.id)
```
jcoyne: I think you can pass `background_job_result: @result` and it will serialize/deserialize for you using GlobalId.

app/jobs/create_virtual_objects_job.rb
```diff
@@ -0,0 +1,34 @@
...
+# Create virtual objects in the background
+class CreateVirtualObjectsJob < ApplicationJob
+  queue_as :default
+
+  attr_reader :background_job_result
+
+  # @param [Array] virtual_objects an array of hashes representing a batch of virtual objects
+  # @param [Integer] background_job_result_id identifier of a background job result to store status info
+  def perform(virtual_objects:, background_job_result_id:)
+    background_job_result = BackgroundJobResult.find(background_job_result_id)
```
jcoyne: Avoid this by using GlobalId.

app/models/background_job_result.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# Database-backed model to hold results of long-running jobs
+class BackgroundJobResult < ApplicationRecord
+  enum status: { pending: 0, processing: 1, complete: 2 }
```
jcoyne: Enums in active record are sort of a pain in the ass.  as it's not understandable to anyone using SQL.  How about using a postgresql ENUM (or even a string column) instead?

spec/support/auth_helper.rb
```diff
@@ -4,4 +4,14 @@ module AuthHelper
   def login
     allow(controller).to receive(:check_auth_token)
   end
+
+  def jwt
```
jcoyne: This really deserves to be its own PR.

mjgiarlo: @jcoyne 	OK: https://github.com/sul-dlss/dor-services-app/pull/424

Will remove this commit when I push changes next.

mjgiarlo:
@jcoyne :speech_balloon: 
> The PR message says you updated openapi.json, but I don't see that in the changed files.

Included now.

app/controllers/virtual_objects_controller.rb
mjgiarlo: I went back and forth on that. Which I'm doing again now. :smile:

app/controllers/virtual_objects_controller.rb
mjgiarlo: @jcoyne 	OK

app/jobs/create_virtual_objects_job.rb
mjgiarlo: K

app/models/background_job_result.rb
mjgiarlo: Changed. Thanks for the tip!

db/migrate/20190917215521_create_background_job_results.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class CreateBackgroundJobResults < ActiveRecord::Migration[5.2]
+  def up
+    execute <<-SQL
+      CREATE TYPE background_job_result_status AS ENUM (
```
jcoyne: Great! We should port this to workflow-service.

openapi.json
```diff
@@ -249,6 +253,45 @@
...
+        ],
+        "summary": "View results of a background job",
+        "description": "Used to allow the application run long-running processes out of the request/response cycle",
+        "operationId": "background_job_results#show",
+        "responses": {
+          "200": {
+            "description": "OK"
+          },
+          "202": {
+            "description": "Accepted"
```
jcoyne: Can you say that "this is returned when the job has been queued, but has not completed processing.

openapi.json
```diff
@@ -249,6 +253,45 @@
...
+      "get": {
+        "tags": [
+          "jobs"
+        ],
+        "summary": "View results of a background job",
+        "description": "Used to allow the application run long-running processes out of the request/response cycle",
+        "operationId": "background_job_results#show",
+        "responses": {
+          "200": {
+            "description": "OK"
```
jcoyne: "This is returned when the job has completed processing successfully"

openapi.json
```diff
@@ -258,12 +301,12 @@
         "description": "Combines a parent object with child objects to create an object like an Atlas (composed of several images)",
         "operationId": "virtual_objects#create",
         "responses": {
-          "204": {
-            "description": "objects created"
-          },
-          "422": {
-            "description": "unable to process request"
+          "201": {
+            "description": "virtual object creation job created"
```
jcoyne: Can you describe the `Location` header here? https://swagger.io/docs/specification/describing-responses/#headers

openapi.json
mjgiarlo: done, pushed

openapi.json
mjgiarlo: done, pushed

openapi.json
mjgiarlo: yep, coming soon

mjgiarlo: Pushed.

PR #425, opened 2019-09-19
Validate OpenAPI specification

mjgiarlo:
Includes:
* Add validation step to CI build
* Add validator badge to README
* Prefer HTTP 204 over 201 when not creating a new resource

## Why was this change made?

To make our spec validation more transparent and more robust.

Yes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
     - image: circleci/postgres:10
       environment:
         POSTGRES_USER: postgres
-        POSTGRES_DB: dor_services_test
+        POSTGRES_DB: dor_services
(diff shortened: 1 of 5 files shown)


mjgiarlo:
@justinlittman :speech_balloon: 
> So there seem to be some other changes mixed in here.

The OpenAPI validator (openapi-enforcer-cli) was complaining about using HTTP 201 without a location header, so I made the code changes to use HTTP 204 instead. 

It's a little problematic that the rules enforced by the command-line validator appear to be somewhat different from the rules enforced by the validator that generates the README badge. I briefly looked into alternatives, but didn't find anything more compelling.

PR #426, opened 2019-09-19
Add embargoReleaseDate to the show route

jcoyne:
## Why was this change made?

So we have a way of getting the embargo date for the object

yes

### DIFF
--- a/spec/requests/show_object_spec.rb
+++ b/spec/requests/show_object_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe 'Get the object' do
     allow(Dor).to receive(:find).and_return(object)
   end
 
-  context 'when the object exists' do
+  context 'when the object exists with minimal metadata' do
     before do
       object.descMetadata.title_info.main_title = 'Hello'
       object.label = 'foo'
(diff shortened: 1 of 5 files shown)


peetucket:
Looks like a test failure (need to check to be sure that embargoMetadata is even present as it won't be for collections)...

jcoyne:
@peetucket okay, I've added a check for that.

spec/requests/show_object_spec.rb
```diff
@@ -22,4 +22,20 @@
...
+    before do
+      object.descMetadata.title_info.main_title = 'Hello'
+      object.label = 'foo'
+      object.embargoMetadata.release_date = DateTime.parse '2019-09-26T07:00:00Z'
+    end
+
+    it 'returns the object' do
+      get '/v1/objects/druid:mk420bs7601',
+          headers: { 'Authorization' => "Bearer #{jwt}" }
+      expect(response).to be_successful
```
mjgiarlo: Should this be `.to have_http_status(:ok)` instead?

jcoyne: You're right that that would be more specific

PR #430, opened 2019-09-19
Document background job result bodies

mjgiarlo:
I started down this road when discovering in dor-services-client that dor-services-app was not reliably returning JSON as output.

Part of this work includes ensuring the output field of background job results is serialized as JSON (and deserialized as a hash with indifferent access because we are used to dealing with hashes with indifferent access in the Rails context).

## Why was this change made?

To add missing detail to the API specification.

Yes

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -263,17 +263,31 @@
         "operationId": "background_job_results#show",
         "responses": {
           "200": {
-            "description": "The background job has completed successfully"
+            "description": "The background job has completed successfully",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/BackgroundJobResultResponse"
+                }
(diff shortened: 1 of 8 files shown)


openapi.json
```diff
@@ -892,6 +906,25 @@
...
       },
+      "BackgroundJobResultResponse": {
+        "type": "object",
+        "properties": {
+          "output": {
+            "type": "object",
+            "description": "output from the job",
+            "properties": {
+              "errors": {
+                "$ref": "#/components/schemas/ErrorResponse"
```
jcoyne: This doesn't seem right unless you have `errors` inside of `errors`

mjgiarlo: @jcoyne good eye. fixed, I think. take another look?

PR #432, opened 2019-09-19
Align docker configuration with workflow-service

mjgiarlo:
## Why was this change made?

Prior to this, the application in containerland was not connecting to its database, and the docker-compose image lacked psql

n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
         PGUSER: circleci-demo-ruby
         RAILS_ENV: test
         NOKOGIRI_USE_SYSTEM_LIBRARIES: true
-    - image: circleci/postgres:10
+    - image: circleci/postgres:9.6-alpine
       environment:
         POSTGRES_USER: postgres
         POSTGRES_DB: dor_services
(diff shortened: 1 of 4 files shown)


mjgiarlo:
I could use some extra :eyes: on this. I looked at this with @aaron-collier and it stumped both of us. Here's the deal. I can't explain why [this build](https://circleci.com/gh/sul-dlss/dor-services-app/136) fails in the way it does and the `workflow-server-rails` build does not. As far as I can tell, the two codebases have near identical `.circleci/config.yml` files (due to changes in this branch). I must be missing something.

The piece I don't get is that I seem to have to explicitly install `postgresql-client` in the `dor-services-app` CircleCI config, but that doesn't need to happen for the workflow service. Apparently?

Maybe @jcoyne and/or @justinlittman have ideas?

justinlittman:
Fixed it (mostly). Comparing against workflow-server-rails was a bit of a red herring, as wfs isn't actually green if you re-run.

There were 2 problems:
* The postgres client had to be installed.
* A newer version of postgres had to be used.

It is now failing due to openapi problems, but the docker db is working.

I'll also fix workflow-server-rails.

justinlittman:
See https://github.com/sul-dlss/workflow-server-rails/pull/258

mjgiarlo:
@justinlittman Awesome, thank you! I should not have assumed the other build was hunky-dorey.

PR #434, opened 2019-09-20
Object#show now returns releaseTags and current_version

jcoyne:
## Why was this change made?

So that Argo doesn't have to be aware of the Fedora 3 data model to display the tags.

Yes


jcoyne:
I believe the tests are working here, but our validation script is failing on warnings.

PR #441, opened 2019-09-23
deprecate sdr#current_version in favor of preservation-client

ndushay:
## Why was this change made?

To use the new preservation-client gem to get current_version from a preservation-catalog ReST endpoint, rather than the Sinatra sdr-services-app.  Also deprecates methods here designed to proxy getting current version from sdr-services-app.

Yes, marked the sdr#current_version endpoint deprecated.

## Reviewers, please

- does this notify Honeybadger appropriately?
- does this have appropriate deprecation warnings?

### DIFF
--- a/app/controllers/sdr_controller.rb
+++ b/app/controllers/sdr_controller.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class SdrController < ApplicationController
+  extend Deprecation
+  self.deprecation_horizon = 'dor-services-app version 4.0'
+
   def cm_inv_diff
     unless %w(all shelve preserve publish).include?(params[:subset])
       render status: :bad_request, plain: "Invalid subset value: #{params[:subset]}"
@@ -25,8 +28,10 @@ class SdrController < ApplicationController
   end
 
   def current_version
+    Honeybadger.notify('dor-services-app deprecated API endpoint `sdr#current_version` called - use preservation-client current_version instead')
     proxy_faraday_response(sdr_client.current_version)
   end
+  deprecation_deprecate current_version: 'use preservation-client current_version in caller instead'
 
   def file_content
     sdr_response = sdr_client.file_content(version: params[:version], filename: params[:filename])
--- a/app/services/sdr_client.rb
+++ b/app/services/sdr_client.rb
@@ -2,6 +2,9 @@
 
 # A client for talking to sdr-services-app
 class SdrClient
+  extend Deprecation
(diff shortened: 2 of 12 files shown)


app/services/sdr_client.rb
```diff
@@ -30,6 +33,7 @@ def metadata(ds_name:)
   end
 
   def current_version(parsed: false)
+    Honeybadger.notify('dor-services-app deprecated method `SDRClient.current_version` called - use preservation-client current_version instead')
```
ndushay: I am confident this is never called except from sdr_controller, but it seemed safest to leave the code in and remove it later.

app/controllers/sdr_controller.rb
```diff
@@ -25,8 +28,10 @@ def ds_metadata
   end
 
   def current_version
+    Honeybadger.notify('dor-services-app deprecated API endpoint `sdr#current_version` called - use preservation-client current_version instead')
```
ndushay: dor-services-client has PR sul-dlss/dor-services-client/pull/101 to deprecate it's method, which will hopefully be the only caller.

app/services/version_service.rb
```diff
@@ -103,7 +102,10 @@ def try_to_get_current_version(assume_accessioned = false)
     # The submitted milestone is part of the accessionWF.
     raise Dor::Exception, 'Object currently being accessioned' if accessioning?
 
-    SdrClient.current_version work.pid
+    Preservation::Client.objects.current_version(work.pid)
+  rescue Preservation::Client::NotFoundError
```
ndushay: This exception was formerly trapped for in SdrClient;  I think it is a nice side effect to have it here -- it's more obvious what is happening.  I also improved the error message, IMO.

config/settings.yml
```diff
@@ -34,6 +34,8 @@ solr:
 
 workflow_url: 'https://workflow.example.com/workflow'
 sdr_url: 'http://user:password@sdr-services.example.com/sdr'
+preservation_catalog:
+  url: 'https://example.org/prescat'
```
ndushay: note that shared_configs already have this for stage and prod.

openapi.json
```diff
@@ -101,8 +101,9 @@
         "tags": [
           "preservation"
         ],
-        "summary": "Get the current version of the object from SDR",
-        "description": "",
+        "summary": "DEPRECATED. Get the current version of the object from SDR",
+        "description": "DEPRECATED.  Use preservation-client gem instead.",
+        "deprecated": true,
```
ndushay: I looked this up and it's the correct thing to do.

PR #443, opened 2019-09-23
Move reset workspace from common-accessioning

jcoyne:
## Why was this change made?
This puts this function behind an API so that the process can be more effectively managed

Yes


app/controllers/workspaces_controller.rb
```diff
@@ -23,6 +23,13 @@ def destroy
     head :no_content
   end
 
+  # Once an object has been transfered to preservation clean up the workspace.
```
mjgiarlo: Typo for `transferred`

PR #445, opened 2019-09-24
Revert "Document background job result bodies"

mjgiarlo:
Reverts sul-dlss/dor-services-app#430


jcoyne:
@mjgiarlo can you say why you are making this change?

PR #450, opened 2019-09-27
Create virtual objects using database-backed approach

### LINKED ISSUE #447: Unrevert database-backed virtual object creation when infra is ready
~Blocked by https://github.com/sul-dlss/operations-tasks/issues/2041~

- [x] https://github.com/sul-dlss/dor-services-app/pull/444
- [x] https://github.com/sul-dlss/dor-services-app/pull/445
- [x] https://github.com/sul-dlss/dor-services-app/pull/446

mjgiarlo:
Fixes #447 

Includes:
* Align docker configuration with workflow-service
* Document background job result bodies: I started down this road when discovering in dor-services-client that dor-services-app was not reliably returning JSON as output. Part of this work includes ensuring the output field of background job results is serialized as JSON (and deserialized as a hash with indifferent access because we are used to dealing with hashes with indifferent access in the Rails context).
* Add ActiveRecord and Postgres as dependencies. This makes it possible to reliably stash information about background jobs in a persistent store, which will be used to run virtual object creation in the background.
* Include `factory_bot_rails` as a dependency which will make it easier to spin
up fixture-like factories when testing with database entities later.
* Create a model and a table for background job results. The model holds any output from the job, an HTTP status code, and a processing status.
* Add factory_bot_rails to the test suite. This allows us to more efficiently test database-backed entities, specifically `BackgroundJobResults`.
* Add new controller and view for background job results.
* Use a background job to create virtual objects. Use the new job from the `CreateVirtualObjectsController`. In the job, record progress and results in the application database

## Why was this change made?

1. Prior to this, the application in containerland was not connecting to its database, and the docker-compose image lacked psql.
1. To add missing detail to the API specification.
1. To be able to process large batches of virtual objects outside of the request/response cycle.

Yes.

### DIFF
--- a/app/jobs/create_virtual_objects_job.rb
+++ b/app/jobs/create_virtual_objects_job.rb
@@ -19,12 +19,8 @@ class CreateVirtualObjectsJob < ApplicationJob
       errors << { parent_id => [e.message] }
     end
 
-    if errors.any?
-      background_job_result.output = { errors: errors }
-      background_job_result.code = 422
-    else
-      background_job_result.code = 200
-    end
+    background_job_result.output = { errors: errors } if errors.any?
+    background_job_result.code = 200
 
     background_job_result.complete!
   end
--- a/app/jobs/create_virtual_objects_job.rb
+++ b/app/jobs/create_virtual_objects_job.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# Create virtual objects in the background
+class CreateVirtualObjectsJob < ApplicationJob
+  queue_as :default
+
+  # @param [Array] virtual_objects an array of hashes representing a batch of virtual objects
+  # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
+  def perform(virtual_objects:, background_job_result:)
+    background_job_result.processing!
+
+    errors = []
+
+    virtual_objects.each do |virtual_object|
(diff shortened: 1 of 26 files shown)


jcoyne:
@mjgiarlo sorry, needs a rebase.

mjgiarlo:
@jcoyne There you go.

app/jobs/create_virtual_objects_job.rb
```diff
@@ -0,0 +1,31 @@
...
+      parent_id, child_ids = virtual_object.values_at(:parent_id, :child_ids)
+      # Update the constituent relationship between the parent and child druids
+      errors << ConstituentService.new(parent_druid: parent_id).add(child_druids: child_ids)
+    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception => e
+      errors << { parent_id => [e.message] }
+    end
+
+    if errors.any?
+      background_job_result.output = { errors: errors }
+      background_job_result.code = 422
```
jcoyne: I don't believe 422 is the right code here.  The request for the status itself was "ok", I think you need to just return JSON that says that the status itself is: "there was a problem...".  422 would indicate, you asked improperly for the status.

mjgiarlo: Changed. Thanks!

PR #454, opened 2019-10-04
Use default queue for sidekiq

mjgiarlo:
What was there was copypasta that I missed. This will allow jobs to process.


jcoyne:
Needs a rubocop fixup, then feel free to merge.

PR #456, opened 2019-10-07
Pin sidekiq to ~> 5.0

jcoyne:
## Why was this change made?

Sidekiq 6 doesn't support daemonization, so it doesn't fit into our current deployment patterns

N/A

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'okcomputer'
 gem 'pg'
 gem 'progressbar' # for the cleaner rake task
 gem 'ruby-cache', '~> 0.3.0'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -34,7 +34,7 @@ gem 'okcomputer'
 gem 'pg'
 gem 'progressbar' # for the cleaner rake task
 gem 'ruby-cache', '~> 0.3.0'
-gem 'sidekiq'
+gem 'sidekiq', '~> 5.2'
```
ndushay: can we get a comment explaining why we're pinning?

PR #457, opened 2019-10-07
Trap for a nil pointer and give a helpful message

jcoyne:
## Why was this change made?

The nil pointer exceptions logged in honeybadger don't make it easy to decide what action needs to be taken.

Yes

### DIFF
--- a/app/services/public_xml_service.rb
+++ b/app/services/public_xml_service.rb
@@ -104,6 +104,13 @@ class PublicXmlService
     # locate and extract the resourceId/fileId elements
     doc = src_item.contentMetadata.ng_xml
     src_resource = doc.at_xpath("//resource[@id=\"#{src_resource_id}\"]")
+
+    unless src_resource
+      raise Dor::DataError, "The contentMetadata of #{object.pid} has an exernalFile "\
+        "reference to #{src_druid}, #{src_resource_id}, but #{src_druid} doesn't have " \
+        "a matching resource node in it's contentMetadata"
+    end
+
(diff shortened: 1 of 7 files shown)


app/services/public_xml_service.rb
```diff
@@ -104,6 +104,13 @@ def add_data_from_src(external_file)
     # locate and extract the resourceId/fileId elements
     doc = src_item.contentMetadata.ng_xml
     src_resource = doc.at_xpath("//resource[@id=\"#{src_resource_id}\"]")
+
+    unless src_resource
+      raise Dor::DataError, "The contentMetadata of #{object.pid} has an exernalFile "\
```
mjgiarlo: s/exernalFile/externalFile/

app/services/public_xml_service.rb
```diff
@@ -104,6 +104,13 @@ def add_data_from_src(external_file)
...
+        "reference to #{src_druid}, #{src_resource_id}, but #{src_druid} doesn't have " \
+        "a matching resource node in it's contentMetadata"
```
mjgiarlo: s/it's/its/

openapi.json
```diff
@@ -353,6 +353,16 @@
         "responses": {
           "200": {
             "description": "OK"
+          },
+          "422": {
+            "description": "There is a problem with the data that makes is impossible to publish"
```
mjgiarlo: need a comma at the end of the line?

jcoyne: Trailing commas are illegal in json.

jcoyne: oh, not trailing here.

PR #464, opened 2019-10-14
use pry for irb

ndushay:
## Why was this change made?

So rails console will be pry, rather than irb.  I find it really useful to be able to "ls xxx", where xxx is an object, to see the available methods and where they come from.  I also really appreciate the color output.

n/a

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -3,5 +3,9 @@
 
 ENV['RACK_ENV'] = ARGV.first unless ARGV.first.nil?
 require File.expand_path(File.dirname(__FILE__) + '/../config/boot')
-require 'irb'
-IRB.start
+
+# require 'irb'
+# IRB.start
+
+require 'pry'
+Pry.start


ndushay:
@mjgiarlo - I want this to happen on the deployed boxes, as well.

mjgiarlo:
@ndushay OK, so you want to ssh to a box and run `bundle exec rails c -e production` and see `pry`? If so, do we need to add `pry` to the `Gemfile`? I don't believe that dependency is accessible when `RAILS_ENV=production`.

ndushay:
@mjgiarlo good point.  I'm now ... busted.

jmartin-sul:
fwiw, i am generally in favor of including pry as a prod dependency so that it is usable on prod rails console.  i wouldn't push hard to reopen this work if others would prefer not to do that.  but i find pry so much nicer than default irb (i'd personally be happy to see `gem 'pry-rails'` in the default dependencies for all our rails projects).

are there significant downsides to making it a prod dependency?  e.g. does it bloat the app appreciably, or have a bad vulnerability history?

jmartin-sul:
also, to get pry for rails console in all environments, i believe all that's needed is to add `gem 'pry-rails'` to `Gemfile`, and to add `.pry_history` to `.gitignore` (well, that's not strictly necessary, but still a good idea).

e.g.:
https://github.com/sul-dlss/preservation_catalog/search?q=pry&unscoped_q=pry
https://github.com/sul-dlss/argo/search?q=pry&unscoped_q=pry

mjgiarlo:
> fwiw, i'm generally in favor of including pry as a prod dependency so that it's usable on prod rails console. i wouldn't push hard to reopen this work if others would prefer not to do that. but i find pry so much nicer than default irb (i'd personally be happy to see `gem 'pry-rails'` in the general/ungrouped dependencies for all our rails projects).
> 
> are there significant downsides to making it a prod dependency? e.g. does it bloat the app appreciably, or have a bad vulnerability history?

I don't think that's problematic, no. Totally fine by me.

PR #466, opened 2019-10-15
ShelvingService provides more details to client when missing content dir

ndushay:
## Why was this change made?

So argo will show something more useful than this when a content dir is missing in the shelve step for common-accessioning.

image

"shelve : Internal Server Error: 500 (Response from dor-services-app did not contain a body. Check honeybadger for dor-services-app for backtraces, and look into adding a `rescue_from` in dor-services-app to provide more details to the client in the future)"

n/a


jmartin-sul:
er, now that i think about it... would a test be appropriate?

PR #467, opened 2019-10-15
Add a wrapping exception with less indirection

jcoyne:
## Why was this change made?

Previously it was unclear where this exception was raised from.  This implementation is a little more straightforward and has tests

No, but this changes the code to conform to the documented return values.

### DIFF
--- a/app/controllers/shelves_controller.rb
+++ b/app/controllers/shelves_controller.rb
@@ -4,11 +4,6 @@
 class ShelvesController < ApplicationController
   before_action :load_item, only: :create
 
-  # exception defined in application.rb;  see https://pdabrowski.com/blog/ruby-rescue-from-errors-with-grace
-  rescue_from(DorServices::ContentDirNotFoundError) do |e|
-    render status: :internal_server_error, plain: e.message
-  end
-
   def create
     if @item.is_a?(Dor::Item)
       ShelvingService.shelve(@item)
@@ -20,5 +15,11 @@ class ShelvesController < ApplicationController
         ]
       }, status: :unprocessable_entity
     end
+  rescue ShelvingService::ContentDirNotFoundError => e
+    render json: {
+      errors: [
+        { title: 'Content directory not found', detail: e.message }
+      ]
+    }, status: :unprocessable_entity
   end
 end
--- a/app/services/shelving_service.rb
+++ b/app/services/shelving_service.rb
@@ -2,6 +2,8 @@
 
 # Push file changes for shelve-able files into stacks
 class ShelvingService
+  class ContentDirNotFoundError < RuntimeError; end
+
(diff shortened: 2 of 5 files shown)


app/controllers/shelves_controller.rb
```diff
@@ -4,11 +4,6 @@
 class ShelvesController < ApplicationController
   before_action :load_item, only: :create
 
-  # exception defined in application.rb;  see https://pdabrowski.com/blog/ruby-rescue-from-errors-with-grace
```
ndushay: i figured this comment was a sufficient explanation for where to find the exception.

I find indirection for the exception declaration, with an explanation provided, a lot less frustrating than indirection for the happy path logic.

app/controllers/shelves_controller.rb
```diff
@@ -4,11 +4,6 @@
...
-  rescue_from(DorServices::ContentDirNotFoundError) do |e|
-    render status: :internal_server_error, plain: e.message
-  end
```
ndushay: we use this approach in a number of other controllers (rescue_from)

jcoyne: rescue_from is okay for a generic error handler, but the handler is not specific to the context in which it is called in.   I like to see that we are handling it in a manner appropriate for the calling context.    I think it's easier to see where that comes from too.

app/controllers/shelves_controller.rb
jcoyne: The previous way of handing this doesn't give us a backtrace to find the source of the error.

jmartin-sul:
mixed feelings.  i like the terseness of #466, but i did have concerns about there being a bit too much magic to figure out when such an error comes up.  the comment in both places mostly quelled that concern for me.

i do for sure like the addition of the test in this PR, and would like to see the test part stay regardless of whatever else happens.

@jcoyne
> The previous way of handing this doesn't give us a backtrace to find the source of the error.

would that not have been provided by the `e.message` in the response?

https://github.com/sul-dlss/dor-services-app/pull/467/files#diff-88dda79bf662662b9af1b2fb711c855fL8-L10

```ruby
  rescue_from(DorServices::ContentDirNotFoundError) do |e|
    render status: :internal_server_error, plain: e.message
  end
```

i think my vote is a strong "yes" for the test, and abstention for the rest.  happy to actually vote yay or nay on the app code changes if things are still tied after further discussion.

jcoyne:
@jmartin-sul okay, I redact what I said about backtraces.  I made a bad assumption due to what I saw in Honeybadger.

ndushay:
I think we are at a "no strong objections" point, and perhaps it's not worth more time spent?

mjgiarlo:
@ndushay WFM. Merging.

PR #469, opened 2019-10-15
Publish objects using a background job

jcoyne:
## Why was this change made?

Objects that have several hundred files may take several minutes to publish, so make this an asynchronous job.  In our testing we found a 600 page book took over 82s to publish

Yes

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -5,9 +5,22 @@ SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SELECT pg_catalog.set_config('search_path', '', false);
 SET check_function_bodies = false;
-SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
+
+--
+-- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
+--
+
+CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
+
+
+--
+-- Name: EXTENSION plpgsql; Type: COMMENT; Schema: -; Owner: -
+--
+
+COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
+
 
 --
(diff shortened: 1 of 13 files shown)


db/structure.sql
```diff
@@ -5,10 +5,23 @@ SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SELECT pg_catalog.set_config('search_path', '', false);
 SET check_function_bodies = false;
-SET xmloption = content;
```
mjgiarlo: Can you explain the changes to this file (other than the stuff below line 24)?

openapi.json
```diff
@@ -317,11 +317,17 @@
         "operationId": "virtual_objects#create",
         "responses": {
           "204": {
```
mjgiarlo: Should this 204 be changed to 201?

openapi.json
```diff
@@ -351,15 +357,14 @@
...
         "operationId": "objects#publish",
         "responses": {
-          "200": {
-            "description": "OK"
-          },
-          "422": {
-            "description": "There is a problem with the data that makes is impossible to publish",
-            "content": {
-              "application/json": {
+          "204": {
```
mjgiarlo: Should this be changed to 201 (created)?

db/structure.sql
jcoyne: not really, this is what happened when I did `db:migrate`

mjgiarlo: Just checking: is this stuff the PG9 stuff you removed recently? If not, no biggie. If so, it may cause us some trouble.

jcoyne: I'm pretty sure this is on pg10 now.

openapi.json
jcoyne: fixed.

openapi.json
jcoyne: fixed.

PR #470, opened 2019-10-15
Avoid looking for release tags in PURL

jcoyne:
## Why was this change made?

This behavior was originally introduced here: https://github.com/sul-dlss/dor-services/commit/439b8084f5f36567bbefeb954f40ebee82268ff1#diff-8200ce146e3aa5427aa3d4f563164195R81-R82 but none of us (including Ben) can understand why this would be necessary.  Since this slows things down and adds fragility, remove it.

N/A.

### DIFF
--- a/app/services/publish_metadata_service.rb
+++ b/app/services/publish_metadata_service.rb
@@ -16,7 +16,10 @@ class PublishMetadataService
   def publish
     return unpublish unless world_discoverable?
 
-    transfer_metadata
+    # Retrieve release tags from identityMetadata and all collections this item is a member of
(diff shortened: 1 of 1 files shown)


app/services/publish_metadata_service.rb
```diff
@@ -16,7 +16,10 @@ def initialize(item)
   def publish
     return unpublish unless world_discoverable?
 
-    transfer_metadata
+    # Retrieve release tags from metadata and PURL
```
peetucket: maybe update this code comment since we no longer update release tags from PURL?

jcoyne: Yes, good call!

PR #476, opened 2019-10-24
Return a helpful message if a directory doesn't exist

jcoyne:
Previously it was just returning 500 with no message

## Why was this change made?
So that common-accessioning has a clue about what went wrong: https://app.honeybadger.io/projects/52894/faults/53522097

yes.

### DIFF
--- a/app/controllers/workspaces_controller.rb
+++ b/app/controllers/workspaces_controller.rb
@@ -21,6 +21,8 @@ class WorkspacesController < ApplicationController
       CleanupService.cleanup_by_druid druid
     end
     head :no_content
+  rescue Errno::ENOENT => e
+    render build_error('Unable to remove directory', e)
   end
 
   # Once an object has been transferred to preservation clean up the workspace.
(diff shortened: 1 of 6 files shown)


app/controllers/workspaces_controller.rb
```diff
@@ -39,4 +41,21 @@ def full_druid_tree?(druid)
     truncate_druid_tree = DruidTools::AccessDruid.new(druid, workspace_root)
     (!Dir.glob(truncate_druid_tree.path).empty? && !Dir.glob(full_druid_tree.path + '*').empty?)
   end
+
+  # JSON-API error response
```
mjgiarlo: Looks good. This same method appears to be defined in VersionsController as well. Worth considering moving it to ApplicationController or too early?

jcoyne: I think too early yet. Need one more before we abstract

mjgiarlo: okie dokie, I buy it.

PR #480, opened 2019-10-28
Upgrade to postgres 11

jcoyne:
You may need to remove the old volume first if you have one.

```
docker-compose down --volumes
```

## Why was this change made?

Because PG 10 is far out of date.

n/a


mjgiarlo:
@jcoyne Other questions in addition to :point_up: :

1. Do we need to (or should we) specify the PG version in `Dockerfile`?
1. Should we regenerate `db/structure.sql` w/ PG 11?

jcoyne:
1. I think we should specify a version.  We've seen that 12 doesn't work right now.  
1. I don't know.

PR #490, opened 2019-10-31
Callback to the workflow service when step is complete

jcoyne:
## Why was this change made?

This avoids the need for common-accessioning to poll for job completion

Depends on https://github.com/sul-dlss/workflow-server-rails/pull/271

### DIFF
--- a/app/jobs/shelve_job.rb
+++ b/app/jobs/shelve_job.rb
@@ -13,16 +13,14 @@ class ShelveJob < ApplicationJob
       item = Dor.find(druid)
       ShelvingService.shelve(item)
     rescue ShelvingService::ContentDirNotFoundError => e
-      background_job_result.output = { errors: [{ title: 'Content directory not found', detail: e.message }] }
-      background_job_result.complete!
+      return LogFailureJob.perform_later(druid: druid,
+                                         background_job_result: background_job_result,
+                                         workflow_process: 'shelve',
+                                         output: { errors: [{ title: 'Content directory not found', detail: e.message }] })
     end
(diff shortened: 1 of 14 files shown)


app/jobs/shelve_job.rb
```diff
@@ -13,16 +13,14 @@ def perform(druid:, background_job_result:)
...
     rescue ShelvingService::ContentDirNotFoundError => e
-      background_job_result.output = { errors: [{ title: 'Content directory not found', detail: e.message }] }
-      background_job_result.complete!
+      return LogFailureJob.perform_later(druid: druid,
+                                         background_job_result: background_job_result,
+                                         workflow_process: 'shelve',
+                                         output: { errors: [{ title: 'Content directory not found', detail: e.message }] })
     end
 
-    # These two lines should be unnecessary, but we are getting:
```
mjgiarlo: Removing these to see if the underlying behavior still happens in the background job context?

app/services/sdr_ingest_service.rb
```diff
@@ -33,11 +33,6 @@ def self.transfer(dor_item)
     bagger.deposit_group('metadata', metadata_dir)
     bagger.create_tagfiles
     verify_bag_structure(bag_dir)
-
```
mjgiarlo: :tada:

app/jobs/shelve_job.rb
jcoyne: I'm pretty sure it only happens when the background job takes a long time.  Now it will be a short job.

PR #492, opened 2019-11-01
Provide error handling for when the directory can't be deleted

jcoyne:
## Why was this change made?
Better errors when a directory can't be deleted.

n/a

### DIFF
--- a/app/controllers/workspaces_controller.rb
+++ b/app/controllers/workspaces_controller.rb
@@ -21,7 +21,7 @@ class WorkspacesController < ApplicationController
       CleanupService.cleanup_by_druid druid
     end
     head :no_content
-  rescue Errno::ENOENT => e
+  rescue Errno::ENOENT, Errno::ENOTEMPTY => e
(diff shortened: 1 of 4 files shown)


app/controllers/workspaces_controller.rb
```diff
@@ -23,6 +23,8 @@ def destroy
     head :no_content
   rescue Errno::ENOENT => e
     render build_error('Unable to remove directory', e)
+  rescue Errno::ENOTEMPTY => e
```
mjgiarlo: Not important, but since the behavior/message is the same on lines 25 and 27, I'm curious why you didn't do:

```ruby
rescue Errno::ENOENT, Errno::ENOTEMPTY => e
  render build_error('Unable to remove directory', e)
```

jcoyne: I was hoping rubocop would do it for me. 😆

mjgiarlo: typical rubocop

PR #499, opened 2019-11-08
Add an API endpoint for handling legacy metadata

jcoyne:
## Why was this change made?

This can be used by accessionWF robots so they don't need to make Fedora calls

yes.

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -13,4 +13,23 @@ class MetadataController < ApplicationController
     service = PublicDescMetadataService.new(@item)
     render xml: service
   end
+
+  # This supports the Legacy Fedora 3 data model. This is used by the accessionWF.
+  def update_legacy_metadata
+    datastream_names = { descriptive: 'descMetadata',
+                         technical: 'technicalMetadata',
+                         content: 'contentMetadata',
+                         rights: 'rightsMetadata' }
+
+    datastream_names.each do |section, datastream_name|
+      values = params[section]
+      next unless values
+
+      LegacyMetadataService.update_datastream_if_newer(datastream: @item.datastreams[datastream_name],
+                                                       updated: Time.zone.parse(values[:updated]),
+                                                       content: values[:content])
+    end
+
+    @item.save!
+  end
 end
--- a/app/services/legacy_metadata_service.rb
+++ b/app/services/legacy_metadata_service.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Operations on the legacy Fedora 3 metadata.
+# This is used by the accessionWF
+class LegacyMetadataService
+  # If the updated value is newer than then createDate of the datastream, then update it.
+  def self.update_datastream_if_newer(datastream:, updated:, content:)
+    datastream.content = content if !datastream.createDate || updated > datastream.createDate
+    return if !datastream.createDate || updated > datastream.createDate
+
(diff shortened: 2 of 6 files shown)


.rubocop.yml
```diff
@@ -14,6 +14,9 @@ AllCops:
 Rails:
   Enabled: true
 
+Lint/AssignmentInCondition:
```
mjgiarlo: Would you consider reverting this change? Assignment in condition in Rubyland always makes me wonder what the author intended. I'd rather leave this enabled, myself. Would be good to hear from other @sul-dlss/infrastructure-team folks.

app/controllers/metadata_controller.rb
```diff
@@ -13,4 +13,24 @@ def descriptive
...
+
+  # This supports the Legacy Fedora 3 data model. This is used by the accessionWF.
+  def update_legacy_metadata
+    datastream_names = { descriptive: 'descMetadata',
+                         technical: 'technicalMetadata',
+                         content: 'contentMetadata',
+                         rights: 'rightsMetadata' }
+
+    datastream_names.each do |section, datastream_name|
+      next unless values = params[section]
```
mjgiarlo: FWIW, I don't think this block would be terrible if you removed the assignment here and just used `params[section]` in all places where `values` is used now. I also wouldn't mind if we replaced this line with two lines, one for the assignment and one for the check.

app/controllers/metadata_controller.rb
```diff
@@ -13,4 +13,24 @@ def descriptive
...
+
+      update_datastream_if_newer(datastream: @item.datastreams[datastream_name],
+                                 updated: Time.zone.parse(values[:updated]),
+                                 content: values[:content])
+    end
+
+    @item.save!
+  end
+
+  delegate :update_datastream_if_newer, to: LegacyMetadataService
```
mjgiarlo: Can you say more about why you're using delegation as the technique to invoke this service object's method instead of calling the method on the object? Is this a pattern you want to see us use more or does it prepare us for a further refactoring or... ?

.rubocop.yml
ndushay: I'm not a big fan of assignment in condition either.

app/controllers/metadata_controller.rb
ndushay: I would prefer either of Mike's two alternatives to the assignment condition

.rubocop.yml
jcoyne: 😢 

okay

app/controllers/metadata_controller.rb
jcoyne: no reason. do you hate it?

mjgiarlo: @jcoyne I don't hate it, but it did surprise me and I thought it worth flagging. I have, I guess, a slight-to-moderate preference for doing this without delegation for the sake of consistency with the rest of the codebase (where we make only light usage of `delegate` and delegate only to an attr/method defined within the class rather than delegating to another class. Tagging in @sul-dlss/infrastructure-team here too.

mjgiarlo:
> @ndushay I think our patterns are more specific than implied by "fedora3", so I prefer to call them something other than fedora3....I'd be up for a better name than "legacy" if you can think of one.

I don't have a preference here FWIW.

ndushay:
I can live with "legacy".  I can't live with failing tests, tho.

PR #507, opened 2019-11-26
Validate that object is valid before saving.

jcoyne:
## Why was this change made?

This catches problems earlier than when this was done in common-accessioning

n/a


ndushay:
this will be an AWESOME change.  yay!

PR #508, opened 2019-11-26
return metadata identifiers in preferred order

peetucket:
## Why was this change made?

Based on a discussion in `#dlss-aaas` about metadata refreshes for objects with both a catkey and a barcode, it may be necessary to prefer they catkey when both are present.  This returns the identifiers in a preferred order based on the known prefixes configured so that when the first is selected it will work as desired.

### DIFF
--- a/app/services/metadata_service.rb
+++ b/app/services/metadata_service.rb
@@ -9,9 +9,14 @@ class MetadataService
   class << self
     @@cache = Cache.new(nil, nil, 250, 300)
 
-    # TODO: Return a prioritized list
+    def known_prefixes
+      handlers.keys
+    end
+
+    # return the identifiers found in the same order of the known prefixes we specified
     def resolvable(identifiers)
-      identifiers.select { |identifier| can_resolve?(identifier) }
+      res_ids = identifiers.select { |identifier| can_resolve?(identifier) }
+      MetadataService.known_prefixes.map { |prefix| res_ids.find { |res_id| res_id.start_with?(prefix.to_s) } }.compact
     end
 
     def fetch(identifier)
--- a/app/services/metadata_service.rb
+++ b/app/services/metadata_service.rb
@@ -13,6 +13,8 @@ class MetadataService
     def resolvable(identifiers)
       res_ids = identifiers.select { |identifier| can_resolve?(identifier) }
       VALID_PREFIXES.map { |prefix| res_ids.find { |res_id| res_id.start_with?(prefix.to_s) } }.compact
+      # NOTE: the purpose of .map here is to ensure we return any resolvable identifiers in the
+      #       preferred order specified above in KNOWN_PREFIXES, so that the .first is the preferred one
     end
 
     def fetch(identifier)
--- a/app/services/metadata_service.rb
+++ b/app/services/metadata_service.rb
@@ -9,14 +9,10 @@ class MetadataService
   class << self
     @@cache = Cache.new(nil, nil, 250, 300)
 
-    def known_prefixes
-      handlers.keys
-    end
-
     # return the identifiers found in the same order of the known prefixes we specified
     def resolvable(identifiers)
       res_ids = identifiers.select { |identifier| can_resolve?(identifier) }
-      MetadataService.known_prefixes.map { |prefix| res_ids.find { |res_id| res_id.start_with?(prefix.to_s) } }.compact
+      VALID_PREFIXES.map { |prefix| res_ids.find { |res_id| res_id.start_with?(prefix.to_s) } }.compact
     end
 
     def fetch(identifier)
--- a/app/services/refresh_metadata_action.rb
+++ b/app/services/refresh_metadata_action.rb
@@ -27,6 +27,6 @@ class RefreshMetadataAction
   def fetch_datastream
     candidates = @object.identityMetadata.otherId.collect(&:to_s)
     metadata_id = MetadataService.resolvable(candidates).first
-    metadata_id.blank? ? nil : MetadataService.fetch(metadata_id.to_s)
+    metadata_id.nil? ? nil : MetadataService.fetch(metadata_id.to_s)
   end
 end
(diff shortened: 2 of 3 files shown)


app/services/refresh_metadata_action.rb
```diff
@@ -27,6 +27,6 @@ def run(datastream)
   def fetch_datastream
     candidates = @object.identityMetadata.otherId.collect(&:to_s)
     metadata_id = MetadataService.resolvable(candidates).first
-    metadata_id.nil? ? nil : MetadataService.fetch(metadata_id.to_s)
+    metadata_id.blank? ? nil : MetadataService.fetch(metadata_id.to_s)
```
peetucket: this ensures an empty array is treated the same as nil

app/services/metadata_service.rb
```diff
@@ -16,9 +16,10 @@ def can_resolve?(identifier)
       handlers.key?(prefix.to_sym)
     end
 
-    # TODO: Return a prioritized list
+    # return the identfiers found in the same order of the known prefixes we specified
     def resolvable(identifiers)
-      identifiers.select { |identifier| can_resolve?(identifier) }
+      res_ids = identifiers.select { |identifier| can_resolve?(identifier) }
+      MetadataService.known_prefixes.map { |prefix| res_ids.find { |res_id| res_id.start_with?(prefix.to_s) } }.compact
```
peetucket: there may be a better way to achieve a sort by `known_prefixes`

mjgiarlo: How would you feel about something like:

```ruby
# may be worth renaming this method... `validate` ?
def resolvable(identifiers)
  identifiers.select { |identifier| can_resolve?(identifier) && known_prefix?(identifier) }.sort_by { |identifier| known_prefixes.index(identifier.split(':').first) }
end

def known_prefix?(identifier)
  known_prefixes.any? { |prefix| identifier.start_with?(prefix.to_s) }
end
```

peetucket: I don't think this actually sorts the resolvable identifiers in the order of known prefixes (at broke the tests I had written).  But I do like your implementation of `known_prefix?`, which implements the existing `can_resolve?` in a more clear way.  I updated the PR to replace `can_resolve?` with this implementation and added some tests for that method.

spec/services/metadata_service_spec.rb
```diff
@@ -7,6 +7,34 @@
...
+    end
+
+    it 'returns an empty array when no resolvable identifiers are found' do
+      expect(described_class.resolvable(['bogus:1234', 'nada:5678'])).to eq([])
+    end
+  end
+
+  describe 'can_resolve?' do
+    it 'returns false for an unknown prefix' do
+      expect(described_class).not_to be_can_resolve('bogus:1234')
```
peetucket: rubocop made me do this, i preferred the older style: `expect(described_class.can_resolve?('bogus:1234')).to be_truthy`

app/services/metadata_service.rb
peetucket: never mind, had to rebase from master and the implementation of `can_resolve?` changed from under me to something else

app/services/refresh_metadata_action.rb
jcoyne: How could this be an array if we call `.first` on the array returned by `.resolvable()`?

app/services/metadata_service.rb
mjgiarlo: You still open to considering :point_up:? or meh?

app/services/refresh_metadata_action.rb
peetucket: ahh, i guess i missed that bit -- will revert to nil?

app/services/metadata_service.rb
peetucket: Based on the recent code changes, it would have to be: 

```
    # return the identifiers found in the same order of the known prefixes we specified
    def resolvable(identifiers)
      identifiers
        .select { |identifier| can_resolve?(identifier) && known_prefix?(identifier) }
        .sort_by { |identifier| VALID_PREFIXES.index(identifier.split(':').first) }
    end

    def known_prefix?(identifier)
      VALID_PREFIXES.any? { |prefix| identifier.start_with?(prefix.to_s) }
    end
```

I'm ambivalent - neither seems super intuitive to me.  What do you think @jcoyne ?

mjgiarlo: Also tagging in @ndushay for her take.

ndushay: There is an ongoing conversation in aaas slack channel in which Ben A asks if anyone still needs barcodes.    Personally, I'm ok with ckey before barcode as long as the channel and Andrew agree.

jcoyne: we might still need them for google books.

mjgiarlo: @peetucket Since this has dragged on a bit (my bad!), I'd be happy to move forward with the code we've already got in this branch as long as you add a comment that the purpose of the `#map` operation here is to sort the identifiers by prefix.

peetucket: sure thing, added a comment

app/services/metadata_service.rb
```diff
@@ -13,6 +13,8 @@ class << self
     def resolvable(identifiers)
       res_ids = identifiers.select { |identifier| can_resolve?(identifier) }
       VALID_PREFIXES.map { |prefix| res_ids.find { |res_id| res_id.start_with?(prefix.to_s) } }.compact
+      # NOTE: the purpose of .map here is to ensure we return any resolvable identifiers in the the
```
mjgiarlo: Typo: `s/the the/the /`

app/services/metadata_service.rb
```diff
@@ -13,6 +13,8 @@ class << self
...
+      #       preferred order specificed above in KNOWN_PREFIXES, so that the .first is the preferred one
```
mjgiarlo: Typo: `s/specificed/specified/`

app/services/metadata_service.rb
peetucket: fixed

app/services/metadata_service.rb
peetucket: this comment was very specificed and led to me fixifying it

mjgiarlo: @peetucket :laughing:

PR #512, opened 2019-12-05
Remove HB alert for deprecated endpoint

mjgiarlo:
## Why was this change made?

Because it is too noisy and we already have an issue to track this. Also, do not want our HB account to be throttled again.

no

### DIFF
--- a/app/controllers/sdr_controller.rb
+++ b/app/controllers/sdr_controller.rb
@@ -19,7 +19,6 @@ class SdrController < ApplicationController
 
   # Deprecated
   def ds_manifest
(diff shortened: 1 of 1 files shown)


jmartin-sul:
> jcoyne should sign off -- we do want to make sure we get all these spots fixed.

to @ndushay's concern, what if we put something in the logs that we could grep for?  we'd still get the info we want, and while it'd be slightly less convenient, it'd be less of a strain on HB.

mjgiarlo:
@jcoyne what would you think of going with @jmartin-sul's idea? logging these to disk rather than to honeybadger?

jcoyne:
@mjgiarlo I considered that option when I wrote it this way. I think we want to be actively notified when this is called, and using a logger won't do that.

PR #514, opened 2019-12-06
Shelve w pres client

ndushay:
DRAFT:  
- ~~must not be deployed until sul-dlss/preservation_catalog/pull/1250 is deployed.~~
- ~~should not be deployed until sul-dlss/common-accessioning/pull/465 is deployed~~

## Why was this change made?

To get rid of  `cm-inv-diff` API endpoint in sdr-services-app.  We use preservation-client `shelve_content_diff` method instead.

yes - marked DEPRECATED there and also in the code.

### DIFF
--- a/app/controllers/sdr_controller.rb
+++ b/app/controllers/sdr_controller.rb
@@ -6,8 +6,10 @@ class SdrController < ApplicationController
 
   MANIFEST_DEPRECATION_MESSAGE = 'Use preservation-client manifest or signature_catalog in caller instead.'
   CURRENT_VERSION_DEPRECATION_MESSAGE = 'Use preservation-client current_version in caller instead.'
+  CM_INV_DIFF_DEPRECATION_MESSAGE = 'Use preservation-client content_inventory_diff or shelve_content_diff in caller instead.'
 
   def cm_inv_diff
+    Honeybadger.notify("dor-services-app deprecated API endpoint `sdr#cm_inv_diff` called. #{CM_INV_DIFF_DEPRECATION_MESSAGE}")
     unless %w(all shelve preserve publish).include?(params[:subset])
       render status: :bad_request, plain: "Invalid subset value: #{params[:subset]}"
       return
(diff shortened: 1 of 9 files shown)


app/controllers/sdr_controller.rb
```diff
@@ -5,6 +5,8 @@ class SdrController < ApplicationController
   self.deprecation_horizon = 'dor-services-app version 4.0'
 
   def cm_inv_diff
+    dep_msg = 'Use preservation-client content_inventory_diff or shelve_content_diff in caller instead.'
```
jcoyne: This endpoint is currently in use by https://github.com/sul-dlss/common-accessioning/blob/78b57069cb9bb30fd712f43d2dec9fa2f739616f/lib/technical_metadata_service.rb#L62

ndushay: yup it's one of the reasons this is a draft PR.  I found that and am working on a PR for it right now.

openapi.json
```diff
@@ -75,8 +75,8 @@
         "tags": [
           "preservation"
         ],
-        "summary": "Get the diff of the content metadata from DOR and SDR",
-        "description": "",
+        "summary": "DEPRECATED. Get the diff between content metadata provided and version in preservation",
+        "description": "DEPRECATED.  Use preservation-client gem instead.",
         "operationId": "sdr#cm_inv_diff",
```
mjgiarlo: Can you add `"deprecated": true,` too?

ndushay: ja, will do.

app/controllers/sdr_controller.rb
ndushay: See https://github.com/sul-dlss/common-accessioning/pull/465

app/controllers/sdr_controller.rb
```diff
@@ -8,6 +8,8 @@ class SdrController < ApplicationController
   CURRENT_VERSION_DEPRECATION_MESSAGE = 'Use preservation-client current_version in caller instead.'
 
   def cm_inv_diff
+    dep_msg = 'Use preservation-client content_inventory_diff or shelve_content_diff in caller instead.'
```
mjgiarlo: Would you consider pulling this string up to line 9 or thereabouts and making it a constant? `CONTENT_DIFF_DEPRECATION_MESSAGE`?

ndushay: sure - one sec.

ndushay: pushed.

PR #520, opened 2019-12-10
Make openapi.json valid JSON

jcoyne:
## Why was this change made?

It was not valid json

yes


mjgiarlo:
@jcoyne Hrm, I thought CI was doing OAS validation for us. Not anymore?

PR #527, opened 2019-12-11
Revert "Revert "Callback when preservation ingest has started""

mjgiarlo:
Reverts sul-dlss/dor-services-app#526


mjgiarlo:
@jcoyne Feel free to make this a real PR and go forward with it once you're ready to coordinate the deploys (w/ common-accessioning and workflow-server-rails).

PR #529, opened 2019-12-12
Pres client v1.0.0

ndushay:
## Why was this change made?

short answer:  to upgrade to preservation-client v1.0.0

longer answer:  preservation-client v1.0.0 has a non-backwards compatible change with previous versions (just in which errors are thrown when something isn't found).  I wanted to go thru all consumers (argo, common-accessioning, and dor-services-app) and take care of any updates 
- separately from Monday's dep updates, and also 
- separately from switching DSA to use pres-client to get signature_catalog.

As it happens, nothing needs to be tweaked with pres-client v1.0.0 here.

I also took the opportunity to tweak something I did recently in the ShelvingService - see inline comments.

n/a


app/services/shelving_service.rb
```diff
@@ -32,10 +32,10 @@ def shelve
   # retrieve the differences between the current contentMetadata and the previously ingested version
   # (filtering to select only the files that should be shelved to stacks)
   def shelve_diff
-    raise Dor::ParameterError, 'Missing Dor::Config.stacks.local_workspace_root' if Dor::Config.stacks.local_workspace_root.nil?
-    raise Dor::Exception, 'Missing contentMetadata datastream' if work.contentMetadata.nil?
-
     @shelve_diff ||= begin
+      raise Dor::ParameterError, 'Missing Dor::Config.stacks.local_workspace_root' if Dor::Config.stacks.local_workspace_root.nil?
+      raise Dor::Exception, 'Missing contentMetadata datastream' if work.contentMetadata.nil?
+
```
ndushay: I realized that there was no need to execute these guard checks every time this method is called.

PR #534, opened 2019-12-17
Add documentation.

jcoyne:
## Why was this change made?
make it easier to understand.

n/a

### DIFF
--- a/app/services/refresh_metadata_action.rb
+++ b/app/services/refresh_metadata_action.rb
@@ -3,7 +3,7 @@
 # Look into identityMetadata for compliant ids and use them to fetch
 # descriptive metadata from Symphony.  Put the fetched value in the descMetadata
 class RefreshMetadataAction
-  # @return [NilClass,Object] returns nil if there was no resolvalble metadata id.
+  # @return [NilClass,Object] returns nil if there was no resolvable metadata id.
   def self.run(object)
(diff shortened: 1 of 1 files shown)


app/services/refresh_metadata_action.rb
```diff
@@ -3,6 +3,7 @@
 # Look into identityMetadata for compliant ids and use them to fetch
 # descriptive metadata from Symphony.  Put the fetched value in the descMetadata
 class RefreshMetadataAction
+  # @return [NilClass,Object] returns nil if there was no resolvalble metadata id.
```
justinlittman: I think resolvable is misspelled.

Also, is this saying "NilClass or Object" or "array of NilClass followed by Object"?

jcoyne: no, it's saying NilClass or Object may be returned.

justinlittman: Thanks. Still typo on "resolvable".

jmartin-sul:
```suggestion
  # @return [NilClass,Object] returns nil if there was no resolvable metadata id.
```

jmartin-sul: took the liberty of merging since this typo was addressed

PR #535, opened 2019-12-17
Differentiate between missing ids and bad data

jcoyne:
## Why was this change made?

This helps understand what can go wrong with this step

yes

### DIFF
--- a/app/controllers/metadata_refresh_controller.rb
+++ b/app/controllers/metadata_refresh_controller.rb
@@ -9,13 +9,19 @@ class MetadataRefreshController < ApplicationController
   end
 
   def refresh
-    status = RefreshMetadataAction.run(@item)
+    status = RefreshMetadataAction.run(identifiers: identifiers,
+                                       datastream: @item.descMetadata)
+    return render status: :unprocessable_entity, plain: "#{@item.pid} had no resolvable identifiers: #{identifiers.inspect}" unless status
     return render status: :internal_server_error, plain: "#{@item.pid} descMetadata missing required fields (<title>)" if missing_required_fields?
 
-    @item.save! if status
+    @item.save!
   end
 
   private
+
+  def identifiers
+    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
+  end
 
   def missing_required_fields?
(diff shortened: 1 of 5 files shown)


app/controllers/metadata_refresh_controller.rb
```diff
@@ -9,14 +9,20 @@ class MetadataRefreshController < ApplicationController
   end
 
   def refresh
-    status = RefreshMetadataAction.run(@item)
+    status = RefreshMetadataAction.run(identifiers: identifiers,
+                                       datastream: @item.descMetadata)
+    return render status: :unprocessable_entity, plain: "#{@item.pid} had no resolvable identifiers: #{identifiers.inspect}" unless status
```
justinlittman: This warren of `ifs` and `unlesses` is a bit confusing, but I leave to you to determine if can be cleaned up.

jcoyne: I can't think of a better way.  Do you have any suggestions?

justinlittman: Well, the `unless status` and `if status` could be combined into an `if status else` (assuming that didn't break the logic). But certainly not a show-stopper.

jcoyne: Oh, I can drop the `if status` at the end because we're already ensuring it's present. I'll make that change.

app/services/refresh_metadata_action.rb
```diff
@@ -25,9 +27,10 @@ def run(datastream)
 
   private
 
+  attr_reader :identifiers, :datastream
+
   def fetch_datastream
-    candidates = @object.identityMetadata.otherId.collect(&:to_s)
-    metadata_id = MetadataService.resolvable(candidates).first
+    metadata_id = MetadataService.resolvable(identifiers).first
```
ndushay: is this another spot where we want to be explicit about which identifier we are using?  Like @peetucket did with pre-assembly re: catkey and barcode?

jcoyne: This is exactly what the MetadataService does.  In any case, changing that is not part of the purpose of this PR.

PR #539, opened 2019-12-18
Test without using node

jcoyne:
## Why was this change made?

* Don't install software we don't need.
* Make the builds faster

n/a


jcoyne:
Node is needed for the openapi spec validation.

PR #540, opened 2019-12-18
Test on the ruby 2.7 build

jcoyne:


jcoyne:
Looks like the config gem version 2.0.0  doesn't support ruby 2.7: https://github.com/railsconfig/config/issues/259

PR #547, opened 2019-12-19
Add route for particular sidekiq queues

jcoyne:
## Why was this change made?

Currently we are prohibited from looking at particular queues in the dashboard.

yes

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -134,7 +134,7 @@
         ]
       }
     },
-    "/queues/stats": {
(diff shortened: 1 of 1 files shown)


openapi.json
```diff
@@ -145,6 +145,17 @@
             "description": "OK"
           }
         }
+        "parameters": [
```
mjgiarlo: Comma missing on prior line?

PR #548, opened 2019-12-20
Handles and/or documents pres client errors.

### LINKED ISSUE #545: trap for preservation-client gem errors in dor-services-app and handle accordingly
dor-services-app uses preservation-client in the following ways:
- shelving API service:
   - uses preservation-client to get content diff for shelved files
- versioning service:
   - uses preservation-client to get current_version
- other?

dor-services-app code should trap for any errors from preservation-client (e.g. can't connect, not found, generic error, response status other than expected (i.e. not a 200) ) and:
- pass along appropriate error code to dor-services-client (which needs to handle it properly)
- notify honeybadger as appropriate

See https://github.com/sul-dlss/preservation-client/blob/master/lib/preservation/client.rb#L30-L39

justinlittman:
closes #545

## Why was this change made?
To make sure that preservation client errors are handled correctly.

No.

### DIFF
--- a/app/jobs/create_virtual_objects_job.rb
+++ b/app/jobs/create_virtual_objects_job.rb
@@ -17,7 +17,7 @@ class CreateVirtualObjectsJob < ApplicationJob
       # Do not add `nil`s to the errors array as they signify successful
       # creation of the virtual object
       errors << result if result.present?
-    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception => e
+    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception, Preservation::Client::Error => e
       errors << { parent_id => [e.message] }
     rescue StandardError => e
       errors << { parent_id => [e.message] }
--- a/app/services/constituent_service.rb
+++ b/app/services/constituent_service.rb
@@ -19,6 +19,7 @@ class ConstituentService
   # subsequent calls will erase the previous changes.
   # @param [Array<String>] child_druids the identifiers of the child objects
   # @raises ActiveFedora::RecordInvalid if AF object validations fail on #save!
+  # @raise [Preservation::Client::Error] if bad response from preservation catalog.
   # @returns [NilClass, Hash] true if successful, hash of errors otherwise (if combinable validation fails)
   def add(child_druids:)
     errors = ItemQueryService.validate_combinable_items(parent: parent_druid, children: child_druids)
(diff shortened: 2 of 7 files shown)


app/jobs/create_virtual_objects_job.rb
```diff
@@ -17,7 +17,7 @@ def perform(virtual_objects:, background_job_result:)
       # Do not add `nil`s to the errors array as they signify successful
       # creation of the virtual object
       errors << result if result.present?
-    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception => e
+    rescue ActiveFedora::ObjectNotFoundError, Rubydora::FedoraInvalidRequest, Dor::Exception, Preservation::Client::Error => e
```
mjgiarlo: I wonder what folks would think about dropping lines 20 and 21 altogether? I'm not sure that they add anything, as all four of these exception classes are subclasses of `StandardError`, so lines 22 and 23 would handle them the exact same way without altering any behavior. See:

```ruby
[1] pry(main)> ActiveFedora::ObjectNotFoundError < StandardError
=> true
[2] pry(main)> Rubydora::FedoraInvalidRequest < StandardError
=> true
[3] pry(main)> Dor::Exception < StandardError
=> true
[4] pry(main)> Preservation::Client::Error < StandardError
=> true
```

app/services/constituent_service.rb
```diff
@@ -19,6 +19,7 @@ def initialize(parent_druid:)
   # subsequent calls will erase the previous changes.
   # @param [Array<String>] child_druids the identifiers of the child objects
   # @raises ActiveFedora::RecordInvalid if AF object validations fail on #save!
+  # @raise [Preservation::Client::Error] if bad response from preservation catalog.
```
mjgiarlo: `s/@raise/@raises/` for consistency w/ prior line (and wouldn't mind seeing the exception in the prior line wrapped in brackets!)

mjgiarlo: On second thought, it looks like there's more usage of `@raises` (14 instances) than `@raise` (8 instances) in this codebase. Maybe best to standardize on `@raises`, assuming both are OK with YARD? Meh.

app/jobs/create_virtual_objects_job.rb
jcoyne: I prefer we keep this explicit, so we have some idea about exactly what we're expecting.

app/services/constituent_service.rb
jcoyne: I think `@raise` is corrrect.

mjgiarlo: Good for a follow-on PR then, to be both consistent and correct? :laughing:

PR #551, opened 2019-12-20
Use the correct yard tag (@raise)

jcoyne:
## Why was this change made?

Comply with Yard syntax

n/a

### DIFF
--- a/app/services/constituent_service.rb
+++ b/app/services/constituent_service.rb
@@ -21,7 +21,7 @@ class ConstituentService
   # @param [Array<String>] child_druids the identifiers of the child objects
   # @raise ActiveFedora::RecordInvalid if AF object validations fail on #save!
   # @raise [Preservation::Client::Error] if bad response from preservation catalog.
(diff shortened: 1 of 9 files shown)


app/services/constituent_service.rb
```diff
@@ -18,7 +18,7 @@ def initialize(parent_druid:)
   # Typically this is only called one time (with a list of all the pids) because
   # subsequent calls will erase the previous changes.
   # @param [Array<String>] child_druids the identifiers of the child objects
-  # @raises ActiveFedora::RecordInvalid if AF object validations fail on #save!
+  # @raise ActiveFedora::RecordInvalid if AF object validations fail on #save!
   # @returns [NilClass, Hash] true if successful, hash of errors otherwise (if combinable validation fails)
```
mjgiarlo: Should we also `s/@returns/@return/` or save that for another PR or meh?

jcoyne: another pr please.

mjgiarlo: @jcoyne :ok:

ndushay:
oh no!  merge conflict!

PR #572, opened 2020-01-14
Check for a valid catkey in the metadata service

### LINKED ISSUE #567: Registration endpoint (Objects#create) should validate catkey
Currently it returns an internal server error if the catkey has a space in it, but it should return a 4xx error if the catkey field doesn't look like a valid catkey.  A valid catkey is "a number, or numbers separated by colons"

## Backtrace

    line 44 of [PROJECT_ROOT]/app/models/symphony_reader.rb: symphony_response
    line 80 of [PROJECT_ROOT]/app/models/symphony_reader.rb: json
    line 84 of [PROJECT_ROOT]/app/models/symphony_reader.rb: bib_record

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/59523546)

aaron-collier:
## Why was this change made?

Fixes #567 

Not entirely sure if this is the right way to resolve this issue, but walking through the code and backtrace it seems to be the point of failure.

### DIFF
--- a/app/services/metadata_service.rb
+++ b/app/services/metadata_service.rb
@@ -41,10 +41,10 @@ class MetadataService
       VALID_PREFIXES.include?(prefix)
     end
 
-    def valid_catkey?(catkey)
-      return true if catkey.is_a? Numeric
+    CATKEY_REGEX = /^\d+(:\d+)*$/
 
-      return true if catkey =~ /[\d+:]+/
+    def valid_catkey?(catkey)
+      CATKEY_REGEX.match?(catkey.to_s)
     end
   end
 end
--- a/app/services/metadata_service.rb
+++ b/app/services/metadata_service.rb
@@ -18,10 +18,10 @@ class MetadataService
     end
 
     def fetch(identifier)
+      return false unless valid_identifier?(identifier)
+
       @@cache.fetch(identifier) do
         (prefix, identifier) = identifier.split(/:/, 2)
-        raise MetadataError, "Unknown metadata prefix: #{prefix}" unless VALID_PREFIXES.include?(prefix)
-        raise MetadataError, "Invalid catkey: #{identifier}" unless valid_catkey?(identifier)
 
         marcxml = MarcxmlResource.find_by(prefix.to_sym => identifier)
         marcxml.mods
@@ -36,15 +36,23 @@ class MetadataService
 
     private
 
+    def valid_identifier?(identifier)
+      raise MetadataError, "Unknown metadata prefix: #{identifier}" unless can_resolve?(identifier)
+      raise MetadataError, "Invalid catkey: #{identifier}" unless valid_catkey?(identifier)
+
+      true
+    end
+
     def can_resolve?(identifier)
       (prefix, _identifier) = identifier.split(/:/, 2)
       VALID_PREFIXES.include?(prefix)
     end
 
-    CATKEY_REGEX = /^\d+(:\d+)*$/
+    CATKEY_REGEX = /^\d+(:\d+)*$/.freeze
 
-    def valid_catkey?(catkey)
-      CATKEY_REGEX.match?(catkey.to_s)
+    def valid_catkey?(identifier)
+      (_prefix, identifier) = identifier.split(/:/, 2)
+      CATKEY_REGEX.match?(identifier.to_s)
     end
   end
 end
--- a/app/services/metadata_service.rb
+++ b/app/services/metadata_service.rb
@@ -21,6 +21,7 @@ class MetadataService
       @@cache.fetch(identifier) do
         (prefix, identifier) = identifier.split(/:/, 2)
         raise MetadataError, "Unknown metadata prefix: #{prefix}" unless VALID_PREFIXES.include?(prefix)
+        raise MetadataError, "Invalid catkey: #{identifier}" unless valid_catkey?(identifier)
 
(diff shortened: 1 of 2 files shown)


app/services/metadata_service.rb
```diff
@@ -39,5 +40,11 @@ def can_resolve?(identifier)
       (prefix, _identifier) = identifier.split(/:/, 2)
       VALID_PREFIXES.include?(prefix)
     end
+
+    def valid_catkey?(catkey)
+      return true if catkey.is_a? Numeric
+
+      return true if catkey =~ /[\d+:]+/
```
mjgiarlo: This regex might also match strings that don't follow the pattern:

```ruby
[1] pry(main)> /[\d+:]+/.match('12345:')
=> #<MatchData "12345:">
[2] pry(main)> /[\d+:]+/.match('12345:234')
=> #<MatchData "12345:234">
[3] pry(main)> /[\d+:]+/.match('12345:234:')
=> #<MatchData "12345:234:">
[4] pry(main)> /[\d+:]+/.match('12345:234 :')
=> #<MatchData "12345:234">
[5] pry(main)> /[\d+:]+/.match('12345: 234 :')
=> #<MatchData "12345:">
```

aaron-collier: Aw shoot...

mjgiarlo: Suggestion coming...

mjgiarlo: I think something like this could work:

```ruby
CATKEY_REGEX = /^\d+(:\d+)*$/

def valid_catkey?(catkey)
  CATKEY_REGEX.match?(catkey.to_s)
end
```

mjgiarlo:
```
[27] pry(main)> match?(3)
=> true
[28] pry(main)> match?(12345)
=> true
[29] pry(main)> match?('123')
=> true
[30] pry(main)> match?('123:')
=> false
[31] pry(main)> match?('123 ')
=> false
[32] pry(main)> match?('123:')
=> false
[33] pry(main)> match?('123: ')
=> false
[34] pry(main)> match?('123:234')
=> true
[35] pry(main)> match?('123:234:576')
=> true
[36] pry(main)> match?('123:234:57ffffff6')
=> false
[37] pry(main)> match?('123:234:57 6')
=> false
[38] pry(main)> match?('123:234:57')
=> true
[39] pry(main)> match?('123:234:57:')
=> false
[40] pry(main)> match?(':123:234:57')
```

app/services/metadata_service.rb
```diff
@@ -18,9 +18,10 @@ def resolvable(identifiers)
     end
 
     def fetch(identifier)
+      return false unless valid_identifier?(identifier)
```
mjgiarlo: It looks like `#valid_identifier?` either returns `true` or raises, so I don't know if `fetch` ever returns `false`. I'm thinking that means the new method below should be a void method that either raises or doesn't. So instead of this line, you'd have

```ruby
validate_identifier!(identifier)
```

And then below you can drop the return value of `true` in the method.

app/services/metadata_service.rb
```diff
@@ -35,9 +36,23 @@ def label_for(identifier)
...
+
     def can_resolve?(identifier)
       (prefix, _identifier) = identifier.split(/:/, 2)
       VALID_PREFIXES.include?(prefix)
     end
+
+    CATKEY_REGEX = /^\d+(:\d+)*$/.freeze
+
+    def valid_catkey?(identifier)
+      (_prefix, identifier) = identifier.split(/:/, 2)
```
mjgiarlo: Now that the same split operation is in three places in this class, I wonder if this PR might be a chance to do this once higher up in the class.

app/services/metadata_service.rb
```diff
@@ -35,9 +36,23 @@ def label_for(identifier)
...
+
+      true
+    end
+
     def can_resolve?(identifier)
       (prefix, _identifier) = identifier.split(/:/, 2)
       VALID_PREFIXES.include?(prefix)
     end
+
+    CATKEY_REGEX = /^\d+(:\d+)*$/.freeze
```
mjgiarlo: Have a minor preference to move this constant up near `VALID_PREFIXES` just to keep like things near one another, but don't care too much about that.

spec/services/metadata_service_spec.rb
```diff
@@ -64,4 +68,18 @@
 
     it { is_expected.to eq 'The isomorphism and thermal properties of the feldspars' }
   end
+
+  describe '#valid_catkey?' do
+    it 'receives a number identifier' do
+      expect(described_class.send(:'valid_catkey?', 'catkey:12345')).to be_truthy
```
mjgiarlo: I don't *believe* you need the single quotes in the first arg. `described_class.send(:valid_catkey?, '...')` should be A-OK. (Symbols can end in a question mark or an exclamation point.)

app/services/metadata_service.rb
```diff
@@ -35,9 +36,23 @@ def label_for(identifier)
...
+
+    def valid_catkey?(identifier)
+      (_prefix, identifier) = identifier.split(/:/, 2)
+      CATKEY_REGEX.match?(identifier.to_s)
```
mjgiarlo: I'm thinking the `#to_s` is not needed here, since the prior line should be splitting a string into an array of two strings.

PR #581, opened 2020-01-21
Adds stub for puma and updates invoke.sh.

justinlittman:
## Why was this change made?
To fix the docker container.

No.


jcoyne:
Looks like it needs `rubocop -a` to fix the formatting

PR #584, opened 2020-01-21
Instrument the code with events

jcoyne:
## Why was this change made?

So that we log important events for an object.

n/a


jcoyne:
@justinlittman exactly.

PR #586, opened 2020-01-22
Pass lane_id when creating workflow in start_preservation_workflow_job.

### LINKED ISSUE #583: Pass on lane id when creating new workflow
So that the child workflow has the same lane id as the parent.

For consideration:
```
$ find app -type f | xargs grep create_workflow_by_name
app/jobs/start_preservation_workflow_job.rb:    Dor::Config.workflow.client.create_workflow_by_name(druid, 'preservationIngestWF', version: version)
app/services/version_service.rb:    Dor::Config.workflow.client.create_workflow_by_name(work.pid, 'versioningWF', version: work.current_version)
```

justinlittman:
closes #583

## Why was this change made?
So that new workflows inherit the lane id from the workflow that created it.

No.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,7 +43,7 @@ gem 'uuidtools', '~> 2.1.4'
 # DLSS/domain-specific dependencies
 gem 'cocina-models', '~> 0.6.0'
 gem 'dor-services', '~> 8.0'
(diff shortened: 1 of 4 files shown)


justinlittman:
Ahhh ... https://github.com/sul-dlss/preservation_robots/pull/191

PR #600, opened 2020-01-27
Add experiment about rights not being provided

jcoyne:
## Why was this change made?
I'm pretty sure this is always sent.  if that is the case, we could make this required and eliminate some conditionals which make the code confusing.

n/a


mjgiarlo:
@jcoyne rubocop'd

PR #603, opened 2020-01-28
Re-adds route for getting MARCXML.

### LINKED ISSUE #594: Expose an API to get the catalog record
So that we can use the catalog record to get the place and date of publication

I believe this used to exist in dor-services-app, but it was removed because we didn't use it.

justinlittman:
closes #594

## Why was this change made?
To re-add a route for getting the MARCXML for an object, so that it can be used to determine the rights of Google books.

Yes.

### DIFF
--- a/app/controllers/marcxml_controller.rb
+++ b/app/controllers/marcxml_controller.rb
@@ -1,10 +1,27 @@
 # frozen_string_literal: true
 
-# Given a barcode, returns a catkey by looking in searchworks
-class MarcxmlController < ApplicationController
-  def catkey
-    marcxml = MarcxmlResource.find_by(barcode: params[:barcode])
+class MarcxmlController < ApplicationController #:nodoc:
+  before_action :set_marcxml_resource
 
-    render plain: marcxml.catkey
+  rescue_from(SymphonyReader::ResponseError) do |e|
+    render status: :internal_server_error, plain: e.message
+  end
+
+  def catkey
+    render plain: @marcxml.catkey
+  end
+
+  def marcxml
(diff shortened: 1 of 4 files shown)


spec/requests/get_marcxml_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+              inds: '41',
+              subfields: [{ code: 'a', data: 'some data' }]
+            }
+          ]
+        }
+      }
+    }
+  end
+
+  # before do
```
jcoyne: Remove

justinlittman: Gone.

app/controllers/marcxml_controller.rb
```diff
@@ -1,10 +1,27 @@
...
+
+  private
+
+  def set_marcxml_resource
+    @marcxml = MarcxmlResource.find_by(**marcxml_resource_params)
+  end
 
-    render plain: marcxml.catkey
+  def marcxml_resource_params
+    params.slice(:barcode, :catkey).to_unsafe_h.symbolize_keys
```
jcoyne: why bother with all this?  why not just return `params`?  or `params.permit(:barcode, :catkey)`?

justinlittman: I was just reverting code, but will fix.

justinlittman:
Turns out Symphony doesn't return 404 when requested MARCXML for a record that doesn't exist:
```
Internal Server Error: 500 (Record not found in Symphony: ) (Dor::Services::Client::UnexpectedResponse)
```

PR #604, opened 2020-01-28
Add provenance metadata to the legacy update endpoint

jcoyne:
## Why was this change made?
This will allow us to refactor the provenance robot so that it doesn't need to hit Fedora directly

yes.


mjgiarlo:
:policeman: :disappointed:

Issue #598, opened 2020-01-27
CocinaMapper should map special_dissemination_wf for APOs


jcoyne:
This can be used by https://github.com/sul-dlss/common-accessioning/blob/master/lib/robots/dor_repo/accession/end_accession.rb#L31-L34

This will allow us to remove that call to Fedora.

wasDisseminationWF is a common value for that field to have

mjgiarlo:
Closed by https://github.com/sul-dlss/dor-services-app/pull/605 ?

PR #605, opened 2020-01-28
Provide administrative properties of an APO via the Cocina API

jcoyne:
## Why was this change made?
This will enable us to decouple common-accessioning from direct Fedora calls

Yes


mjgiarlo:
@jcoyne should this issue have closed https://github.com/sul-dlss/dor-services-app/issues/598?

PR #609, opened 2020-01-30
Support for collections with release tags

jcoyne:
## Why was this change made?
This is fixed by upgrading cocina-models. Previously trying to convert a collection with release tags would result in an error:

```
       undefined method `releaseTags' for #<Cocina::Models::Collection::Administrative>
```

yes


jcoyne:
@justinlittman yes, previously it included a dependency bump to cocina-models, but it's since been rebased. This test failed before the version bump, so seems like a good test to have.

PR #611, opened 2020-01-30
Refactor events to supply additional attributes

jcoyne:
## Why was this change made?

Version events were not supplying host.  All jobs now specify which system invoked them.

n/a

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -4,7 +4,7 @@
 class EventsController < ApplicationController
   def create
     params.require(:event_type)
-    EventFactory.create(druid: params[:object_id], event_type: params[:event_type], data: params[:data])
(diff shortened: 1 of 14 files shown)


app/controllers/events_controller.rb
```diff
@@ -4,7 +4,7 @@
 class EventsController < ApplicationController
   def create
     params.require(:event_type)
-    EventFactory.create(druid: params[:object_id], event_type: params[:event_type], data: params[:data])
+    Event.create!(druid: params[:object_id], event_type: params[:event_type], data: params[:data])
```
mjgiarlo: Can you say more about this change? I'm fuzzy on when to use one or the other.

jcoyne: This is the API endpoint, so the request to this endpoint will presumably be supplying it's own values for host and invoked by.  We can't do that when using EventFactory.

PR #634, opened 2020-02-05
Allow Cocina::Mapper to build APOs with parent APOs

mjgiarlo:
Connects to sul-dlss/common-accessioning#556

## Why was this change made?

To fix a production bug with APOs asserting parent APOs.

no

### DIFF
--- a/spec/requests/show_object_spec.rb
+++ b/spec/requests/show_object_spec.rb
@@ -94,6 +94,7 @@ RSpec.describe 'Get the object' do
       before do
         object.descMetadata.title_info.main_title = 'Hello'
         object.label = 'foo'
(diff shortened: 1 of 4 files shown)


spec/requests/show_object_spec.rb
```diff
@@ -111,6 +111,7 @@
         expect(json['structural']).to eq({})
         expect(json['administrative']['default_object_rights']).to match '<rightsMetadata>'
         expect(json['administrative']['registration_workflow']).to be_nil
+        expect(json['administrative']['hasAdminPolicy']).to be_nil
```
jcoyne: This is probably not a good test

mjgiarlo: @jcoyne Updated.

PR #635, opened 2020-02-06
Use model properties rather than add_value

jcoyne:
## Why was this change made?

Using model properties will help us be able to change the implementation without having to change the interface.

n/a

### DIFF
--- a/spec/services/registration_service_spec.rb
+++ b/spec/services/registration_service_spec.rb
@@ -179,7 +179,6 @@ RSpec.describe RegistrationService do
 
     RSpec.shared_examples 'common registration' do
       it 'produces a registered object' do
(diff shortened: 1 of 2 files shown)


spec/services/registration_service_spec.rb
```diff
@@ -360,55 +359,34 @@
...
+        expect(klass).to have_received(:new)
+          .with(
+            hash_including(admin_policy_object_id: apo.id,
+                           source_id: 'sul:SOMETHING-www.example.org')
+          )
         expect(item.rightsMetadata.content).to be_equivalent_to apo.defaultObjectRights.content
         expect(EventFactory).to have_received(:create)
       end
 
-      context 'when source_id has one colon' do
```
jcoyne: Moved here: https://github.com/sul-dlss/dor-services/pull/685

PR #651, opened 2020-02-10
Allow registration requests to specify an optional value for abstract

### LINKED ISSUE #632: Allow registration to accept abstract when metadata_source is label.
That will allow argo to get rid of this nonsense:
https://github.com/sul-dlss/argo/blob/master/app/forms/collection_form.rb#L29

mjgiarlo:
Fixes #632

## Why was this change made?

To allow for upstream apps to pass abstracts to dor-services-app instead of doing this directly in DOR (in support of decoupling from Fedora).

no

### DIFF
--- a/app/models/registration_request.rb
+++ b/app/models/registration_request.rb
@@ -11,6 +11,7 @@ class RegistrationRequest
   # @option params [String] :metadata_source
   # @option params [String] :rights
   # @option params [String] :collection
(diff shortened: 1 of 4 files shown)


mjgiarlo:
@jcoyne :speech_balloon: 
> Can you update the openapi.yml too?

Thanks for the reminder! Ready for another review.

PR #656, opened 2020-02-11
Add endpoint for displaying release tags

mjgiarlo:
Connects to sul-dlss/google-books#283
Connects to sul-dlss/dor_indexing_app#336

## Why was this change made?

Includes:

* Extract release tags action from objects controller into its own controller
* Add new `#show` action to `ReleaseTagsController`
* Use Committee for validating the `ReleaseTagsController#create` action
* Favor Rails built-in params handling to custom (outdated) handling
* Consolidate release tags-related functionality into `ReleaseTags` service
* Update OpenAPI specification

Yes.

### DIFF
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -176,7 +176,7 @@ module Dor
     private
 
     def released_for
-      Dor::ReleaseTags::IdentityMetadata.for(@druid_obj).released_for({})
+      ::ReleaseTags.for(item: @druid_obj)
     end
 
     def dor_items_for_constituents
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -76,25 +76,6 @@ class ObjectsController < ApplicationController
     proxy_faraday_response(response)
   end
 
-  # You can post a release tag as JSON in the body to add a release tag to an item.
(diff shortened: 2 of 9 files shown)


app/models/dor/update_marc_record_service.rb
```diff
@@ -176,7 +176,7 @@ def released_to_searchworks?
     private
 
     def released_for
-      Dor::ReleaseTags::IdentityMetadata.for(@druid_obj).released_for({})
+      ::ReleaseTags.for(item: @druid_obj)
```
ndushay: teaching opportunity:  what does `::` at the beginning mean?

mjgiarlo: Ah! Namespace resolution operator. Without it, Ruby looks for `Dor::ReleaseTags` because the containing module is `Dor`. Prefixing a constant with `::` lets Ruby know to search in the global namespace.

ndushay: i believe i've asked that before, and will probably ask again.  thanks for the larnin'.   

I will file this under "precious ruby-ism" ... weird syntax that means something in Ruby that a non-ruby developer wouldn't know.

mjgiarlo: oh, yes, totally! :sparkles:

PR #657, opened 2020-02-11
Adds support to setting embargo when registering.

justinlittman:
refs https://github.com/sul-dlss/sdr-api/issues/122

## Why was this change made?
Allow an embargo to be set when creating an object.

Yes.


jcoyne:
@justinlittman conflicts

jcoyne:
Is this still necessary?   This is no longer the path that google books take, right?

justinlittman:
Duh. Good catch.

PR #658, opened 2020-02-11
Allow creating an object with a Cocina model

jcoyne:
## Why was this change made?

So that we can support creating with more values.

yes


Gemfile.lock
```diff
@@ -1,3 +1,11 @@
+PATH
+  remote: ../cocina-models
```
mjgiarlo: Does this file need to be updated? This looks like cruft from earlier when you might have been pointing at a local path?

app/services/cocina/object_creator.rb
```diff
@@ -0,0 +1,134 @@
...
+    def create(params)
+      obj = case params[:type]
+            when *Cocina::Models::DRO::TYPES
+              build_dro(params)
+            when *Cocina::Models::Collection::TYPES
+              build_collection(params)
+            when *Cocina::Models::AdminPolicy::TYPES
+              build_apo(params)
+            else
+              raise "Unknown type #{params[:type]}"
```
mjgiarlo: It looks like adding coverage for this line would make the build happier.

app/services/cocina/object_creator.rb
```diff
@@ -0,0 +1,134 @@
...
+            end
+      if validate(obj)
+        af_model = create_from_model(obj)
+
+        # This will rebuild the cocina model from fedora, which shows we are only returning persisted data
+        return Mapper.build(af_model)
+      end
+
+      raise 'not valid'
```
mjgiarlo: It looks like adding coverage for this line would make the build happier.

app/services/cocina/object_creator.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    def create_from_model(obj)
+      af_object = if obj.admin_policy?
+                    create_apo(obj)
+                  elsif obj.dro?
+                    create_dro(obj)
+                  elsif obj.collection?
+                    create_collection(obj)
+                  else
+                    raise "unsupported type #{obj.type}"
```
mjgiarlo: It looks like adding coverage for this line would make the build happier.

app/services/cocina/object_creator.rb
```diff
@@ -0,0 +1,134 @@
...
+    end
+
+    def catkey_for(obj)
+      obj.identification.catalogLinks&.find { |l| l.catalog == 'symphony' }&.catalogRecordId
+    end
+
+    def content_type_tag(type)
+      tag = case type
+            when Cocina::Models::Vocab.image
+              'Image'
```
mjgiarlo: It looks like only this branch is tested.

openapi.yml
```diff
@@ -700,50 +700,11 @@ paths:
...
-                  type: array
-                  items:
-                    type: string
-                    enum:
-                      - descMetadata
-                  description: If 'descMetadata' is provided, then record is seeded from symphony
-                  example: ['descMetadata']
-              required:
-                - admin_policy
+              oneOf:
```
mjgiarlo: Nice change!

openapi.yml
```diff
@@ -871,6 +832,8 @@ components:
         - administrative
         - identification
         - structural
+    Description:
+      type: object
```
mjgiarlo: Does this object require any structure or have any required props?

app/services/cocina/object_creator.rb
jcoyne: @mjgiarlo  Is exhaustive code coverage required? I don't think those would be useful tests.

openapi.yml
jcoyne: Not until @arcadiafalcone tells us what it's going to look like.

app/services/cocina/object_creator.rb
mjgiarlo: No, not required. These branches are all simple enough, and in fact... testing these exhaustively will increase the drag when refactoring these later (if we want to replace the conditionals with polymorphism, for example).

PR #662, opened 2020-02-13
Refactor object create to use request models

jcoyne:
## Why was this change made?
Using the appropriate models.

n/a

### DIFF
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -3,11 +3,6 @@
 module Cocina
   # Given a Cocina model, create an ActiveFedora model.
   class ObjectCreator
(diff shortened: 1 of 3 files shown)


app/services/cocina/object_creator.rb
```diff
@@ -3,10 +3,6 @@
 module Cocina
   # Given a Cocina model, create an ActiveFedora model.
   class ObjectCreator
-    # We don't persist this druid, but it passes the validation for constructing a model.
-    # This allows us to validate the model before minting a pid
-    DUMMY_DRUID = 'druid:ab123cd4567'
-
     # TODO: We shouldn't need to provide a bogus externalIdentifier
```
justinlittman: Cruft.

PR #666, opened 2020-02-13
Adds Embargo Service and uses when doing a cocina deposit.

### LINKED ISSUE #659: Add embargo support to new cocina deposit
blocked by #658

justinlittman:
closes #659

## Why was this change made?
So that embargo can be set when doing a cocina deposit.

No.

### DIFF
--- a/app/services/cocina/mapper.rb
+++ b/app/services/cocina/mapper.rb
@@ -37,11 +37,9 @@ module Cocina
         label: item.label,
         version: item.current_version,
         administrative: build_administrative,
-        description: build_descriptive,
         identification: build_identification,
         access: build_access
       }.tap do |props|
-        props[:access][:embargo] = { releaseDate: item.embargoMetadata.release_date.iso8601 } if item.embargoMetadata.release_date
         props[:identification][:catalogLinks] = [{ catalog: 'symphony', catalogRecordId: item.catkey }] if item.catkey
 
         unless item.is_a?(Dor::Etd) || item.contentMetadata.new?
@@ -49,6 +47,9 @@ module Cocina
             contains: build_filesets(item.contentMetadata, version: item.current_version, id: item.pid)
           }
         end
+
+        description = build_descriptive
+        props[:description] = description unless description.nil?
       end
     end
 
@@ -64,6 +65,8 @@ module Cocina
         access: build_access
       }.tap do |props|
         props[:identification][:catalogLinks] = [{ catalog: 'symphony', catalogRecordId: item.catkey }] if item.catkey
+        description = build_descriptive
+        props[:description] = description unless description.nil?
       end
     end
 
@@ -73,9 +76,11 @@ module Cocina
         type: Cocina::Models::Vocab.admin_policy,
         label: item.label,
         version: item.current_version,
-        administrative: build_apo_administrative,
-        description: build_descriptive
-      }
+        administrative: build_apo_administrative
+      }.tap do |props|
+        description = build_descriptive
+        props[:description] = description unless description.nil?
+      end
     end
 
     private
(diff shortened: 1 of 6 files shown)


app/services/cocina/object_creator.rb
```diff
@@ -77,6 +77,7 @@ def create_dro(obj)
                     label: obj.label).tap do |item|
         item.descMetadata.mods_title = obj.description.title.first.titleFull
         item.identityMetadata.tag = content_type_tag(obj.type)
+        EmbargoService(item: item, release_date: obj.access.embargo.releaseDate, access: obj.access.embargo.access) if obj.access.embargo
```
mjgiarlo: Should this be `EmbargoService.embargo(...)`?

mjgiarlo: Related: I wonder if this class needs more/better test coverage. Assuming :point_up: is the case, I'd expect the test suite to bark about this.

justinlittman: There's no test for object_creator. https://github.com/sul-dlss/dor-services-app/issues/667

app/services/cocina/mapper.rb
```diff
@@ -176,6 +172,31 @@ def build_release_tags
...
+      {}.tap do |access|
+        embargo = build_embargo
+        access[:embargo] = embargo unless embargo.empty?
+      end
+    end
+
+    def build_embargo
+      return {} unless item.embargoMetadata.release_date
+
+      {}.tap do |embargo|
```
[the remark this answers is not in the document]
justinlittman: Fixed.

app/services/cocina/mapper.rb
```diff
@@ -114,10 +115,9 @@ def build_identification
     end
 
     def build_descriptive
-      case item
-      when Dor::Etd
+      if item == Dor::Etd
         { title: [{ primary: true, titleFull: item.properties.title.first }] }
-      else
+      elsif item.full_title
```
jcoyne: I believe that all items (that are not Etds) should have a full_title.

app/services/cocina/mapper.rb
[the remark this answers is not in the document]
jcoyne: doesn't appear to be.

app/services/cocina/mapper.rb
justinlittman: Descriptive isn't required.

jcoyne: I'm curious as to why you needed this change in this PR.   While descriptive is not currently required, I think we want to move to requiring it.

justinlittman:
```
  1) Cocina::Mapper when item is a Dor::Etd builds the object
     Failure/Error: klass.new(props)
     
     Dry::Struct::Error:
       [Cocina::Models::Description::Title.new] nil (NilClass) has invalid type for :titleFull violates constraints (type?(String, nil) failed)
     # ./app/services/cocina/mapper.rb:29:in `build'
     # ./app/services/cocina/mapper.rb:11:in `build'
     # ./spec/services/cocina/mapper_spec.rb:6:in `block (2 levels) in <top (required)>'
     # ./spec/services/cocina/mapper_spec.rb:127:in `block (3 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # Dry::Types::ConstraintError:
     #   nil violates constraints (type?(String, nil) failed)
     #   ./app/services/cocina/mapper.rb:29:in `build'
```

jcoyne: I'm not sure why the ETD isn't on this branch: https://github.com/sul-dlss/dor-services-app/pull/666/files#diff-c03d1fa455324b14d721e59ca497ab9eR127

justinlittman: Because I was testing object equality, not instance of a class. Fixed.

PR #671, opened 2020-02-19
Add an endpoint that can replace all extant uses of dor-fetcher

jcoyne:
## Why was this change made?

This allows consumers to determine the members of a collection. Adding this will allow us to close down the dor-fetcher-service and retire the dor-fetcher client.

yes

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -427,6 +427,41 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+  '/v1/objects/{object_id}/members':
+    get:
+      tags:
+        - objects
+      summary: List the members of this collection
+      description: ''
+      operationId: 'members#index'
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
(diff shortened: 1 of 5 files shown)


openapi.yml
```diff
@@ -427,6 +427,23 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+  '/v1/objects/{object_id}/query/members':
```
justinlittman: Why `/v1/objects/{object_id}/query/members` instead of `/v1/objects/{object_id}/members`?

openapi.yml
```diff
@@ -427,6 +427,23 @@ paths:
...
+    get:
+      tags:
+        - objects
+      summary: List the members that this collection has
```
justinlittman: List the members of this collection

openapi.yml
jcoyne: I was looking at paralleling the collections method in that controller already.  I think your option is valid too.

PR #685, opened 2020-02-21
Map checksums to cocina model

### LINKED ISSUE #680: Cocina mapper should support hasMessageDigests for files

jcoyne:
## Why was this change made?
Fixes #680

n/a

### DIFF
--- a/app/services/cocina/dro_structural_builder.rb
+++ b/app/services/cocina/dro_structural_builder.rb
@@ -45,11 +45,19 @@ module Cocina
 
     def build_files(file_nodes, version:, parent_id:)
       file_nodes.map do |node|
+        md5 = node.xpath('checksum[@type="MD5"]').text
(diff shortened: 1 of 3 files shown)


app/services/cocina/dro_structural_builder.rb
```diff
@@ -45,11 +45,19 @@ def build_filesets(content_metadata_ds, version:, id:)
...
-          label: node['id'],
-          version: version
+          {
+            externalIdentifier: "#{parent_id}/#{node['id']}",
+            type: Cocina::Models::Vocab.file,
+            label: node['id'],
+            version: version,
+            hasMessageDigests: []
+          }.tap do |attrs|
+            attrs[:hasMessageDigests] << { type: 'sha1', digest: sha1 } if sha1
```
justinlittman: Really don't like how this could allow multiple sha1s (with and without different digests), but that is a cocina modeling issue.

PR #699, opened 2020-02-25
Returns 502 from symphony error. Stop logging to HB.

### LINKED ISSUE #694: Handle 500 from Symphony in Object#create
This happens consistently and causes the error to ripple down to google-books.

At the very least, should not HB alert.

See:
* https://app.honeybadger.io/projects/50568/faults/59538379
* https://app.honeybadger.io/projects/67990/faults/60731529

justinlittman:
closes #694

## Why was this change made?
So that a correct error was returned when Symphony failed. And stop spamming HB.

Yes

### DIFF
--- a/app/models/symphony_reader.rb
+++ b/app/models/symphony_reader.rb
@@ -52,7 +52,6 @@ class SymphonyReader
       errmsg = "Record not found in Symphony: #{@catkey}"
     else
       errmsg = "Got HTTP Status-Code #{resp.status} retrieving #{@catkey} from Symphony: #{resp.body}"
-      Honeybadger.notify(errmsg)
     end
 
     raise ResponseError, errmsg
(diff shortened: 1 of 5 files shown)


spec/models/symphony_reader_spec.rb
```diff
@@ -94,30 +94,6 @@
...
-              }
-            ]
-          }
-        end
-
-        before do
-          stub_request(:get, format(Settings.catalog.symphony.json_url, catkey: catkey)).to_return(status: 403, body: err_body.to_json)
-        end
-
-        it 'raises ResponseError and notifies Honeybadger' do
```
ndushay: doesn't it still raise a ResponseError?

app/models/symphony_reader.rb
```diff
@@ -52,7 +52,6 @@ def symphony_response
       errmsg = "Record not found in Symphony: #{@catkey}"
     else
       errmsg = "Got HTTP Status-Code #{resp.status} retrieving #{@catkey} from Symphony: #{resp.body}"
-      Honeybadger.notify(errmsg)
     end
 
     raise ResponseError, errmsg
```
ndushay: does it not raise a ResponseError here?

spec/models/symphony_reader_spec.rb
justinlittman: You're right. I was excessive it my pruning of this test. Restored.

Issue #702, opened 2020-02-26
[ÜBER] Create an endpoint for assembly to support reaccessioning


jcoyne:
This can be used by pre-assembly/lyberservices-scripts and Goobi (after it supports token header)

Fixes sul-dlss/argo#1781

Steps to take, in order:

- [x] update dor-services-app to add a new endpoint (sul-dlss/dor-services-app#703)
- [x] update dor-services-client to expose this endpoint (sul-dlss/dor-services-client#147)
- [x] cut new release of dor-services-client
- [x] update lybservices-scripts to use this new endpoint via dor-services-client (sul-dlss/lyberservices-scripts#59)
- [x] update pre-assembly to use this new endpoint via dor-services-client (sul-dlss/pre-assembly#622 (in a branch now, ready for testing in #739)
- [x] test new object and existing object accessioning with both pre-assembly ~~and lybservices-scripts~~
- [x] coordinate with Goobi to get their changes done (#745 and sul-dlss/sul-goobi#258)

Background, relevant code from pre-assembly:
```ruby
      create_new_version if openable?
      initialize_assembly_workflow

    # Call web service to add assemblyWF to the object in DOR.
    def initialize_assembly_workflow
      workflow_client.create_workflow_by_name(druid.druid, 'assemblyWF', version: current_object_version)
    end

    def workflow_client
      logger = Logger.new(Settings.workflow.logfile, Settings.workflow.shift_age)
      Dor::Workflow::Client.new(url: Settings.workflow_url, logger: logger, timeout: Settings.workflow.timeout)
    end

    def openable?
      version_client.openable?
    end

    def object_client
      @object_client ||= Dor::Services::Client.object(druid.druid)
    end

    def version_client
      object_client.version
    end

    # When reaccessioning, we need to first open and close a version without kicking off accessionWF
    def create_new_version
      version_client.open(
        significance: 'major',
        description: 'pre-assembly re-accession',
        opening_user_name: bundle.bundle_context.user.sunet_id
      )
      version_client.close(start_accession: false)
    end
```

PR #703, opened 2020-02-26
implement endpoint for starting accessioning

peetucket:
## Why was this change made?

part of #702 to allow for reaccessioning via Goobi
when done, return to the checklist in #702 

yes

If so, please confirm:

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -50,6 +50,28 @@ class ObjectsController < ApplicationController
     render json: Cocina::Mapper.build(@item)
   end
 
+  # Initialize specified workflow (assemblyWF by default), and also version if needed
+  # called by pre-assembly, goobi and lybservices-scripts to kick off accessioning for a new or existing object
+  # params can optionally include the information needed to open a new version as well as the workflow to start (which defaults to assemblyWF)
+  # @option opts [Boolean] :assume_accessioned If true, does not check whether object has been accessioned.
+  # @option opts [String] :significance set significance (major/minor/patch) of version change
+  # @option opts [String] :description set description of version change
+  # @option opts [String] :opening_user_name add opening username to the events datastream
+  # @option opts [String] :workflow the workflow to start (defaults to 'assemblyWF')
+  def start_accession
+    params[:workflow] ||= default_start_accession_workflow # default if not specified
+
+    # if this is an existing versionable object, open and close it without starting accessioning
+    if VersionService.can_open?(@item, params)
+      VersionService.open(params)
+      VersionService.close(params.merge(start_accession: false))
+    end
+
+    # initialize workflow
+    workflow_client.create_workflow_by_name(@item.pid, params[:workflow], version: @item.current_version)
+    head :created
+  end
+
   # called from Argo, the accessionWF and from the releaseWF.
   # Takes an optional 'workflow' argument, which will call back to
   # the 'publish-complete' step of that workflow if provided
@@ -86,12 +108,20 @@ class ObjectsController < ApplicationController
 
   private
 
+  def workflow_client
+    WorkflowClientFactory.build
+  end
+
   def proxy_faraday_response(response)
     render status: response.status, content_type: response.headers['Content-Type'], body: response.body
   end
 
   def create_params
     params.except(:action, :controller).to_unsafe_h.merge(body_params)
+  end
+
+  def default_start_accession_workflow
+    'assemblyWF'
   end
 
   def body_params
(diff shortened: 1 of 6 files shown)


app/controllers/objects_controller.rb
```diff
@@ -50,6 +50,28 @@ def show
     render json: Cocina::Mapper.build(@item)
   end
 
+  # Initialize specified workflow (assemblyWF by default), and also version if needed
+  # called by pre-assembly, goobi and lybservices-scripts to kick off accessioning for a new or existing object
+  # params can optionally include the information needed to open a new version as well as the workflow to start (which defaults to assemblyWF)
+  # @option opts [Boolean] :assume_accessioned If true, does not check whether object has been accessioned.
```
jcoyne: This method doesn't have any options.

app/controllers/objects_controller.rb
```diff
@@ -50,6 +50,28 @@ def show
...
+  # @option opts [String] :significance set significance (major/minor/patch) of version change
+  # @option opts [String] :description set description of version change
+  # @option opts [String] :opening_user_name add opening username to the events datastream
+  # @option opts [String] :workflow the workflow to start (defaults to 'assemblyWF')
+  def start_accession
+    params[:workflow] ||= default_start_accession_workflow # default if not specified
```
jcoyne: can this be `workflow = params[:workflow] || default_start_accession_workflow`?

config/routes.rb
```diff
@@ -27,6 +27,7 @@
         post 'preserve'
         post 'update_marc_record'
         post 'notify_goobi'
+        post 'start_accession'
```
jcoyne: Not a big deal, but how about just `accession`?

openapi.yml
```diff
@@ -175,6 +175,32 @@ paths:
...
+    post:
+      tags:
+        - objects
+      summary: Start accessioning or re-accession an object to DOR
+      description: ''
+      operationId: 'objects#start_accession'
+      responses:
+        '201':
+          description: Accessiong/re-accessioning started
+      parameters:
```
jcoyne: Are significance, description, opening_user_name missing?

openapi.yml
```diff
@@ -175,6 +175,32 @@ paths:
...
+        - name: id
+          in: path
+          description: ID of object
+          required: true
+          schema:
+            $ref: '#/components/schemas/Druid'
+        - name: workflow
+          in: query
+          description: which workflow should be started
```
jcoyne: add a note about the default.

app/controllers/objects_controller.rb
ndushay: yeah, maybe reword indicated they are params passed to the controller?

app/controllers/objects_controller.rb
```diff
@@ -94,6 +120,10 @@ def create_params
     params.except(:action, :controller).to_unsafe_h.merge(body_params)
   end
 
+  def default_start_accession_workflow
```
ndushay: I'm 'meh' on calling this out as a method;  as a former java-head, i would define it as a string near the method that uses it, or just use it directly in the `||=` statement.  But leaving it as is works too.

spec/requests/start_accession_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      allow(VersionService).to receive(:can_open?).and_return(false)
+    end
+
+    it 'does not open or close a version and starts default workflow' do
+      post "/v1/objects/#{druid}/start_accession",
+           headers: { 'Authorization' => "Bearer #{jwt}" }
+      expect(response).to be_successful
+      expect(workflow_client).to have_received(:create_workflow_by_name).with(object.pid, default_start_accession_workflow, version: '1')
+      expect(VersionService).not_to have_received(:open)
+      expect(VersionService).not_to have_received(:close)
```
ndushay: i would probably split the "what it does re: version" into a separate test and ignore it in this test, but meh - no biggie.

app/controllers/objects_controller.rb
peetucket: will update comment to indicate those params are passed through to versioning (they are useful so an app like pre-assembly can specify then).

app/controllers/objects_controller.rb
peetucket: sounds good

config/routes.rb
peetucket: sure

openapi.yml
peetucket: done

openapi.yml
peetucket: added

PR #704, opened 2020-02-26
Add endpoint for showing and creating administrative tags

### LINKED ISSUE #688: Create an endpoint for creating user tags
In the future, we don't want to persist user tags in DOR, but we still need a way to create them.
Add a new endpoint for this purpose. For now this will still write into identityMetadata, but will be changed to a database in the future.

### LINKED ISSUE #689: Create an endpoint for reading user tags
In the future user tags will not be stored in DOR, so we will create an API for fetching them.  So given an object druid, return the list of tags for that object. Currently this will read from identityMetadata, but in the future this will be changed to read from a db.

mjgiarlo:
Fixes #688
Fixes #689

Also includes:
* Run linter before test suite (fail fast)

## Why was this change made?

To begin decoupling administrative tags from the current Fedora implementation which winds up preserving tags (we don't want this). This sets up the ability to persist tags somewhere other than Fedora, such as a database.

Yes.

No. It introduces a brand new endpoint.


spec/services/administrative_tags_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  let(:tags) { ['Foo : Bar', 'Bar : Baz : Quux'] }
+
+  describe '.for' do
+    it 'returns the array of administrative tags as XML nodes' do
+      expect(described_class.for(item: item)).to eq(tags)
+    end
+  end
+
+  describe '.create' do
+    it 'creates a plain directory in the workspace when passed no source directory' do
```
jcoyne: huh?

app/services/administrative_tags.rb
```diff
@@ -0,0 +1,21 @@
...
+    item.identityMetadata.ng_xml.search('//tag').map(&:content)
+  end
+
+  # Add one or more administrative tags for an item
+  #
+  # @param item [Dor::Item]  the item to create administrative tag(s) for
+  # @param tags [Array<String>] a non-empty array of tags (strings)
+  # @return [Array<Nokogiri::XML::Node>]
+  def self.create(item:, tags:)
+    tags.map { |tag| Dor::TagService.add(item, tag) }
```
jcoyne: Is the plan here to make a new persistence for tags in another PR?

jcoyne: Yeah, that's exactly what I said in the ticket. 🙄

spec/services/administrative_tags_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+RSpec.describe AdministrativeTags do
+  before do
+    described_class.create(item: item, tags: tags)
+  end
+
+  let(:item) { Dor::Item.new(pid: 'druid:aa123bb7890') }
+  let(:tags) { ['Foo : Bar', 'Bar : Baz : Quux'] }
+
+  describe '.for' do
+    it 'returns the array of administrative tags as XML nodes' do
```
jcoyne: Strings!

mjgiarlo: @jcoyne PLEASE SIR MAY I HAVE ANOTHER

PR #705, opened 2020-02-27
Sets default rights for an item.

### LINKED ISSUE #700: ObjectCreator should add the default rights from the APO
Ref https://github.com/sul-dlss/google-books/issues/267#issuecomment-591504724

justinlittman:
closes #700

## Why was this change made?
So that rights are set when creating an item.

No.

No.

If so, please confirm:

### DIFF
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -10,16 +10,14 @@ module Cocina
     def create(params, event_factory:)
       obj = Cocina::Models.build_request(params)
 
-      if validate(obj)
-        af_model = create_from_model(obj)
+      # Validate will raise an error if not valid.
+      validate(obj)
+      af_model = create_from_model(obj)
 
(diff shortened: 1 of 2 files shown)


app/services/cocina/object_creator.rb
```diff
@@ -125,5 +124,16 @@ def content_type_tag(type, direction)
             end
       "Process : Content Type : #{tag}"
     end
+
+    def change_access(item, access)
+      # 'world', 'stanford', 'location-based', 'citation-only', 'dark'
+      raise 'location-based access not implemented' if access == 'location-based'
+
+      # See https://github.com/sul-dlss/dor-services/blob/master/lib/dor/datastreams/rights_metadata_ds.rb
+      rights_type = access == 'citation-only' ? 'none' : access
```
jcoyne: Is this right?  For citation-only, shouldn't the discover access be "world"?

PR #709, opened 2020-02-27
Store administrative tags in the database

### LINKED ISSUE #679: [ÜBER] Create a data store for user tags
Because user tags should not end up in MOABs or be on PURL.

- [x] [Write tags into this data store (vs. identityMetadata)](https://github.com/sul-dlss/dor-services-app/pull/709)
- [x] [Support endpoint in dor-services-client](https://github.com/sul-dlss/dor-services-client/pull/154)
- [x] [Index the tags from the new data store](https://github.com/sul-dlss/dor_indexing_app/pull/358)
- [x] [Create/update/destroy/list tags via dor-services API in Argo](https://github.com/sul-dlss/argo/pull/1857)
- [x] [Write integration test testing interaction between Argo, DSA, and DIA](https://github.com/sul-dlss/infrastructure-integration-test/pull/34)
- [x] [Point Hydrus at the new API](https://github.com/sul-dlss/hydrus/pull/410)
- [x] [Point gis-robot-suite at the new API](https://github.com/sul-dlss/gis-robot-suite/pull/261)
- [x] [Remove dead related code from dor-services gem](https://github.com/sul-dlss/dor-services/pull/696)
- [x] Migrate tags from Fedora 3 to the new store (see [question below in the comments](https://github.com/sul-dlss/dor-services-app/issues/679#issuecomment-594098784))

mjgiarlo:
Connects to #679

## Why was this change made?

This change allows reading administrative tags from the database OR Fedora, and only writes them TO the database, which will give us some time to pull off the migration of tags from Fedora to the database, making a gradual transition.

Yes.

I don't think it does. Tested in stage and in this integration test: https://github.com/sul-dlss/infrastructure-integration-test/pull/34

### DIFF
--- a/app/controllers/administrative_tags_controller.rb
+++ b/app/controllers/administrative_tags_controller.rb
@@ -14,7 +14,32 @@ class AdministrativeTagsController < ApplicationController
   end
 
   def create
-    AdministrativeTags.create(item: @item, tags: params.require(:administrative_tags))
+    AdministrativeTags.create(item: @item,
+                              tags: params.require(:administrative_tags),
+                              replace: params[:replace])
+  rescue ActiveRecord::RecordInvalid => e
+    render status: :conflict, plain: e.message
+  else
     head :created
+  end
+
+  def update
+    AdministrativeTags.update(item: @item,
+                              current: CGI.unescape(params[:id]),
+                              new: params.require(:administrative_tag))
+  rescue ActiveRecord::RecordNotFound => e
+    render status: :not_found, plain: e.message
+  rescue ActiveRecord::RecordInvalid => e
+    render status: :conflict, plain: e.message
+  else
+    head :no_content
+  end
+
+  def destroy
+    AdministrativeTags.destroy(item: @item, tag: CGI.unescape(params[:id]))
+  rescue ActiveRecord::RecordNotFound => e
+    render status: :not_found, plain: e.message
+  else
+    head :no_content
   end
 end
--- a/openapi.yml
+++ b/openapi.yml
@@ -342,7 +342,7 @@ paths:
               schema:
                 type: array
                 items:
-                  type: string
+                  $ref: '#/components/schemas/AdministrativeTag'
         '404':
           description: Object not found
       parameters:
@@ -355,7 +355,7 @@ paths:
     post:
       tags:
         - objects
-      summary: Create one or more administrative tags
+      summary: Create (or replace) one or more administrative tags
       description: ''
       operationId: 'administrative_tags#create'
       responses:
@@ -363,6 +363,10 @@ paths:
           description: Created
         '400':
           description: Bad request
+        '404':
+          description: Object not found
+        '409':
+          description: Request would result in a duplicate tag
       parameters:
         - name: object_id
           in: path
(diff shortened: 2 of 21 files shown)


spec/rails_helper.rb
```diff
@@ -56,8 +58,16 @@
   config.include AuthHelper
 
   config.before :suite do
+    DatabaseCleaner.strategy = :transaction
+    DatabaseCleaner.clean_with(:truncation)
```
jcoyne: Why do we need this?  Aren't the transaction's that rails does by default sufficient?

mjgiarlo: @jcoyne why do we need DBCleaner at all? Or why do we need line 62?

jcoyne: Why DBCleaner at all.

mjgiarlo: @jcoyne Because without it, these [FactoryBot creates](https://github.com/sul-dlss/dor-services-app/pull/709/files#diff-feb40e3c3103b6ae222ee1b3da2ae088R11-R12) keep adding rows which causes [this test](https://github.com/sul-dlss/dor-services-app/pull/709/files#diff-feb40e3c3103b6ae222ee1b3da2ae088R17) to fail due to:

```
  1) AdministrativeTags.for with matching rows in the database returns administrative tags from the database
     Failure/Error: expect(described_class.for(item: item_with_db_tags)).to eq(['Foo : Bar', 'Bar : Baz : Quux'])
     
       expected: ["Foo : Bar", "Bar : Baz : Quux"]
            got: ["Foo : Bar", "Bar : Baz : Quux", "Foo : Bar", "Bar : Baz : Quux", "Foo : Bar", "Bar : Baz : Quux", "Foo : Bar", "Bar : Baz : Quux", "Foo : Bar", "Bar : Baz : Quux", "Foo : Bar", "Bar : Baz : Quux"]
     
       (compared using ==)
     # ./spec/services/administrative_tags_spec.rb:17:in `block (4 levels) in <top (required)>'
```

I'm open to a different solution if you'd prefer one.

ndushay:
@mjgiarlo this needs love.

mjgiarlo:
@ndushay :speech_balloon: 
> @mjgiarlo this needs love.

All the admin tag-related stuff is in-flight and nearing the top of my queue.

jcoyne:
@mjgiarlo this change is blocked by https://github.com/sul-dlss/dor_indexing_app/pull/358 ,right?

mjgiarlo:
@jcoyne :speech_balloon: 
> @mjgiarlo this change is blocked by [sul-dlss/dor_indexing_app#358](https://github.com/sul-dlss/dor_indexing_app/pull/358) ,right?

Yep.

app/controllers/administrative_tags_controller.rb
```diff
@@ -8,13 +8,35 @@ class AdministrativeTagsController < ApplicationController
     render status: :not_found, plain: e.message
   end
 
+  rescue_from(ActiveRecord::RecordNotFound) do |e|
```
jcoyne: I'm not clear what could trigger this.  Is it possible to wrap the action that does it with a rescue?

app/controllers/administrative_tags_controller.rb
```diff
@@ -8,13 +8,35 @@ class AdministrativeTagsController < ApplicationController
...
+    render status: :not_found, plain: e.message
+  end
+
+  rescue_from(ActiveRecord::RecordInvalid) do |e|
+    render status: :conflict, plain: e.message
```
jcoyne: Same here.

openapi.yml
```diff
@@ -379,11 +383,119 @@ paths:
...
+      description: ''
+      operationId: 'administrative_tags#update (PATCH)'
+      responses:
+        '204':
+          description: Updated
+        '400':
+          description: Bad request
+        '404':
+          description: Object or tag not found
+        '409':
```
jcoyne: Is it important for the client to know about this?  Can we just create the any tags that have been provided that don't already exist?

app/controllers/administrative_tags_controller.rb
mjgiarlo: @jcoyne This can be triggered by the `#create` and `#update` operations below, if passed a druid and tag combo that already exist in the database.

app/controllers/administrative_tags_controller.rb
mjgiarlo: @jcoyne This can be triggered by `#update` if it is asked to update a tag/druid combo that does not already exist.

openapi.yml
mjgiarlo: @jcoyne This is for an update operation for a single tag, when the client asks to update a tag to a druid/tag combo that already exists.

app/controllers/administrative_tags_controller.rb
jcoyne: What I'm saying is that this code doesn't make this readily apparent.  Can we move the rescue into that action, so that it's easier to see what might be calling this?

spec/requests/administrative_tags_spec.rb
```diff
@@ -99,5 +123,206 @@
                                     '/properties/administrative_tags [] contains fewer than min items"}]}')
       end
     end
+
+    context 'when one or more tags already exist' do
+      # NOTE: Yep, this is pretty gross. Faking ActiveModel::Errors here. Why is it gross?
```
jcoyne: Would it be possible to just create an existing record so that the request would cause a validation error?

openapi.yml
jcoyne: Okay, that makes sense.  You can't replace `Content type : video` with `Registered by : Mike` if `Registered by : Mike` already exists.

mjgiarlo: @jcoyne *exactamundo*!

spec/requests/administrative_tags_spec.rb
mjgiarlo: @jcoyne :speech_balloon: 
> Would it be possible to just create an existing record so that the request would cause a validation error?

Possible, yeah. I'd prefer not to make that change now if I can avoid it.

jcoyne: If the alternative is "pretty gross", why not now?

mjgiarlo: @jcoyne I'm reluctant to change the interaction behavior just because the test looks gross. At least there are a couple comments. That way if someone else has a brain-wave about how to do the testing smarter, the expectation is documented here.

app/controllers/administrative_tags_controller.rb
mjgiarlo: gotcha. done.

app/controllers/administrative_tags_controller.rb
mjgiarlo: Done

PR #712, opened 2020-02-28
image files populate presentation attribute in cocina-model

### LINKED ISSUE #692: Cocina mapper should support presentation element for files that are images.
currently height and width are in contentMetadata for images

ndushay:
## Why was this change made?

When a file has image height and width, they need to be mapped to the cocina model.

fixes #692 

n/a

DOES this need to be tested in stage?  There is already a ticket to add an integration test to that suite:  sul-dlss/infrastructure-integration-test#23

If so, please confirm:

### DIFF
--- a/app/services/cocina/dro_structural_builder.rb
+++ b/app/services/cocina/dro_structural_builder.rb
@@ -49,6 +49,9 @@ module Cocina
         # The old google books use these upcased versions. See https://argo.stanford.edu/view/druid:dd116zh0343
         md5 = node.xpath('checksum[@type="md5"]').text.presence || node.xpath('checksum[@type="MD5"]').text
         sha1 = node.xpath('checksum[@type="sha1"]').text.presence || node.xpath('checksum[@type="SHA-1"]').text
+        height = node.xpath('imageData/@height').text.presence
+        width = node.xpath('imageData/@width').text.presence
+
         Cocina::Models::File.new(
           {
             externalIdentifier: "#{parent_id}/#{node['id']}",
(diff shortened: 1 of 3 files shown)


app/services/cocina/dro_structural_builder.rb
```diff
@@ -57,8 +60,10 @@ def build_files(file_nodes, version:, parent_id:)
             size: node['size'],
             hasMimeType: node['mimetype'],
             version: version,
-            hasMessageDigests: []
+            hasMessageDigests: [],
+            presentation: {}
```
jcoyne: I think we should omit this if it doesn't have height or width.

ndushay: will do.  Thanks for the feedback

ndushay:
@jcoyne I believe I have fixed your concern.

Note that the code only creates presentation attribute when both height and width are present -- does it need to accommodate only one of those attributes being present?

spec/services/cocina/mapper_spec.rb
```diff
@@ -83,22 +83,20 @@
         expect(cocina_model.structural.contains.size).to eq 2
         folder1 = cocina_model.structural.contains.first
         expect(folder1.label).to eq 'Folder 1'
-        expect(folder1.structural.contains.first.label).to eq 'folder1PuSu/story1u.txt'
```
ndushay: this line is a duplicate

spec/services/cocina/mapper_spec.rb
```diff
@@ -83,22 +83,20 @@
...
 
         file1 = folder1.structural.contains.first
         expect(file1.label).to eq 'folder1PuSu/story1u.txt'
         expect(file1.size).to eq 7888
         expect(file1.hasMimeType).to eq 'text/plain'
         expect(file1.hasMessageDigests.first.digest).to eq '61dfac472b7904e1413e0cbf4de432bda2a97627'
         expect(file1.hasMessageDigests.first.type).to eq 'sha1'
-        true
```
ndushay: this true is unnec.

PR #717, opened 2020-03-02
Pin iso-639 to 0.2.x

aaron-collier:
## Why was this change made?

Pin iso-639 until we upgrade to ruby 2.6

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,8 @@ gem 'dry-schema', '~> 1.4'
 gem 'faraday', '~> 1.0'
 gem 'faraday_middleware', '~> 1.0.0.rc1' # dependency of dor-workflow-client. remove when release > 0.14.0
 gem 'honeybadger'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -29,6 +29,8 @@ gem 'dry-schema', '~> 1.4'
 gem 'faraday', '~> 1.0'
 gem 'faraday_middleware', '~> 1.0.0.rc1' # dependency of dor-workflow-client. remove when release > 0.14.0
 gem 'honeybadger'
+# iso-639 0.3.0 isn't compatible with ruby 2.5.  This declaration can be dropped when we upgrade to 2.6
```
ndushay: really helpful to include link about breakage to tell us when we can remove pinning.
```
# see https://github.com/alphabetum/iso-639/issues/12
# iso-639 is used by dor-services gem via stanford-mods gem
```

PR #719, opened 2020-03-02
Map copyright and use and reproduction statement to Cocina model

jcoyne:
## Why was this change made?

So that GB can add these properties.

ref sul-dlss/google-books#331
ref sul-dlss/google-books#333

yes

If so, please confirm:

### DIFF
--- a/app/services/cocina/access_builder.rb
+++ b/app/services/cocina/access_builder.rb
@@ -15,6 +15,8 @@ module Cocina
       { access: access_rights }.tap do |access|
         embargo = build_embargo
         access[:embargo] = embargo unless embargo.empty?
+        access[:useAndReproductionStatement] = item.rightsMetadata.use_statement.first if item.rightsMetadata.use_statement.first.present?
+        access[:copyright] = item.rightsMetadata.copyright.first if item.rightsMetadata.copyright.first.present?
       end
     end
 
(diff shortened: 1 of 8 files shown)


app/services/cocina/access_builder.rb
```diff
@@ -15,6 +15,8 @@ def build
       { access: access_rights }.tap do |access|
         embargo = build_embargo
         access[:embargo] = embargo unless embargo.empty?
+        access[:useAndReproductionStatement] = item.rightsMetadata.use_statement.first if item.rightsMetadata.use_statement.first.present?
+        access[:copyright] = item.rightsMetadata.copyright.first if item.rightsMetadata.use_statement.first.present?
```
mjgiarlo: `item.rightsMetadata.use_statement.first.present?` should be `item.rightsMetadata.copyright.first.present?` ?

mjgiarlo: Would be nice if this were caught by the test suite. Consider adding a test that covers this?

openapi.yml
```diff
@@ -855,6 +855,12 @@ components:
             - 'location-based'
             - 'citation-only'
             - 'dark'
+        copyright:
+          description: The human readable reuse and reproduction statement that applies
```
mjgiarlo: It looks like these descriptions should be swapped. Also, `s/reuse/use/`

PR #721, opened 2020-03-03
lower case openapi enum for versioning significance values

peetucket:
## Why was this change made?

openapi enums are case sensitive, and the significance should come in as lower case (and currently does from argo, and new accession method from dor-services-app)

yes

If so, please confirm:

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -210,9 +210,9 @@ paths:
           schema:
             type: string
             enum:
-              - Minor
-              - Major
-              - Admin
+              - minor
+              - major
+              - admin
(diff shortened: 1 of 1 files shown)


peetucket:
> I like the lowercase variants, but it would be better if we could deprecate the others and standardize on one set. Also great to show that both are compatible with the code.

I just looked, and argo uses lowered significance when you manually open a version.  I agree and think it is safe to enforce lower only in this new method (and it fits with my new PRs for pre-assembly and lyberservices-scripts).  adjusted the PR

PR #726, opened 2020-03-06
Fix for reaccession issues

peetucket:
## Why was this change made?

to fix a regression introduced in #703 

fixes sul-dlss/argo#1843

basically I inadvertently renamed one of the permitted params in the "close" method to match my new endpoint name, when it should have been left alone ... renaming it caused the existing parameter to be ignored, triggering accessionWF in all cases (even when you didn't want to be triggered with a close version)

n/a

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -61,7 +61,7 @@ class VersionsController < ApplicationController
     params.permit(
       :description,
       :significance,
-      :accession,
+      :start_accession,
       :user_name,
       :version_num
     ).to_h.symbolize_keys
--- a/openapi.yml
+++ b/openapi.yml
@@ -759,6 +759,13 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+        - name: start_accession
+          in: query
+          schema:
+            type: boolean
+            default: true
+          required: false
+          description: Indicates if accessionWF should be started after closing the version
   '/v1/objects/{object_id}/versions':
     post:
       tags:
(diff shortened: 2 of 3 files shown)


app/controllers/versions_controller.rb
```diff
@@ -61,7 +61,7 @@ def close_params
     params.permit(
       :description,
       :significance,
-      :accession,
+      :start_accession,
```
ndushay: d'oh!

jcoyne: Can you add this to the openapi.yml?  I think this would go here? https://github.com/sul-dlss/dor-services-app/blob/master/openapi.yml#L188

peetucket: good idea - added this parameter to the 'close' method (a slightly different spot than you mentioned) but the one that is relevant

openapi.yml
```diff
@@ -759,6 +759,11 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+        - name: start_accession
+          type: boolean
+          in: path
+          description: Indicates if accessionWF should be started after closing the version
+          default: true
```
jcoyne: I don't believe this is valid, should be in the schema property.

openapi.yml
```diff
@@ -759,6 +759,11 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+        - name: start_accession
+          type: boolean
```
jcoyne: I don't believe this is valid, should be in the schema property.

jcoyne: See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameterObject

ndushay:
can we make the openapi.yml update a separate ticket?  this is affecting production processing.

peetucket:
fixed openapi

PR #727, opened 2020-03-06
Allow collections to be set by the create endpoint

jcoyne:
And returned by the show endpoint.

## Why was this change made?

Fixes https://github.com/sul-dlss/google-books/issues/391

yes

yes

it will be tested on stage.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1,3 +1,4 @@
+
 openapi: 3.0.0
 info:
   description: Backend API for the Stanford digital repository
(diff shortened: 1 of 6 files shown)


openapi.yml
```diff
@@ -1366,6 +1366,8 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Sequence'
+        isMemberOf:
```
justinlittman: A description in the openapi would be appreciated, because I can't say it is intuitive to have collection membership in structural.

jcoyne: I don't know that there's a way to do that given that `$ref` can't have a sibling.
https://swagger.io/docs/specification/using-ref/#sibling

justinlittman: It can have a sibling, it will just be ignored. However, having a description will still help someone reading the openapi.

jcoyne: Okay

PR #731, opened 2020-03-10
Call Dor::EmbargoService directly

jcoyne:
## Why was this change made?

Rather than the indirection of Embargoable.  The embargoable methods will be deprecated in dor-services

n.a

no

### DIFF
--- a/app/services/embargo_release_service.rb
+++ b/app/services/embargo_release_service.rb
@@ -55,14 +55,81 @@ class EmbargoReleaseService
     Honeybadger.notify "Unable to release embargo for: #{druid}", backtrace: e.backtrace
   end
 
-  def self.release
+  def release_all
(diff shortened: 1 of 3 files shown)


mjgiarlo:
@sul-dlss/infrastructure-team In the future, when we move code around (which we ought to keep doing), I think we need to consider putting more testing in place. This PR had at least two bugs that went undiscovered for months:

1. `self.release`, the class method that was being invoked in the rake task, was changed to `release_all`, an instance method. As a result, the rake task was never succeeding.
2. `release_20_pct_vis_embargo` is an undefined method. should have been `release_20_pct_vis`.

Neither of these were caught by the test suite and perhaps even more problematically, no alerts have been sent, apparently. Just FYI as I chase down these ETD embargo-related bugs!

PR #761, opened 2020-03-27
[Draft] Maps descriptive metadata for ETDs to Cocina model

justinlittman:
## Why was this change made?
To map descriptive metadata for ETDs to Cocina model.

No.

So if/when to deploy this needs to be considered, as there is the possibility that this will encounter some ETDs that cause this to error.

### DIFF
--- a/app/services/cocina/etd_description_builder.rb
+++ b/app/services/cocina/etd_description_builder.rb
@@ -1,0 +1,689 @@
+# frozen_string_literal: true
+
+module Cocina
+  # builds the Description subschema for ETDs
+  class EtdDescriptionBuilder
+    def self.build(item)
+      new(item).build
+    end
+
+    def initialize(item)
+      @item = item
+    end
+
+    def build
+      {
+        title: build_title,
+        contributor: build_contributors,
+        event: [build_submission, build_conferral, build_publication, build_copyright],
+        form: build_form,
+        language: build_language,
+        note: build_notes,
+        identifier: build_identifier,
+        purl: build_purl,
+        url: build_url,
+        marcEncodedData: build_marc_encoded_data,
+        adminMetadata: build_admin_metadata
+      }
+    end
+
+    private
+
+    attr_reader :item
+
+    def properties
+      @properties ||= item.datastreams['properties']
+    end
+
+    def readers
+      @readers ||= Nokogiri::XML(item.datastreams['readers'].content)
+    end
+
+    def build_title
+      [{ value: item.title, status: 'primary' }]
+    end
+
+    def build_contributors
+      [].tap do |contributors|
+        contributors << build_author
+        contributors.concat(build_advisors)
+        contributors.concat(build_readers)
+      end
+    end
+
+    def build_author
+      build_person(item.name, item.suffix).tap do |person|
+        person[:status] = 'primary'
+        person[:role] = [
+          {
+            value: 'author',
+            code: 'aut',
+            uri: 'http://id.loc.gov/vocabulary/relators/aut',
+            source: {
+              code: 'marcrelator',
+              uri: 'http://id.loc.gov/vocabulary/relators/'
+            }
+          },
+          {
+            value: 'author',
+            uri: 'http://rdaregistry.info/Elements/a/P50195',
+            source: {
+              uri: 'http://www.rdaregistry.info/Elements/a/'
+            }
+          }
+        ]
+      end
+    end
+
+    def build_advisors
+      readers.search('//reader[readerrole ="Advisor" or readerrole ="Co-Adv" or readerrole = "Dissertation Co-Advisor" or readerrole = "Co-Adv"]').map do |reader_elem|
+        build_advisor(reader_elem)
+      end
+    end
+
+    def build_advisor(reader_elem)
+      build_person(reader_elem.at('name').text, reader_elem.at('suffix').text).tap do |advisor|
+        advisor[:role] = [
+          {
+            value: reader_elem.at('readerrole').text,
+            source: {
+              value: 'ETD reader roles'
+            }
+          },
+          {
+            value: 'degree supervisor',
+            code: 'dgs',
+            uri: 'http://id.loc.gov/vocabulary/relators/dgs',
+            source: {
+              code: 'marcrelator',
+              uri: 'http://id.loc.gov/vocabulary/relators/'
+            }
+          },
+          {
+            value: 'degree supervisor',
+            uri: 'http://rdaregistry.info/Elements/a/P50091',
+            source: {
+              uri: 'http://www.rdaregistry.info/Elements/a/'
+            }
+          }
+        ]
+      end
+    end
+
+    def build_readers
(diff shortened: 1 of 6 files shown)


mjgiarlo:
Ooooooh.

jcoyne:
This is great. At some point we're going to need to move this into hydra_etd, so that we can submit this in the SDR-API.

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+    def build_contributors
+      [].tap do |contributors|
+        contributors << build_author
+        contributors.concat(build_advisors)
+        contributors.concat(build_readers)
+      end
+    end
+
+    def build_author
+      # Author has structuredValue, other people do not.
```
arcadiafalcone: The structuredValue attribute applies to all agents where a name suffix is present, not just the author.

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+    end
+
+    def build_notes
+      [
+        {
+          value: item.abstract,
+          type: 'summary'
+        },
+        {
+          value: "Submitted to the #{item.schoolname}."
```
arcadiafalcone: This should be the either the department name or the school name depending on the value - see the logic here: https://github.com/sul-dlss/common-accessioning/blob/263a336aae19f995a53c9c28d2bf79a08a562a78/lib/robots/dor_repo/etd_submit/submit_marc.rb#L110

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+        type: 'copyright',
+        date: [
+          {
+            value: item.degreeconfyr,
+            encoding: %w[edtf w3cdtf marc iso8601]
+          }
+        ],
+        structuredValue: [
+          {
+            value: "©#{item.degreeconfyr}",
```
arcadiafalcone: This should be the year based on the submission date, not the conferral date.

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+        ]
+      }
+    end
+
+    def build_copyright
+      {
+        type: 'copyright',
+        date: [
+          {
+            value: item.degreeconfyr,
```
arcadiafalcone: This should be the year based on the submission date, not the conferral date.

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+        ]
+      }
+    end
+
+    def build_publication
+      {
+        type: 'publication',
+        date: [
+          {
+            value: item.degreeconfyr,
```
arcadiafalcone: This should be the current year.

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+            type: 'publication statement place',
+            standard: ['RDA']
+          },
+          {
+            value: '[Stanford University]',
+            type: 'publication statement publisher',
+            standard: ['RDA']
+          },
+          {
+            value: item.degreeconfyr,
```
arcadiafalcone: This should be the current year.

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+        {
+          value: 'm        d',
+          type: '006'
+        },
+        {
+          value: 'cr un',
+          type: '007'
+        },
+        {
+          value: "170607t#{item.degreeconfyr}#{item.degreeconfyr}cau     om    000 0 eng d",
```
arcadiafalcone: The first date should be the publication year (i.e., the current year) and the second date should be the copyright year (i.e., the year the thesis was submitted).

app/services/cocina/etd_description_builder.rb
justinlittman: Gotcha. Fixed.

app/services/cocina/etd_description_builder.rb
justinlittman: That logic does not align with the example. The output from that logic is "Submitted to the School of Business Administration", whereas the example has "Submitted to the School of Business". The descMetadata for the item has "Submitted to the Graduate School of Business".

app/services/cocina/etd_description_builder.rb
justinlittman: Fixed.

app/services/cocina/etd_description_builder.rb
justinlittman: Fixed.

app/services/cocina/etd_description_builder.rb
justinlittman: This code is being used for a real-time mapping between Fedora and Cocina. Thus, this would always be whatever the current year is at the time the mapping is run, e.g., if run now would be 2020. If run for the same item next year would be 2021.

In the future, will want to set this to the current year (i.e., when this is run once at ingest time).

Where should be get this value in the meantime (i.e., for mapping)? Possibly get from descMetadata < originInfo < dateIssued?

arcadiafalcone: Got it. In that case use the submission date.

app/services/cocina/etd_description_builder.rb
arcadiafalcone: This took some teasing out. What the logic should be is:
If the item.schoolname contains "Business", the note should be "Submitted to the Graduate School of Business."
If the item.schoolname contains "Law", the note should be "Submitted to the School of Law."
If the item.schoolname contains "Education", the note should be "Submitted to the Graduate School of Education."
In all other cases, the note should be "Submitted to the Department of <item.department>."
I'll update the example to match this.

app/services/cocina/etd_description_builder.rb
```diff
@@ -0,0 +1,663 @@
...
+
+    def readers
+      @readers ||= Nokogiri::XML(item.datastreams['readers'].content)
+    end
+
+    def build_title
+      [{ value: item.title, status: 'primary' }]
+    end
+
+    def build_contributors
```
arcadiafalcone: I just realized that the school/department should be one of the contributors. I'll update the examples to reflect this. The school vs. department logic is the same as that of the "Submitted to X" note as commented below.

app/services/cocina/etd_description_builder.rb
justinlittman: Fixed.

app/services/cocina/etd_description_builder.rb
justinlittman: Fixed.

app/services/cocina/etd_description_builder.rb
justinlittman: Fixed using submission date.

app/services/cocina/etd_description_builder.rb
justinlittman: Fixed.

PR #762, opened 2020-03-27
add ETD APO to seed objects

### LINKED ISSUE #756: Add ETD APO object to stage seed data (useful for integration tests)
See  https://github.com/sul-dlss/dor-services-app/blob/master/lib/tasks/cleaner.rake#L29

the ETD APO object is  https://argo-stage.stanford.edu/view/druid:bx911tp9024

Andrew imported it to stage by exporting the FOXML from prod and adding to stage.

ndushay:
## Why was this change made?

Closes #756

So infrastructure-integration-tests for ETDs have the correct APO object.

na

na


lib/tasks/seeds/druid:bx911tp9024.xml
```diff
@@ -0,0 +1,1062 @@
...
+            <format>MODS</format>
+            <source>symphony</source>
+          </descMetadata>
+          <relationships xmlns:fedora="info:fedora/fedora-system:def/relations-external#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
+            <fedora:isMemberOf rdf:resource="info:fedora/druid:sw909tc7852">
+            </fedora:isMemberOf>
+            <fedora:isMemberOfCollection rdf:resource="info:fedora/druid:sw909tc7852">
+            </fedora:isMemberOfCollection>
+          </relationships>
+          <accessioning>
```
jcoyne: woah, I've never seen an APO like this before.

ndushay: it's probably super old.

PR #765, opened 2020-03-31
Fixed sdrPreserve and shelve mapping for files.

justinlittman:
refs https://github.com/sul-dlss/google-books/issues/462

## Why was this change made?
Fix mapping of sdrPreserve and shelve.

No.

No.


justinlittman:
Duh.

PR #769, opened 2020-03-31
Adds access validation when creating objects, publishing, and shelving.

### LINKED ISSUE #768: Validate that shelve and publish are no when object rights are dark
If an object rights are set to dark, the shelve and publish attributes are set to false for all files.

justinlittman:
closes #768
closes #769

## Why was this change made?
To make sure that publish and shelve match the access for an object.

No.

Yes.

### DIFF
--- a/app/jobs/shelve_job.rb
+++ b/app/jobs/shelve_job.rb
@@ -11,6 +11,17 @@ class ShelveJob < ApplicationJob
 
     begin
       item = Dor.find(druid)
+
+      validator = validator_for?(item)
+      unless validator.valid?
+        return LogFailureJob.perform_later(druid: druid,
+                                           background_job_result: background_job_result,
+                                           workflow: 'accessionWF',
+                                           workflow_process: 'shelve-complete',
+                                           output: { errors: [{ title: 'Access mismatch',
+                                                                detail: "Not all files have dark access and/or are unshelved when item access is dark: #{validator.invalid_filenames}" }] })
+      end
+
       ShelvingService.shelve(item)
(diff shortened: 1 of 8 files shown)


app/jobs/shelve_job.rb
```diff
@@ -26,4 +35,9 @@ def perform(druid:, background_job_result:)
                                 background_job_result: background_job_result,
                                 workflow_process: 'shelve-complete')
   end
+
+  def item_valid?(item)
+    model = Cocina::Mapper.build(item)
+    ValidateDarkService.valid?(model)
```
jcoyne: It might be nice to have a `validate` method here that returns a dry-monad that explains why it's not valid.

justinlittman: I opted not to use dry-monads since not currently a dependency, but accomplished similar by enhancing ValidateDarkService.

app/services/validate_dark_service.rb
```diff
@@ -0,0 +1,47 @@
...
+  # @param [Cocina::Models::DRO,Cocina::Models::RequestDRO] item to be validated
+  def initialize(item)
+    @item = item
+  end
+
+  def valid?
+    invalid_files.empty?
+  end
+
+  def invalid_files
```
jcoyne, line 14: Why is this a public method?

justinlittman: Why not? It is a potentially useful method for some other user of the service that needs access to the Cocina Files, rather than just the filenames.

jcoyne: Since this is not a library, no big deal.  If it was a library, a bigger public interface makes the code harder to change.

PR #770, opened 2020-03-31
Remove the export directory when doing the workspace reset

jcoyne:
## Why was this change made?

Fixes https://github.com/sul-dlss/common-accessioning/issues/907
also
Fixes sul-dlss/common-accessioning/issues/321

no

no

### DIFF
--- a/app/services/reset_workspace_service.rb
+++ b/app/services/reset_workspace_service.rb
@@ -4,14 +4,11 @@
 class ResetWorkspaceService
   class DirectoryAlreadyExists < StandardError; end
 
-  class BagAlreadyExists < StandardError; end
-
   # @raise [DirectoryAlreadyExists] if the archived directory already exists
-  # @raise [BagAlreadyExists] if the bag for this version already exists
   # @raise [Errno::ENOENT] if the directory doesn't exist
   def self.reset(druid:, version:)
     reset_workspace_druid_tree(druid:, version:, workspace_root: Settings.stacks.local_workspace_root)
-    reset_export_bag(druid:, version:, export_root: Settings.sdr.local_export_home)
+    remove_export_bag(druid:, export_root: Settings.sdr.local_export_home)
   end
 
   # @raises [Errno::ENOENT] if the directory doesn't exist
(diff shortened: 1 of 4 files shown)


mjgiarlo:
@jcoyne Confirming that this is on hold until tested in stage (by @andrewjbtw)?

jcoyne:
@mjgiarlo yes

mjgiarlo:
@jcoyne do you know if this was tested?

jcoyne:
I can't remember.

mjgiarlo:
@jcoyne :speech_balloon: 
> I can't remember.

ok, we'll bug/tag AB in the new year.

ndushay (commented):
Can this be moved forward?  

Does it have any relationship to sul-dlss/common-accessioning#321?  sul-dlss/common-accessioning#770?  sul-dlss/common-accessioning#658 ?

app/services/reset_workspace_service.rb
```diff
@@ -26,14 +23,13 @@ def self.reset_workspace_druid_tree(druid:, version:, workspace_root:)
     FileUtils.mv(druid_tree_path, "#{druid_tree_path}_v#{version}")
   end
 
-  def self.reset_export_bag(druid:, version:, export_root:)
```
ndushay, line 29: can you explain what the thinking was as to resetting the export bag vs. removing it and why it is now appropriate to remove it?

ndushay: so this is the symlink left behind?

jcoyne: I'm not sure what "this" is you are referring to. This ticket just removes the `/dor/export` directory for the item which I believe satisfies the expected outcome:

> I am not sure what the correct cleanup state is supposed to be, since I'm not sure if there is supposed to be something left behind in /dor/export, such as a log. But I don't think the folder is supposed to contain broken symlinks.

ndushay:
@andrewjbtw do you want to test this?

andrewjbtw:
@ndushay yes - and I think I'm finally at a point this week when I can spend time testing more things.

andrewjbtw:
Does this PR also stop renaming the export directory to the version-specifc name like `{druid}_v{#}`? I can't tell if that directory is not being created now or if it's being created and then immediately deleted. 

I'm not sure there's much utility in renaming the export directory with the version name, so it's probably not an issue to stop creating it. I just want to make sure I understand what's changing. Two years ago I thought there might be some reason to keep that directory but it doesn't seem to serve any function. Maybe at one point renaming helped keep processes from conflicting with each other.

ndushay:
I have no idea where the versioned directories for export came from.  Could they be related to making the bags for exporting to preservation??   @jcoyne do you have any idea?

jcoyne:
It gets rid of the renaming part. I couldn't understand what the purpose of that was either.

ndushay:
@andrewjbtw so this PR gets rid of the versioned directories;  is that ok? 

@jcoyne there are now merge conflicts.

andrewjbtw:
Yes, I think it's fine to not have the version directory. I think this is good to go. I had some test items get stuck in accessioning but my conclusion is that the issue they've run into are unrelated to this change since they happened in the preservationIngestWF long before there's any cleanup.

ndushay (approved):

ndushay:
This now has merge conflicts, but you have ✅ after you address that.

PR #780, opened 2020-04-02
Handle objects that don't provide access subschema

### LINKED ISSUE #776: [dor-services-app/stage] NoMethodError: undefined method `access' for nil:NilClass
## Backtrace

    line 89 of [PROJECT_ROOT]/app/services/cocina/object_creator.rb: block in create_dro
    line 86 of [PROJECT_ROOT]/app/services/cocina/object_creator.rb: tap
    line 86 of [PROJECT_ROOT]/app/services/cocina/object_creator.rb: create_dro

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/62496152)

jcoyne:
## Why was this change made?

This indicates they should use the default provided by the APO

Fixes #776

no

yes, it fixes it.

### DIFF
--- a/app/services/validate_dark_service.rb
+++ b/app/services/validate_dark_service.rb
@@ -14,7 +14,7 @@ class ValidateDarkService
   def invalid_files
     @invalid_files ||= begin
       [].tap do |invalid_files|
-        next unless item.access.access == 'dark'
+        next unless item.access&.access == 'dark'
 
         files.each do |file|
           invalid_files << file if file.administrative.shelve || file.access.access != 'dark'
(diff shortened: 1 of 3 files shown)


app/services/validate_dark_service.rb
```diff
@@ -14,7 +14,7 @@ def valid?
   def invalid_files
     @invalid_files ||= begin
       [].tap do |invalid_files|
-        next unless item.access.access == 'dark'
+        next unless item.access&.access == 'dark'
```
justinlittman, line 17: access has a default of dark, so I don't think this is necessary but certainly won't hurt.

jcoyne: I think it only defaults if access is provided at all.

PR #788, opened 2020-04-03
Turn off validation of dark until pre-assembly can make valid objects

### LINKED ISSUE #787: Comment out  access validators until preassembly produces correct data
Back out https://github.com/sul-dlss/dor-services-app/pull/769/files

Until this is done: https://github.com/sul-dlss/pre-assembly/issues/638

jcoyne:
## Why was this change made?

Fixes #787

n/a

n/a


ndushay:
FYI - pre-assembly DOES now make dark objects correctly ... but why validate these and no others?

PR #790, opened 2020-04-03
PreserveJob should retry when there are errors

jcoyne:
## Why was this change made?

Ref https://github.com/sul-dlss/google-books/issues/474

no

no


jcoyne:
@ndushay yes, sidekiq retries any uncaught errors and Andrew indicates that a retry sometimes fixes the problem in 474

PR #798, opened 2020-04-08
Adds new workers for dor-services-app

### LINKED ISSUE #796: Fix Hydrus testing suite
See https://github.com/sul-dlss/hydrus/pull/413

The issue? Spec tests non-deterministically fail (between 2 and 9 different failures each run) due to the DSC registration call timing out. Note that this is using the "legacy" registration method in DSA, so recentish changes to Cocina registrations should not be at fault.

Funnily enough, if I run the whole suite a couple times locally, the second run succeeds. smth smth caching smth? I think we could use another set of :eyes: on this one.

@jcoyne asks if it could be related to not having messaging in the docker-compose service cluster

### LINKED ISSUE #797: Setup capistrano for second dor-services-app worker.
The server has already been provisioned by operations now needs to be added to capistrano

jermnelson:
## Why was this change made?
Fixes #797 

n/a

### DIFF
--- a/config/deploy/prod.rb
+++ b/config/deploy/prod.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
-server 'dor-services-prod.stanford.edu', user: 'dor_services', roles: %w(web app)
+server 'dor-services-prod.stanford.edu', user: 'dor_services', roles: %w(web app worker)
+server 'dor-services-worker-prod-a.stanford.edu', user: 'dor_services', roles: %w(app worker)
 
 Capistrano::OneTimeKey.generate_one_time_key!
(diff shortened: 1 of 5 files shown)


config/deploy/prod.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
 server 'dor-services-prod.stanford.edu', user: 'dor_services', roles: %w(web app)
```
mjgiarlo, line 3: Should prod also include the worker role?

config/deploy/prod.rb
```diff
@@ -1,5 +1,7 @@
...
+# Need app role to use shared_configs but passenger will fail but we're ignoring
```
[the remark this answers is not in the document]
jermnelson: No passenger isn't complaining now, so kruft and removing

config/deploy/prod.rb
jermnelson: Yep, fixed.

PR #809, opened 2020-04-14
Restart sidekiq using systemd

mjgiarlo:
BLOCKED by [related PR](https://github.com/sul-dlss/puppet/pull/5267): sets `profile::systemd_sidekiq::sidekiq_count: 3` in the dor-services-app hiera to keep 3 sidekiq processes running (the default is 5), and allows restarting sidekiq via sudo.

As part of this, remove capistrano-sidekiq dependency (because incompatible with our systemd-ified sidekiq configuration in puppet) and create a new task for restarting sidekiq.

Note that this PR also removes worker responsibility from our web boxes, which
cannot use sidekiq with systemd until they undergo OS upgrades. As we don't
currently need the extra capacity, we'll run workers on dedicated worker boxes
only (using systemd).

## Why was this change made?

To keep pace with Sidekiq.

No.

No.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -25,9 +25,8 @@ require 'capistrano/bundler'
 require 'capistrano/honeybadger'
 require 'capistrano/passenger'
 require 'capistrano/rails/migrations'
-require 'capistrano/sidekiq'
 require 'dlss/capistrano'
 require 'whenever/capistrano'
 
 # Loads custom tasks from `lib/capistrano/tasks' if you have any defined.
-Dir.glob('lib/capistrano/tasks/*.cap').each { |r| import r }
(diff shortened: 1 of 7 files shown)


jcoyne:
@mjgiarlo what if we remove the `:worker` role from `dor-services-{env}` and only run the workers on the worker box for now. We don't really need the extra capacity at this point as we're not running google books.

mjgiarlo:
@jcoyne :speech_balloon: 
> @mjgiarlo what if we remove the `:worker` role from `dor-services-{env}` and only run the workers on the worker box for now. We don't really need the extra capacity at this point as we're not running google books.

Sounds good to me. Done.

PR #811, opened 2020-04-15
[Draft] Run workers on web boxes too

mjgiarlo:
This was blocked by these boxes being on CentOS 6.

## Why was this change made?

To keep pace with sidekiq and increase worker capacity.

no

no


mjgiarlo:
Now that DSA has two dedicated worker boxes, we will not need to re-enable workers on the web boxes. Closing.

PR #813, opened 2020-04-17
Prevent honeybadger from thinking admin tags are honeybadger tags

mjgiarlo:
## Why was this change made?

Cleans up the HB alert and removes possibly confusing information.

no

no


mjgiarlo:
@jermnelson :taco:

PR #814, opened 2020-04-17
Create an Update API endpoint

jcoyne:
## Why was this change made?

So we can update items using the cocina model.

yes.

yes


app/services/cocina/object_updater.rb
```diff
@@ -88,18 +81,20 @@ def update_dro
...
         raise Dor::ParameterError, "Not all files have dark access and/or are unshelved when item access is dark: #{validator.invalid_filenames}" unless validator.valid?
       end
 
-      raise Dor::ParameterError, "Identifer on the query and in the body don't match" if item.pid != obj.externalIdentifier
+      raise Dor::ParameterError, "Identifier on the query and in the body don't match" if item.pid != obj.externalIdentifier
 
       # Validate APO exists (this raises an error if it doesn't)
       Dor.find(obj.administrative.hasAdminPolicy)
+
+      # Can't currently roundtrip desc metadata, including title.
```
jcoyne, line 96: Please document why we can't round trip desc metadata including title.

app/services/administrative_tags.rb
```diff
@@ -69,6 +77,16 @@ def content_type
       .map { |tag| tag.split(' : ').last }
   end
 
+  # @return [Array<String>] an array of tags (strings), possibly empty
+  def project
```
jcoyne, line 81: It was my understanding that project tag was one that needed to be persisted in the XML and shouldn't be in the database.

mjgiarlo: @jcoyne I don't know the answer to this. Maybe @andrewjbtw does?

Current state: all tags are persisted to the database. There is no special handling for project or process (or any other) tags at the moment.

justinlittman:
I've completed my testing with Google Books remediate.

app/services/administrative_tags.rb
jcoyne: Andrew said this is fine. 👍

app/services/cocina/object_updater.rb
```diff
@@ -0,0 +1,183 @@
...
+        raise Dor::ParameterError, "Not all files have dark access and/or are unshelved when item access is dark: #{validator.invalid_filenames}" unless validator.valid?
+      end
+
+      raise Dor::ParameterError, "Identifier on the query and in the body don't match" if item.pid != obj.externalIdentifier
+
+      # Validate APO exists (this raises an error if it doesn't)
+      Dor.find(obj.administrative.hasAdminPolicy)
+
+      # Can't currently roundtrip desc metadata, including title.
+      # Note that title is the only desc metadata field handled by the mapper. However, the mapped title is composed from
```
jcoyne, line 97: 👍

app/services/cocina/object_updater.rb
```diff
@@ -0,0 +1,183 @@
...
+      # several MODS fields which makes writing back to the MODS problematic.
+      raise NotImplementedError, 'Updating descriptive metadata not supported' if obj.description.title.first.value != TitleMapper.build(item)
```
mjgiarlo, line 99: I don't need to care that much about this, but could we consider a custom error type here? AFAIK, `NotImplementedError` is typically used to signal functionality that only works on certain Rubies or OS platforms. I acknowledge that some of our other codebases use `NotImplementedError` to signal that a method has not yet been written or overridden, so this code isn't exactly inconsistent w/ those codebases.

Again, I don't need to care that much about this if y'all aren't inclined to make this change, but I thought it worth raising at least for edification purposes. cc: @sul-dlss/infrastructure-team

app/controllers/objects_controller.rb
```diff
@@ -3,6 +3,10 @@
 class ObjectsController < ApplicationController
   before_action :load_item, except: [:create]
 
+  rescue_from(NotImplementedError) do |e|
```
mjgiarlo, line 6: See other comment re: using `NotImplementedError`.

app/controllers/objects_controller.rb
```diff
@@ -3,6 +3,10 @@
...
+    render status: :not_implemented, plain: e.message
```
mjgiarlo, line 7: I wonder if we should be returning a different HTTP status here. 501 indicates that the request method (HTTP verb) is unrecognized. How about a 422 or 500, depending on whether we think this situation is caused by the client or the server?

app/services/cocina/object_updater.rb
```diff
@@ -0,0 +1,183 @@
...
+        AdministrativeTags.update(item: item, current: existing_tag, new: new_tag)
+      end
+    end
+
+    def tag_starting_with(item, prefix)
+      AdministrativeTags.for(item: item).each do |tag|
+        return tag if tag.start_with?(prefix)
+      end
+      nil
+    end
```
mjgiarlo, line 134: I wonder if instead of adding the tag logic to the updater, you'd be open to moving these smarts into the AdministrativeTags service, perhaps as a new flag (`create_if_missing` or some such) to `AdministrativeTags#update`? We may (?) need this code in other contexts, eventually. 

Then the code here could be changed to:

```ruby
    def add_tags(item, obj)
      AdministrativeTags.update(item: item, 
                                current: AdministrativeTags.content_type(item: item).first,
                                new: content_type_tag(obj.type, obj.structural&.hasMemberOrders&.first&.viewingDirection),
                                create_if_missing: true)
      return unless obj.administrative.partOfProject

      AdministrativeTags.update(item: item,
                                current: AdministrativeTags.project(item: item).first,
                                new: "Project : #{obj.administrative.partOfProject}",
                                create_if_missing: true) 
    end
```

Improvement? Or worse? Or meh?

openapi.yml
```diff
@@ -956,6 +956,44 @@ paths:
         '502':
           description: Error connecting to Symphony
   '/v1/objects/{id}':
+    patch:
```
mjgiarlo, line 959: Does this also need `put` or does `patch` suffice?

app/services/cocina/object_updater.rb
justinlittman: This was how I initially implemented it. However, for some reason that I couldn't figure out the `rescue_from` was misbehaving. Happy to try again / pair if someone feels strongly about it.

mjgiarlo: @justinlittman happy to pair

app/controllers/objects_controller.rb
justinlittman: From https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html: 501 means "The server does not support the functionality required to fulfill the request."

openapi.yml
justinlittman: In what sense? `patch` works fine from dor-services-client.

mjgiarlo: @justinlittman Rails sets up both PATCH and PUT by default when using resourceful routes for update operations. It's not a problem if DSC uses PATCH instead of PUT, even if it means we expose a PUT that is not covered by the OAS.

app/controllers/objects_controller.rb
mjgiarlo: @justinlittman Yeah, but also:

> This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource.

We recognize the request method here, methinks?

justinlittman:
@mjgiarlo the request specs are actually preferable because they test both the creating / updating and the mapping (a 2-fer).

app/services/cocina/object_updater.rb
justinlittman: So I started writing this code. However, need to add conditionals for when there is not an existing content type or project tag (e.g., AdministrativeTags.content_type(item: item) returns []). Also, AdministrativeTags.content_type(item: item) and AdministrativeTags.project(item: item) don't return tags (as the prefix is removed). By the time that you take all of this into account, the difference is meh to worse.

app/controllers/objects_controller.rb
justinlittman: So let's go with the 422. It tells the client that it is doing something wrong.

app/services/cocina/object_updater.rb
mjgiarlo: @justinlittman Sounds good to me, agreed. Thanks for considering.

app/controllers/objects_controller.rb
mjgiarlo: @justinlittman :shipit:

PR #817, opened 2020-04-20
Adds prioritization for workflow-related jobs.

### LINKED ISSUE #816: Add prioritization to Preserve, Publish, Shelve
These should accept lane-id, which should determine queue.

justinlittman:
closes #816

## Why was this change made?
To support prioritization so that Google Books have lower priority than other items.

No.

Yes.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -84,16 +84,18 @@ class ObjectsController < ApplicationController
   def publish
     result = BackgroundJobResult.create
     EventFactory.create(druid: params[:id], event_type: 'publish_request_received', data: { background_job_result_id: result.id })
+    queue = params['lane-id'] == 'low' ? :low : :default
 
-    PublishJob.perform_later(druid: params[:id], background_job_result: result, workflow: params[:workflow])
+    PublishJob.set(queue: queue).perform_later(druid: params[:id], background_job_result: result, workflow: params[:workflow])
     head :created, location: result
   end
 
   def preserve
     result = BackgroundJobResult.create
     EventFactory.create(druid: params[:id], event_type: 'preserve_request_received', data: { background_job_result_id: result.id })
+    queue = params['lane-id'] == 'low' ? :low : :default
 
-    PreserveJob.perform_later(druid: params[:id], background_job_result: result)
+    PreserveJob.set(queue: queue).perform_later(druid: params[:id], background_job_result: result)
     head :created, location: result
   end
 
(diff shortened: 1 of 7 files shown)


justinlittman:
image

app/controllers/shelves_controller.rb
```diff
@@ -8,8 +8,9 @@ def create
     if @item.is_a?(Dor::Item)
       result = BackgroundJobResult.create
       EventFactory.create(druid: @item.pid, event_type: 'shelve_request_received', data: { background_job_result_id: result.id })
+      lane_id = workflow_client.process(pid: @item.pid, workflow_name: 'accessionWF', process: 'shelve').lane_id
```
jcoyne, line 11: Rather than querying the workflow service, can we have the caller pass this in?

app/controllers/objects_controller.rb
```diff
@@ -84,16 +84,18 @@ def accession
   def publish
     result = BackgroundJobResult.create
     EventFactory.create(druid: params[:id], event_type: 'publish_request_received', data: { background_job_result_id: result.id })
+    lane_id = workflow_client.process(pid: params[:id], workflow_name: 'accessionWF', process: 'publish').lane_id
```
jcoyne, line 87: also here

app/controllers/objects_controller.rb
```diff
@@ -84,16 +84,18 @@ def accession
...
 
-    PublishJob.perform_later(druid: params[:id], background_job_result: result, workflow: params[:workflow])
+    PublishJob.set(queue: lane_id).perform_later(druid: params[:id], background_job_result: result, workflow: params[:workflow])
     head :created, location: result
   end
 
   def preserve
     result = BackgroundJobResult.create
     EventFactory.create(druid: params[:id], event_type: 'preserve_request_received', data: { background_job_result_id: result.id })
+    lane_id = workflow_client.process(pid: params[:id], workflow_name: 'accessionWF', process: 'sdr-ingest-transfer').lane_id
```
jcoyne, line 96: and here

app/controllers/shelves_controller.rb
justinlittman: That was my first thought as well, except (1) the caller (a robot) was going to have to make exactly this call to get the lane and (2) dor-services-app is already aware of the workflow service. Given this, this approach seemed like the simplest / clearest.

jcoyne: @justinlittman I think it's better for the caller to get that information before making the API call.

1. I'd prefer if DSA had no knowledge of the workflow system, but right now settiling for limited knowledge
2. Don't make an API call if you don't have all the info the API needs.
3. We can modify lybercore so that the robot know what queue it's running in.

PR #822, opened 2020-04-22
Stop sending workflows to preservation

jcoyne:
## Why was this change made?
New objects don't make a workflows datastream.

no


jcoyne:
Looks like we have a special case for workflows to hit the api.

PR #825, opened 2020-04-24
Updated to cocina-models-0.32.0.

justinlittman:
## Why was this change made?
Update to latest cocina-models.

Yes.

To be tested.


justinlittman:
Yes, still necessary.

PR #827, opened 2020-04-24
Tweaks to descriptive metadata for collections.

justinlittman:
## Why was this change made?
So that descreptive metadata for collections works as expected by hydrus.

No

Testing in hydrus.

### DIFF
--- a/app/services/cocina/mapper.rb
+++ b/app/services/cocina/mapper.rb
@@ -127,7 +127,7 @@ module Cocina
         # This is for etds that haven't yet gone through the other-metadata workflow step
         { title: [{ status: 'primary', value: item.properties.title.first }] }
       else
-        { title: [{ status: 'primary', value: item.full_title }] }
+        { title: [{ status: 'primary', value: item.full_title || item.label }] }
       end
     end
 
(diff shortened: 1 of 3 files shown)


app/services/cocina/mapper.rb
```diff
@@ -127,7 +127,7 @@ def build_descriptive
         # This is for etds that haven't yet gone through the other-metadata workflow step
         { title: [{ status: 'primary', value: item.properties.title.first }] }
       else
-        { title: [{ status: 'primary', value: item.full_title }] }
+        { title: [{ status: 'primary', value: item.full_title || item.label }] }
```
jcoyne, line 130: I'm not sure I follow what's happening here. Why do we need to look for label?

jcoyne: Shouldn't `ObjectCreator#add_description` ensured this was set?

app/services/cocina/object_creator.rb
```diff
@@ -133,17 +133,29 @@ def catkey_for(obj)
...
+        add_description_from_catkey(item)
       elsif obj.description
         item.descMetadata.mods_title = obj.description.title.first.value
       else
         item.descMetadata.mods_title = obj.label
       end
     end
 
+    def add_collection_description(item)
+      # Hydrus created collections don't expect a description.
```
jcoyne, line 145: I'm willing to argue that the existing behavior is a bug, but I'm willing to let it slide for now, if we write an issue to address this.

app/services/cocina/mapper.rb
jcoyne: Can we make this happen only if it's a Hydrus::Collection?  That makes it clearer as to what the use case is.

PR #833, opened 2020-04-28
newer ruby for circle build

ndushay:
## Why was this change made?

preparing to upgrade ruby

na

no


jmartin-sul:
feel free to wait for codeclimate if you'd like, or to merge if it looks like it's not going to return a result.

PR #839, opened 2020-04-30
Generate external identifiers for resources without an ID

mjgiarlo:
**HOLD** until further testing. This is the un-reverted version of https://github.com/sul-dlss/dor-services-app/pull/836.

Fixes sul-dlss/was-registrar-app#221

## Why was this change made?

This should help us map and accession WAS seed and crawl objects. One assumption
in this PR I want to call out for reviewers: We prefer a base index of 1 over a base index of 0. True?

No

I will deploy to stage and 1) run integration specs, and 2) test that some test objects map cleanly to the Cocina model in the Argo UI, do not generate HB alerts, and go through `accessionWF` cleanly.

Can you help me identify 3-5 such objects, @andrewjbtw? They should have `contentMetadata` with ID-less resources, and should be in the error state as described in the ticket. cc: @justinlittman

### DIFF
--- a/app/services/cocina/dro_structural_builder.rb
+++ b/app/services/cocina/dro_structural_builder.rb
@@ -31,12 +31,12 @@ module Cocina
     attr_reader :item
 
     def build_filesets(content_metadata_ds, version:, id:)
-      content_metadata_ds.ng_xml.xpath('//resource').map do |resource_node|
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@justinlittman sorry for the delay. this is now merged.

PR #843, opened 2020-05-01
do not error out on duplicate tag creation

### LINKED ISSUE #841: Allow creating a tag that already exists
ETDs does this and then fails when it tries to create the same tag the second time.

peetucket:
## Why was this change made?

fixes #841 

n/a

### DIFF
--- a/app/services/administrative_tags.rb
+++ b/app/services/administrative_tags.rb
@@ -90,7 +90,6 @@ class AdministrativeTags
   # @param tags [Array<String>] a non-empty array of tags (strings)
   # @param replace [Boolean] replace current tags? default: false
   # @return [Array<AdministrativeTag>]
-  # @raise [ActiveRecord::RecordInvalid] if any druid/tag rows are duplicates
   def create(tags:, replace: false)
     ActiveRecord::Base.transaction do
       AdministrativeTag.where(druid: item.pid).destroy_all if replace
--- a/app/services/administrative_tags.rb
+++ b/app/services/administrative_tags.rb
@@ -96,7 +96,7 @@ class AdministrativeTags
       AdministrativeTag.where(druid: item.pid).destroy_all if replace
 
       tags.map do |tag|
-        AdministrativeTag.create!(druid: item.pid, tag: tag)
+        AdministrativeTag.find_or_create_by!(druid: item.pid, tag: tag)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


peetucket:
I'm not sure the test I altered is really substantial due to the mocking.  It would be nice to confirm the number of tags being created for example.

app/services/administrative_tags.rb
```diff
@@ -90,13 +90,12 @@ def project
   # @param tags [Array<String>] a non-empty array of tags (strings)
   # @param replace [Boolean] replace current tags? default: false
   # @return [Array<AdministrativeTag>]
-  # @raise [ActiveRecord::RecordInvalid] if any druid/tag rows are duplicates
```
mjgiarlo, line 93: @peetucket Yanked this line out too since I don't believe we'll see it anymore.

spec/services/administrative_tags_spec.rb
```diff
@@ -3,8 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe AdministrativeTags do
-  let(:item_with_db_tags) { Dor::Item.new(pid: 'druid:aa123bb7890') }
-  let(:item_without_db_tags) { Dor::Item.new(pid: 'druid:bc234dg8901') }
+  let(:item) { Dor::Item.new(pid: 'druid:aa123bb7890') }
```
mjgiarlo, line 6: I should have renamed this variable to the simpler `item` when I removed the db migration code. Took the opportunity to do that now.

spec/services/administrative_tags_spec.rb
```diff
@@ -79,159 +78,152 @@
...
 
       it 'returns an empty array' do
-        expect(described_class.project(item: item_with_db_tags)).to eq([])
+        expect(described_class.project(item: item)).to eq([])
       end
     end
   end
 
   describe '#create' do
-    before do
```
mjgiarlo, line 165: I... can't remember why I avoided manipulating the database before. Seems like the tests are better when we *do* manipulate the database. See the improved tests below.

peetucket: agreed, confirming the database changes is a nice way to confirm the expected behavior

mjgiarlo:
@peetucket :speech_balloon: 
> I'm not sure the test I altered is really substantial due to the mocking. It would be nice to confirm the number of tags being created for example.

Agreed!

PR #847, opened 2020-05-04
Fix seed objects

mjgiarlo:
## Why was this change made?

This allows the base APOs and collections to be loaded via `rake seed`

No

No.


ndushay:
thanks mike.

mjgiarlo:
@ndushay thank you for finding them!

PR #854, opened 2020-05-07
[HOLD] Normalize administrative tags

jcoyne:
## Why was this change made?
There are 29,822 unique tags and 8,139,563 tag instances, so normalizing the tags would offer a significant space savings. This would also make it possible for us to provide autocomplete on tags (currently too slow).  This step was started in #853 which will allow the migration to go smoothly.

Currently holding to deploy #855 and build the new table:

```
    AdministrativeTag.all.find_each do |at|
      at.update(tag_label: TagLabel.find_or_create_by!(tag: at.tag))
    end
```

or
```ruby
 ActiveRecord::Base.connection.execute('UPDATE administrative_tags a SET tag_label_id = (SELECT id from tag_labels b WHERE a.tag = b.tag)')
```

no

no

### DIFF
--- a/app/models/administrative_tag.rb
+++ b/app/models/administrative_tag.rb
@@ -6,10 +6,7 @@ class AdministrativeTag < ApplicationRecord
 
   belongs_to :tag_label
 
-  validates :tag, format: {
-    with: VALID_TAG_PATTERN,
-    message: 'must be a series of 2 or more strings delimited with space-padded colons, e.g., "Registered By : mjgiarlo : now"'
-  }, uniqueness: {
+  validates :tag_label_id, uniqueness: {
     scope: :druid,
     message: 'has already been assigned to the given druid (no duplicate tags for a druid)'
   }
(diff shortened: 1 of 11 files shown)


db/migrate/20200507202950_add_tag_label_id_to_administrative_tags.rb
```diff
@@ -0,0 +1,5 @@
+class AddTagLabelIdToAdministrativeTags < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :administrative_tags, :tag_label, index: true
```
justinlittman, line 3: No foreign key?

jcoyne: add_reference  is a shortcut for creating a column, index and foreign key at the same time.

jcoyne: Oops, I see i'm missing the `foreign_key: true` option.

PR #859, opened 2020-05-08
Add search for tags

jcoyne:
## Why was this change made?

This will enable https://github.com/sul-dlss/argo/issues/165

yes

yes, but will be tested when the corresponding argo pr is created.

### DIFF
--- a/app/controllers/administrative_tags_controller.rb
+++ b/app/controllers/administrative_tags_controller.rb
@@ -2,7 +2,7 @@
 
 # Administrative tags controller (nested resource under objects)
 class AdministrativeTagsController < ApplicationController
-  before_action :load_item
+  before_action :load_item, only: %i[create index update destroy]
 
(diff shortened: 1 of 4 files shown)


app/controllers/administrative_tags_controller.rb
```diff
@@ -13,6 +13,11 @@ def index
     render json: AdministrativeTags.for(item: @item)
   end
 
+  def search
+    results = TagLabel.where("tag like ?", "#{params[:q]}%").limit(10).pluck(:tag)
```
justinlittman, line 17: ordering?

justinlittman, line 17: Should limit be an optional param?

jcoyne: Currently we only need 10. YAGNI

jcoyne: Not part of the use case yet.

PR #870, opened 2020-05-13
Update the sdr-ingest-transfer step rather than preservation-ingest-i…

jcoyne:
…nitiated

## Why was this change made?

preservation-ingest-initiated is going away https://github.com/sul-dlss/workflow-server-rails/pull/381 which will enable this step to be reset.

-  infra-integration-tests

none

### DIFF
--- a/app/jobs/preserve_job.rb
+++ b/app/jobs/preserve_job.rb
@@ -10,7 +10,7 @@ class PreserveJob < ApplicationJob
     LogFailureJob.perform_later(druid: opts[:druid],
                                 background_job_result: opts[:background_job_result],
                                 workflow: 'accessionWF',
-                                workflow_process: 'preservation-ingest-initiated',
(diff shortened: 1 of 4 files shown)


app/jobs/preserve_job.rb
```diff
@@ -10,7 +10,7 @@ class PreserveJob < ApplicationJob
     LogFailureJob.perform_later(druid: opts[:druid],
                                 background_job_result: opts[:background_job_result],
                                 workflow: 'accessionWF',
-                                workflow_process: 'preservation-ingest-initiated',
+                                workflow_process: 'sdr-ingest-transfer',
```
ndushay, line 13: i strongly prefer "preservation" as it is less confusing than "sdr" which means a lot of different things.

jcoyne: I'm not going to disagree, but this name is already in the system.

PR #874, opened 2020-05-19
Update URL path by removing v1

jgreben:
## Why was this change made?
The `v1` part of the SymWS URL is deprecated and will be removed in the next version of SymWS.

I did not find a test in this repo that references this specific URL path.

`config/setting.yml`


jcoyne:
@joshschneider if we remove the v1 right now, will the service work or do we need to coordinate this change with the upgrade?

justinlittman:
Other Josh: @jgreben ^^

PR #875, opened 2020-05-20
get the catkey from symphony web services instead of SW

peetucket:
## Why was this change made?

to fix sul-dlss/google-books#544

This uses symphony web services to fetch a catkey given a barcode, instead of using searchworks, which should lead to more results.  It also switches the existing dor-services-app endpoint that only fetches a catkey given a barcode (i.e. no marcxml returned) so that it doesn't fetch the marcxml as well, which makes it have to call only one API call instead of two.

IMPORTANT: this change requires updates to shared_configs at the same time:
- [ ] sul-dlss/shared_configs#1358  (dor-services-app-prod)
- [ ] sul-dlss/shared_configs#1362  (dor-services-app-stage cleanup)
- [ ] sul-dlss/shared_configs#1360  (dor-services-app-qa)

deployed to dor-services-stage; tested calls from gbooks-stage and argo-stage and confirmed they worked as expected

All changes are internal to dor-services-app, it should be seamless to consumers.


app/controllers/marcxml_controller.rb
```diff
@@ -1,26 +1,20 @@
...
 class MarcxmlController < ApplicationController #:nodoc:
-  before_action :set_marcxml_resource
-
   rescue_from(SymphonyReader::ResponseError) do |e|
     render status: :internal_server_error, plain: e.message
   end
 
   def catkey
-    render plain: @marcxml.catkey
+    render plain: SymphonyReader.new(**marcxml_resource_params).fetch_catkey
```
jcoyne, line 9: Can we be specific about what is getting passed here?
```suggestion
render plain: SymphonyReader.new(barcode: params.fetch(:barcode)).fetch_catkey
```

app/models/marcxml_resource.rb
```diff
@@ -6,11 +6,7 @@ def self.find_by(catkey: nil, barcode: nil)
     if catkey
       new(catkey: catkey)
     elsif barcode
-      barcode_search_url = format(Settings.catalog.barcode_search_url, barcode: barcode)
-      response = Faraday.get(barcode_search_url)
-      catkey = JSON.parse(response.body)['id']
-
-      new(catkey: catkey)
+      new(catkey: SymphonyReader.new(barcode: barcode).fetch_catkey)
```
jcoyne, line 9: Since `MarcxmlResource` creates a `SymphonyReader`, why not just refactor `MarcxmlResource#marc_record` to use a barcode rather than making 2 requests?

app/controllers/marcxml_controller.rb
peetucket: yes, i can do something like this instead: `SymphonyReader.new(barcode: marcxml_resource_params[:barcode]).fetch_catkey`

app/models/marcxml_resource.rb
peetucket: i just pushed a commit that does some refactoring, is it what you had in mind?  it doesn't eliminate the API call (which we need to do to fetch a catkey if provided a barcode) but it does eliminate creating two `SymphonyReader`s and better encapsulates that logic

app/controllers/marcxml_controller.rb
```diff
@@ -1,26 +1,20 @@
...
 class MarcxmlController < ApplicationController #:nodoc:
-  before_action :set_marcxml_resource
-
   rescue_from(SymphonyReader::ResponseError) do |e|
     render status: :internal_server_error, plain: e.message
   end
 
   def catkey
-    render plain: @marcxml.catkey
+    render plain: SymphonyReader.new(barcode: marcxml_resource_params[:barcode]).fetch_catkey
```
jcoyne, line 9: you can use `params` rather than `marcxml_resource_params` here.  That's less indirection.

app/models/symphony_reader.rb
```diff
@@ -1,20 +1,31 @@
...
   def self.client
     Faraday.new(headers: Settings.catalog.symphony.headers)
   end
 
-  def initialize(catkey:)
+  def initialize(catkey: nil, barcode: nil)
     @catkey = catkey
+    @barcode = barcode
+
+    raise ArgumentError, 'Must supply either a catkey or barcode' if @barcode.nil? && @catkey.nil?
```
jcoyne, line 17: I don't think we need this validation here since you check in fetch_barcode_response and fetch_marc_response

app/controllers/marcxml_controller.rb
```diff
@@ -1,26 +1,20 @@
...
   end
 
   def marcxml
-    render xml: @marcxml.marcxml
+    render xml: MarcxmlResource.new(**marcxml_resource_params).marcxml
```
jcoyne, line 13: writing this out as `catkey: params[:catkey], barcode: params[:barcode]` would be a lot simpler.

app/models/symphony_reader.rb
peetucket: removed

app/controllers/marcxml_controller.rb
peetucket: changed

app/controllers/marcxml_controller.rb
peetucket: changed

app/models/marcxml_resource.rb
jcoyne: yep.

PR #878, opened 2020-05-21
Ditch uniqueness validation in favor of DB constraint

mjgiarlo:
Fixes sul-dlss/argo#2064

Would like @jcoyne to review before this is merged.

## Why was this change made?

Prefer uniqueness constraints in the database over application validations, which are prone to race conditions. As part of this work, add a new honeybadger alert if AR validation exceptions are still raised which we would no longer expect.

I'm not sure how to test a race condition fix such as this one.

None.

### DIFF
--- a/app/controllers/administrative_tags_controller.rb
+++ b/app/controllers/administrative_tags_controller.rb
@@ -24,6 +24,12 @@ class AdministrativeTagsController < ApplicationController
                               tags: params.require(:administrative_tags),
                               replace: params[:replace])
   rescue ActiveRecord::RecordInvalid => e
+    Honeybadger.notify('[SURPRISE] AdministrativeTags.create raised AR::RecordInvalid!',
+                       context: {
+                         druid: params[:object_id],
+                         tags: params.require(:administrative_tags),
+                         replace: params[:replace]
+                       })
     render status: :conflict, plain: e.message
   else
     head :created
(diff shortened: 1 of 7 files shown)


mjgiarlo:
I'm not sure the slight drop in coverage in this PR is worth attempting to cover.

app/controllers/administrative_tags_controller.rb
```diff
@@ -36,6 +42,12 @@ def update
   rescue ActiveRecord::RecordNotFound => e
     render status: :not_found, plain: e.message
   rescue ActiveRecord::RecordInvalid => e
+    Honeybadger.notify('[SURPRISE] AdministrativeTags.update raised AR::RecordInvalid!',
+                       context: {
+                         druid: params[:object_id],
+                         current: CGI.unescape(params[:id]),
```
jcoyne, line 48: You shouldn't need to unescape here, the webserver would have done this.

mjgiarlo: That's copypasta of the args a few lines up. I'd prefer to avoid making that change in this PR if that's OK.

mjgiarlo: @jcoyne :point_up: OK with this?

jcoyne: sure.

PR #880, opened 2020-05-22
Associate catch block with the method that uses it 

jcoyne:
## Why was this change made?

This makes it easier to understand the dataflow.

Test suite

n/a

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -24,14 +24,12 @@ class ObjectsController < ApplicationController
     raise e
   end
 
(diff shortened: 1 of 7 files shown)


jcoyne:
@justinlittman Thank you.  I've updated this to now move the catch block close to that method so the dataflow is easier to understand.

jcoyne:
@justinlittman I don't believe that returning text was a considered decision. We have many routes that return JSON API strutured errors and I believe we should migrate all routes to do so.

aaron-collier:
@justinlittman can you re-review this to clear/update your request? Thanks.

PR #883, opened 2020-05-26
Trap for the case when an object does not have a title

jcoyne:
and display a more helpful error. 

## Why was this change made?
 The current error:
```
Cocina::Models::ValidationError: #/components/schemas/DescriptiveBasicValue/properties/value does not allow null values
```
is rather opaque as to what the problem is.

Test suite

openapi.yml and generated docs.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -45,6 +45,10 @@ class ObjectsController < ApplicationController
     models = ActiveFedora::ContentModel.models_asserted_by(@item)
     @item = @item.adapt_to(Etd) if models.include?('info:fedora/afmodel:Etd')
     render json: Cocina::Mapper.build(@item)
+  rescue Cocina::Mapper::MissingTitle
+    json_api_error(status: :unprocessable_entity,
+                   title: 'Missing title',
+                   message: "All objects are required to have a title, but #{params[:id]} appears to be malformed as a title cannot be found.")
   end
 
   # Initialize specified workflow (assemblyWF by default), and also version if needed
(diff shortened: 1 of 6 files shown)


app/controllers/objects_controller.rb
```diff
@@ -47,6 +47,18 @@ def show
     models = ActiveFedora::ContentModel.models_asserted_by(@item)
     @item = @item.adapt_to(Etd) if models.include?('info:fedora/afmodel:Etd')
     render json: Cocina::Mapper.build(@item)
+  rescue Cocina::Mapper::MissingTitle
+    render status: :unprocessable_entity,
```
justinlittman, line 51: All other errors from this controller return text.

jcoyne: @justinlittman I don't believe that returning text was a considered decision. We have many routes that return JSON API structured errors and I believe we should migrate all routes to do so. I have migrated another in #880

justinlittman: I don't disagree, but think (1) it is problematic to have some return text and others JSON and (2) am concerned that this might break consumers that assumed all responses were text.

justinlittman: Happy to discuss.

jcoyne: The only consumer of an error body is here: https://github.com/sul-dlss/dor-services-client/blob/8e6d25b302e29f101e652b68784de1ab8ad7871f/lib/dor/services/client/response_error_formatter.rb#L27 and it makes no difference to it.

jcoyne: Regarding:
> (1) it is problematic to have some return text and others JSON 

I wholeheartedly agree, but this is already the situation we find ourselves in and this PR works toward fixing that.

aaron-collier:
@justinlittman are you good on this for now? Your request is blocking merge. Thanks.

ndushay:
now it needs a rebase.

PR #884, opened 2020-05-26
Update the cocina model

jcoyne:
## Why was this change made?
adjust openapi.yml and tests for the new models

test suite

openapi.yml


jcoyne:
holding until this can be merged: https://github.com/sul-dlss/dor-services-client/pull/175

PR #890, opened 2020-05-27
Derive type from contentType

jcoyne:
## Why was this change made?
Previously we were deriving this from tags, but tags are not repository items, we should be using the contentType from contentMetadata instead.

This adds support for geo and webarchive_seed.

Test suite

n/a

### DIFF
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -98,7 +98,7 @@ module Cocina
           apply_default_access(item)
         end
 
-        item.contentMetadata.content = ContentMetadataGenerator.generate(druid: pid, object: obj) if obj&.structural&.contains
+        item.contentMetadata.content = ContentMetadataGenerator.generate(druid: pid, object: obj)
 
         add_identity_metadata(obj, item, 'item')
       end
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -147,7 +147,7 @@ module Cocina
     end
 
     def add_dro_tags(pid, obj)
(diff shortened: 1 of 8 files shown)


app/services/cocina/object_creator.rb
```diff
@@ -158,27 +158,6 @@ def add_collection_tags(pid, obj)
...
-              'Map'
-            when Cocina::Models::Vocab.media
-              'Media'
-            when Cocina::Models::Vocab.manuscript
-              'Manuscript'
-            when Cocina::Models::Vocab.book
-              short_dir = direction == 'right-to-left' ? 'rtl' : 'ltr'
-              "Book (#{short_dir})"
-            else
-              Cocina::Models::Vocab.object
```
mjgiarlo, line 177: @jcoyne Yeah, this looks like the potential culprit for https://github.com/sul-dlss/hydrus/issues/435. This would return a URI instead of a label.

app/services/cocina/to_fedora/process_tag.rb
```diff
@@ -0,0 +1,29 @@
...
+                'Map'
+              when Cocina::Models::Vocab.media
+                'Media'
+              when Cocina::Models::Vocab.manuscript
+                'Manuscript'
+              when Cocina::Models::Vocab.book
+                short_dir = direction == 'right-to-left' ? 'rtl' : 'ltr'
+                "Book (#{short_dir})"
+              else
+                Cocina::Models::Vocab.object
```
mjgiarlo, line 23: Should this be 'File'? This looks like the Hydrus bug.

PR #892, opened 2020-05-27
[Refactor] Move goobi methods into the concrete class pertaining to Goobi

jcoyne:
## Why was this change made?
Makes the code easer to understand.

Test suite

n/a

### DIFF
--- a/app/models/dor/goobi.rb
+++ b/app/models/dor/goobi.rb
@@ -25,6 +25,8 @@ module Dor
       end
     end
 
+    private
+
     # We send all tags to Goobi, but "DPG : Workflow : xxx" is the one tag that Goobi uses
     def goobi_xml_tags
       goobi_tag_list.map(&:to_xml).join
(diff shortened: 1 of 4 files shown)


app/models/dor/goobi.rb
```diff
@@ -49,5 +51,53 @@ def xml_request
...
+        GoobiTag.new(name: tag_split[0], value: tag_split[1])
+      end
+    end
+
+    # returns true or false depending if the specially defined goobi DPG ocr tag is present in the object
+    # @return [boolean]
+    def goobi_ocr_tag_present?
+      @goobi_ocr_tag_present ||= begin
+        dpg_goobi_ocr_tag = 'DPG : OCR : TRUE'
+        AdministrativeTags.for(pid: @druid_obj.id).any? { |tag| tag.casecmp(dpg_goobi_ocr_tag).zero? } # case insensitive compare
```
justinlittman, line 80: Would be clearer with `casecmp?`.

jcoyne: I'm just moving the code here to a different location, not trying to change it's behavior.

Issue #782, opened 2020-04-03
Cocina registration should accept an abstract in the descriptive metadata


jcoyne:
Argo passes an abstract when registering a collection.
https://github.com/sul-dlss/dor-services-app/pull/761 might offer a clue for how to do this.

justinlittman:
See https://github.com/sul-dlss/cocina-models/blob/master/spec/fixtures/description/etd.json#L420

ndushay:
Is this how this work is envisioned?

- change to cocina-models to allow abstract as an attribute of description
- dor-services-app will set the abstract attribute of description in the cocina-model when creating a new object
- argo will pass in the abstract attribute to dor-services-app when a collection is registered (already being done)

jcoyne:
@ndushay I don't believe any change will be required to cocina-models.  @justinlittman has pointed out we can add it as a `note` (see https://github.com/sul-dlss/cocina-models/blob/master/lib/cocina/models/description.rb#L11)  I believe the other parts are correct.

ndushay:
I would think we would want to separate it from note - I'll check w arcadia.

ndushay:
slack'ed with Arcadia;  she says go with note of type 'summary', as in https://github.com/sul-dlss/cocina-models/blob/master/spec/fixtures/description/etd.json#L420 as was indicated above.

>this is how it currently stands in the modeling work i've been doing:
```
{
  "note": [
    {
      "value": "This is an abstract.",
      "type": "summary"
    }
  ]
}
```

ndushay:
PRs merged;  dor-services-app fix deployed;  argo fix will be deploy with next week dep updates.

PR #893, opened 2020-05-28
ObjectCreator.create retains description summary note (abstract)

ndushay:
## Why was this change made?

Part of #782 - Argo collection registration takes a collection abstract, and we've been dropping it on the floor.  This allows dsa to retain it.

added tests to test suite.

na


app/services/cocina/mapper.rb
```diff
@@ -131,9 +131,17 @@ def build_identification
...
 
+    # rubocop:disable Rails/DynamicFindBy
     def build_descriptive
-      { title: [{ status: 'primary', value: TitleMapper.build(item) }] }
+      desc = { title: [{ status: 'primary', value: TitleMapper.build(item) }] }
+
+      # collections are registered with abstracts
+      return desc if item.descMetadata.find_by_terms(:abstract).blank?
+
+      desc[:note] = [{ type: 'summary', value: item.descMetadata.find_by_terms(:abstract).text }]
```
ndushay, line 141: this is an Om datastream, so we want `find_by_terms`, not the thing rubocop is asking for.

app/services/content_metadata_generator.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 # Builds the contentMetadata xml from cocina filesets
-# rubocop:disable Metrics/ClassLength
```
ndushay, line 4: this was flagged by rubocop as no longer needed.

app/services/cocina/mapper.rb
```diff
@@ -131,9 +131,17 @@ def build_identification
       end
     end
 
+    # rubocop:disable Rails/DynamicFindBy
     def build_descriptive
-      { title: [{ status: 'primary', value: TitleMapper.build(item) }] }
+      desc = { title: [{ status: 'primary', value: TitleMapper.build(item) }] }
+
+      # collections are registered with abstracts
+      return desc if item.descMetadata.find_by_terms(:abstract).blank?
```
jcoyne, line 139: I find this easier to read if we avoid the low-level api (e.g. find_by_terms):

```suggestion
      return desc if item.descMetadata.abstract.blank?
```

app/services/cocina/mapper.rb
jcoyne, line 141:
```suggestion
      desc[:note] = [{ type: 'summary', value: item.descMetadata.abstract.first }]
```

ndushay: I'd swear I had it this way and the tests didn't pass ... in any case I'm glad to make this change and have it work.

ndushay:
@jcoyne thanks for the suggestion - I started out that way but must've done something wrong.  Glad for that fix!

PR #895, opened 2020-05-28
[Refactor] Use upstream accessor methods

jcoyne:


app/models/dor/service_item.rb
```diff
@@ -6,14 +6,7 @@ class ServiceItem
     # this is a class level method so it can be used on aribtrary druids (e.g. collection the item is associated with) without having to instantiate the object
     # look in identityMetadata/otherId[@name='catkey']
     def self.get_ckey(object)
```
justinlittman, line 8: Can't we get rid of this method altogether?

jcoyne: Yes, I want to get rid of several of these methods, but in a different PR.

PR #904, opened 2020-05-29
[Refactor] Move remaining builder classes into the FromFedora module

jcoyne:
## Why was this change made?

So that we have a consistent way of constructing all sub-schemas.  No implementations have been changed, this just moves classes to different namespaces and renames some methods (e.g. `build` -> `props`).

Test suite

n/a


jcoyne:
@ndushay because it's building a hash of properties.

PR #916, opened 2020-06-08
remove unncessary content/download files methods and associated tests

### LINKED ISSUE #877: Remove /v1/objects/{id}/contents and /v1/objects/{id}/contents/{path}
See https://github.com/sul-dlss/dor-services-client/issues/174

peetucket:
## Why was this change made?

Fixes #877, now that we no longer use that functionality in argo and it can be removed.  ~~Pending sul-dlss/argo#2066 merge and deploy~~ <--- now done

Follow up with removing from dor-services-client too: https://github.com/sul-dlss/dor-services-client/pull/176

Hasn't been tested yet...should run integration tests

The openapi spec was updated.


peetucket:
I suppose it is breaking something that used to be there.  We could bring up at the 11am meeting tomorrow.

jcoyne:
Yep, I added it to the agenda.

peetucket:
We decided to not bump the version, so I think this PR is ready.

PR #918, opened 2020-06-11
version_service: can't close version if active assemblyWF

### LINKED ISSUE #2093: Implement subject name mappings
See "Name subject with name type" in https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/subject_name_spec.rb

ndushay:
## Why was this change made?

Closes sul-dlss/argo/issues/2093

 versions should not be closable if there is an active assemblyWF.

wrote rspec tests; test suite

n/a

### DIFF
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -38,7 +38,7 @@ class VersionService
     vmd_ds.sync_then_increment_version sdr_version
     vmd_ds.save unless work.new_record?
 
-    WorkflowClientFactory.build.create_workflow_by_name(work.pid, 'versioningWF', version: work.current_version)
+    workflow_client.create_workflow_by_name(work.pid, 'versioningWF', version: work.current_version)
 
     return if (opts.keys & open_options_requiring_work_save).empty?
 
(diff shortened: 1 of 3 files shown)


app/services/version_service.rb
```diff
@@ -142,4 +143,12 @@ def accessioning?
   def open_options_requiring_work_save
     [:opening_user_name, :significance, :description]
   end
+
+  def active_assembly_wf?
```
jcoyne, line 147: Having this much shared knowlege between dor-services-app and workflow troubles me.  Why make workflow a separate service if we have conditional logic about how that workflow can proceed outside of the workflow service? Furthermore, we are adding coupling between to supposedly independent workflows.  I think this would be less confusing if we could figure out how to represent this inside the workflows themselves?

ndushay: To quote a colleague:  "ok, but that's not what THIS PR is about."  :-)

ndushay: putting this another way -- we already merged this logic into argo to fix a reported bug.   I'm definitely up for discussing a better way to address the problem but right now we have code doing this in Argo and this PR was meant to align things ... should they remain unaligned until we have a better solution?

jcoyne: I think it was premature to do it in argo until we had a solution. I had set up a meeting to discuss this, but the work proceeded anyway.  I'm concerned that by putting it in code it makes it harder to make the changes we ought to make on an architectural level.

ndushay: fixing a production problem reported by the SDR manager is the urgency.

jcoyne: Since we have the band-aid on argo, is there a way to trigger the original production problem?  I thought that would prevent a user from getting to that state.

PR #919, opened 2020-06-11
[Refactor] Call ObjectCreator and ObjectUpdate with the cocina model

jcoyne:
## Why was this change made?

This relieves these already overburdened classes of the responsibility of how to build the cocina models. No functional changes

Test suite

n/a


config/settings/development.yml
```diff
@@ -1 +1,3 @@
-fedora_url: 'http://fedoraAdmin:fedoraAdmin@localhost:8983/fedora'
+fedora_url: http://fedoraAdmin:fedoraAdmin@localhost:8983/fedora
+solr:
```
justinlittman, line 2: ?

jcoyne: This allows us the run in development mode and connect to the proper server.

PR #929, opened 2020-06-11
[Hold] Cleanup all files in the workspace

jcoyne:
## Why was this change made?

Previously, objects originating in pre-assembly, which creates symlinks from /dor/assembly to /dor/workspace were getting cleaned up properly. However, objects created via the sdr-api, which put real files in /dor/workspace were leaving around a versioned directory the first time the cleanup service was called, and removing that on the second call to the cleanup service.  This ensures it is cleaned up correctly the first time.

Test suite. Andrew will sign-off on stage

n/a


andrewjbtw:
Looks good. I created a Hydrus item and it also gets the appropriate cleanup, while leaving the /data/hydrus-files space untouched.

PR #930, opened 2020-06-12
[Refactor] Switch all validation errors to use the ValidationError class

jcoyne:
No functional changes

## Why was this change made?

For consistency

Test suite

n/a


app/validators/cocina/validate_dark_service.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module Cocina
```
ndushay, line 3: is the idea that this would move to cocina gem, eventually?

jcoyne: There is no "cocina" gem, so no.

PR #945, opened 2020-06-29
Update dependencies

cbeer:


Gemfile
```diff
@@ -60,7 +60,7 @@ group :test, :development do
   gem 'rack-console'
   gem 'rspec-rails'
   gem 'rspec_junit_formatter'
-  gem 'rubocop', '~> 0.85'
+  gem 'rubocop', '~> 0.85.1'
```
jcoyne, line 63: Pinning to a minor version was intentional in https://github.com/sul-dlss/dor-services-app/pull/935

mjgiarlo: in the process of changing the approach per slack chat

PR #946, opened 2020-06-29
Pin rubocop to minor, not patch, and regenerate config

mjgiarlo:
@sul-dlss/infrastructure-team will discuss this on Tuesday, June 29, 2020.

## Why was this change made?

Was a part of #945

Tests

None


jcoyne:
https://github.com/sul-dlss/sdr-api/blob/master/Gemfile#L41

PR #954, opened 2020-07-10
Expect that ETDs have a sourceId

jcoyne:
## Why was this change made?

Since https://github.com/sul-dlss/hydra_etd/pull/469 was merged

Test suite and tested on stage.

n/a

### DIFF
--- a/app/services/cocina/from_fedora/identification.rb
+++ b/app/services/cocina/from_fedora/identification.rb
@@ -15,15 +15,15 @@ module Cocina
       end
 
       def props
-        case item
-        when Dor::Etd
-          # ETDs don't have source_id, but we can use the dissertationid (in otherId) for this purpose
-          { sourceId: item.otherId.find { |id| id.start_with?('dissertationid:') } }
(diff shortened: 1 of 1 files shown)


app/services/cocina/from_fedora/identification.rb
```diff
@@ -15,15 +15,15 @@ def initialize(item)
...
-        end
+        return {} if item.is_a? Dor::Collection
+
+        return { sourceId: item.source_id } if item.source_id
+
+        # ETDs post Summer 2020 have a source id, but legacy ones don't.  In that case look for a dissertation_id.
+        dissertation = item.otherId.find { |id| id.start_with?('dissertationid:') }
+        raise "unable to resolve a sourceId for #{item.pid}" unless dissertation
+
+        { sourceId: dissertation }
```
ndushay, line 26: do we want it to start with "dissertationId: " ?

jcoyne: This is what was already in use (old line 21)

PR #957, opened 2020-07-13
Add the ability to set and modify the agreement via the cocina API

jcoyne:
## Why was this change made?

This will support ETDs using the Cocina API to set the agreement.

Test suite

n/a


jcoyne:
@ndushay it will be checked via the integration tests.

ndushay:
in a perfect world, wouldn't it be encumbent on you, as the PR submitter, to run the appropriate integration tests?  They may be delayed for a few days ...

PR #974, opened 2020-08-18
support controlled digital lending

### LINKED ISSUE #969: Add new CDL option (controlled digital lending) as an option for object rights
See sul-dlss/argo#2200 for full details. Basically we need a new option for rights and the ability to generate this option in rightsMetadata. It will be available for selection in Argo.

- [ ] New rights option needs to be added here: 
  - https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/to_fedora/access.rb#L12-L22
  - https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/from_fedora/access.rb#L44-L52
- [x] Bump dor-services to 9.6 or better
- [ ] Make updates to dor-rights-auth:  sul-dlss/dor-rights-auth/issues/44  this is needed to fix the from fedora mapping and the associated test
- [x] Write a test for the to fedora mapping (note, there are no tests at all yet)

peetucket:
## Why was this change made?

Fixes #969 
Part of sul-dlss/argo#2200
Follow up with sul-dlss/argo#2208

Unit tests

The consul page on rights was updated to reflect the new option: https://consul.stanford.edu/display/chimera/Rights+metadata+--+the+rightsMetadata+datastream


app/services/cocina/to_fedora/access.rb
```diff
@@ -17,6 +17,8 @@ def self.apply(item, access)
                         'none'
                       when 'dark'
                         'dark'
+                      when 'cdl-stanford-nd'
```
jcoyne, line 20: Can you add a test that shows this branch is a valid option?

peetucket: yeah, I noticed there are no tests for the `to_fedora` mappings, is that just an oversight and should we create a new spec for this?

peetucket, line 20: needs to match the new value from cocina models (pending PR)

Gemfile
```diff
@@ -47,8 +47,9 @@ gem 'uuidtools', '~> 2.1.4'
 gem 'whenever', require: false
 
 # DLSS/domain-specific dependencies
-gem 'cocina-models', '~> 0.34.1'
-gem 'dor-services', '~> 9.5'
+gem 'cocina-models', '~> 0.36'
+gem 'dor-services', '~> 9.6'
+gem 'dor-rights-auth', '>= 1.5.0' # required for new CDL rights
```
peetucket, line 52: like this? @jcoyne

jcoyne: This is fine. I think just running `bundle update dor-rights-auth` would have done the trick too.

spec/services/cocina/from_fedora/dro_access_spec.rb
```diff
@@ -84,6 +84,32 @@
...
+                <group rule="no-download">stanford</group>
+              </cdl>
+            </machine>
+          </access>
+        </rightsMetadata>
+      XML
+    end
+
+    it 'builds the hash' do
+      expect(access).to eq(access: 'controlled digital lending', controlled_digital_lending: true, download: 'controlled digital lending')
```
peetucket, line 109: so our object level access is 'controlled digital lending', we also have the attribute set to true here, which allows for something more hybrid in the future (like access is 'stanford', but controlled_digital_lending == true)

jcoyne, line 109: I'm pretty sure that download is supposed to be only `true` or `false`, not a string.

peetucket: from the other specs, it looks like 'download' can take a number of other string values, e.g. https://github.com/sul-dlss/dor-services-app/blob/master/spec/services/cocina/from_fedora/dro_access_spec.rb#L111

but maybe download should be 'none' ? (as it is for citation-only objects)

jcoyne: Sorry, `none` is the correct value: https://github.com/sul-dlss/cocina-models/blob/master/openapi.yml#L761-L769

peetucket: so the download attribute comes from here: https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/from_fedora/access.rb#L30-L35

so i think the most correct thing to do is go back to dor-rights-auth and parse the rights-object correctly so that it can vary in the future based on the no-download rule existing in the CDL clause (or not)

but the faster and simpler method is to simply add another condition here so that CDL is always no download (just like citation-only and dark) https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/from_fedora/access.rb#L33

peetucket: something like this in dor-rights-auth could detect this rule and allow us to map it here ?  https://github.com/sul-dlss/dor-rights-auth/pull/48

peetucket: I' having second thoughts about the dor-rights-auth parsing of the no-download rule ... we may be going down a rabbit hole there unnecessarily for what we really need to do right now.  Pushed a simpler commit to dor-services-app to deal with the donwload='none'.

PR #977, opened 2020-08-19
Get imageData from techmd service

### LINKED ISSUE #975: imageData in contentMetadata is lost on cocina roundtrip
Before
https://sul-dor-test.stanford.edu/fedora/objects/druid:cc243mg0841/datastreams/contentMetadata/content?asOfDateTime=2020-07-28T16:34:22.823Z

After
https://sul-dor-test.stanford.edu/fedora/objects/druid:cc243mg0841/datastreams/contentMetadata/content?asOfDateTime=2020-08-18T21:14:12.145Z

It is necessary for https://github.com/sul-dlss/purl/blob/27b881991f2dfc53dd5f151f7f67726a2d355e93/app/models/content_metadata.rb#L125-L127

jcoyne:
Unfortunately, we don't have the technicalMetadata service fully populated, so if new "metadata-only" versions are made, we would lose the dataImage tag.  One possibility is to create stub records in the techMd for every druid that is not already represented using contentMetadata as a source.

at present there are 363291 druids represented in the techmd service:

```
DroFile.distinct.count('druid')
   (27024.4ms)  SELECT COUNT(DISTINCT "dro_files"."druid") FROM "dro_files"
 => 363291
```

We have 1341143 images, 802251 books, 244543 maps in Argo (~2.1 million objects)

So we have about 1.8 million objects that would need to be transferred to the techmd service.

This could be done by calling this rake task for each object:
```
RAILS_ENV=production bin/rake techmd:generate_for_moab['druid:bb000kg4251']
```
or by calling `MoabProcessingService.process(druid)`

## Why was this change made?
Fixes #975

The only reason contentMetadata needs imageData is so that purl can generate IIIF info.  This allows us to get this info from tech-metadata service and will allow us to remove it from the assembly workflow.

### DIFF
--- a/app/services/publish/content_metadata_with_image_sizes.rb
+++ b/app/services/publish/content_metadata_with_image_sizes.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Publish
+  # For presentation, we need to know the image dimensions
+  # see https://github.com/sul-dlss/purl/blob/27b881991f2dfc53dd5f151f7f67726a2d355e93/app/models/content_metadata.rb#L125-L127
+  # We retrieve this data from the technical metadata service.
+  class ContentMetadataWithImageSizes
+    def initialize(content_metadata)
+      @content_metadata = content_metadata
+    end
+
+    def to_xml
+      content_metadata.ng_xml.xpath('//contentMetadata/resource/file').each do |file|
+        next unless image? file['mimetype']
+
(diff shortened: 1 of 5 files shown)


andrewjbtw:
The technical metadata service isn't generating technical metadata for non-preserved files, so it won't have values for most of the JP2s because most JP2s are only sent to Stacks. We would need to have it start running on those files.

jcoyne:
Superseded by #980

PR #978, opened 2020-08-19
Move the ContentMetadataGenerator to the Cocina::ToFedora module

jcoyne:
No functional changes.

## Why was this change made?

This place is reflective of the classes purpose

Test suite.


app/services/cocina/to_fedora/content_metadata_generator.rb
```diff
@@ -0,0 +1,152 @@
+# frozen_string_literal: true
+
+module Cocina
+  module ToFedora
+    # Builds the contentMetadata xml from cocina filesets
+    class ContentMetadataGenerator
+      VALID_THREE_DIMENSION_EXTENTIONS = ['.obj'].freeze
```
mjgiarlo, line 7: Typo? `EXTENSIONS` instead of `EXTENTIONS`

jcoyne: That's a preexisting condition: https://github.com/sul-dlss/dor-services-app/blob/5338bfed53f5b3c5b9478a3251e67e2a4041dbb5/app/services/content_metadata_generator.rb#L5  Can you make a new issue for that?

mjgiarlo: @jcoyne Oh, I didn't realize that. I did a browser find and didn't see old references, but that's bc GitHub hid the diff. :unamused:

No biggie. Approving.

PR #1008, opened 2020-08-20
MARC2MODS XSLT updates

arcadiafalcone:
## Why was this change made?

New 3.7 XSLT released by LC + local cleanup/customization

Running the XSLT against fixture records

n/a


aaron-collier (dismissed):
HOLD until after CDL?

jcoyne (approved):

PR #1027, opened 2020-08-29
Update cocina-models to support multiple collections

jcoyne:
Hold until sdr-client and dor-services-client have been updated to support multiple collections and google-books, sdr-api, and argo have been similarly updated.

## Why was this change made?

Ref https://github.com/sul-dlss/argo/issues/2233


peetucket:
Fixed one rubocop issue and re-ran autogen to get the other complaint about `Assignment Branch Condition size`

peetucket (approved):

app/services/cocina/object_creator.rb
```diff
@@ -74,7 +74,7 @@ def create_dro(obj)
       Dor::Item.new(pid: pid,
                     admin_policy_object_id: obj.administrative.hasAdminPolicy,
                     source_id: obj.identification.sourceId,
-                    collection_ids: [obj.structural&.isMemberOf].compact,
+                    collection_ids: Array.wrap(obj.structural&.isMemberOf).compact,
```
peetucket, line 77: do you know if we need to make this change anywhere else in our code?

jcoyne: I don't know of any.

PR #1033, opened 2020-09-01
fix file download for flipbook by fixing hasMemberOrders construction

jmartin-sul:
## Why was this change made?

fix a production issue where files cannot be downloaded because `hasMemberOrders` was not correctly constructed for the `'Book (flipbook, ltr)'` content type.

unit tests, so far.  will also test on stage before deploying to prod.

n/a, fix for regression to intended behavior

### DIFF
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -4,6 +4,11 @@ module Cocina
   module FromFedora
     # builds the Structural subschema for Cocina::Models::DRO from a Dor::Item
     class DroStructural
+      VIEWING_DIRECTION_FOR_CONTENT_TYPE = {
+        'Book (ltr)' => 'left-to-right',
+        'Book (rtl)' => 'right-to-left'
+      }.freeze
+
       def self.props(item, type:)
         new(item, type: type).props
       end
@@ -58,12 +63,8 @@ module Cocina
         else
           # Fallback to using tags.  Some books don't have bookData nodes in contentMetadata XML.
           # When we migrate from Fedora 3, we don't need to look this up from AdministrativeTags
-          case AdministrativeTags.content_type(pid: item.id).first
-          when 'Book (ltr)'
-            [{ viewingDirection: 'left-to-right' }]
-          when 'Book (rtl)'
-            [{ viewingDirection: 'right-to-left' }]
-          end
+          content_type = AdministrativeTags.content_type(pid: item.id).first
+          [{ viewingDirection: VIEWING_DIRECTION_FOR_CONTENT_TYPE[content_type] }] if VIEWING_DIRECTION_FOR_CONTENT_TYPE[content_type]
         end
       end
 
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -6,6 +6,7 @@ module Cocina
     class DroStructural
       VIEWING_DIRECTION_FOR_CONTENT_TYPE = {
         'Book (ltr)' => 'left-to-right',
+        'Book (flipbook, ltr)' => 'left-to-right',
         'Book (rtl)' => 'right-to-left'
       }.freeze
 
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -13,20 +13,16 @@ module Cocina
         @type = type
       end
 
-      # rubocop:disable Metrics/AbcSize
       def props
         {}.tap do |structural|
-          structural[:hasMemberOrders] = create_member_order if type == Cocina::Models::Vocab.book
+          has_member_orders = build_has_member_orders
+          structural[:hasMemberOrders] = has_member_orders if has_member_orders.present?
 
           contains = build_filesets(item.contentMetadata, version: item.current_version.to_i, id: item.pid)
           structural[:contains] = contains if contains.present?
 
-          sequence = build_sequence(item.contentMetadata)
-          if sequence.present?
-            structural[:hasMemberOrders] ||= [{}]
-            structural[:hasMemberOrders].first[:members] = sequence
-          end
           structural[:hasAgreement] = item.identityMetadata.agreementId.first unless item.identityMetadata.agreementId.empty?
+
           begin
             # Note that there is a bug with item.collection_ids that returns [] until item.collections is called. Below side-steps this.
             structural[:isMemberOf] = item.collections.first.id if item.collections.present?
@@ -36,11 +32,20 @@ module Cocina
           end
         end
       end
-      # rubocop:enable Metrics/AbcSize
 
       private
 
       attr_reader :item, :type
+
+      def build_has_member_orders
+        ret_val = create_member_order if type == Cocina::Models::Vocab.book
+        sequence = build_sequence(item.contentMetadata)
+        if sequence.present?
+          ret_val ||= [{}]
+          ret_val.first[:members] = sequence
+        end
+        ret_val if defined? ret_val
+      end
 
       def create_member_order
         reading_direction = item.contentMetadata.ng_xml.xpath('//bookData/@readingDirection').first&.value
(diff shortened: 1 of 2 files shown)


jmartin-sul:
e.g.

https://argo.stanford.edu/view/druid:by619wx9035

#### dor-services-app

* https://app.honeybadger.io/projects/50568/faults/65002733/fcbb388c-ebe1-11ea-8c39-13875088279f?page=0
* `GET` https://dor-services-prod.stanford.edu/v1/objects/druid:by619wx9035 

> Cocina::Models::ValidationError: #/components/schemas/DROStructural/properties/hasMemberOrders does not allow null values
>
> Full Backtrace
> 
>     [GEM_ROOT]/gems/cocina-models-0.37.0/lib/cocina/models/validator.rb:12 :in `rescue in validate`
>     [GEM_ROOT]/gems/cocina-models-0.37.0/lib/cocina/models/validator.rb:7 :in `validate`
>     [GEM_ROOT]/gems/cocina-models-0.37.0/lib/cocina/models/dro.rb:41 :in `new`
>     [PROJECT_ROOT]/app/services/cocina/mapper.rb:37 :in `build`
>     [PROJECT_ROOT]/app/services/cocina/mapper.rb:16 :in `build`
>     [PROJECT_ROOT]/app/controllers/objects_controller.rb:47 :in `show`

#### argo (error bubbles up from DSA)

* https://app.honeybadger.io/projects/49894/faults/65949933/fc9c3ffe-ebe1-11ea-8707-4a3fae742b3d?page=0
*  `GET` https://argo.stanford.edu/items/druid:by619wx9035/files?id=00003050_0001.TIF 

> Dor::Services::Client::UnexpectedResponse: Internal Server Error: 500 (Response from dor-services-app did not contain a body. Check honeybadger for dor-services-app for backtraces, and look into adding a `rescue_from` in dor-services-app to provide more details to the client in the future)
> 
> Full Backtrace
> 
>     [GEM_ROOT]/gems/dor-services-client-6.10.0/lib/dor/services/client/versioned_service.rb:34 :in `raise_exception_based_on_response!`
>     [GEM_ROOT]/gems/dor-services-client-6.10.0/lib/dor/services/client/object.rb:61 :in `find`
>     [PROJECT_ROOT]/app/controllers/files_controller.rb:106 :in `load_resource`
>     [GEM_ROOT]/gems/activesupport-5.2.4.3/lib/active_support/callbacks.rb:426 :in `block in make_lambda`
>     [GEM_ROOT]/gems/activesupport-5.2.4.3/lib/active_support/callbacks.rb:198 :in `block (2 levels) in halting`
>     [GEM_ROOT]/gems/actionpack-5.2.4.3/lib/abstract_controller/callbacks.rb:34 :in `block (2 levels) in <module:Callbacks>`

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -4,6 +4,12 @@ module Cocina
   module FromFedora
     # builds the Structural subschema for Cocina::Models::DRO from a Dor::Item
     class DroStructural
+      VIEWING_DIRECTION_FOR_CONTENT_TYPE = {
+        'Book (ltr)' => 'left-to-right',
+        'Book (flipbook, ltr)' => 'left-to-right',
+        'Book (rtl)' => 'right-to-left'
```
aaron-collier, line 10: Didn't we also see `Book (flipbook, rtl)`? If not, no biggie.

jmartin-sul: i don't recall seeing that myself, but i only queried the one druid that we were all looking at yesterday wrt this issue (by619wx9035).  though looking at the other one andrew mentioned, that's also `Book (flipbook, ltr)`, it seems (in the tags, no `bookData` stanza in `contentMetadata` afaict for that one either).

maybe @mjgiarlo queried and found `Book (flipbook, rtl)` examples while we were all clumping on this, and i missed it while i was looking at something else?

either way, easy change, happy to add in this PR or a follow on.

andrewjbtw: You can see the list of content types in use under the "Process" tag in the Argo facet sidebar: https://argo.stanford.edu/catalog?f%5Bexploded_tag_ssim%5D%5B%5D=Process+%3A+Content+Type

jmartin-sul: i'd also be happy to query admin tags for other viewing direction/content type mappings, though i'd be inclined to make any changes that came out of that in a follow on PR.

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -53,12 +64,8 @@ def create_member_order
...
           # Fallback to using tags.  Some books don't have bookData nodes in contentMetadata XML.
           # When we migrate from Fedora 3, we don't need to look this up from AdministrativeTags
-          case AdministrativeTags.content_type(pid: item.id).first
-          when 'Book (ltr)'
-            [{ viewingDirection: 'left-to-right' }]
-          when 'Book (rtl)'
-            [{ viewingDirection: 'right-to-left' }]
-          end
+          content_type = AdministrativeTags.content_type(pid: item.id).first
+          [{ viewingDirection: VIEWING_DIRECTION_FOR_CONTENT_TYPE[content_type] }] if VIEWING_DIRECTION_FOR_CONTENT_TYPE[content_type]
```
mjgiarlo, line 68: Probably for @andrewjbtw: should we have a DEFAULT viewing direction for cases that fall through the cracks? (Presumably left to right?) This PR fixes flipbooks but there may be other cases that pass through this method without a viewing direction. Or should those cases not having a viewing direction?

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -36,12 +38,21 @@ def props
...
         end
       end
-      # rubocop:enable Metrics/AbcSize
 
       private
 
       attr_reader :item, :type
 
+      def build_has_member_orders
+        ret_val = create_member_order if type == Cocina::Models::Vocab.book
```
mjgiarlo, line 47: Would you consider renaming this variable? (Something like `member_orders`?)

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -36,12 +38,21 @@ def props
...
+        sequence = build_sequence(item.contentMetadata)
+        if sequence.present?
+          ret_val ||= [{}]
+          ret_val.first[:members] = sequence
+        end
+        ret_val if defined? ret_val
```
mjgiarlo, line 53: I believe the `ret_val` will always be defined at this point due to the assignment on line 47. (Even if the conditional returns false, the variable will be defined. It will be `nil`.)

jmartin-sul: indeed!  good catch.

app/services/cocina/from_fedora/dro_structural.rb
jmartin-sul: ok, added `Book (flipbook, rtl)` and the other `Process : Content Type` tags relevant to reading direction, as discussed on slack.

app/services/cocina/from_fedora/dro_structural.rb
jmartin-sul: done!

app/services/cocina/from_fedora/dro_structural.rb
jmartin-sul: simplified as suggested

jmartin-sul:
@mjgiarlo 💬 
> LGTM. Do you think the individual commits are meaningful or should we squash when merging?

i did try to keep the commits semantically meaningful, and usually have a slight pref for small focused commits.  but also the last couple commits were messy little follow-ons, and ideally i'd have made this PR like 3 commits instead of 6, and i don't think it's worth 20 minutes of rebase wrangling to get a shiny text book example, so...

up to you!  whatever you'd prefer.

jmartin-sul:
buried now in a resolved thread, but @andrewjbtw posted a useful link, for the `Process : Content Type` facet in argo:  https://argo.stanford.edu/catalog?f%5Bexploded_tag_ssim%5D%5B%5D=Process+%3A+Content+Type

mjgiarlo:
@jmartin-sul I squash-merged. Thanks!

PR #1038, opened 2020-09-02
ETDs now have mods title and no longer rely on properties.

jcoyne:
## Why was this change made?

The ETD project has recently undergone updates where the properties are stored in a database and not fedora.  The object is given a proper mods title when it is registered in the SDR.

Test suite.


ndushay (approved):
WFM.

PR #1055, opened 2020-09-08
adjust collection member solr query to correctly return collection members

peetucket:
## Why was this change made?

Fixes sul-dlss/argo#2243

Basically, the collection member solr query wasn't returning anything, preventing collection member release from working correctly.

Tested on -stage, by running a call from argo-stage.

### DIFF
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -5,7 +5,7 @@ class MembersController < ApplicationController
   # Return the published members of this collection
   def index
     solr_params = {
-      q: "is_member_of_collection_ssim:\"#{params[:id]}\" published_dttsim:[* TO *]",
+      q: "is_member_of_collection_ssim:\"#{ActiveFedora::Base.internal_uri(params[:object_id])}\" published_dttsim:[* TO *]",
       wt: :json,
       fl: 'id,objectType_ssim',
       rows: 100_000_000
--- a/spec/requests/members_for_collection_spec.rb
+++ b/spec/requests/members_for_collection_spec.rb
@@ -7,6 +7,17 @@ RSpec.describe 'Get the members' do
     allow(ActiveFedora::SolrService.instance.conn).to receive(:get).and_return(solr_response)
   end
 
+  let(:druid) { 'druid:mk420bs7601' }
+
+  let(:solr_params) do
+    {
+      fl: 'id,objectType_ssim',
+      q: "is_member_of_collection_ssim:\"#{ActiveFedora::Base.internal_uri(druid)}\" published_dttsim:[* TO *]",
+      rows: 100_000_000,
+      wt: :json
+    }
+  end
+
   let(:solr_response) do
     {
       'response' => {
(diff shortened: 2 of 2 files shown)


app/controllers/members_controller.rb
```diff
@@ -5,7 +5,7 @@ class MembersController < ApplicationController
   # Return the published members of this collection
   def index
     solr_params = {
-      q: "is_member_of_collection_ssim:\"#{params[:id]}\" published_dttsim:[* TO *]",
+      q: "is_member_of_collection_ssim:\"info:fedora/#{params[:id]}\" published_dttsim:[* TO *]",
```
jcoyne, line 8: How about: `ActiveFedora::Base.internal_uri(params[:id])` Even better might be to do:
```
_query_:"{!raw f=is_member_of_collection_ssim}#{ActiveFedora::Base.internal_uri(params[:id])}"
```

peetucket: Didn't realize there was an internal_uri method, that seems like a natural change, but I'm not sure I get that `!raw` syntax.  How does that incorporate the other parts of the query?

jcoyne: The raw query parser does no analysis or transformation: https://lucene.apache.org/solr/guide/6_6/other-parsers.html#OtherParsers-RawQueryParser

`_query_` is a magic field for the dismax parser: https://lucene.apache.org/solr/guide/6_6/the-extended-dismax-query-parser.html

peetucket: I just used `ActiveFedora::Base.internal_uri(params[:id])` but kept the query syntax - its not clear to me what the advantage is of using the `_query_` magic field?

jcoyne: If you don't use the `_query_` field you can't escape to use the raw parser, you'll use the default dismax parser.

app/controllers/members_controller.rb
```diff
@@ -4,8 +4,9 @@
 class MembersController < ApplicationController
   # Return the published members of this collection
   def index
+    pid = params[:object_id] || params[:id]
```
jcoyne, line 7: Why do we need to look at both of these?

jcoyne: It looks to me like `object_id` is the only valid choice:
```
$ bin/rails routes -g members
        Prefix Verb URI Pattern                              Controller#Action
object_members GET  /v1/objects/:object_id/members(.:format) members#index {:format=>:json}
```

jcoyne (changes requested):
It looks to me like `spec/requests/members_for_collection_spec.rb` is missing an assertion. Can you add it please?

app/controllers/members_controller.rb
peetucket: ah, when looking for that and missed it, but noticed it wasn't working with params[:id], fixed it by changing to params[:object_id] (and it works now)

peetucket:
> It looks to me like `spec/requests/members_for_collection_spec.rb` is missing an assertion. Can you add it please?

what assertion are you thinking of?

jcoyne:
The one that says we call solr with the correct query (contains the id).

peetucket:
adjusted the member spec to confirm correct solr params are sent

jcoyne (approved):

PR #1063, opened 2020-09-11
Ensure that Hydrus objects are able to be updated

jcoyne:
## Why was this change made?

Because they don't have titles, so they are a special case

Fixes https://github.com/sul-dlss/argo/issues/2248

Test suite.


ndushay (approved):
oy.

PR #1074, opened 2020-09-22
Include a high-level rights characterization in the 856 sent to Symphony

cbeer:
## Why was this change made?

For controlled digital lending, Searchworks needs to know (from the MARC record)  whether there's an SDR object with CDL  rights available. We've heard other use cases for additional rights information (e.g. "show me all the world accessible data")

- [x] With tests?
- [x] confirming with libsys that their processes will handle the new data
- [x] Vitus: "I don't see any problem with sending that type of information in a |x of the managed purl"


mjgiarlo (approved):
LGTM! Thanks for adding the tests.

jmartin-sul (approved):

PR #1075, opened 2020-09-23
Report a sensible error if an object is missing a required sourceId

jcoyne:
## Why was this change made?

This will show a message in argo that one can use rather than having to dig through honeybadger to find out why the request to dor-services-app failed.

### DIFF
--- a/spec/requests/show_object_spec.rb
+++ b/spec/requests/show_object_spec.rb
@@ -51,7 +51,7 @@ RSpec.describe 'Get the object' do
       end
 
       it 'returns the object' do
-        get '/v1/objects/druid:mk420bs7601',
+        get '/v1/objects/druid:bc123df4567',
             headers: { 'Authorization' => "Bearer #{jwt}" }
         expect(response).to have_http_status(:ok)
         expect(response_model).to eq expected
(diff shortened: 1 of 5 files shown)


spec/requests/show_object_spec.rb
```diff
@@ -251,6 +251,31 @@
       end
     end
 
+    context 'when the object exists without a sourceId' do
+      let(:object) do
+        Dor::Item.new(pid: 'druid:bc123df4567',
```
jmartin-sul, line 256: i found the differing druid between here...

spec/requests/show_object_spec.rb
```diff
@@ -251,6 +251,31 @@
...
+      end
+
+      let(:expected) do
+        {
+          errors: [{ detail: 'unable to resolve a sourceId for druid:bc123df4567', status: '422', title: 'Missing sourceId' }]
+        }
+      end
+
+      it 'returns the error' do
+        get '/v1/objects/druid:mk420bs7601',
```
jmartin-sul, line 272: ...and here confusing.  but that also doesn't seem unique to this new test, and i think it's unsurprising that it'd work given the lack of `.with` in the `allow(Dor).to receive(:find).and_return(object)`?

anyway, wouldn't hold up the PR over it, but assuming this is an actual inconsistency and not my misunderstanding, it'd at least be a nice bit of backlog tech debt to ticket.

jmartin-sul (approved):
one small nitpick assuming i'm understanding things correctly, wouldn't hold up merging over it so feel free to merge if you'd like.  happy to file a follow on ticket or something if my confusion was justified.

spec/requests/show_object_spec.rb
jcoyne: Fixed the test

jmartin-sul: nice, thanks!

PR #1076, opened 2020-09-23
Extract a ToFedora::DROAccess class

jcoyne:
This adds no new functionality

## Why was this change made?

This removes a number of conditionals and duplicate code by adding a ToFedora::DROAccess which subclasses ToFedora::Access.  This subclass only handles DROAccess so it can have fewer conditionals.  This makes the code easier to follow.

Test suite.


peetucket (approved):
LGTM

PR #1079, opened 2020-09-24
Update uniform title mapping

jcoyne:
## Why was this change made?
Ref https://github.com/sul-dlss-labs/cocina-descriptive-metadata/pull/18

### DIFF
--- a/spec/services/cocina/from_fedora/descriptive/titles_spec.rb
+++ b/spec/services/cocina/from_fedora/descriptive/titles_spec.rb
@@ -256,9 +256,67 @@ RSpec.describe Cocina::FromFedora::Descriptive::Titles do
             "status": 'primary'
           },
           {
-            "value": 'Shakespeare, William, 1564-1616. Hamlet',
+            "structuredValue": [
+              {
+                "value": 'Shakespeare, William, 1564-1616',
+                "type": 'name'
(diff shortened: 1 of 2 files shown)


spec/services/cocina/from_fedora/descriptive/titles_spec.rb
```diff
@@ -256,14 +256,72 @@
...
+              }
+            ],
             "type": 'uniform',
             "uri": 'http://id.loc.gov/authorities/names/n80008522'
           }
         ]
       end
     end
 
+    context 'when there are uniform titles with multiple name part elements' do
```
ndushay, line 276: wouldn't we also want a test for when it's a single part?  or is there one already not showing up in the github diff?

ndushay (commented):
Question about spec for another branch in case statement.

spec/services/cocina/from_fedora/descriptive/titles_spec.rb
jcoyne: One already exists: https://github.com/sul-dlss/dor-services-app/blob/master/spec/services/cocina/from_fedora/descriptive/titles_spec.rb#L242

ndushay (approved):

PR #1084, opened 2020-09-30
Script to test our descriptive cocina mapping on real data

jcoyne:
Requires:
- [x] https://github.com/sul-dlss/sul-solr-configs/pull/230
- [x] #1052

Tested on stage.


mjgiarlo (approved):
LOVELY

PR #1085, opened 2020-09-30
Check for data loss via okcomputer endpoint

tallenaz:
## Why was this change made?

ops wants to check for data loss, similarly to argo, prescat, and sul-pub

deployed to qa briefly to verify a passing check

DevOps docs change is upcoming.


ndushay (commented):
LGTM but i don't know how soon we'll have data in the three environments.

justinlittman (approved):

PR #1086, opened 2020-10-01
Copy the role attribute from Fedora to use in Cocina

jcoyne:
## Why was this change made?
Fixes https://github.com/sul-dlss/argo/issues/2262

Tested on stage.

### Before
Screen Shot 2020-10-01 at 1 50 08 PM

### After
Screen Shot 2020-10-01 at 1 49 55 PM

### DIFF
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -97,10 +97,22 @@ module Cocina
         end
       end
 
+      def digests(node)
+        [].tap do |digests|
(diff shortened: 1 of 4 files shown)


app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -97,18 +97,30 @@ def build_sequence(content_metadata_ds)
         end
       end
 
+      def digests(node)
```
jcoyne, line 100: Had to extract this method to keep under the ABC threshold.

ndushay (approved):
Looks good.  thanks for the refactor, too.

Issue #1098, opened 2020-10-01
[dor-services-app/prod] KeyError: key not found: nil


jcoyne:
## Backtrace

    line 94 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/titles.rb: fetch
    line 94 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/titles.rb: block in structured_name
    line 94 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/titles.rb: map

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/68290956)

PR #1099, opened 2020-10-01
Handle namePart nodes that are unlabeled

jcoyne:
## Why was this change made?

Fixes https://github.com/sul-dlss/dor-services-app/issues/1098 which will allow https://argo.stanford.edu/view/druid:qx518tq4049 to proceed.


ndushay (approved):

ndushay:
haha - there WAS a way to sneak through that case statement with an error, but I didn't spot it yesterday.  hahahahaha.

jcoyne:
The behavior present in https://argo.stanford.edu/view/druid:qx518tq4049 was not specified in the test cases: https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_titleInfo.txt#L238-L246, so it slipped through.

PR #1134, opened 2020-10-14
Correct types for contributor names in cocina

jcoyne:
## Why was this change made?

We were using incorrect types for names in cocina.


ndushay (commented):
failing specs...

ndushay (approved):

Issue #1095, opened 2020-10-01
Map cocina descriptive relatedItem


jcoyne:
See: https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_relatedItem.txt

Fix this pending test (and add any that are missing):
```
Cocina::ToFedora::Descriptive::RelatedResource when it has a related item without type TODO: Waiting to hear back from Arcadia about https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_relatedItem.txt#L51-L67
```

PR #1138, opened 2020-10-15
Map MODS relatedItem to cocina descriptive

jcoyne:
## Why was this change made?
See #1095


ndushay (approved):
see slack channel comment:  "question:   if we have code that is essentially translating cocina models into xml, and translating xml into cocina models ... should that translation code be in its own gem, instead of being another thing piled into the dor-services-app?    The dor-services-app needs to take the (mods xml) from DOR and turn it into cocina models, and vice versa ... but that seems like a good use of a gem, no?"

jcoyne:
@ndushay Possibly, but currently we have no other consumers, so that seems like extra trouble to manage the dependency.

PR #1180, opened 2020-10-16
Add constituent resource type

### LINKED ISSUE #1174: Cocina mapping error: key not found: "constituent"
```
druid:xd664hq0928 error. key not found: "constituent"
```

jcoyne:
## Why was this change made?
Fixes #1174


mjgiarlo (commented):
worth adding a test or nah?

jcoyne:
@mjgiarlo I don't think so, we are sample testing, which is what is finding this, but typically you don't test that all the values of the enum exist.  Or it would be a test that would be `a = a`

mjgiarlo (approved):

PR #1196, opened 2020-10-20
Handle missing coordinates.

### LINKED ISSUE #1184: Cocina Mapping Error: undefined method `text' for nil:NilClass
```
druid:rt706cv5800 error. undefined method `text' for nil:NilClass
```

justinlittman:
closes #1184

## Why was this change made?
To handle cases in which coordinates is missing from a geo subject.

Unit

NA

### DIFF
--- a/spec/services/cocina/from_fedora/descriptive/subject_spec.rb
+++ b/spec/services/cocina/from_fedora/descriptive/subject_spec.rb
@@ -611,6 +611,23 @@ RSpec.describe Cocina::FromFedora::Descriptive::Subject do
     end
   end
 
+  context 'with a cartographic subject missing coordinates' do
+    let(:xml) do
+      <<~XML
+        <subject>
+          <cartographics>
+            <scale>1:22,000,000</scale>
+            <projection>Conic proj</projection>
+          </cartographics>
+        </subject>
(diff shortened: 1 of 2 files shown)


spec/services/cocina/from_fedora/descriptive/subject_spec.rb
```diff
@@ -584,6 +584,23 @@
...
+          <cartographics>
+            <scale>1:22,000,000</scale>
+            <projection>Conic proj</projection>
+          </cartographics>
+        </subject>
+      XML
+    end
+
+    it 'builds the cocina data structure' do
+      expect(build).to eq []
```
jcoyne, line 600: This would result in data loss? Shoudn't it transform the data it has at least?

jcoyne (changes requested):

spec/services/cocina/from_fedora/descriptive/subject_spec.rb
jcoyne: Ah, those are in form. Nothing is lost.

jcoyne (approved):

jcoyne:
Please rebase.

PR #1205, opened 2020-10-21
[HOLD] Updates cocina models.

justinlittman:
## Why was this change made?
Update cocina models.

NA

openapi

### DIFF
--- a/app/services/cocina/to_fedora/access.rb
+++ b/app/services/cocina/to_fedora/access.rb
@@ -37,7 +37,7 @@ module Cocina
         when 'dark'
           'dark'
         else
-          access.download == 'none' ? "#{access.access}-nd" : access.access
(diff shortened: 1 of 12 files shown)


app/services/cocina/to_fedora/access.rb
```diff
@@ -37,7 +37,7 @@ def rights_type
         when 'dark'
           'dark'
         else
-          access.download == 'none' ? "#{access.access}-nd" : access.access
+          access.respond_to?(:download) && access.download == 'none' ? "#{access.access}-nd" : access.access
```
jcoyne, line 40: why would it not respond to download?

mjgiarlo (approved):

app/services/cocina/to_fedora/access.rb
justinlittman: `DROAccess` has download; `Access` does not.

PR #1222, opened 2020-10-22
Add missing related resource type

### LINKED ISSUE #1214: Cocina map error: key not found: "otherVersion"
In prod: https://app.honeybadger.io/projects/50568/faults/68951965

jcoyne:
## Why was this change made?

Fixes #1214

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/related_resource.rb
+++ b/app/services/cocina/to_fedora/descriptive/related_resource.rb
@@ -12,6 +12,7 @@ module Cocina
           'succeeded by' => 'succeeding',
           'reviewed by' => 'reviewOf',
           'has original version' => 'original',
+          'has version' => 'otherVersion',
           'has part' => 'constituent',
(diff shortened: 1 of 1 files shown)


app/services/cocina/to_fedora/descriptive/related_resource.rb
```diff
@@ -12,6 +12,7 @@ class RelatedResource
           'succeeded by' => 'succeeding',
           'reviewed by' => 'reviewOf',
           'has original version' => 'original',
+          'has version' => 'otherVersion',
```
ndushay, line 15: seems we should have tests for these, no?

jcoyne: What would the test say? "There is a hash that has these values?"  That's not a great test.

justinlittman (approved):
Agree that this is essentially config and doesn't merit a unique test.

PR #1233, opened 2020-10-22
Revert cocina-models to 0.41.0.

justinlittman:
## Why was this change made?
0.42.0 has a problematic handling of classification subjects.

Unit

NA


mjgiarlo (approved):

app/services/cocina/from_fedora/descriptive/subject.rb
```diff
@@ -6,11 +6,12 @@ class Descriptive
       # Maps subject nodes from MODS to cocina
       class Subject
         NODE_TYPE = {
-          'temporal' => 'time',
-          'topic' => 'topic',
-          'geographic' => 'place',
+          'classification' => 'classification',
```
aaron-collier, line 9: I just reordered this list. I'm not sure classification is necessary here, and can drop if not needed.

Issue #1248, opened 2020-10-26
[dor-services-app/prod] NoMethodError: undefined method `[]' for nil:NilClass


justinlittman:
## Backtrace

    line 111 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/admin_metadata.rb: build_language
    line 20 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/admin_metadata.rb: block in build
    line 19 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/admin_metadata.rb: tap

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/68952285)

Issue #1240, opened 2020-10-23
[dor-services-app/prod] NoMethodError: undefined method `text' for nil:NilClass


jcoyne:
## Backtrace

    line 110 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/admin_metadata.rb: build_language
    line 20 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/admin_metadata.rb: block in build
    line 19 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/admin_metadata.rb: tap

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/69059367)

PR #1256, opened 2020-10-27
Implement language mappings to_ and from_fedora per Arcadia's examples

### LINKED ISSUE #1090: Map cocina descriptive language
When I call `Cocina::ToFedora::Descriptive.transform` with:
```
{
  "language": [
    {
      "value": "English",
      "code": "eng",
      "uri": "http://id.loc.gov/vocabulary/iso639-2/eng",
      "source": {
        "code": "iso639-2b",
        "uri": "http://id.loc.gov/vocabulary/iso639-2/"
      }
    }
  ]
}
```

It should have `<language>` nodes in the descMetadata

https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_language.txt

### LINKED ISSUE #1216: Cocina map: translated language
It should be able to round trip this xml as found on stage in druid:hc607hj7585
```xml
  <language>
    <languageTerm authority="iso639-2b" type="code">eng</languageTerm>
  </language>
  <language objectPart="translation">
    <languageTerm authority="iso639-2b" type="code">fre</languageTerm>
  </language>
```

currently it's omitting `objectPart`:
```xml
  <language>
    <languageTerm type="code" authority="iso639-2b">eng</languageTerm>
  </language>
  <language>
    <languageTerm type="code" authority="iso639-2b">fre</languageTerm>
  </language>
```

ndushay:
**NOTE:  we have a partial mapping for language;  this may actually reduce the number of errors ... not sure if this should be held as a "new mapping" or not.**

## Why was this change made?

Fixes #1216 -- what to do with the language element `objectPart` attribute
Fixes #1090 -- mapping the language element.   (@jcoyne had this ticket. I believe he did PR #1130, which was a beginning, and this PR does the rest)

This PR implements all of Arcadia's examples* in https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_language.txt for both "from_fedora" and "to_fedora" .

Note that the known examples that were failing in prod are covered by the specs:

#1248 - from gf100ky7916 (quick fix with PR #1249), which has

```xml
<language>
    <languageTerm type="text">Latin</languageTerm>
    <languageTerm type="code" authority="iso639-2b">lat</languageTerm>
  </language>
```

specs have langTerm of type text analogous to above, and langTerm type code analogous to above, and "text" and "code" langTerms occurring together and apart.

#1240 - from md951vt4966 (quick fix with PR #1246), which has 

```xml
<language>
    <languageTerm type="code" authority="iso639-2b">eng</languageTerm>
  </language>
```
specs now cover this case.

bb012xz4244, from honeybadger which has 

```xml
  <language>
    <languageTerm authority="iso639-2b" authorityURI="http://id.loc.gov/vocabulary/iso639-2" valueURI="http://id.loc.gov/vocabulary/iso639-2/eng" type="code">eng</languageTerm>
    <languageTerm authority="iso639-2b" authorityURI="http://id.loc.gov/vocabulary/iso639-2" valueURI="http://id.loc.gov/vocabulary/iso639-2/eng" type="text">English</languageTerm>
  </language>
```
specs now cover this case.

* Arcadia still needs to do a PR to address the problem I encountered https://github.com/sul-dlss/dor-services-app/issues/1090#issuecomment-715532802 -- my specs reflect her answer https://github.com/sul-dlss/dor-services-app/issues/1090#issuecomment-716613678.

image

Wrote functional specs directly off Arcadia's examples in https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_language.txt

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/language.rb
+++ b/app/services/cocina/from_fedora/descriptive/language.rb
@@ -6,11 +6,17 @@ module Cocina
       # Maps languages
       class Language
         LANG_XPATH = '//mods:language'
-        LANG_TEXT_XPATH = './mods:languageTerm[@type="text"]/text()'
-        LANG_TEXT_AUTHORITY_URI_XPATH = './mods:languageTerm[@type="text"]/@authorityURI'
-        LANG_TEXT_VALUE_URI_XPATH = './mods:languageTerm[@type="text"]/@valueURI'
-        LANG_CODE_XPATH = './mods:languageTerm[@type="code"]/text()'
-        LANG_CODE_AUTHORITY_XPATH = './mods:languageTerm[@type="code"]/@authority'
+        LANG_STATUS_XPATH = './@status'
+        OBJECT_PART_XPATH = './@objectPart'
+        DISPLAY_LABEL_XPATH = './@displayLabel'
+        LANGUAGE_TERM_XPATH = "#{LANG_XPATH}/mods:languageTerm"
+        SCRIPT_TERM_XPATH = "#{LANG_XPATH}/mods:scriptTerm"
+        LANG_TERM_TEXT_XPATH = './mods:languageTerm[@type="text"]/text()'
+        LANG_TERM_CODE_XPATH = './mods:languageTerm[@type="code"]/text()'
+        LANG_TERM_CODE_AUTHORITY_XPATH = './mods:languageTerm[@type="code"]/@authority'
+        LANG_TERM_TEXT_AUTHORITY_XPATH = './mods:languageTerm[@type="text"]/@authority'
+        TEXT_AUTHORITY_URI_XPATH = './*[@type="text"]/@authorityURI' # can be for languageTerm or scriptTerm
+        TEXT_VALUE_URI_XPATH = './*[@type="text"]/@valueURI' # can be for languageTerm or scriptTerm
 
         # @param [Nokogiri::XML::Document] ng_xml the descriptive metadata XML
         # @return [Hash] a hash that can be mapped to a cocina model
@@ -25,10 +31,12 @@ module Cocina
         def build
           [].tap do |langs|
             languages.each do |lang|
-              langs << { code: language_code_for(lang),
-                         value: language_text_for(lang),
-                         uri: language_uri_for(lang),
-                         source: language_source_for(lang) }.reject { |_k, v| v.blank? }
+              attribs = {}
+              attribs = lang_term_attributes_for(lang) if language_term?(lang)
+              attribs[:status] = language_status_for(lang) if language_status_for(lang).present?
+              attribs[:script] = script_term_attributes_for(script_term_nodes(lang)) if script_term?(lang)
+
+              langs << attribs
             end
           end
         end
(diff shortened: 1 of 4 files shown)


app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -37,28 +45,89 @@ def build
...
+            source: language_source_for(lang)
+          }.reject { |_k, v| v.blank? }
+        end
+
+        def script_term_attributes_for(script_term_nodes)
+          return if script_term_nodes.blank?
+
+          code, value, authority = nil
+          script_term_nodes.each do |script_term_node|
+            code ||= script_term_node.xpath('text()').to_s if script_term_node.xpath('@type').to_s == 'code'
```
jcoyne, line 64: I find that using the methods on the node is easier to understand. What do you think of this?
```suggestion
            code ||= script_term_node.content if script_term_node['type'] == 'code'
```

app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -37,28 +45,89 @@ def build
...
+            value ||= script_term_node.xpath('text()').to_s if script_term_node.xpath('@type').to_s == 'text'
```
jcoyne, line 65:
```suggestion
            code ||= script_term_node.content if script_term_node['type'] == 'text'
```

app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -37,28 +45,89 @@ def build
...
+            authority ||= script_term_node.xpath('@authority').to_s
```
jcoyne, line 66:
```suggestion
            authority ||= script_term_node['authority']
```

app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -37,28 +45,89 @@ def build
...
+          end
+          source = { code: authority } if authority
+          {
+            code: code,
+            value: value,
+            source: source
+          }.reject { |_k, v| v.blank? }
```
jcoyne, line 73:
```suggestion
          }.compact
```

app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -37,28 +45,89 @@ def build
...
-          lang.xpath(LANG_TEXT_VALUE_URI_XPATH, mods: DESC_METADATA_NS).to_s
+        # this can be present for type text and/or code, but we only want one.
+        def language_value_uri_for(lang)
+          result = lang.xpath(TEXT_VALUE_URI_XPATH, mods: DESC_METADATA_NS).to_s
+          result = lang.xpath('./*/@valueURI', mods: DESC_METADATA_NS).to_s if result.blank?
+          result
+        end
+
+        def language_applies_to(lang)
+          [value: lang.xpath(OBJECT_PART_XPATH, mods: DESC_METADATA_NS).to_s] if lang.xpath(OBJECT_PART_XPATH, mods: DESC_METADATA_NS).to_s.present?
```
jcoyne, line 96:
```suggestion
          value = lang.xpath(OBJECT_PART_XPATH, mods: DESC_METADATA_NS).to_s
          [value: value] if value.present?
```

app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -6,11 +6,17 @@ class Descriptive
...
       class Language
         LANG_XPATH = '//mods:language'
-        LANG_TEXT_XPATH = './mods:languageTerm[@type="text"]/text()'
-        LANG_TEXT_AUTHORITY_URI_XPATH = './mods:languageTerm[@type="text"]/@authorityURI'
-        LANG_TEXT_VALUE_URI_XPATH = './mods:languageTerm[@type="text"]/@valueURI'
-        LANG_CODE_XPATH = './mods:languageTerm[@type="code"]/text()'
-        LANG_CODE_AUTHORITY_XPATH = './mods:languageTerm[@type="code"]/@authority'
+        LANG_STATUS_XPATH = './@status'
+        OBJECT_PART_XPATH = './@objectPart'
+        DISPLAY_LABEL_XPATH = './@displayLabel'
```
jcoyne, line 11: Can we move these strings that are only used one time into the method that uses them?  That makes it easier to read because you don't have to refer back here for the definitions.  I recognize there was an established pattern here, but I'm not a huge fan.  It would also be okay if we want to do that in another PR. I mostly want to hear your thoughts on this idea.

app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -37,28 +45,89 @@ def build
...
 
+        def lang_term_attributes_for(lang)
+          {
+            code: language_code_for(lang),
+            value: language_text_for(lang),
+            uri: language_value_uri_for(lang),
+            appliesTo: language_applies_to(lang),
+            displayLabel: language_display_label(lang),
+            source: language_source_for(lang)
+          }.reject { |_k, v| v.blank? }
```
jcoyne, line 56:
```suggestion
          }.compact
```

app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -37,28 +45,89 @@ def build
...
+          code = lang.xpath(LANG_TERM_TEXT_AUTHORITY_XPATH, mods: DESC_METADATA_NS).to_s if code.blank? && language_term?(lang)
+          # this can be present for type text and/or code, but we only want one.
+          uri = lang.xpath(TEXT_AUTHORITY_URI_XPATH, mods: DESC_METADATA_NS).to_s
+          uri = lang.xpath('./*/@authorityURI', mods: DESC_METADATA_NS).to_s if uri.blank?
           {
-            code: lang.xpath(LANG_CODE_AUTHORITY_XPATH, mods: DESC_METADATA_NS).to_s,
-            uri: lang.xpath(LANG_TEXT_AUTHORITY_URI_XPATH, mods: DESC_METADATA_NS).to_s
+            code: code,
+            uri: uri
           }.reject { |_k, v| v.blank? }
```
jcoyne, line 117: I recognize this is not yours, but `.compact` here would be great too.

app/services/cocina/to_fedora/descriptive/language.rb
```diff
@@ -26,16 +26,52 @@ def write
 
         attr_reader :xml, :languages
 
+        # rubocop:disable Metrics/AbcSize
         def write_basic(language)
-          xml.language do
+          top_attributes = {}
+          top_attributes[:displayLabel] = language.displayLabel if language.displayLabel
+          top_attributes[:objectPart] = applies_to_first_value(language.appliesTo) if applies_to_first_value(language.appliesTo)
```
jcoyne, line 33:
```suggestion
          applies_to = applies_to_first_value(language.appliesTo)
          top_attributes[:objectPart] = applies_to if applies_to
```

app/services/cocina/to_fedora/descriptive/language.rb
```diff
@@ -26,16 +26,52 @@ def write
...
             if language.value
-              xml.languageTerm language.value, type: 'text'
-            else
-              attributes = { type: 'code' }
-              attributes[:authority] = language.source.code
+              attributes[:type] = 'text'
+              xml.languageTerm language.value, attributes
+            end
+
+            if language.code
```
jcoyne, line 46: should this be an `else if`?

jcoyne (changes requested):

justinlittman:
@ndushay The problem that was causing the errors has been fixed, so I would suggest that this is a new mapping.

app/services/cocina/to_fedora/descriptive/language.rb
ndushay: no - a language can have both a code and a value or it can have just one of them.

app/services/cocina/from_fedora/descriptive/language.rb
ndushay: I slightly prefer the constants to call out the hardcoded xpath expressions, but if they are only used once, i prefer them to be near the method using them.   I haven't liked this pattern of having them all together at the top.  I like the idea of addressing this as a separate PR.   Perhaps for ALL the from_fedora mapping classes at once, so we can abolish the yucky pattern and create one we like.

app/services/cocina/from_fedora/descriptive/language.rb
ndushay: nice.  I was following an established pattern for this as well!

app/services/cocina/from_fedora/descriptive/language.rb
ndushay: I get errors using compact here, e.g.

```
 5) Cocina::FromFedora::Descriptive::Language with language code only builds the cocina data structure
     Failure/Error:
       expect(build).to eq [
         {
           "code": 'eng',
           "source": {
             "code": 'iso639-2b'
           }
         }
       ]
     
       expected: [{:code=>"eng", :source=>{:code=>"iso639-2b"}}]
            got: [{:code=>"eng", :source=>{:code=>"iso639-2b", :uri=>""}}]
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -[{:code=>"eng", :source=>{:code=>"iso639-2b"}}]
       +[{:code=>"eng", :source=>{:code=>"iso639-2b", :uri=>""}}]
       
     # ./spec/services/cocina/from_fedora/descriptive/language_spec.rb:71:in `block (3 levels) in <top (required)>'
```

app/services/cocina/from_fedora/descriptive/language.rb
ndushay: thanks!  this is nicer!

app/services/cocina/from_fedora/descriptive/language.rb
ndushay: this one does NOT cause errors. yay!

app/services/cocina/from_fedora/descriptive/language.rb
ndushay: yes, nicer!

app/services/cocina/to_fedora/descriptive/language.rb
ndushay: 👍

app/services/cocina/from_fedora/descriptive/language.rb
jcoyne: Ah, I forgot about the empty string case.  Compact is only for the nil values.

jcoyne (approved):

ndushay:
@jcoyne I took all of your suggestions that didn't cause test failures.

@justinlittman I think this is exactly the edge case that is confusing:  we have solved the language mapping bugs we have run across SO FAR, but the additional mappings in this PR are for anticipated issues with language mappings.

I would consider this "new mappings" if the language element hadn't been mapped at all yet ... but since it is partially mapped ... doesn't this solve problems rather than create them?   I thought there were no errors if we didn't map a MODS element at all, but if we map it partially, then what?

justinlittman:
I've tested this and it produces no new errors.

PR #1259, opened 2020-10-28
Adds scripts for creating cache and validating mapping to Cocina.

justinlittman:
## Why was this change made?
Provide tooling for testing Cocina mappings.

Locally.

README

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,3 +101,62 @@ To spin up a local rails console:
 To spin up a local development server:
 
   `bundle exec rails s`
+
+## Test Cocina mappings
+Given the tremendous heterogeneity of existing Fedora data, it is helpful to test mappings against actual production data. However, due to the slowness of retrieving from Fedora, creating a local cache of Fedora data is helpful.
+
+Several tools (described below) assist with creating a local cache and performing the testing.
+
+These tools are best run on a server within the network. (Currently installed on `sdr-deploy.stanford.edu`).
+
+### Setup
+1. Create a `settings.local.yml` containing:
+
+        ssl:
+          cert_file: "tls/certs/dor-services-prod-dor-prod.crt"
+          key_file: "tls/private/dor-services-prod-dor-prod.key"
+        
+        fedora_url: 'https://sul-dor-prod.stanford.edu/fedora'
+        
+        solr:
+          url: 'https://sul-solr.stanford.edu/solr/argo3_prod'
+2. Copy certificates locally:
+
+        scp -r <user from puppet>@<dor serices production host>:/etc/pki/tls .
+
+### Create a list of random druids
+```
+bin/generate-druid-list 100000
+```
+
+This will create `druids.txt`, containing a list of druids. If the file already exists, the existing druids will be used and new druids will be added to it.
+
+Note that the druids are unique.
+
+### Seed the cache
+```
+bin/generate-cache 100000
+```
+
+Using the druids from `druids.txt`, this will retrieve the item from Fedora and store the objects, datastreams, and disseminations in the `cache` directory.
+
+### Validate mapping to Cocina from Fedora
+```
+bin/validate-to-cocina 100000
+```
+
+Using the druids from `druids.txt` and the cache, this will map the Fedora item to the Cocina model and record any errors.
+
+Errors are returned ordered by the number of items that raised that error. For example:
+```
+41 of 500 (8.2%)
+Error: undefined method `text' for nil:NilClass (32 errors)
+Examples: druid:qb322bg3331, druid:vd938vg1826, druid:kc373zp2312, druid:gx497tb8747, druid:zq244qv7198, druid:pd738nx3263, druid:qm576pd0390, druid:tp743yv4651, druid:md775pw3650, druid:xw600sw0934
+Error: key not found: "Other version"
+Did you mean?  "otherVersion" (4 errors)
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -101,3 +101,62 @@ To spin up a local rails console:
...
+
+        ssh -L 8443:sul-solr.stanford.edu:443 dor_services@dor-services-prod.stanford.edu
+2. Create the list.
+
+        bin/generate-druid-list 100000
+
+This will create `druids.txt`, containing a list of druids. If the file already exists, the existing druids will be used and new druids will be added to it.
+
+Note that the druids are unique.
+
```
ndushay, line 137: perhaps include "remove Solr Tunnel" as a step -- we lost the Argo production index due to a tunnel left open in the past ...

README.md
```diff
@@ -101,3 +101,62 @@ To spin up a local rails console:
...
+          key_file: "tls/private/dor-services-prod-dor-prod.key"
+        
+        fedora_url: 'https://sul-dor-prod.stanford.edu/fedora'
+        
+        solr:
+          url: 'https://sul-solr.stanford.edu/solr/argo3_prod'
+2. Connect to VPN (if offsite).
+3. Copy certificates locally:
+
+        scp -r dor_services@dor-services-prod.stanford.edu:/etc/pki/tls .
```
ndushay, line 124: i suspect this is fine, but we've had some ops folks suggest it's better if we don't get this specific in a public repo

README.md
```diff
@@ -101,3 +101,62 @@ To spin up a local rails console:
...
+```
+bin/validate-to-cocina 100000
+```
+
+Using the druids from `druids.txt` and the cache, this will map the Fedora item to the Cocina model and record any errors.
+
+Errors are returned ordered by the number of items that raised that error. For example:
+```
+41 of 500 (8.2%)
+Error: undefined method `text' for nil:NilClass (32 errors)
```
ndushay, line 155: this is a great start.  In the future, maybe we can dig a little deeper and determine which class is giving the nil error (e.g. adminMetadata?  contributor?  subject?)

lib/rubydora_patch.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require 'active_support/gzip'
+
+# Patch Rubydora::RestApiClient to inject cache
```
ndushay, line 5: maybe "... to allow local caching of DOR content (for validating cocina mappings)" ?

ndushay (commented):
Looks pretty awesome.  I'm thinking there should be tests for the rubydora patch.

I'm making this a comment rather than a "request changes" so I can be easily overruled on my suggestions.

README.md
justinlittman: Yup. Added.

README.md
justinlittman: Obscured.

lib/rubydora_patch.rb
justinlittman: Done.

README.md
```diff
@@ -101,3 +101,63 @@ To spin up a local rails console:
...
+1. Create a `settings.local.yml` containing:
+
+        ssl:
+          cert_file: "tls/certs/dor-services-prod-dor-prod.crt"
+          key_file: "tls/private/dor-services-prod-dor-prod.key"
+        
+        fedora_url: 'https://sul-dor-prod.stanford.edu/fedora'
+        
+        solr:
+          url: 'https://sul-solr.stanford.edu/solr/argo3_prod'
```
ndushay, line 120: it might just be argo_prod now??   I think cbeer renamed it and argo3_prod is somehow a link?   not sure.

bin/generate-druid-list
```diff
@@ -0,0 +1,31 @@
...
+sample_size = Integer(ARGV[0])
+REQUEST_SIZE = 250
+
+OpenSSL::SSL::VERIFY_PEER = OpenSSL::SSL::VERIFY_NONE
+druids = Set.new
+
+druids.merge(File.read('druids.txt').split) if File.exist?('druids.txt')
+
+while druids.length < sample_size
+  results = ActiveFedora::SolrService.query('*:*', sort: 'random asc', fl: 'id', rows: REQUEST_SIZE)
```
ndushay, line 21: add `fq: '-has_model_ssim:"info:fedora/afmodel:Part"'` as a query argument here (to avoid ETD part objects)

ndushay: also, if you add wt=csv, then you'll get back something even easier to parse:

https://sul-solr-b.stanford.edu/solr/argo_prod/select?q=*:*&sort=random%20asc&fl=id&rows=10&fq=-has_model_ssim:%22info:fedora/afmodel:Part%22&wt=csv

```
id
druid:bk427kz4094
druid:yk877xp5181
druid:zz155sv2614
druid:qy805wh8168
druid:sw089sm8303
druid:ks499fs9755
druid:wx897vx0942
druid:xh465yn6146
druid:jv693gg2872
druid:nr368fk3363
```

ndushay:
Aside from my suggestions for the Solr query, I'd say let's get this merged and tweak it further in additional PRs.

ndushay (changes requested):
Happy to merge with the Solr query changes

README.md
justinlittman: Not sure. This is what was in prod config.

justinlittman:
@ndushay This is ready for re-review at your convenience.

ndushay (approved):

PR #1282, opened 2020-10-28
fix subject name mapping from fedora when name has no type

### LINKED ISSUE #1265: mapping validation error: key not found: nil (subject mapping)
## Backtrace

    line 88 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/subject.rb: fetch
    line 88 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/subject.rb: simple_item
    line 38 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/subject.rb: block in build

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/69242449)

ndushay:
## Why was this change made?

Fixes #1265 

Also updates references to mapping specification for xit statements.

Was able to reproduce the error locally with new unit test, and the fix addressed it.

### DIFF
--- a/spec/services/cocina/from_fedora/descriptive/subject_spec.rb
+++ b/spec/services/cocina/from_fedora/descriptive/subject_spec.rb
@@ -529,6 +529,33 @@ RSpec.describe Cocina::FromFedora::Descriptive::Subject do
     context 'with a name-title subject with additional terms including genre subdivision, authority for terms' do
       xit 'TODO https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_subject.txt#L507'
     end
+
+    context 'without name type' do
+      let(:xml) do
+        <<~XML
+          <subject authority="naf" authorityURI="http://id.loc.gov/authorities/names" valueURI="http://id.loc.gov/authorities/names/n81070667">
+            <name>
+              <namePart>Stanford University. Libraries.</namePart>
+            </name>
+          </subject>
+        XML
+      end
+
+      it 'builds the cocina data structure' do
+        expect(build).to eq [
+          {
+            source:
+              {
+                code: 'naf',
+                uri: 'http://id.loc.gov/authorities/names'
+              },
+            type: 'name',
+            uri: 'http://id.loc.gov/authorities/names/n81070667',
+            value: 'Stanford University. Libraries.'
+          }
+        ]
+      end
+    end
   end
 
   context 'with a geographic subject subdivision' do
--- a/spec/services/cocina/from_fedora/descriptive/subject_spec.rb
+++ b/spec/services/cocina/from_fedora/descriptive/subject_spec.rb
@@ -428,14 +428,17 @@ RSpec.describe Cocina::FromFedora::Descriptive::Subject do
           }
         ]
       end
+
+      # Example 2 w dates
+      xit 'TODO https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_subject.txt#L224'
     end
 
     context 'with additional terms and authority for terms' do
-      xit 'TODO https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_subject.txt#L247'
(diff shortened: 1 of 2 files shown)


spec/services/cocina/from_fedora/descriptive/subject_spec.rb
```diff
@@ -524,7 +527,32 @@
     end
 
     context 'with a name-title subject with additional terms including genre subdivision, authority for terms' do
-      xit 'TODO https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_subject.txt#L429'
+      xit 'TODO https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_subject.txt#L507'
+    end
+
+    context 'without name type' do
```
jcoyne, line 533: Did you run this by @arcadiafalcone?  Did she agree that this is not a data error? Does an example need to be added into https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_subject.txt?

jcoyne (changes requested):
After talking with Arcadia this sounds like a data error, so maybe we shouldn't apply this change.

ndushay:
I am closing this PR, as you were on top of spotting this as a data error rather than a mapping error!   Thank you.

ndushay:
Hrmm - I need to look at more examples -- it may be the one I chose to use as a spec was a data error but I don't think they all are.

ndushay:
@jcoyne fixed the example in the spec to be data that has the problem, but isn't a subject masquerading as a name.

spec/services/cocina/from_fedora/descriptive/subject_spec.rb
```diff
@@ -524,7 +527,34 @@
...
+      let(:xml) do
+        <<~XML
+          <subject authority="naf" authorityURI="http://id.loc.gov/authorities/names" valueURI="http://id.loc.gov/authorities/names/n81070667">
+            <name>
```
jcoyne, line 537: Have we asked @arcadiafalcone if this shouldn't have `type="organization"` for this to be valid?

jcoyne (changes requested):

spec/services/cocina/from_fedora/descriptive/subject_spec.rb
ndushay: nope, I haven't talked to @arcadiafalcone .  We clearly have data like this, tho.  JLitt trapped 55 errors in his random sample ...

This fix avoids errors ... but whether it also needs to notify honeybadger for bad data ... I couldn't say.

ndushay:
@jcoyne ready for re-review

spec/services/cocina/from_fedora/descriptive/subject_spec.rb
justinlittman: From @arcadiafalcone:
> ideally type="personal" would be present, but it's not required. i would like to remediate these, but it's valid mods to not have it.

So perhaps we should HB alert when no type?

arcadiafalcone: Yes, alert for it but let it pass. 64 out of 373 nil_key errors I analyzed were bad data for reasons other than the missing type, and the rest were just missing the attribute. Missing the type doesn't make the records invalid MODS, but it will cause problems generating data in other schemas down the line and can be a symptom of something else wrong with the data.

jcoyne (approved):

ndushay (commented):
Just pushed commit that alerts honeybadger when subject name has no type attribute.

spec/services/cocina/from_fedora/descriptive/subject_spec.rb
ndushay: Just pushed commit that alerts honeybadger when subject name has no type attribute.

ndushay:
@jcoyne @justinlittman this now alerts Honeybadger when there is no type attribute on a name subject.

jcoyne (approved):

justinlittman:
Verified that this reduces errors.

PR #1297, opened 2020-10-30
Use progressbar for validate-to-cocina

jcoyne:
## Why was this change made?
Show pretty progress.


mjgiarlo (commented):
:sparkles: 

Would like @justinlittman's input before merging. OK with a visual representation over numerical?

jcoyne:
@justinlittman great, I made that change.

PR #1298, opened 2020-10-30
add info on how to test mappings on sdr-deploy

peetucket:
## Why was this change made?

Add some info to the readme on how to run cocina model validations on a server.


peetucket:
have another look @ndushay ?

ndushay (approved):

Issue #1274, opened 2020-10-28
mapping validation error: key not found: "#N/A" (subject?)


justinlittman:
NOTE:  Arcadia says below (https://github.com/sul-dlss/dor-services-app/issues/1274#issuecomment-718993850):

>For error handling, treat "#N/A" as blank and report the error. I expect this error will primarily affect authority attributes but it may turn up other places as well.

Two actions needed by engineer:

1. the mapping code is modified to drop the attribute when its value is #N/A and send Honeybadger a notification (see https://github.com/sul-dlss/dor-services-app/pull/1282/files# for an example notification)
2. assign this ticket back to Arcadia once the error handling is added to the code and she'll follow through with remediation.

**Do not have PR close this ticket -- mention the ticket # but don't use github magic words in PR description**

------ original ticket below -----

Backtrace

    line 88 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/subject.rb: fetch
    line 88 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/subject.rb: simple_item
    line 38 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/subject.rb: block in build

PR #1300, opened 2020-10-30
Handle subject with #N/A authority or subject-name with #N/A types

jcoyne:
## Why was this change made?

Ref #1274

On deploy server:
```
795 of 350000 (0.22714285714285715%)%)
```


ndushay (approved):
we have N/A in two diff spots? yuck.

Issue #1290, opened 2020-10-29
mapping validation error:  (contributor) ./mods:role/mods:roleTerm[@type="code"] is missing required authority attribute


ndushay:
NOTE:  Arcadia says below (https://github.com/sul-dlss/dor-services-app/issues/1290#issuecomment-718997876):

> If there's a valueURI without an authority attribute, that's valid and doesn't need to be reported. If the value is three letters, let it pass without the authority attribute and report it as an error. If the value is not three letters, something unpredictable is wrong with the data and it should fail.

Two actions needed by engineer:

1. the mapping code is modified per Arcadia's instructions, and if requested, sends Honeybadger a notification (see https://github.com/sul-dlss/dor-services-app/pull/1282/files# for an example notification)
2. assign this ticket back to Arcadia once the error handling is added to the code and she'll follow through with data remediation.

**Do not have PR close this ticket -- mention the ticket # but don't use github magic words in PR description**

------ original ticket below -----

JLitt's most recent run of validator (from fedora to cocina mappings) for 325K random prod objects:

```
Error: ./mods:role/mods:roleTerm[@type="code"] is missing required authority attribute (19 errors)
druid:cr391gq7466
druid:ds919sg5985
druid:jq959xy6710
druid:nf676xt4970
druid:jn867nx7451
druid:nc018xn7579
druid:tr882hg3507
druid:tr870pd7774
druid:zc221rc0493
druid:ds846xd1044
druid:fw863hr6225
druid:bs056js0016
druid:rx675br5706
druid:mw965fm4175
druid:sb590rq6970
druid:fj037zf2211
druid:ck132wb6800
druid:my683pp9971
druid:jm834bv7649
```

It's possible the mapping needed for this error is related to the missing mappings in #1251

arcadiafalcone:
If there's a valueURI without an authority attribute, that's valid and doesn't need to be reported. If the value is three letters, let it pass without the authority attribute and report it as an error. If the value is not three letters, something unpredictable is wrong with the data and it should fail.

PR #1301, opened 2020-10-30
Fixes handling of role code.

justinlittman:
addresses problem in #1290

## Why was this change made?
To satisfy the requirement:
```
If there's a valueURI without an authority attribute, that's valid and doesn't need to be reported. If the value is three letters, let it pass without the authority attribute and report it as an error. If the value is not three letters, something unpredictable is wrong with the data and it should fail.
```

Unit

NA


app/services/cocina/from_fedora/descriptive/contributor.rb
```diff
@@ -110,7 +110,6 @@ def names
 
         # rubocop:disable Metrics/AbcSize
         # rubocop:disable Metrics/CyclomaticComplexity
-        # rubocop:disable Metrics/PerceivedComplexity
```
jcoyne, line 113: 👏

app/services/cocina/from_fedora/descriptive/contributor.rb
```diff
@@ -134,13 +133,23 @@ def roles_for(name)
...
 
         def type_for(type)
           Honeybadger.notify('[DATA ERROR] Contributor type incorrectly capitalized', { tags: 'data_error' }) if type.downcase != type
           ROLES.fetch(type.downcase)
         end
+
+        def check_code(role_code, role_authority)
+          return if role_code.nil? || role_authority
+
+          if role_code&.content.present? && role_code.content.size == 3
```
jcoyne, line 146: I don't think you need `&.` here because you've already checked that it's not nil.

jcoyne (changes requested):

app/services/cocina/from_fedora/descriptive/contributor.rb
```diff
@@ -134,13 +133,23 @@ def roles_for(name)
...
+            Honeybadger.notify('[DATA ERROR] Contributor role code is missing authority', { tags: 'data_error' })
+            return
+          end
+
+          raise Cocina::Mapper::InvalidDescMetadata, "Contributor role code is missing and has unexpected value: #{role_code&.content}"
```
jcoyne, line 151: Nil check isn't needed here either.

jcoyne (changes requested):

ndushay:
Please assign #1290 to Arcadia after this is merged.

app/services/cocina/from_fedora/descriptive/contributor.rb
justinlittman: Fixed.

app/services/cocina/from_fedora/descriptive/contributor.rb
justinlittman: Fixed.

.rubocop_todo.yml
```diff
@@ -23,30 +23,30 @@ Lint/SuppressedException:
...
 # Configuration parameters: IgnoredMethods.
 Metrics/AbcSize:
   Max: 38
 
-# Offense count: 4
+# Offense count: 7
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 119
+  Max: 118
```
jcoyne, line 34: 👏

.rubocop_todo.yml
```diff
@@ -23,30 +23,30 @@ Lint/SuppressedException:
...
 # Configuration parameters: CountComments, CountAsOne, ExcludedMethods.
 Metrics/MethodLength:
   Max: 43
 
-# Offense count: 8
+# Offense count: 16
 # Configuration parameters: IgnoredMethods.
 Metrics/PerceivedComplexity:
-  Max: 13
+  Max: 15
```
jcoyne, line 49: 😢  Can you add an exception on one method rather than increasing the threshold for all methods?

jcoyne (changes requested):

jcoyne (approved):

PR #1302, opened 2020-10-30
ObjectsController#show: handle all Cocina::Mapper.build exceptions as 422

### LINKED ISSUE #1284: when showing an item that has a descriptive conversion error, return an explanation
Return a 422 (or something else?) with a JSON-API formatted error, and above all, avoid returning a generic 500 error with no explanation

This will show a more helpful message in the workflow service.

jmartin-sul:
## Why was this change made?

Handle unexpected errors from `Cocina::Mapper.build` so that an informative 422 is returned (with a ~~backtrace filtered to dor-services-app code~~full backtrace because my first attempt at filtering was hacky and this seemed useful as-is).  This will hopefully help future `Cocina::Mapper` debugging.

closes #1284

* supported by a bit more hierarchy for `Cocina::Mapper` exceptions
* `ObjectsController#json_api_error`: param for optional meta field in JSON-API error
  * `#show`: use meta field for returning backtrace
* test the new functionality
* update openapi spec for new json api error field

unit tests

hopefully code comments and new test suffice

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
 class ObjectsController < ApplicationController
   before_action :load_item, except: [:create]
 
@@ -61,6 +62,11 @@ class ObjectsController < ApplicationController
     json_api_error(status: :unprocessable_entity,
                    title: 'Invalid descMetadata',
                    message: e)
+  rescue Cocina::Mapper::UnexpectedBuildError => e
+    json_api_error(status: :unprocessable_entity,
+                   title: 'Unexpected Cocina::Mapper.build error',
+                   message: e.cause,
+                   meta: { backtrace: e.cause.backtrace })
   end
 
   # Initialize specified workflow (assemblyWF by default), and also version if needed
@@ -125,7 +131,7 @@ class ObjectsController < ApplicationController
 
   private
 
-  def json_api_error(status:, message:, title: nil)
+  def json_api_error(status:, message:, title: nil, meta: nil)
     status_code = Rack::Utils.status_code(status)
     render status: status,
            content_type: 'application/vnd.api+json',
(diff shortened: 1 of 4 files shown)


jmartin-sul:
CI sussed out some brittle code, pushing a fix momentarily...

app/controllers/objects_controller.rb
```diff
@@ -135,7 +140,9 @@ def json_api_error(status:, message:, title: nil)
                  'status': status_code.to_s,
                  'title': title || Rack::Utils::HTTP_STATUS_CODES[status_code],
                  'detail': message
-               }
+               }.tap do |h|
+                 h[:backtrace] = backtrace.join("\n") if backtrace
```
jcoyne, line 144: This field is not compliant with JSON-API (see https://jsonapi.org/format/1.1/#error-objects) maybe push it into the `detail`.

jcoyne (changes requested):

app/controllers/objects_controller.rb
jcoyne: It looks like `meta` might be another option (but you should then add it to the openapi.yml.

jmartin-sul:
ok @jcoyne, i switched to returning a more structured backtrace array wrapped in the `meta` field, cleaned up the test to take advantage of the structure, and attempted to update the openapi spec to match the change.  ready for another look, i think.

app/services/cocina/mapper.rb
```diff
@@ -3,17 +3,24 @@
...
     # Raised when this object is missing a sourceID, so it can't be mapped to cocina.
-    class MissingSourceID < StandardError; end
+    class MissingSourceID < BuildError; end
 
     # Raised when assumptions about descMetadata are violated
-    class InvalidDescMetadata < StandardError; end
+    class InvalidDescMetadata < BuildError; end
+
+    # Raised on unexpected mapper failures, so .build errors can be rescued without rescuing all of StandardError
+    class UnexpectedBuildError < BuildError; end
```
jcoyne, line 23: Why not just rescue StandardError, since you wrap all StandardErrors with this class?

jcoyne (changes requested):

app/services/cocina/mapper.rb
jmartin-sul: i didn't want to catch errors from either of the lines above `Cocina::Mapper.build` since the ticket seemed to be asking for handling of "descriptive conversion" which i read (maybe incorrectly?) as being about the `.build` call.

happy to simplify and just catch `StandardError`.

jcoyne (approved):

Issue #1293, opened 2020-10-29
mapping validation error: key not found: "naf"


ndushay:
NOTE:  Arcadia says (https://github.com/sul-dlss/dor-services-app/issues/1292#issuecomment-719002741):

> ignore the invalid type attribute and report the error.

Two actions needed by engineer:

1. the mapping code is modified per Arcadia's instructions, and if requested, sends Honeybadger a notification (see https://github.com/sul-dlss/dor-services-app/pull/1282/files# for an example notification)
2. assign this ticket back to Arcadia once the error handling is added to the code and she'll follow through with data remediation.

**Do not have PR close this ticket -- mention the ticket # but don't use github magic words in PR description**

------ original ticket below -----

from JLitt's most recent run of validator (from fedora to cocina mappings) for 325K random prod objects:

```
Error: key not found: "naf" (1 errors)
druid:pw260mt1867
```

arcadiafalcone:
This is the same problem as in #1292.

ndushay:
code greased for thes by PR #1309

Issue #1292, opened 2020-10-29
mapping validation error: key not found: "primary"


ndushay:
NOTE:  Arcadia says below (https://github.com/sul-dlss/dor-services-app/issues/1292#issuecomment-719002741):

> ignore the invalid type attribute and report the error.

Two actions needed by engineer:

1. the mapping code is modified per Arcadia's instructions, and if requested, sends Honeybadger a notification (see https://github.com/sul-dlss/dor-services-app/pull/1282/files# for an example notification)
2. assign this ticket back to Arcadia once the error handling is added to the code and she'll follow through with data remediation.

**Do not have PR close this ticket -- mention the ticket # but don't use github magic words in PR description**

------ original ticket below -----

from JLitt's most recent run of validator (from fedora to cocina mappings) for 325K random prod objects:

```
Error: key not found: "primary" (3 errors)
druid:rw667tg8009
druid:md049mq5771
druid:wq793rc8271
```

~~This *may* have to do with "primary" belonging on "language" element rather than "languageTerm" element.  See https://github.com/sul-dlss/dor-services-app/issues/1090#issuecomment-715532802 and sul-dlss-labs/cocina-descriptive-metadata/issues/99~~

arcadiafalcone:
This isn't connected to the language issue, but rather an invalid name type. To handle, ignore the invalid type attribute and report the error.

ndushay:
Examples: druid:rw667tg8009, druid:md049mq5771, druid:wq793rc8271, druid:sd836bn3355, druid:tj585kp4051, druid:bj241ch3330

ndushay:
code greased for thes by PR #1309

PR #1309, opened 2020-11-03
name mapping fix for unrecognized type attribute value

ndushay:
## Why was this change made?

To add mapping grease for #1292 (`<name>` has invalid type attribute of 'primary') and #1293 (`<name>` has invalid type attribute of 'naf')

- wrote unit tests for these cases
- with from-fedora-to-cocina validator

    ```
    $ diff results-master-branch.txt results-name-type-unrec-branch.txt 
    1c1
    < 723 of 350000 (0.20657142857142857%)
    ---
    > 719 of 350000 (0.20542857142857143%)
    722,725d721
    < Error: key not found: "primary" (3 errors)
    < druid:rw667tg8009
    < druid:md049mq5771
    < druid:wq793rc8271
    731,732d726
    < Error: key not found: "naf" (1 errors)
    < druid:pw260mt1867
    ```

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/contributor.rb
+++ b/app/services/cocina/from_fedora/descriptive/contributor.rb
@@ -155,7 +155,12 @@ module Cocina
         end
 
         def type_for(type)
+          unless Contributor::ROLES.keys.include?(type.downcase)
+            Honeybadger.notify("[DATA ERROR] Contributor type unrecognized '#{type}'", { tags: 'data_error' })
+            return
+          end
           Honeybadger.notify('[DATA ERROR] Contributor type incorrectly capitalized', { tags: 'data_error' }) if type.downcase != type
+
           ROLES.fetch(type.downcase)
         end
       end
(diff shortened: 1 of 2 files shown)


.rubocop_todo.yml
```diff
@@ -23,27 +23,27 @@ Lint/SuppressedException:
...
 # Configuration parameters: IgnoredMethods.
 Metrics/AbcSize:
   Max: 38
 
-# Offense count: 4
+# Offense count: 9
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 119
+  Max: 122
```
jcoyne, line 34: Please don't increase this for every class in the project. Just add an exception for the class you are working on if you don't want to refactor.

jcoyne (changes requested):

.rubocop_todo.yml
ndushay: i figured in this one case, because it was only going up by 3, it made sense.  If it had been a bigger increase, I would have done as you suggested.  I can put the exception in the class.

I suspect the _todo file hadn't been regenned in some time, as a bunch of other occurrence counts went up.  :-(

ndushay:
@jcoyne redid the rubocop commit.

jcoyne (approved):
Thank you!

PR #1317, opened 2020-11-04
do not add nil for nameIdentifiers source

### LINKED ISSUE #1310: FromFedora mapping error: #/components/schemas/Source/properties/code does not allow null values
```
Error: #/components/schemas/Source/properties/code does not allow null values (3 errors)
```

Examples: druid:dz692fn7184, druid:qw833gg0782, druid:jb294tg1139

peetucket:
## Why was this change made?

Fix #1310 

New unit test; on the console

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/contributor.rb
+++ b/app/services/cocina/from_fedora/descriptive/contributor.rb
@@ -98,8 +98,10 @@ module Cocina
           identifier = name.xpath('mods:nameIdentifier', mods: DESC_METADATA_NS).first
           return unless identifier
 
+          identifier_type = identifier['type']
+          source = { code: identifier_type } if identifier_type
           type = 'URI' if URI::DEFAULT_PARSER.make_regexp.match?(identifier.text)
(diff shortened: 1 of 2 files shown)


.rubocop_todo.yml
```diff
@@ -31,7 +31,7 @@ Metrics/AbcSize:
 # Offense count: 4
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 119
+  Max: 120
```
jcoyne, line 34: Would you mind adding the exception to the one file that needs it, rather than raising the bar globally?

jcoyne (changes requested):

.rubocop_todo.yml
peetucket: Sure, disabled this cop for only that single class

peetucket: Oh actually it looks like that was done in a separate PR.  I rebased and got it in.

jcoyne (approved):

PR #1323, opened 2020-11-04
Remove double sidekiq config block

mejackreed:
While doing some sidekiq sleuthing I say this.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -3,7 +3,3 @@
 Sidekiq.configure_server do |config|
   config.redis = { url: Settings.redis_url }
 end
-
-Sidekiq.configure_client do |config|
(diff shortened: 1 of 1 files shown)


jcoyne (approved):
Thanks.

config/initializers/sidekiq.rb
```diff
@@ -3,7 +3,3 @@
 Sidekiq.configure_server do |config|
   config.redis = { url: Settings.redis_url }
 end
-
-Sidekiq.configure_client do |config|
```
jcoyne, line 7: This is for the client, and the other is for the server, so is it really double?

mjgiarlo: @mejackreed thoughts on this?

mejackreed: Oops! I guess I missed that :)

PR #1329, opened 2020-11-06
Adds support for overwriting to cache. Changes validate-cocina-roundt…

justinlittman:
…rip to use cache and write results to file.

## Why was this change made?
To enhance ability to test mappings.

Locally.

README

### DIFF
--- a/bin/generate-cache
+++ b/bin/generate-cache
@@ -4,14 +4,15 @@
 require_relative '../config/environment'
 require_relative '../lib/fedora_cache'
 
-if ARGV.length != 1
-  puts 'Usage: bin/generate-cache <SAMPLE_SIZE>'
+if ARGV.length != 2
+  puts 'Usage: bin/generate-cache <SAMPLE_SIZE> <OVERWRITE: true/false>'
   exit(false)
 end
 
 sample_size = Integer(ARGV[0])
+overwrite = ARGV[1] == 'true'
 
-cache = FedoraCache.new
+cache = FedoraCache.new(overwrite: overwrite)
 
 druids = File.read('druids.txt').split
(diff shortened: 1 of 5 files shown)


bin/generate-cache
```diff
@@ -4,14 +4,15 @@
 require_relative '../config/environment'
 require_relative '../lib/fedora_cache'
 
-if ARGV.length != 1
-  puts 'Usage: bin/generate-cache <SAMPLE_SIZE>'
+if ARGV.length != 2
+  puts 'Usage: bin/generate-cache <SAMPLE_SIZE> <OVERWRITE: true/false>'
```
ndushay, line 8: should it have a default value of false?

bin/validate-cocina-roundtrip
```diff
@@ -2,47 +2,167 @@
...
 
 require_relative '../config/environment'
+require_relative '../lib/fedora_cache'
+require 'set'
 
-def round_tripped_xml(item)
-  title_builder = Cocina::FromFedora::Descriptive::TitleBuilderStrategy.find(label: item.label)
-  desc_props = Cocina::FromFedora::Descriptive.props(title_builder: title_builder, mods: item.descMetadata.ng_xml)
+if ARGV.length != 2
+  puts 'Usage: bin/validate-cocina-roundtrip <SAMPLE_SIZE> <SHUFFLE: true/false>'
```
ndushay, line 9: should shuffle (which might be renamed random?) have a default value?

ndushay (approved):
looks great.  Just a question on whether the 2nd arguments should have a default (and if so, would we still pass in a boolean, or would we pass in "overwrite" or even use a switch)

PR #1331, opened 2020-11-06
Add directions on generating a full list of druids.

justinlittman:
## Why was this change made?
To document how to generate a full list of druids.

NA

README.


README.md
```diff
@@ -133,6 +133,16 @@ This will create `druids.txt`, containing a list of druids. If the file already
 
 Note that the druids are unique.
 
+### Create a list of all druids
+Alternatively, a comprehensive list of druids can be generated by querying Fedora's DB.
+
+On the Fedora server, get the MySQL password from `~/.fedora.my.cnf`.
+
+```
+mysql --user=fedora --password fedora -e "select doPID from doRegistry where doPID like 'druid:%';" -N | cut -c 1-17 > druids.txt
```
ndushay, line 142: is the "cut" part getting rid of noise lines in the output?  Also wondering if this is more of a wiki thing since it's not really part of this app, but a part of Fedora.

ndushay (changes requested):

README.md
justinlittman: It is getting rid of table formatting noise.

Which wiki would you recommend?

ndushay: wiki for this app would be fine with me, or ???   @jmartin-sul @jcoyne @mjgiarlo have any feelings about where this documentation should live?

mjgiarlo: I think it's OK for the README given it's related to the prior section, but I don't have strong feelings on this.

jcoyne: 👎 to wikis. Put it in the repo (not necessarily the readme)

mjgiarlo:
@ndushay are you cool with this going forward until we have a better place to stash this info?

ndushay (approved):
whatever.  I dunno where to stash it.  devopsdocs?  readme on sul-dor-xxx ?

PR #1354, opened 2020-11-10
Object registration/deposit wired up to use ToFedora mappings

ndushay:
part of sul-dlss/happy-heron/issues/240

## Why was this change made?

As part of H2 work -- we want h2 data with cocina descriptive metadata to be mapped into Fedora.

I wrote missing tests for code before I made changes (e.g. item vs collection object, different ways title was getting set, the way we set the abstract for collection records coming from Argo), and I made sure those tests passed along with tests showing that incoming descriptive metadata was being mapped.

Also deployed to qa to test.

### DIFF
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -108,6 +108,8 @@ module Cocina
       obj.identification&.catalogLinks&.find { |l| l.catalog == 'symphony' }&.catalogRecordId
     end
 
+    # @param [Dor::[Item|Collection|APO]] item
+    # @param [Cocina:Models::Request[DOR|Collection|xxx]] obj
     # @raises SymphonyReader::ResponseError if symphony connection failed
     def add_description(item, obj)
       # Hydrus doesn't set description. See https://github.com/sul-dlss/hydrus/issues/421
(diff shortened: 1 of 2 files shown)


app/services/cocina/object_creator.rb
```diff
@@ -120,13 +122,11 @@ def add_description(item, obj)
...
       elsif obj.description
-        item.descMetadata.mods_title = obj.description.title.first.value
+        item.descMetadata.content = Cocina::ToFedora::Descriptive.transform(obj.description).to_xml
+        item.descMetadata.content_will_change!
       else
         item.descMetadata.mods_title = obj.label
       end
-
-      # collections registered via Argo have abstracts, which appear here as note of type summary
-      item.descMetadata.abstract = obj.description.note.first.value if obj.description&.note&.first&.type == 'summary'
```
ndushay, line 129: this is covered by the ToFedora mapping code (I wrote a test to prove it)

app/services/cocina/object_creator.rb
```diff
@@ -120,13 +122,11 @@ def add_description(item, obj)
         item.label = truncate_label(label)
         item.objectLabel = label
       elsif obj.description
-        item.descMetadata.mods_title = obj.description.title.first.value
+        item.descMetadata.content = Cocina::ToFedora::Descriptive.transform(obj.description).to_xml
+        item.descMetadata.content_will_change!
       else
         item.descMetadata.mods_title = obj.label
```
ndushay, line 128: this still works (I wrote a test to prove it)

spec/services/cocina/object_creator_spec.rb
```diff
@@ -3,10 +3,11 @@
 require 'rails_helper'
 
 RSpec.describe Cocina::ObjectCreator do
-  subject(:item) { described_class.create(request, persister: persister) }
+  subject(:result) { described_class.create(request, persister: persister) }
```
ndushay, line 6: it's sometimes a collection, or an APO ...

ndushay (commented):
Do we need a test for an APO object to make sure they don't have a problem with the change?

jcoyne (approved):

ndushay:
With this branch deployed in stage, Andrew tested various methods of creating objects and signed off;  I ran the infrastructure-integration tests and all ran green except h2, which i have manually confirmed works.  (a PR to fix h2 integration test is coming)

mjgiarlo (approved):

PR #1363, opened 2020-11-11
Add support for providing a list of druids to generate-cache.

justinlittman:
## Why was this change made?
Make it easier to generate cache for a list of druids or a single druid.

Locally

README


ndushay (approved):
can you add this to the README, too?   Not sure if other examples are warranted ...

(from slack)
```
$ bin/generate-cache -d druid:bh164hd2167
druid:bh164hd2167 (1)
$ bin/validate-cocina-roundtrip -d druid:bh164hd2167
```

justinlittman:
@ndushay Added to README.

ndushay:
thank you!

PR #1370, opened 2020-11-11
Map related-to null type for H2 related works and links

mjgiarlo:
## Why was this change made?

Connects to https://app.honeybadger.io/projects/50568/faults/69893421 which is causing all deposits to fail in stage.

CI

None

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/related_resource.rb
+++ b/app/services/cocina/to_fedora/descriptive/related_resource.rb
@@ -14,6 +14,7 @@ module Cocina
           'in series' => 'series',
           'part of' => 'host',
           'preceded by' => 'preceding',
+          'related to' => nil, # 'related to' is a null type by design
           'reviewed by' => 'reviewOf',
           'referenced by' => 'isReferencedBy',
           'references' => 'references',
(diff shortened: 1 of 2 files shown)


app/services/cocina/to_fedora/descriptive/related_resource.rb
```diff
@@ -33,7 +33,8 @@ def initialize(xml:, related_resources:)
         def write
           Array(related_resources).each do |related|
             attributes = {}
-            attributes[:type] = TYPES.fetch(related.type) if related.type
+            # NOTE: 'related to' is a null type, in effect
```
jcoyne, line 36: Could we add a 'related to' type into the map that returns a nil? I feel like this should be in that mapping.  Then call `attributes.compact` on line 39.

jcoyne (changes requested):

app/services/cocina/to_fedora/descriptive/related_resource.rb
mjgiarlo: sure.

spec/services/cocina/to_fedora/descriptive/related_resource_spec.rb
```diff
@@ -117,6 +117,71 @@
     end
   end
 
+  context 'when it has a related item with the generic "related to" type (related link from H2)' do
```
ndushay, line 120: is the the same as the next context statement, or am I missing the difference?

app/services/cocina/to_fedora/descriptive/related_resource.rb
mjgiarlo: done. take another look?

spec/services/cocina/to_fedora/descriptive/related_resource_spec.rb
mjgiarlo: not the same. just very similar: related link vs. related work

jcoyne (approved):

Issue #1367, opened 2020-11-11
To Fedora error: undefined method `value' for nil:NilClass


justinlittman:
For druid:bh164hd2167:
```
Error: undefined method `value' for nil:NilClass

Expected XML:
<?xml version="1.0" encoding="UTF-8"?>
<mods xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.loc.gov/mods/v3" version="3.3" xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-3.xsd">
  <identifier type="local" displayLabel="reference_id">sul:MOA0407</identifier><!-- from CSV file note MDToolkit -->
  <identifier type="local" displayLabel="norwich_id">NOR </identifier>
  <identifier type="local" displayLabel="object_id">10407</identifier>
  <titleInfo type="main">
    <title>A Map of the Albert N'yanza and of the Routes leading to its Discovery in 1864 by Sir Samuel White Baker.</title>
  </titleInfo>
  <titleInfo type="translated">
    <title/>
  </titleInfo>
  <originInfo>
    <publisher>Macmillan &amp; Co.</publisher>
    <place>
      <placeTerm type="text">London</placeTerm>
    </place>
    <dateCreated>1864-65 ?</dateCreated>
  </originInfo>
  <physicalDescription>
    <extent>33 cm. x 49 cm.</extent>
    <note displayLabel="color">Colored</note>
  </physicalDescription>
    <subject>
      <cartographics>
        <scale/>
      </cartographics>
    </subject>
  <subject>
    <geographic>Africa--Maps</geographic>
  </subject>  	
  <subject>
    <hierarchicalGeographic authority="lcsh"><continent>Africa</continent></hierarchicalGeographic>
  </subject>
    <subject>
      <geographicCode authority="marcgac">fc</geographicCode>
    </subject>
    <subject>
      <geographic authority="marcgac">Africa, Central</geographic>
    </subject>
    <subject>
      <geographic authority="lcsh">Central Africa</geographic>
    </subject>
  <subject>
    <temporal>19th c.</temporal>
  </subject>
  <typeOfResource>cartographic</typeOfResource>
  <name type="personal">
    <namePart>Baker, Samuel White, Sir, 1821-1893</namePart>
    <role>
      <roleTerm type="code" authority="marcrelator">ctg</roleTerm>
[...]

PR #1374, opened 2020-11-12
Adds mapping of note displayLabel attr to cocina. Ignore blank notes …

justinlittman:
…when mapping to cocina.

refs #1367

## Why was this change made?
Improve mapping of notes to cocina.

Unit

NA

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/notes.rb
+++ b/app/services/cocina/from_fedora/descriptive/notes.rb
@@ -31,10 +31,11 @@ module Cocina
         end
 
         def notes
-          set = ng_xml.xpath('//mods:note', mods: DESC_METADATA_NS)
+          set = ng_xml.xpath('//mods:note', mods: DESC_METADATA_NS).select { |node| node.text.present? }
           set.map do |node|
             { value: node.text }.tap do |attributes|
               attributes[:type] = node[:type] if node[:type]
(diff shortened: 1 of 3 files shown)


justinlittman:
Holding as a draft until clarification on how to handle blank notes.

app/services/cocina/from_fedora/descriptive/notes.rb
```diff
@@ -33,10 +33,13 @@ def abstract
         def notes
           set = ng_xml.xpath('//mods:note', mods: DESC_METADATA_NS)
           set.map do |node|
+            next nil if node.text.blank?
```
mjgiarlo, line 36: Perhaps six-of-one-half-dozen-of-the-other, but you could also consider a different iterator here instead of injecting and then stripping out `nil` values, like:

```ruby
ng_xml.xpath('//mods:note', mods: DESC_METADATA_NS).select { |node| node.text.present? }.map do |node|
  # ...
end
```

Non-blocking suggestion.

mjgiarlo (approved):

app/services/cocina/from_fedora/descriptive/notes.rb
justinlittman: Thanks for the suggestion. Made the change.

PR #1439, opened 2020-11-18
[Refactor] .each is fine when you aren't using the index

ndushay:
## Why was this change made?

I noticed this on the way to doing something else.  I think I'm the guiltiest party, too.

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/note.rb
+++ b/app/services/cocina/to_fedora/descriptive/note.rb
@@ -8,9 +8,9 @@ module Cocina
         # @params [Nokogiri::XML::Builder] xml
         # @params [Array<Cocina::Models::DescriptiveValue>] notes
         def self.write(xml:, notes:)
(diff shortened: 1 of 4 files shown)


aaron-collier (approved):

app/services/cocina/to_fedora/descriptive/note.rb
```diff
@@ -8,9 +8,9 @@ class Note
         # @params [Nokogiri::XML::Builder] xml
         # @params [Array<Cocina::Models::DescriptiveValue>] notes
         def self.write(xml:, notes:)
-          Array(notes).each_with_index do |note, alt_rep_group|
+          Array(notes).each_with_index do |note, index|
```
jcoyne, line 11: I think that `alt_rep_group` is a much more descriptive name than `index`

ndushay: it is more descriptive than index, but ... in the context of the method that uses it that way, not here.

Issue #1404, opened 2020-11-16
Map relatedItem form


justinlittman:
For druid:cy979mw6316, expected:
```
   <mods:relatedItem type="original">
      <mods:physicalDescription>
         <mods:form authority="marcform">print</mods:form>
         <mods:extent>v. ; 24 cm.</mods:extent>
      </mods:physicalDescription>
   </mods:relatedItem>
```

but received:

```
<relatedItem type="original">
    <physicalDescription>
      <extent>v. ; 24 cm.</extent>
    </physicalDescription>
  </relatedItem>
 ```

Issue #1403, opened 2020-11-16
Map relatedItem identifier


justinlittman:
For druid:cy979mw6316, expected:
```
   <mods:relatedItem type="otherFormat">
      <mods:titleInfo>
         <mods:title>[Serial set]</mods:title>
      </mods:titleInfo>
      <mods:name>
         <mods:namePart>United States. Congress.</mods:namePart>
      </mods:name>
      <mods:identifier type="local">(OCoLC)4978913</mods:identifier>
   </mods:relatedItem>

```

but received:

```
  <relatedItem type="otherFormat">
    <titleInfo>
      <title>[Serial set]</title>
    </titleInfo>
    <name>
      <namePart>United States. Congress.</namePart>
    </name>
  </relatedItem>
```

Note the missing identifier.

justinlittman:
Depends on https://github.com/sul-dlss-labs/cocina-descriptive-metadata/issues/142

Issue #1412, opened 2020-11-16
Map location


justinlittman:
See https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_location.txt

For example, druid:dp434sf5680

justinlittman:
Re-enable `Cocina::ToFedora::Descriptive::RelatedResource when it has a related item with the generic "related to" type (related link from H2) builds the xml` once completed.

justinlittman:
And re-enable `Cocina::ToFedora::Descriptive::RelatedResource when it has a related item without title builds the xml`.

PR #1442, opened 2020-11-19
When mapping from fedora, re-use builders to build related items.

justinlittman:
refs #1403, #1404

## Why was this change made?
`mods:relatedItems` can contain all of the same elements as `mods:mods`. This changes the mapping so that use the same builders for both.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/from_fedora/descriptive.rb
+++ b/app/services/cocina/from_fedora/descriptive.rb
@@ -6,22 +6,9 @@ module Cocina
     class Descriptive
       DESC_METADATA_NS = Dor::DescMetadataDS::MODS_NS
 
-      BUILDERS = {
-        note: Notes,
-        language: Language,
-        contributor: Contributor,
-        event: Descriptive::Event,
-        subject: Subject,
-        form: Form,
-        identifier: Identifier,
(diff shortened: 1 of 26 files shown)


justinlittman:
Before:
```
Status (n=100000):
  Success:   36758 (36.758%)
  Different: 62590 (62.59%)
  To Cocina error:     41 (0.041%)
  To Fedora error:     1 (0.001%)
  Missing:     610 (0.61%)
```

After:
```
Status (n=100000):
  Success:   36579 (36.579%)
  Different: 55679 (55.679%)
  To Cocina error:     7131 (7.131%)  
  To Fedora error:     1 (0.001%)
  Missing:     610 (0.61%)
```

I'm not concerned about the minor decrease in success. It is probably from dropping the support for `relatedItem location` (https://github.com/sul-dlss/dor-services-app/pull/1442/files#diff-350da82e48685b015cf1845599d0c5b4a67fd46e69d2db399cdc1df66ca5aa7bL44), which we'll get back with https://github.com/sul-dlss/dor-services-app/issues/1412.

jcoyne (approved):

PR #1443, opened 2020-11-19
When mapping to fedora, re-use writers to write related items.

justinlittman:
## Why was this change made?
`mods:relatedItems` can contain most of the same elements as `mods:mods`. This changes the mapping so that use the same writers for both.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/to_fedora/descriptive.rb
+++ b/app/services/cocina/to_fedora/descriptive.rb
@@ -6,6 +6,7 @@ module Cocina
     # Fedora 3 data model descMetadata
     class Descriptive
       # @param [Cocina::Models::Description] descriptive
+      # @param [string] druid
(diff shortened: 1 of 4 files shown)


justinlittman:
Before:
```
Status (n=100000):
  Success:   36579 (36.579%)
  Different: 55679 (55.679%)
  To Cocina error:     7131 (7.131%)  
  To Fedora error:     1 (0.001%)
  Missing:     610 (0.61%)
```

After:
```
Status (n=100000):
  Success:   37277 (37.277%)
  Different: 54981 (54.981%)
  To Cocina error:     7131 (7.131%)
  To Fedora error:     1 (0.001%)
  Missing:     610 (0.61%)
```

mjgiarlo (approved):

Issue #1375, opened 2020-11-12
contributor mappings FromFedora and ToFedora used by h2


ndushay:
See sul-dlss-labs/cocina-descriptive-metadata/pull/125/files

and

https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/h2_cocina_mappings/h2_to_cocina_contributor.txt

sul-dlss/happy-heron/issues/240 is in progress so h2 will put the right stuff in the cocina-model sent, ultimately, to DSA.

jcoyne:
I'm unsure what needs to happen in DSA for this. https://github.com/sul-dlss-labs/cocina-descriptive-metadata/pull/125 is a change to h2_cocina_mappings which shouldn't require any DSA modifications.

ndushay:
Maybe nothing but I want to take a careful look.  It'll be splendid if we already have all the mappings we need for H2.

ndushay:
So I'm down to 3 failing tests from the h2 cocina mappings.  ...

PR #1447, opened 2020-11-19
more H2 contrib mappings

ndushay:
## Why was this change made?

Part of #1375 - ensuring that all the mappings needed for contributors by h2 are coded here.

Mapping changes in this PR 
- Cocina contributor type `event` added but it is mapped to `corporate` in MODS, which already has a mapping to Cocina `organization` (in other words, it's not a reciprocal mapping)
- accommodates a cocina contributor having multiple roles.
- DataCite event data should not become a cocina `form`

There are 3 failing specs but I want to put those in a separate PR to avoid potential merging problems, as there is an open PR on `originInfo` mappings, which are implicated in these failing specs.

I used https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/h2_cocina_mappings/h2_to_cocina_contributor.txt as my source for the two new spec files added;  I was glad for all the existing specs so I knew whether my changes were messing up anything else.

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/contributor.rb
+++ b/app/services/cocina/from_fedora/descriptive/contributor.rb
@@ -6,12 +6,12 @@ module Cocina
     class Descriptive
       # Maps contributors
       class Contributor
+        # key: MODS, value: cocina
         ROLES = {
           'personal' => 'person',
           'corporate' => 'organization',
           'family' => 'family',
-          'conference' => 'conference',
-          'event' => 'event'
+          'conference' => 'conference'
         }.freeze
 
         NAME_PART = {
@@ -171,9 +171,8 @@ module Cocina
           authority = ng_role.xpath(ROLE_AUTHORITY_XPATH, mods: DESC_METADATA_NS).first
           authority_uri = ng_role.xpath(ROLE_AUTHORITY_URI_XPATH, mods: DESC_METADATA_NS).first
           authority_value = ng_role.xpath(ROLE_AUTHORITY_VALUE_XPATH, mods: DESC_METADATA_NS).first
-          marcrelator = marc_relator_role?(authority, authority_uri, authority_value)
 
-          check_code(code, authority)
+          check_role_code(code, authority)
 
           {}.tap do |role|
             if authority&.content.present?
@@ -187,6 +186,7 @@ module Cocina
 
             role[:uri] = authority_value&.content
             role[:code] = code&.content
+            marcrelator = marc_relator_role?(authority, authority_uri, authority_value)
             role[:value] = normalized_role_value(text.content, marcrelator) if text
 
             if role[:code].blank? && role[:value].blank?
@@ -209,7 +209,7 @@ module Cocina
           ROLES.fetch(type.downcase)
         end
 
-        def check_code(role_code, role_authority)
+        def check_role_code(role_code, role_authority)
           return if role_code.nil? || role_authority
 
           if role_code.content.present? && role_code.content.size == 3
--- a/app/services/cocina/from_fedora/descriptive/contributor.rb
+++ b/app/services/cocina/from_fedora/descriptive/contributor.rb
@@ -10,7 +10,8 @@ module Cocina
           'personal' => 'person',
           'corporate' => 'organization',
           'family' => 'family',
-          'conference' => 'conference'
+          'conference' => 'conference',
+          'event' => 'event'
         }.freeze
 
         NAME_PART = {
@@ -91,7 +92,7 @@ module Cocina
             contributor_hash = name_authority_uri(name, contributor_hash)
             contributor_hash[:type] = type_for(name['type']) if name['type'].present?
             contributor_hash[:status] = name['usage'] if name['usage']
-            roles = [roles_for(name)]
+            roles = roles_for(name)
             contributor_hash[:role] = roles unless roles.flatten.empty? || contributor_hash[:name].blank?
             contributor_hash[:note] = notes_for(name)
             contributor_hash[:identifier] = identifier_for(name)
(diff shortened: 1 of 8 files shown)


app/services/cocina/to_fedora/descriptive/contributor.rb
```diff
@@ -5,7 +5,10 @@ module ToFedora
     class Descriptive
       # Maps contributors from cocina to MODS XML
       class Contributor
-        NAME_TYPE = FromFedora::Descriptive::Contributor::ROLES.invert.freeze
+        name_types = FromFedora::Descriptive::Contributor::ROLES.invert
+        # one way mapping:  MODS 'corporate' already maps to Cocina 'organization'
+        name_types['event'] = 'corporate'
+        NAME_TYPE = name_types.freeze
```
ndushay, line 11: this is another funky thing, but there's another class that uses NAME_TYPE as a class variable, so ... my attempts to refactor this died.   Again, would rather have the working code merged and deal with the refactors after ...

app/services/cocina/to_fedora/descriptive/contributor.rb
```diff
@@ -5,7 +5,10 @@ module ToFedora
     class Descriptive
       # Maps contributors from cocina to MODS XML
       class Contributor
-        NAME_TYPE = FromFedora::Descriptive::Contributor::ROLES.invert.freeze
+        name_types = FromFedora::Descriptive::Contributor::ROLES.invert
+        # one way mapping:  MODS 'corporate' already maps to Cocina 'organization'
+        name_types['event'] = 'corporate'
```
mjgiarlo, line 10: So we want to modify the `FromFedora::Descriptive::Contributor::ROLES` constant for use in the `ToFedora::Descriptive::Contributor` class but not in the FromFedora one?

app/services/cocina/to_fedora/descriptive/form.rb
```diff
@@ -69,6 +69,7 @@ def write_physical_description(forms)
         end
 
         def write_basic(form)
+          return nil if form.source&.value&.match?(/DataCite/i)
```
ndushay, line 72: yes, this is needed from h2 work.  🤮

spec/services/cocina/from_fedora/descriptive_spec.rb
```diff
@@ -301,42 +301,48 @@
...
           type: 'person',
-          role: [{
-            value: 'degree supervisor',
-            code: 'ths',
-            source: {
-              code: 'marcrelator',
-              uri: 'http://id.loc.gov/vocabulary/relators/'
+          role: [
+            { value: 'degree supervisor' },
+            {
```
ndushay, line 306: here the multiple roles for a single contributor affect pre-existing specs ...

app/services/cocina/to_fedora/descriptive/contributor.rb
ndushay: yes - that is what I mean by "one way mapping" -- the FROM hash would have one key going to two different values.  This is nasty ... but it does work.  I tried to clarify with the comment.

app/services/cocina/to_fedora/descriptive/contributor.rb
```diff
@@ -56,21 +59,41 @@ def name_attributes(contributor)
...
-                xml.roleTerm role.value, attributes
-              end
-              if role.code.present?
-                attributes[:type] = 'code'
-                xml.roleTerm role.code, attributes
-              end
+          mr_roles_xml = marcrelator_roles_xml(contributor)
+          return mr_roles_xml if mr_roles_xml.present?
+
+          Array(contributor.role).each { |role| xml_role(role) unless role.value&.match?(/conference/i) }
```
mjgiarlo, line 67: Do you intend to return the whole `Array(contributor.role)` here or a separate array of xml_role info where role.value doesn't match conference? I thought the return value of `#each` was the array it was invoked on. So maybe you want to return `Array(contributor.role)` but you also want to cause side effects? I'm a bit confused on this one.

app/services/cocina/to_fedora/descriptive/contributor.rb
```diff
@@ -56,21 +59,41 @@ def name_attributes(contributor)
...
+        end
+
+        MARC_RELATOR_PIECE = 'id.loc.gov/vocabulary/relators'
+
+        def marcrelator_roles_xml(contributor)
+          result = []
+          contributor.role.select do |role|
```
mjgiarlo, line 74: I'm a bit puzzled by using `#select` with a block that doesn't seem to have a condition, but that does an append side effect.

app/services/cocina/to_fedora/descriptive/contributor.rb
ndushay: I want an Array, that is comprised of contributor roles, to call xml_role for each role EXCEPT one case.  I will try it right now with just "contributor.role" and see if that works.

app/services/cocina/to_fedora/descriptive/contributor.rb
ndushay: i think this evolved as I wrote it.  I'll do an each.

app/services/cocina/to_fedora/descriptive/contributor.rb
mjgiarlo: OK. would you consider this small tweak to what you have?

```ruby
NAME_TYPE = Cocina::FromFedora::Descriptive::Contributor::ROLES.invert.merge('event' => 'corporate').freeze
```

I think the comment is OK.

app/services/cocina/from_fedora/descriptive/contributor.rb
```diff
@@ -143,44 +144,59 @@ def names
...
-          role_authority_uri = name.xpath(ROLE_AUTHORITY_URI_XPATH, mods: DESC_METADATA_NS).first
-          role_authority_value = name.xpath(ROLE_AUTHORITY_VALUE_XPATH, mods: DESC_METADATA_NS).first
+        ROLE_CODE_XPATH = './mods:roleTerm[@type="code"]'
+        ROLE_TEXT_XPATH = './mods:roleTerm[@type="text"]'
+        ROLE_AUTHORITY_XPATH = './mods:roleTerm/@authority'
+        ROLE_AUTHORITY_URI_XPATH = './mods:roleTerm/@authorityURI'
+        ROLE_AUTHORITY_VALUE_XPATH = './mods:roleTerm/@valueURI'
+        MARC_RELATOR_PIECE = 'id.loc.gov/vocabulary/relators'
+
+        # shameless green
```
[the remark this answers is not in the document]
mjgiarlo: I know you will follow this up in a forthcoming PR, so no worries.

app/services/cocina/to_fedora/descriptive/contributor.rb
ndushay: hahaha - i couldn't for the life of me remember the `.merge` thing and was doing contortions and it seemed so wrong ....

mjgiarlo (approved):
Please test against a sufficiently large dataset in prod (sdr-deploy) and then feel free to merge if the results look good.

ndushay:
oh my.  very slightly worse.  what do I do?

### Validations Before:
(roundtrip

```
Status (n=500000):
  Success:   182906 (36.5812%)
  Different: 313852 (62.7704%)
  To Cocina error:     203 (0.0406%)
  To Fedora error:     85 (0.017%)
  Missing:     2954 (0.5908%)
```

### Validations After
```
Status (n=500000):
  Success:   181909 (36.3818%)
  Different: 314919 (62.9838%)
  To Cocina error:     206 (0.0412%)
  To Fedora error:     12 (0.0024%)
  Missing:     2954 (0.5908%)
```

justinlittman:
I've gotten bit before by not being rebased off the latest master, which throws off the comparison.

ndushay:
thanks, Justin.  I did notice (before I saw your comment) that I wasn't testing my branch against the same master, so I fixed that and adjusted stats above.

app/services/cocina/from_fedora/descriptive/contributor.rb
```diff
@@ -193,37 +209,26 @@ def type_for(type)
...
-        def check_code(role_code, role_authority)
+        def check_role_code(role_code, role_authority)
           return if role_code.nil? || role_authority
 
           if role_code.content.present? && role_code.content.size == 3
             Honeybadger.notify('[DATA ERROR] Contributor role code is missing authority', { tags: 'data_error' })
             return
           end
 
-          raise Cocina::Mapper::InvalidDescMetadata, "Contributor role code is missing and has unexpected value: #{role_code.content}"
```
ndushay, line 204: this error message didn't make sense, so i tweaked it.

PR #1456, opened 2020-11-20
H2 contrib publisher mappings

### LINKED ISSUE #1375: contributor mappings FromFedora and ToFedora used by h2

ndushay:
## Why was this change made?

Closes #1375 - changes to DSA needed for h2 publisher contributors (but of benefit to all our MODS data - cough cough)

- unit tests
    - any tweaks to existing tests were confirmed as correct by Arcadia
    - got H2 contributor publication mapping tests to pass

- validation results of mapping change on prod data are in a comment below https://github.com/sul-dlss/dor-services-app/pull/1456#issuecomment-731394857 -- IMPROVEMENT!!!

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/event.rb
+++ b/app/services/cocina/to_fedora/descriptive/event.rb
@@ -32,9 +32,9 @@ module Cocina
         def write
           Array(events).each_with_index do |event, count|
             attributes = {}
-            attributes[:eventType] = EVENT_TYPE.fetch(event.type) if events.size > 1 && event.type
+            attributes[:eventType] = EVENT_TYPE.fetch(event.type) if event.type
             if translated?(event)
-              TranslatedEvent.write(xml: xml, event: event, count: count)
+              TranslatedEvent.write(xml: xml, event: event, count: count, event_type: event.type)
             else
               write_basic(event, attributes)
             end
--- a/app/services/cocina/to_fedora/descriptive/event.rb
+++ b/app/services/cocina/to_fedora/descriptive/event.rb
@@ -74,7 +74,10 @@ module Cocina
           xml.send(note.type, note.value, attributes)
         end
 
+        # the only contributors legal for MODS are ones with role publisher
         def contributor(contributor)
+          return unless contributor_role_publisher?(contributor)
+
           attributes = {}
           name = contributor.name.first
           if name.valueLanguage
@@ -82,7 +85,8 @@ module Cocina
             attributes[:script] = name.valueLanguage.valueScript.code
             attributes[:transliteration] = name.standard.value if name.standard
           end
-          xml.send(contributor.role.first.value, name.value, attributes)
+
+          xml.publisher(name.value, attributes)
         end
 
         def location(location)
@@ -123,6 +127,21 @@ module Cocina
           dates.each do |date|
             date_tag(date, event_type, point: date.type)
           end
+        end
+
+        MARC_RELATOR_PIECE = 'id.loc.gov/vocabulary/relators'
+
+        # prefer marcrelator publisher role
+        def contributor_role_publisher?(contributor)
+          return true if contributor.role.any? { |role| role.value.match?(/publisher/i) && role_is_marcrelator?(role) }
+
+          contributor.role.any? { |role| role.value&.match?(/publisher/i) }
+        end
+
+        def role_is_marcrelator?(role)
+          role.source&.code == 'marcrelator' ||
+            role.source&.uri&.include?(MARC_RELATOR_PIECE) ||
+            role.uri&.include?(MARC_RELATOR_PIECE)
         end
       end
     end
--- a/app/services/cocina/to_fedora/descriptive/translated_event.rb
+++ b/app/services/cocina/to_fedora/descriptive/translated_event.rb
@@ -8,14 +8,16 @@ module Cocina
         # @params [Nokogiri::XML::Builder] xml
         # @params [Cocina::Models::Event] event
         # @params [Integer] count
-        def self.write(xml:, event:, count:)
-          new(xml: xml, event: event, count: count).write
+        # @params [String] event_type - see Cocina::ToFedora::Descriptive::Event::EVENT_TYPE
+        def self.write(xml:, event:, count:, event_type:)
+          new(xml: xml, event: event, count: count, event_type: event_type).write
         end
 
-        def initialize(xml:, event:, count:)
+        def initialize(xml:, event:, count:, event_type:)
           @xml = xml
           @event = event
           @count = count
+          @event_type = event_type
           @groups = {}
         end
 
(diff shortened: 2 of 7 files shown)


app/services/cocina/to_fedora/descriptive/event.rb
```diff
@@ -32,9 +32,9 @@ def initialize(xml:, events:)
         def write
           Array(events).each_with_index do |event, count|
             attributes = {}
-            attributes[:eventType] = EVENT_TYPE.fetch(event.type) if events.size > 1 && event.type
```
ndushay, line 35: I spoke with Arcadia about this and it is fine to add eventType if we have it;  it was not avail until MODS 3.5 or something, but anything running through this code is ok.

However, it may affect roundtrip comparisons???

app/services/cocina/to_fedora/descriptive/event.rb
```diff
@@ -74,15 +74,19 @@ def note(note)
           xml.send(note.type, note.value, attributes)
         end
 
+        # the only contributor we can use for MODS are ones with role publisher
         def contributor(contributor)
+          return unless contributor_role_publisher?(contributor)
```
ndushay, line 79: also confirmed this with Arcadia.  Other contributor types don't have a valid MODS element.  We should only get to EVENT from contributor for publisher in any case.

app/services/cocina/to_fedora/descriptive/event.rb
```diff
@@ -74,15 +74,19 @@ def note(note)
...
+
           attributes = {}
           name = contributor.name.first
```
ndushay, line 82: assuming one contributor lalalalalala

ndushay:
## Validation Results

### Roudntrip Before

```
Status (n=250000):
  Success:   92806 (37.1224%)
  Different: 155621 (62.2484%)
  To Cocina error:     98 (0.0392%)
  To Fedora error:     6 (0.0024%)
  Missing:     1469 (0.5876%)
```

### Roundtrip After

```
```

NOTE:  eventType attribute added to events will cause roundtrip to fail, but this change has been approved by Arcadia

app/services/cocina/to_fedora/descriptive/event.rb
ndushay: I have added a normalization to roundtrip validations that accounts for this.

app/services/cocina/to_fedora/descriptive/contributor.rb
```diff
@@ -57,7 +57,7 @@ def name_attributes(contributor)
           end
         end
 
-        # return marcrelator roles only if any are present, otherwise return other roles
+        # write marcrelator roles only if any are present, otherwise write other roles
```
jcoyne, line 60: Why did you choose to change this description?  It does look like it's returning those, isn't it?

app/services/cocina/to_fedora/descriptive/event.rb
```diff
@@ -74,15 +74,19 @@ def note(note)
...
           if name.valueLanguage
             attributes[:lang] = name.valueLanguage.code
             attributes[:script] = name.valueLanguage.valueScript.code
             attributes[:transliteration] = name.standard.value if name.standard
           end
-          xml.send(contributor.role.first.value, name.value, attributes)
+
+          xml.send('publisher', name.value, attributes)
```
jcoyne, line 89: why do send if we know the name of the method?
```suggestion
          xml.publisher(name.value, attributes)
```

app/services/cocina/to_fedora/descriptive/event.rb
```diff
@@ -124,6 +128,21 @@ def date_range(dates, event_type)
...
           end
         end
+
+        MARC_RELATOR_PIECE = 'id.loc.gov/vocabulary/relators'
+
+        # prefer marcrelator publisher role
+        def contributor_role_publisher?(contributor)
+          return true if contributor.role.any? { |role| role.value.match?(/publisher/i) && role_is_marcrelator?(role) }
+
+          contributor.role.any? { |role| role.value&.content&.match?(/publisher/i) }
```
jcoyne, line 138: It looks like this line needs a test.

app/services/cocina/to_fedora/descriptive/event.rb
```diff
@@ -124,6 +128,21 @@ def date_range(dates, event_type)
             date_tag(date, event_type, point: date.type)
           end
         end
+
+        MARC_RELATOR_PIECE = 'id.loc.gov/vocabulary/relators'
+
+        # prefer marcrelator publisher role
+        def contributor_role_publisher?(contributor)
```
jcoyne, line 135: Could this take the role itself, rather than the whole publisher?

jcoyne (changes requested):

app/services/cocina/to_fedora/descriptive/contributor.rb
ndushay: stupidity?   fixing.

app/services/cocina/to_fedora/descriptive/event.rb
ndushay: good point.  fixing.

app/services/cocina/to_fedora/descriptive/event.rb
ndushay: no, because we have to check all roles, potentially.  This code happened because the first role had value H2 'Publisher' rather than marcrelator 'publisher.'   :-P

app/services/cocina/to_fedora/descriptive/event.rb
ndushay: good catch.  line had a bug.  test added.

app/services/cocina/to_fedora/descriptive/event.rb
jcoyne: I guess what I meant was the roles array.

ndushay: what does that actually gain us?  passed by reference anyway ...

mjgiarlo:
@jcoyne mind having another look at this?

app/services/cocina/to_fedora/descriptive/event.rb
jcoyne: It makes the method more testable and the code easier to understand because you know that this method only depends on this one property rather than a whole object.

jcoyne (approved):
Approved, with a suggestion.

PR #1459, opened 2020-11-20
[Refactor] Refactor language mapping

jcoyne:
## Why was this change made?

Remove constants that were only called once, having to reference the top of the file made it harder to read the code.  Also removed unnecessary queries.

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/language.rb
+++ b/app/services/cocina/from_fedora/descriptive/language.rb
@@ -5,19 +5,6 @@ module Cocina
     class Descriptive
       # Maps languages
       class Language
-        LANG_XPATH = 'mods:language'
-        LANG_STATUS_XPATH = './@status'
-        OBJECT_PART_XPATH = './@objectPart'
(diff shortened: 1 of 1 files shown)


app/services/cocina/from_fedora/descriptive/language.rb
```diff
@@ -74,61 +62,36 @@ def script_term_attributes_for(script_term_nodes)
...
-        end
-
-        def script_term?(lang)
-          script_term_nodes(lang).to_s.present?
-        end
-
-        def script_term_nodes(lang)
-          lang.xpath(SCRIPT_TERM_XPATH, mods: DESC_METADATA_NS)
-        end
+        def script_term_nodes(lang); end
```
mjgiarlo, line 94: We still need this method?

jcoyne: nope, removed.

mjgiarlo (approved):
LGTM

PR #1461, opened 2020-11-20
Remove puts and revert test changes

justinlittman:
## Why was this change made?
Remove spurious puts.

NA

NA


jcoyne (approved):

jcoyne (changes requested):
It looks like this has now expanded itself beyond "Remove puts"

justinlittman:
Yes, this is cleanup my mistaken commits.

jcoyne (approved):

PR #1467, opened 2020-11-23
validate-to-cocina, validate-to-fedora: provide option to automatically give result files unique descriptive names

jmartin-sul:
## Why was this change made?

When running tests or reports that produce output files, I find it nice when output is easy to associate with the input that produced it, so I added an option to do that for the cocina validation results, since I was running against a number of branches in a row Friday evening.  Figured others might find it useful, and it was quick to write so I thought it'd save me time in the long run.  Old behavior is still the default.

used the new `-u` flag on friday while running validations.

Screen Shot 2020-11-20 at 7 28 42 PM

`validate-to-cocina` and `validate-to-fedora` help messages

### DIFF
--- a/bin/validate-to-cocina
+++ b/bin/validate-to-cocina
@@ -5,11 +5,12 @@ require_relative '../config/environment'
 require_relative '../lib/fedora_cache'
 require 'optparse'
 
-options = { overwrite: false, random: false }
+options = { overwrite: false, random: false, unique_filename: false }
 parser = OptionParser.new do |option_parser|
   option_parser.banner = 'Usage: bin/validate-to-cocina [options]'
 
   option_parser.on('-sSAMPLE', '--sample SAMPLE', Integer, 'Sample size, otherwise all druids.')
+  option_parser.on('-u', '--unique-filename', 'Result file named for branch and runtime') { options[:unique_filename] = true }
   option_parser.on('-h', '--help', 'Displays help.') do
     puts option_parser
     exit
@@ -50,6 +51,24 @@ end
 
 def summary_line_for(label, count)
   "#{label}: #{count} of #{@sample_size} (#{100 * count.to_f / @sample_size}%)\n"
+end
+
+def branch_name
+  `git rev-parse --abbrev-ref HEAD`.strip
+end
+
+def short_commit_hash
+  `git rev-parse --short HEAD`.strip
+end
+
+def now_str
+  DateTime.now.utc.iso8601.to_s.gsub(':', '-')
+end
+
+def results_file_name(use_unique_filename)
+  return 'results.txt' unless use_unique_filename
+
+  "results_to-cocina_#{now_str}_#{branch_name}_#{short_commit_hash}.txt"
 end
 
 druids = File.read('druids.txt').split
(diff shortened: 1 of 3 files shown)


bin/validate-to-cocina
```diff
@@ -5,11 +5,12 @@ require_relative '../config/environment'
 require_relative '../lib/fedora_cache'
 require 'optparse'
 
-options = { overwrite: false, random: false }
+options = { overwrite: false, random: false, unique_filename: false }
 parser = OptionParser.new do |option_parser|
   option_parser.banner = 'Usage: bin/validate-to-cocina [options]'
 
   option_parser.on('-sSAMPLE', '--sample SAMPLE', Integer, 'Sample size, otherwise all druids.')
+  option_parser.on('-u', '--unique-filename', 'Result file named for branch and runtime') { options[:unique_filename] = true }
```
justinlittman, line 13: Is `{ options[:unique_filename] = true }` necessary?

bin/validate-to-cocina
```diff
@@ -107,7 +126,7 @@ aggregated_data_error_results = aggregated_errors_for(results_by_data_error)
 print_results(aggregated_error_results, 'Error')
 print_results(aggregated_data_error_results, 'Data error')
 
-File.open('results.txt', 'w') do |file|
+File.open(results_file_name(options), 'w') do |file|
```
justinlittman, line 129: Why pass it all of options and not just `options[:unique_filename]`?

bin/validate-to-fedora
```diff
@@ -5,11 +5,12 @@ require_relative '../config/environment'
 require_relative '../lib/fedora_cache'
 require 'optparse'
 
-options = { overwrite: false, random: false }
+options = { overwrite: false, random: false, unique_filename: false }
 parser = OptionParser.new do |option_parser|
   option_parser.banner = 'Usage: bin/validate-to-fedora [options]'
 
   option_parser.on('-sSAMPLE', '--sample SAMPLE', Integer, 'Sample size, otherwise all druids.')
+  option_parser.on('-u', '--unique-filename', 'Result file named for branch and runtime') { options[:unique_filename] = true }
```
justinlittman, line 13: Same comment as above.

bin/validate-to-fedora
```diff
@@ -99,7 +118,7 @@ aggregated_error_results = aggregated_errors_for(results_by_error)
 
 print_results(aggregated_error_results, 'To Fedora error')
 
-File.open('results.txt', 'w') do |file|
+File.open(results_file_name(options), 'w') do |file|
```
justinlittman, line 121: Same comment as above.

justinlittman (changes requested):
Also, please add new flag to options listed in README.

bin/validate-to-cocina
jmartin-sul: My recollection is that in the bit of testing I did, I couldn't get `parser.parse!(into: options)` to set a value for `unique-filename` based solely on the presence of the flag, and taking a param value seemed unnecessary, hence the manual setting of `options[:unique_filename]`.  That did bug me a bit, but I also only spent 10 or 15 minutes trying to avoid what I did here (using [`FalseClass` for type coercion with a default](https://ruby-doc.org/stdlib-2.7.2/libdoc/optparse/rdoc/OptionParser.html#class-OptionParser-label-Type+Coercion), iirc).

bin/validate-to-cocina
jmartin-sul: good point, fixed

bin/validate-to-fedora
jmartin-sul: same reply as that one

bin/validate-to-fedora
jmartin-sul: also fixed

jmartin-sul:
made readme and arg passing touchups, replied to the question about manually setting the options value, and re-tested using same invocations as in the issue description.  ready for another look, i think...

justinlittman (approved):

PR #1507, opened 2020-11-24
Add missing valid authority codes

arcadiafalcone:
## Why was this change made?
Valid authority codes missing from the validation list were causing many false positive errors.

n/a

n/a


ndushay (approved):
my commit is just syntax stuff from Arcadia's commit, so merging.

PR #1516, opened 2020-11-25
Normalizes authorityURI with trailing slash.

### LINKED ISSUE #1484: Mapping: For names, normalize authorityURI
For example, druid:cz548fn3478.

Should be same as https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/from_fedora/descriptive/subject.rb#L69-L73

### LINKED ISSUE #1490: Mapping: For placeTerms, normalize authorityURI
For example, druid:kt455nh2818.

Should be same as https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cocina/from_fedora/descriptive/subject.rb#L69-L73

justinlittman:
closes #1484
closes #1490

## Why was this change made?
Normalizes authorityURI with trailing slash.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/subject.rb
+++ b/app/services/cocina/from_fedora/descriptive/subject.rb
@@ -60,7 +60,7 @@ module Cocina
 
         def source_attrs(subject, attrs = {})
           if subject[:valueURI]
-            source = { code: code_for(subject), uri: uri_for(subject[:authorityURI]) }.compact
+            source = { code: code_for(subject), uri: AuthorityUri.normalize(subject[:authorityURI]) }.compact
             attrs[:source] = source unless source.empty?
             attrs[:uri] = subject[:valueURI]
           elsif subject[:authority]
@@ -68,12 +68,6 @@ module Cocina
             attrs[:source] = source unless source.empty?
           end
           attrs.compact
-        end
-
-        def uri_for(authority_uri)
-          return "#{authority_uri}/" if ['http://id.loc.gov/authorities/names', 'http://id.loc.gov/authorities/subjects'].include?(authority_uri)
-
-          authority_uri
         end
 
         def code_for(subject)
--- a/spec/services/cocina/from_fedora/descriptive/event_spec.rb
+++ b/spec/services/cocina/from_fedora/descriptive/event_spec.rb
@@ -564,7 +564,7 @@ RSpec.describe Cocina::FromFedora::Descriptive::Event do
       <<~XML
         <originInfo>
         <place>
-          <placeTerm type="text" authority="naf" authorityURI="http://id.loc.gov/authorities/names/" valueURI="http://id.loc.gov/authorities/names/n50046557">Stanford (Calif.)</placeTerm>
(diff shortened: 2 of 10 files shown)


justinlittman:
Before:
```
Status (n=100000):
  Success:   39908 (39.908%)
  Different: 59420 (59.42%)
  To Cocina error:     33 (0.033%)
  To Fedora error:     29 (0.029%)
  Missing:     610 (0.61%)
```

After:
```
Status (n=100000):
  Success:   49463 (49.463%)
  Different: 49865 (49.865%)
  To Cocina error:     33 (0.033%)
  To Fedora error:     29 (0.029%)
  Missing:     610 (0.61%)
```

GOAL!

app/services/cocina/from_fedora/descriptive/subject.rb
```diff
@@ -70,12 +70,6 @@ def source_attrs(subject, attrs = {})
           attrs.compact
         end
 
-        def uri_for(authority_uri)
-          return "#{authority_uri}/" if ['http://id.loc.gov/authorities/names', 'http://id.loc.gov/authorities/subjects'].include?(authority_uri)
-
-          authority_uri
-        end
-
```
ndushay, line 78: this is a logic change, as now it takes any of the authority uris ... is that ok?  Is there a test for when it's some other authority?

ndushay (changes requested):
You have one logic difference in this PR that you might want to address

spec/services/cocina/from_fedora/descriptive/event_spec.rb
```diff
@@ -564,7 +564,7 @@
       <<~XML
         <originInfo>
         <place>
-          <placeTerm type="text" authority="naf" authorityURI="http://id.loc.gov/authorities/names/" valueURI="http://id.loc.gov/authorities/names/n50046557">Stanford (Calif.)</placeTerm>
+          <placeTerm type="text" authority="naf" authorityURI="http://id.loc.gov/authorities/names" valueURI="http://id.loc.gov/authorities/names/n50046557">Stanford (Calif.)</placeTerm>
```
ndushay, line 567: you changed these NOT to have slashes to show normalization is working, yes?  Because I think we do want the slashes in the result.

app/services/cocina/from_fedora/descriptive/subject.rb
justinlittman: Yes, the list of authorityURIs expanded. These are only the one I know about. I believe this is OK.

Sprinkled throughout the unit tests are various different authorityURIs that are tested.

spec/services/cocina/from_fedora/descriptive/event_spec.rb
justinlittman: Correct. I removed it here to make sure it was added.

app/services/cocina/from_fedora/descriptive/subject.rb
ndushay: what I'm trying to say is that the subject authorities here only allows for names and subjects and nothing else, while your changes will take anything in the normalized list

ndushay: I am wrong.

ndushay (approved):

PR #1576, opened 2020-12-01
Prefer `Array#uniq` over set operations

mjgiarlo:
And make `Cocina::ToFedora::Descriptive::Subject#all_same_authority?` return a boolean since its value is not used otherwise (i.e., as a string)

## Why was this change made?

Because it was suggested in code review, which I didn't notice before merging.

CI

None

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/subject.rb
+++ b/app/services/cocina/to_fedora/descriptive/subject.rb
@@ -97,10 +97,7 @@ module Cocina
         end
 
         def all_same_authority?(structured_value)
-          return nil if structured_value.nil?
-          return nil unless Set.new(structured_value.map { |value| authority_for(value) }).size == 1
-
-          authority_for(structured_value.first)
+          Array(structured_value).map { |value| authority_for(value) }.uniq.size == 1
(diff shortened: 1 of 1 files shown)


app/services/cocina/to_fedora/descriptive/subject.rb
```diff
@@ -97,10 +97,10 @@ def structured_attributes_for(subject)
         end
 
         def all_same_authority?(structured_value)
-          return nil if structured_value.nil?
-          return nil unless Set.new(structured_value.map { |value| authority_for(value) }).size == 1
+          return false if structured_value.nil?
+          return false unless structured_value.map { |value| authority_for(value) }.uniq.size == 1
 
-          authority_for(structured_value.first)
+          true
```
jcoyne, line 103:
```suggestion
          Array(structured_value).map { |value| authority_for(value) }.uniq.size == 1
```

mjgiarlo: Dig it. Done.

jcoyne (approved):

PR #1578, opened 2020-12-01
Reports if node count mismatch.

### LINKED ISSUE #1577: Report if node count mismatch
Otherwise, there is no indication why druid:qd366kh0954 is not equivalent.

justinlittman:
closes #1577

## Why was this change made?
A mismatch was reported, but the reason was not given.

Locally.

NA


justinlittman:
Now reports:
```
Node count mismatch between original and roundtripped.
```

jcoyne (approved):

PR #1581, opened 2020-12-01
search all url nodes for primary designation when normalizing PURLs

### LINKED ISSUE #1518: Mapping error: PURL mistakenly has primary display usage
PURL should only have `usage="primary display"` when there is no other `<url>` with `usage="primary display"` .

See, for example, druid:cy979mw6316.

This is wrong:
```
<location>
  <physicalLocation>Stanford University Libraries</physicalLocation>
  <url usage="primary display">http://infoweb.newsbank.com/?db=SERIAL</url>
  <url>http://web.lexis-nexis.com/congcomp/form/cong/s_pubadvanced.html?srcboxes=SSMaps&amp;srcboxes=SerialSet</url>
  <url>http://purl.access.gpo.gov/GPO/LPS839</url>
  <url usage="primary display">http://purl.stanford.edu/cy979mw6316</url>
</location>
```

peetucket:
## Why was this change made?

Fixes #1518 (incorrect PURL normalization with primary usage designation).  Did some refactoring to make this more clear (hopefully)

Added new unit test

### DIFF
--- a/app/services/cocina/mods_normalizer.rb
+++ b/app/services/cocina/mods_normalizer.rb
@@ -134,12 +134,25 @@ module Cocina
     end
 
     def normalize_purl
-      ng_xml.root.xpath('//mods:location', mods: Cocina::FromFedora::Descriptive::DESC_METADATA_NS).each do |location_node|
-        url_nodes = location_node.xpath('mods:url', mods: Cocina::FromFedora::Descriptive::DESC_METADATA_NS)
-        purl_node = url_nodes.find { |url_node| Cocina::FromFedora::Descriptive::Location::PURL_REGEX.match(url_node.text) }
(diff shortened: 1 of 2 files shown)


db/structure.sql
```diff
@@ -318,3 +318,5 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20200507202950'),
 ('20200507224637'),
 ('20200521153735');
+
```
mjgiarlo, line 321: Cruft?

mjgiarlo (approved):
Maybe some DB cruft but :+1:

db/structure.sql
peetucket: yeah, not sure where that came from, i had to bring my environment up to date and this got changed and committed by mistake -- i'll just revert his file back to what it was

PR #1586, opened 2020-12-02
Map access condition.

### LINKED ISSUE #1556: Map accessCondition
See https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_accessCondition.txt

For example, druid:wz014rh6670.

justinlittman:
closes #1556

## Why was this change made?
Map access condition.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/location.rb
+++ b/app/services/cocina/from_fedora/descriptive/location.rb
@@ -7,6 +7,12 @@ module Cocina
       class Location
         PURL_REGEX = %r{^https?://purl.stanford.edu/}.freeze
 
+        ACCESS_CONDITION_TYPES = {
+          'restriction on access' => 'access restriction',
+          'restrictionOnAccess' => 'access restriction',
+          'useAndReproduction' => 'use and reproduction'
+        }.freeze
+
         # @param [Nokogiri::XML::Element] resource_element mods or relatedItem element
         # @param [Cocina::FromFedora::Descriptive::DescriptiveBuilder] descriptive_builder
         # @return [Hash] a hash that can be mapped to a cocina model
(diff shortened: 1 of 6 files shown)


justinlittman:
Before:
```
Status (n=100000):
  Success:   80604 (80.604%)
  Different: 18724 (18.724%)
  To Cocina error:     33 (0.033%)
  To Fedora error:     29 (0.029%)
  Missing:     610 (0.61%)
```

After:
```
Status (n=100000):
  Success:   80613 (80.613%)
  Different: 18715 (18.715%)
  To Cocina error:     33 (0.033%)
  To Fedora error:     29 (0.029%)
  Missing:     610 (0.61%)
```

app/services/cocina/from_fedora/descriptive/location.rb
```diff
@@ -24,6 +30,7 @@ def build
             access[:physicalLocation] = physical_locations if physical_locations.present?
             access[:accessContact] = access_contact if access_contact.present?
             access[:url] = url if url.present?
+            access[:note] = note if note.present?
```
mjgiarlo, line 33: not necessary for this PR, but I thought I'd ask: should we toss a `#compact` after the `#tap` so we can ditch the `if` checks? We seem to be moving in that direction generally in DSA mappings.

mjgiarlo (approved):
LGTM

app/services/cocina/from_fedora/descriptive/location.rb
justinlittman: Yeah, let me make that change.

justinlittman: Oh, those are empty array, not nils so compact won't work.

Issue #1751, opened 2020-12-18
Normalize PURL within relatedItem


justinlittman:
For druid:rs426bc0685:

```
Original node:
<relatedItem type="otherVersion" displayLabel="Associated Essay">
  <titleInfo>
    <title>essay68buildingGrandCanyonPart3fredHarveyCo.pdf</title>
  </titleInfo>
  <location>
    <url>http://purl.stanford.edu/vt726fn1198</url>
  </location>
</relatedItem>
Best guess roundtripped node:
<relatedItem type="otherVersion" displayLabel="Associated Essay">
  <titleInfo>
    <title>essay68buildingGrandCanyonPart3fredHarveyCo.pdf</title>
  </titleInfo>
  <location>
    <url usage="primary display">http://purl.stanford.edu/vt726fn1198</url>
  </location>
</relatedItem>
```

This should follow the same rules as adding usage=primary display for the base `<mods>`.

PR #1772, opened 2020-12-18
[DRAFT] Added tests for #1751 but none fail due to purl/url behavior

mjgiarlo:
Connects to #1751.


mjgiarlo:
@justinlittman @jcoyne I honestly can't recall why I created this PR but I'm inclined to close/delete it unless y'all see some value here.

PR #1802, opened 2021-01-08
Ignores and notifies title type.

### LINKED ISSUE #1781: Handle title type
For druid:fj753pb2958:

```
<titleInfo>
  <title type="main">Monaco Grand Prix</title>
</titleInfo>
Best guess roundtripped node:
<titleInfo>
  <title>Monaco Grand Prix</title>
</titleInfo>
```

A data error should be reported when title has a type attribute and the attribute should be removed in normalization.

justinlittman:
closes #1781

## Why was this change made?
Ignores and notifies title type.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/title_builder.rb
+++ b/app/services/cocina/from_fedora/descriptive/title_builder.rb
@@ -25,6 +25,8 @@ module Cocina
             return nil
           end
 
+          Honeybadger.notify('[DATA ERROR] Title with type', { tags: 'data_error' }) if children_with_type?(children)
+
           # If a displayLabel only with no title text element
           # Note: this is an error condition,
           # exceptions documented at: https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_value_dependencies.txt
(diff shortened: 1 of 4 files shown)


justinlittman:
Before:
```
Status (n=100000):
  Success:   96946 (96.946%)
  Different: 2393 (2.393%)
  To Cocina error:     51 (0.051%)
  To Fedora error:     0 (0.0%)
  Missing:     610 (0.61%)
```

After:
```
Status (n=100000):
  Success:   96982 (96.982%)
  Different: 2357 (2.357%)
  To Cocina error:     51 (0.051%)
  To Fedora error:     0 (0.0%)
  Missing:     610 (0.61%)
```

app/services/cocina/from_fedora/descriptive/title_builder.rb
```diff
@@ -43,9 +47,7 @@ def build
         def simple_title?(children)
           return false if children.map(&:name) != ['title'] || children.size > 1
```
jcoyne, line 44: Can you return the expression here?

jcoyne (changes requested):

app/services/cocina/from_fedora/descriptive/title_builder.rb
justinlittman: But this code is more confusing and difficult to grok.

app/services/cocina/from_fedora/descriptive/title_builder.rb
```diff
@@ -25,6 +25,10 @@ def build
             return nil
           end
 
+          children.each do |child|
+            Honeybadger.notify('[DATA ERROR] Title with type', { tags: 'data_error' }) if child.name == 'title' && child[:type].present?
```
mjgiarlo, line 29: If multiple children titles have types, do you think we need individual HB alerts? I wonder if it might suffice to send a single alert if *any* children are typed titles.

app/services/cocina/from_fedora/descriptive/title_builder.rb
mjgiarlo: I prefer the explicit booleans here.

mjgiarlo (approved):

app/services/cocina/from_fedora/descriptive/title_builder.rb
justinlittman: Agree. Fixed.

jcoyne (approved):

PR #1819, opened 2021-01-12
Add a rabbitmq message when creating an object

jcoyne:
## Why was this change made?
https://github.com/sul-dlss/happy-heron/issues/797

### DIFF
--- a/app/services/notifications/object_created.rb
+++ b/app/services/notifications/object_created.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module Notifications
+  # Send a message to a RabbitMQ exchange that an item has been created.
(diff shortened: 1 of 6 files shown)


app/services/notifications/object_created.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def initialize(model:, channel:)
+      @model = model
+      @channel = channel
+    end
+
+    def publish
+      message = { model: model.to_h }
+      exchange = channel.topic('sdr.objects.created')
```
mjgiarlo, line 30: I might like to see something like `'sdr.objects.created'` in configuration eventually but I don't think it's necessary now since we're still kicking tires.

mjgiarlo (approved):

PR #1821, opened 2021-01-13
[HOLD] Adds DataErrorNotifier as an abstraction for notifications about data…

### LINKED ISSUE #1817: Change strategy for reporting data errors
As described in https://github.com/sul-dlss/argo/issues/2356, raising an error for data errors is problematic.

justinlittman:
closes #1817

## Why was this change made?
Adds DataErrorNotifier as an abstraction for notifications about data errors. Changes so that errors are not deliberately raised by mapping from Fedora.

Unit

NA


jcoyne (approved):

jcoyne (changes requested):
Should `lib/data_error_notifier.rb` be removed?

jcoyne (approved):

PR #1825, opened 2021-01-13
to_fedora mapping specs get shared_example

ndushay:
## Why was this change made?

To make it easier to read and maintain to_fedora mapping specs by getting rid of noise in them.  Note that the rest of the specs will be converted to use the shared_example as part of the cocina-descriptive-md alignment work.

unit tests

### DIFF
--- a/spec/services/cocina/to_fedora/descriptive/contributor_spec.rb
+++ b/spec/services/cocina/to_fedora/descriptive/contributor_spec.rb
@@ -1,40 +1,24 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'support/mods_mapping_spec_helper'
 
-# numbered examples from https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_name.txt
+# numbered examples refer to https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_name.txt
 RSpec.describe Cocina::ToFedora::Descriptive::Contributor do
-  subject(:xml) { writer.to_xml }
-
+  # see spec/support/mods_mapping_spec_helper.rb for how writer is used in shared examples
   let(:writer) do
     Nokogiri::XML::Builder.new do |xml|
-      xml.mods('xmlns' => 'http://www.loc.gov/mods/v3',
-               'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance',
-               'version' => '3.6',
-               'xsi:schemaLocation' => 'http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-6.xsd') do
+      xml.mods(mods_attributes) do
         described_class.write(xml: xml, contributors: contributors, titles: titles, id_generator: Cocina::ToFedora::Descriptive::IdGenerator.new)
       end
     end
   end
-  let(:mods_element_open) do
-    <<~XML
-      <mods xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xmlns="http://www.loc.gov/mods/v3" version="3.6"
-        xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-6.xsd">
-    XML
-  end
-  let(:mods_element_close) { '</mods>' }
   let(:titles) { nil }
 
   context 'when contributors is nil' do
     let(:contributors) { nil }
 
-    it 'builds the xml' do
-      expect(xml).to be_equivalent_to <<~XML
-        #{mods_element_open}
-        #{mods_element_close}
-      XML
-    end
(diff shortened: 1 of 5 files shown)


spec/support/mapping_spec_helper.rb
```diff
@@ -0,0 +1,51 @@
...
+          described_class.write(xml: xml, notes: notes, id_generator: Cocina::ToFedora::Descriptive::IdGenerator.new)
+        when described_class == Cocina::ToFedora::Descriptive::RelatedResource
+          described_class.write(xml: xml, related_resources: resources, druid: 'druid:vx162kw9911', id_generator: Cocina::ToFedora::Descriptive::IdGenerator.new)
+        when described_class == Cocina::ToFedora::Descriptive::Title
+          described_class.write(xml: xml, titles: titles, contributors: contributors, id_generator: Cocina::ToFedora::Descriptive::IdGenerator.new)
+        when described_class == Cocina::ToFedora::Descriptive::Subject
+          described_class.write(xml: xml, subjects: subjects, id_generator: Cocina::ToFedora::Descriptive::IdGenerator.new)
+        else
+          p "no builder for described_class: #{described_class}"
+        end
```
mjgiarlo, line 37: We can dump the new rubocop exclusion and lot of duplication here by doing using `case`/`when` more idiomatically like so:

```ruby
case described_class
when Cocina::ToFedora::Descriptive::Access
when Cocina::ToFedora::Descriptive::AdminMetadata
# ...
```

Would that work here or is there a reason you went with a bare `case`?

spec/support/mapping_spec_helper.rb
```diff
@@ -0,0 +1,51 @@
...
+RSpec.shared_examples 'cocina to MODS' do |expected_xml|
+  subject(:xml) { writer.to_xml }
+
+  let(:writer) do
+    Nokogiri::XML::Builder.new do |xml|
+      xml.mods('xmlns' => 'http://www.loc.gov/mods/v3',
+               'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance',
+               'version' => '3.6',
+               'xsi:schemaLocation' => 'http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-6.xsd') do
+        case
```
jcoyne, line 12: To me, this case statement obscures a lot of variability that would better be expressed in the individual tests themselves.  Instead of doing this can you make a method like:

```ruby
def build_mods
  "#{MODS_HEADER}#{yield}#{MODS_FOOTER}"
 end
 ```
 
 Then you can do:
 ```
 xml = build_mods do
  <<~XML
      <classification authority="ddc" edition="11">683</classification>
      <classification authority="ddc" edition="12">684</classification>
    XML
    end
  ```

File: spec/support/mapping_spec_helper.rb (hunk 1 above)
Comment (mjgiarlo) [reply spec/support/mapping_spec_helper.rb:37]: Btw, really like the overall reduction in duplication that shared_specs brings!

Comment (ndushay) [reply spec/support/mapping_spec_helper.rb:37]: Because when i tried doing what you suggest, it didn't work.   '===' gave false, but '==' gives true.   Happy if there's a  solution that works!

File: spec/support/mapping_spec_helper.rb (hunk 2 above)
Comment (ndushay) [reply spec/support/mapping_spec_helper.rb:12]: i don't think there WAS variability for this in the tests themselves.  Except I saw one that called out MODS 3.3 which I will of course leave alone.

File: spec/support/mapping_spec_helper.rb (hunk 1 above)
Comment (mjgiarlo) [reply spec/support/mapping_spec_helper.rb:37]: @ndushay Ah, OK, because the class of a class is `Class` not the class. Gotcha! That makes sense.

File: spec/support/mapping_spec_helper.rb (hunk 2 above)
Comment (ndushay) [reply spec/support/mapping_spec_helper.rb:12]: if you mean the params passed in to `described_class.write`, maybe I can add an argument to the shared_examples for that -- would that work for you?

Comment (jcoyne) [reply spec/support/mapping_spec_helper.rb:12]: The variability I'm complaining about is the "case" statement.  I think it's very hard to understand why that all needs to be in this shared example.  Putting that in the individual tests for each of those classes is the idiomatic way of instantiating the class under test.

Comment (ndushay) [reply spec/support/mapping_spec_helper.rb]: @jcoyne and I spoke and the new approach I have since pushed up works for him.

File: .rubocop.yml [hunk 3]
```diff
@@ -53,6 +53,10 @@ Style/AsciiComments:
   Exclude:
     - 'spec/services/cocina/from_fedora/descriptive/contributor_spec.rb'
 
+Style/EmptyCaseCondition:
+  Exclude:
+    - 'spec/support/mapping_spec_helper.rb'
```
jcoyne, line 58: Is this still necessary? I don't see any case condition i that file.

jcoyne (changes requested):

.rubocop.yml
ndushay: whoops!  will remove.

jcoyne (approved):

PR #1840, opened 2021-01-15
Adds report for finding items with parts.

### LINKED ISSUE #1804: Generate report of objects with part element in descMD
I would like a report of all SDR objects for which the descMD contains the element <part>with or without attributes at any level of the MODS hierarchy.

The output should include the following:
object druid
APO druid
collection druid (if present)
catkey (if present)

CSV preferred.

justinlittman:
closes #1804

## Why was this change made?
Because Arcadia required it.

sdr-deploy

README

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,21 @@ In addition, detailed results for each item with a difference are provided in an
 
 Note that the location of the cache can be set with `FEDORA_CACHE` environment variable.
 
+### Run Fedora / MODS reports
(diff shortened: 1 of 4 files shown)


jcoyne:
Where the PR asks for Why, you filled in what.  Why do we want this report?

justinlittman:
Description updated.

jcoyne (approved):

PR #1892, opened 2021-01-23
new approach to mods <--> cocina mapping specs (leaner, maintainable)

ndushay:
Needs sign off by @arcadiafalcone and @justinlittman, and at least one other developer.

## Why was this change made?

To make it easier:
- for Arcadia to add/update cocina mappings and get them coded (by having them live in dor-services-app)
- for Arcadia to know exactly which of her mappings are and are not implemented
- to maintain our code against Arcadia's mappings by removing duplication - we use shared examples to execute directly off Arcadia's mappings
- no more tickets back and forth between sul-dlss-labs/cocina-descriptive-metadata and this github repo

All the changes were to rspec code, except some yard style documentation added to a few cocina mapping methods.  Code coverage should not go down.


spec/services/cocina/from_fedora/descriptive/notes_spec.rb
```diff
@@ -133,100 +133,7 @@
     end
   end
 
-  context 'with a single abstract' do
-    let(:xml) do
-      <<~XML
-        <abstract>This is an abstract.</abstract>
```
ndushay, line 139: removed specs correspond to new mappings in `spec/services/cocina/mapping/descriptive/mods/`

spec/services/cocina/from_fedora/descriptive/notes_spec.rb
```diff
@@ -133,100 +133,7 @@
...
-          "value": 'This is a synopsis.',
-          "type": 'summary',
-          "displayLabel": 'Synopsis'
-        }
-
-      ]
-    end
-  end
-
+  # abstract extra test
```
ndushay, line 136: tagging @arcadiafalcone

spec/services/cocina/to_fedora/descriptive/subject_classification_spec.rb
```diff
@@ -19,120 +19,4 @@
```
ndushay, line 19: tagging @justinlittman @arcadiafalcone this is the only spec remaining in this file:  when cocina has no subjects.  Do we want to keep a test like this, and if so, perhaps not in a file specific to subject classification?

spec/support/mods_mapping_spec_helper.rb
```diff
@@ -24,3 +24,111 @@
     XML
   end
 end
```
ndushay, line 26: the previous shared example code was for an older approach to this problem and will be going away as we use this new approach.

justinlittman (approved):

spec/services/cocina/mapping/descriptive/mods/abstract_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+                }
+              ]
+            }
+          ]
+        }
+      end
+    end
+  end
+
+  describe 'display label' do
```
arcadiafalcone, line 80: Should this be "abstract with display label" instead of just "display label," to fit the pattern of the other tests?

spec/services/cocina/to_fedora/descriptive/subject_classification_spec.rb
arcadiafalcone: If you want to keep it, I think it makes sense to have a file with all of the nil tests together.

spec/support/mods_mapping_spec_helper.rb
```diff
@@ -24,3 +24,111 @@
...
+    it 'cocina Description maps to normalized MODS' do
+      # the starting MODS is normalized to address discrepancies found against MODS roundtripped to data store (Fedora)
+      #  and back, per Arcadia's specifications.  E.g., removal of empty nodes and attributes; addition of eventType to
+      #  originInfo nodes.
+      expect(actual_xml).to be_equivalent_to Cocina::ModsNormalizer.normalize(mods_ng_xml: orig_mods_ng, druid: local_druid)
+    end
+  end
+end
+
+def ng_mods_for(snippet)
```
arcadiafalcone, line 112: Could we make this MODS 3.7 instead of 3.6? There is one originInfo mapping where it makes a difference, but otherwise I'd prefer to use the most recent version as the default.

spec/services/cocina/mapping/descriptive/mods/abstract_spec.rb
justinlittman: For what it is worth, I have been copying the titles exactly from the original examples (including capitalization).

spec/services/cocina/mapping/descriptive/mods/subject_classification_spec.rb
```diff
@@ -0,0 +1,140 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'MODS subject classification <--> cocina mappings' do
+  describe 'with authority' do
```
arcadiafalcone, line 6: Ditto above, making the label "classification with authority" for readability, with the same applying to the other examples.

spec/support/mods_mapping_spec_helper.rb
ndushay: Let's bump the MODS version to 3.7 as a separate PR as it involves touching code as well as specs.

spec/services/cocina/mapping/descriptive/mods/abstract_spec.rb
ndushay: With or without the number?

spec/services/cocina/to_fedora/descriptive/subject_classification_spec.rb
ndushay: removed.

spec/services/cocina/mapping/descriptive/mods/abstract_spec.rb
justinlittman: Without.

spec/services/cocina/mapping/descriptive/mods/subject_classification_spec.rb
ndushay: used your descriptions, except added context where you didn't have it (e.g. when it was just "Display label")

spec/support/mods_mapping_spec_helper.rb
ndushay: @justinlittman should mods_normalizer also be changed to 3.7?

spec/services/cocina/mapping/descriptive/mods/abstract_spec.rb
ndushay: example descriptions redone.

spec/support/mods_mapping_spec_helper.rb
ndushay: working on this in separate PR.

spec/services/cocina/from_fedora/descriptive/notes_spec.rb
jcoyne: What does this mean?

jcoyne (changes requested):

spec/services/cocina/from_fedora/descriptive/notes_spec.rb
ndushay: It means we have a test that Arcadia's examples doesn't have, so we will keep it in our code but let Arcadia know.   We will likely have some of these as we go along, and wanted to separate conversion of our specs to the new way from addressing all of these.

arcadiafalcone (approved):

spec/services/cocina/from_fedora/descriptive/notes_spec.rb
jcoyne: Would you mind updating the comment to say just that.

ndushay: done.

jcoyne (approved):

PR #1894, opened 2021-01-24
Move table of contents test

justinlittman:
## Why was this change made?
Align table of contents with new approach for mapping specs.

Unit

NA


.rubocop.yml
```diff
@@ -37,6 +37,7 @@ Metrics/ClassLength:
 RSpec/DescribeClass:
   Exclude:
     - 'spec/requests/**/*'
+    - 'spec/services/cocina/mapping/descriptive/mods/**/*'
```
jcoyne, line 40: Is this necessary for all the files?  I find that this really helps me navigate where I find the tests.

jcoyne (changes requested):

.rubocop.yml
justinlittman: In this particular part of the tests, the spec files align with the MODS elements not the ruby mapping classes. This is both how Arcadia organizes her work and makes sense since the examples often cross multiple ruby mapping classes.

In other parts of the mapping tests that do align with ruby mapping classes this rule is still in place.

spec/services/cocina/from_fedora/descriptive/notes_spec.rb
```diff
@@ -133,100 +133,7 @@
...
-          "value": 'This is a synopsis.',
-          "type": 'summary',
-          "displayLabel": 'Synopsis'
-        }
-
-      ]
-    end
-  end
-
+  # abstract extra test
```
jcoyne, line 136: I'm not sure what this means.

jcoyne:
It doesn't seem to me that the description of this PR accurately summarizes the changes seen here.

justinlittman:
Uggh, sorry @jcoyne. This was meant to be a draft. The PR is actually much smaller, but needs a rebase after #1892 is merged.

jcoyne (approved):

PR #1901, opened 2021-01-25
Use Mods 3.7 instead of Mods 3.6

ndushay:
## Why was this change made?

Arcadia asked for it

### mods 3.7 branch

```
Status (n=100000):
  Success:   98076 (98.076%)
  Different: 1286 (1.286%)
  To Cocina error:     28 (0.028%)
  To Fedora error:     0 (0.0%)
  Missing:     610 (0.61%)
```

### main branch

```
Status (n=100000):
  Success:   98076 (98.076%)
  Different: 1286 (1.286%)
  To Cocina error:     28 (0.028%)
  To Fedora error:     0 (0.0%)
  Missing:     610 (0.61%)
```

### DIFF
--- a/app/services/cocina/mods_normalizer.rb
+++ b/app/services/cocina/mods_normalizer.rb
@@ -2,6 +2,7 @@
 
 module Cocina
   # Normalizes a Fedora MODS document, accounting for differences between Fedora MODS and MODS generated from Cocina.
+  # these adjustments have been approved by our metadata authority, Arcadia.
   class ModsNormalizer
     MODS_NS = Cocina::FromFedora::Descriptive::DESC_METADATA_NS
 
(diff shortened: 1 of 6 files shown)


jcoyne:
Do you think we need to concurrently change the version in dor-services? Is there any chance this would change the functionality of the datastream object? https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/desc_metadata_ds.rb#L12

ndushay:
@arcadiafalcone do you think we should upgrade to mods 3.7 in non-cocina contexts (e.g. Argo) ?  Justin C also asks another good question above.

arcadiafalcone:
It would be good to have 3.7 be the default for all new content coming into the SDR.

arcadiafalcone (approved):

justinlittman (approved):

PR #1903, opened 2021-01-25
mods genre cocina mapping specs reworked

ndushay:
## Why was this change made?

reduce ongoing costs of mods/cocina mappings 

changed spec code only;  coverage should be the same


jcoyne (approved):

spec/services/cocina/from_fedora/descriptive/form_spec.rb
```diff
@@ -128,86 +128,9 @@
...
-              "code": 'lcgft',
-              "uri": 'http://id.loc.gov/authorities/genreForms/'
-            }
-          }
-        ]
-      end
-    end
-
-    context 'without valueURI' do
+    # MODS abstract mapping spec not in /mappings/mods/genre_spec.rb
```
arcadiafalcone, line 131: Shouldn't this be "genre mapping spec"?

ndushay: yes - fixed in a later PR.

PR #1907, opened 2021-01-26
mods identifier <--> cocina mapping specs reworked

ndushay:
## Why was this change made?

reduce ongoing costs of mods/cocina mappings

changes to specs only

### DIFF
--- a/spec/services/cocina/from_fedora/descriptive/identifier_spec.rb
+++ b/spec/services/cocina/from_fedora/descriptive/identifier_spec.rb
@@ -19,34 +19,6 @@ RSpec.describe Cocina::FromFedora::Descriptive::Identifier do
     XML
   end
 
-  context 'with an identifier that is from Standard Identifier Scheme' do
-    let(:xml) do
-      <<~XML
-        <identifier type="isbn">1234 5678 9203</identifier>
-      XML
-    end
-
-    it 'builds the cocina data structure' do
-      expect(build).to eq [
-        {
-          "value": '1234 5678 9203',
-          "type": 'ISBN',
-          "note": [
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

spec/services/cocina/from_fedora/descriptive/identifier_spec.rb
```diff
@@ -19,34 +19,8 @@
...
-                "uri": 'http://id.loc.gov/vocabulary/identifiers/'
-              }
-            }
-          ]
-        }
-      ]
-    end
-  end
-
+  # MODS identifier mapping spec not in /mappings/mods/identifier_spec.rb
```
justinlittman, line 22: I'd suggest that these comments be omitted, as they will take another PR to remove later. I don't think comments are a good way to track this.

justinlittman (approved):

spec/services/cocina/from_fedora/descriptive/identifier_spec.rb
ndushay: How do you envision we'll deal w these?  Arcadia go through all the remaining from_ and to_ fedora specs and pull what she wants into the new style mapping ones?   And then devs go through what remains?

I'm open to not using the comments and I'm open to removing the existing ones I've put in this PR and merged ones ... but I'd like to work out a plan w Arcadia.

ndushay: Per slack discussion, I am mollified.  I have removed the offending comments from this PR and will remove the others I have added elsewhere in a separate PR.

PR #1908, opened 2021-01-26
xit formerly passing record_info mods mapping spec until marc -> mods uses MODS 3.7

ndushay:
## Why was this change made?

to fix broken build

the build now passes.


mjgiarlo (approved):

ndushay:
merging now to fix build.  Still want Arcadia and JLitt to be aware.

PR #1914, opened 2021-01-26
Adds mapping tests for subject for H2.

justinlittman:
## Why was this change made?
Move tests from CDM repo to specs.

Unit

NA

### DIFF
--- a/spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
@@ -1,0 +1,310 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'MODS subject <--> cocina mappings for H2' do
+  describe 'FAST topic' do
+    it_behaves_like 'MODS cocina mapping' do
+      let(:mods) do
+        <<~XML
+          <subject authority="fast">
+            <topic authority="fast" authorityURI="http://id.worldcat.org/fast/" valueURI="http://id.worldcat.org/fast/1009447">Marine biology</topic>
+          </subject>
+        XML
+      end
+
+      let(:cocina) do
+        {
+          "subject": [
+            {
+              "value": 'Marine biology',
+              "type": 'topic',
+              "uri": 'http://id.worldcat.org/fast/1009447',
+              "source": {
+                "code": 'fast',
+                "uri": 'http://id.worldcat.org/fast/'
+              }
+            }
+          ]
+        }
+      end
+    end
+  end
+
+  describe 'FAST personal name' do
+    it_behaves_like 'MODS cocina mapping' do
+      let(:mods) do
+        <<~XML
+          <subject authority="fast">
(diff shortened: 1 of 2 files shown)


spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
```diff
@@ -0,0 +1,310 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'MODS subject <--> cocina mappings for H2' do
```
jcoyne, line 5: I don't understand why we need tests for h2 because h2 is just using cocina.

jcoyne (approved):

spec/support/mods_mapping_spec_helper.rb
```diff
@@ -42,7 +45,7 @@
 
   let(:local_errors) { defined?(errors) ? errors : [] }
 
-  context 'when mapping from MODS (to cocina)' do
```
ndushay, line 45: the parens visually set apart the "to cocina" and "from cocina"  from the  "from MODS" and "to MODS" for easier skimming

spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
```diff
@@ -0,0 +1,310 @@
...
+              "source": {
+                "code": 'fast',
+                "uri": 'http://id.worldcat.org/fast/'
+              }
+            }
+          ]
+        }
+      end
+
+      # type will be submitted as "event" by H2. However, this will be lost in roundtrip to Cocina.
```
ndushay, line 178: not lost, so much as changed to "topic".   I wonder if @arcadiafalcone actually wants this to be changed.

spec/support/mods_mapping_spec_helper.rb
```diff
@@ -115,6 +118,29 @@
       expect(actual_xml).to be_equivalent_to Cocina::ModsNormalizer.normalize(mods_ng_xml: orig_mods_ng, druid: local_druid)
     end
   end
+
+  # For H2, descriptive metadata is submitted as cocina, which is then mapped to and persisted as MODS. This verifies
+  # that the MODS can then be roundtripped back to cocina (including allowing for lost data when a source_cocina is provided).
+  context 'when mapping to cocina from MODS from cocina' do
```
ndushay, line 124: if Arcadia believes we need this, I'm ok w it.  I would like her explicit sign off, unless you've agreed verbally already ...

ndushay (changes requested):
just some questions and a request to keep parens.

spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
justinlittman: It is how Arcadia organizes her examples. H2 is just cocina, but it has some specific examples, e.g., FAST.

ndushay: @jcoyne because Arcadia separated them.  We are following her specs here.   But JLitt and I agree it's really more about cocina -> MODS -> cocina rather than about H2.

PR #1916, opened 2021-01-26
Mapping tests for subject for H2

justinlittman:
…to MODS to COCINA (for H2)

## Why was this change made?
* Add tests for H2 subjects.
* Add testing for cocina-first mapping.

Unit

NA

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,7 +37,7 @@ Metrics/ClassLength:
 RSpec/DescribeClass:
   Exclude:
     - 'spec/requests/**/*'
-    - 'spec/services/cocina/mapping/descriptive/mods/**/*'
+    - 'spec/services/cocina/mapping/descriptive/**/*'
 
 RSpec/ExampleLength:
   Max: 8
--- a/spec/services/cocina/mapping/descriptive/h2/subject_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/h2/subject_spec.rb
@@ -1,0 +1,294 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Cocina --> MODS mappings for FAST subjects' do
+  describe 'FAST topic' do
+    it_behaves_like 'cocina MODS mapping' do
+      let(:mods) do
+        <<~XML
+          <subject authority="fast">
+            <topic authority="fast" authorityURI="http://id.worldcat.org/fast/" valueURI="http://id.worldcat.org/fast/1009447">Marine biology</topic>
+          </subject>
+        XML
+      end
+
+      let(:cocina) do
+        {
+          "subject": [
+            {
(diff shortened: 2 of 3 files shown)


spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
```diff
@@ -0,0 +1,294 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'MODS subject <--> cocina mappings for H2' do
```
ndushay, line 5: should this read `Cocina <--> MODS mappings for H2 subjects' or even `Cocina <--> MODS mappings for FAST subjects`?

ndushay (approved):
LGTM;

@arcadiafalcone do you want these h2 specs:
- to keep h2 in the spec file name and description
- to be in the same directory as the other MODS specs?

ndushay (changes requested):
oops - meant to request changes so Arcadia can answer

spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
arcadiafalcone: They're really Cocina --> MODS mappings at this point - they're not roundtripped back to Cocina. `Cocina --> MODS mappings for FAST subjects` is probably the most accurate.

arcadiafalcone: Let's keep H2 in the file names and descriptions -- these mappings are really dependent on the structure of the H2 metadata entry form. Having them in a different directory would be preferable.

justinlittman: OK. Moved them to an `h2` directory and used the description that you provided.

ndushay: ugh - should it be a subdirectory of MODS?  since it's still MODS?

ndushay (approved):
one last question for @arcadiafalcone about whether h2 should be a subdirectory for MODS and one for @justinlittman double checking these aren't already in our old style subject specs.

Approving with those caveats.

spec/services/cocina/mapping/descriptive/mods/subject_h2_spec.rb
arcadiafalcone: It's primarily an H2>Cocina mapping though - the MODS is secondary. I don't think it should be a subdirectory of MODS.

justinlittman: We can always change later if `H2` is no longer appropriate.

justinlittman:
I think contributor and form are the only area with old style tests for H2.

PR #1920, opened 2021-01-27
Clean up duplicate form tests.

justinlittman:
## Why was this change made?
Remove duplicate tests.

Unit

NA


spec/services/cocina/from_fedora/descriptive/form_spec.rb
```diff
@@ -81,113 +81,6 @@
...
-          <subject altRepGroup="1">
-            <cartographics>
-              <scale>&#x6BD4;&#x4F8B;&#x5C3A; 1:650,000.</scale>
-            </cartographics>
-          </subject>
-        XML
-      end
-
-      it 'builds the cocina data structure' do
-        expect(build).to eq [
```
jmartin-sul, line 174: i can't seem to find this cocina is fragment as an expectation in this new branch, though i do see the XML fragment and an expectation for an empty array as a result here:  https://github.com/sul-dlss/dor-services-app/blob/clean_up_form/spec/services/cocina/from_fedora/descriptive/subject_spec.rb#L345-L365

is this me not finding the right thing, or misunderstanding the de-duping here, or the de-duping being too aggressive, or ...?

jmartin-sul (commented):
one question, might just be me missing something, overall LGTM

spec/services/cocina/from_fedora/descriptive/form_spec.rb
justinlittman: Good catch. For some reason this was labeled as 19c even though 19c doesn't exist. Putting it back so Arcadia can sort it out.

jcoyne: 19c is https://github.com/sul-dlss-labs/cocina-descriptive-metadata/pull/331/files

justinlittman: OK, so I merged 19c and added as a test in `subject_cartographic_spec.rb` and deleted here.

jmartin-sul (approved):
LGTM, thanks for the revisit!

PR #1926, opened 2021-01-27
Add mapping tests for related resource for H2.

justinlittman:
## Why was this change made?
Align tests.

Unit

NA

### DIFF
--- a/spec/services/cocina/mapping/descriptive/h2/related_item_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/h2/related_item_spec.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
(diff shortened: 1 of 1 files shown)


jcoyne (approved):

spec/services/cocina/mapping/descriptive/h2/related_item_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+                  "value": 'https://www.example.com/paper.html'
+                }
+              ]
+            }
+          }
+        ]
+      }
+    end
+
+    xit 'broken'
```
jcoyne, line 86: Is this supposed to be here?

justinlittman: Yes.

PR #1930, opened 2021-01-27
mods name <--> cocina mapping specs refactored

ndushay:
## Why was this change made?

To reduce ongoing costs of mods/cocina mappings

it was all test code changes; there should be no coverage decrease (and there's less overall test code!)


spec/support/mods_mapping_spec_helper.rb
```diff
@@ -66,7 +66,15 @@
...
     it 'MODS maps to expected cocina' do
-      expect(actual_cocina_props).to be_deep_equal(cocina)
+      if cocina
+        expect(actual_cocina_props).to be_deep_equal(cocina)
+      else
+        actual_cocina_props.each_key do |key|
+          return true if actual_cocina_props[key].blank?
+
+          expect(actual_cocina_props[key] == [{}]).to eq true
+        end
```
ndushay, line 76: calling this out for @justinlittman to review.  the change you made seemed to require me to put nil in the spec itself, but needed this fix here as well.   It's not elegant, but it works.  Perhaps you have a better solution.

ndushay:
I am not finding which code is no longer covered by tests ...

spec/support/mods_mapping_spec_helper.rb
```diff
@@ -66,7 +66,15 @@
...
 
     it 'MODS maps to expected cocina' do
-      expect(actual_cocina_props).to be_deep_equal(cocina)
+      if cocina
+        expect(actual_cocina_props).to be_deep_equal(cocina)
+      else
+        actual_cocina_props.each_key do |key|
+          return true if actual_cocina_props[key].blank?
+
+          expect(actual_cocina_props[key] == [{}]).to eq true
```
mjgiarlo, line 75: I'm curious if `expect(actual_cocina_props[key]).to eq [{}]` would work just as well here (and below).

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
```diff
@@ -0,0 +1,1553 @@
...
+          }
+        ]
+      }
+    end
+  end
+
+  describe 'Name with language' do
+    xit 'TODO: implement "both type and status should be siblings to name, as they apply to the contributor entity."'
+
+    let(:druid) { 'cp049zn0898' } # adapted from ...
```
justinlittman, line 1488: This should just a comment `# adapted from druid:cp049zn0898`.

The `let(:druid)` is for when the druid is needed for the mapping.

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
```diff
@@ -0,0 +1,1553 @@
...
+          # NOTE: strictly speaking, the following warning isn't true ...
+          Notification.new(msg: 'Missing name/namePart element')
+          # FIXME: we're not warning about missing roleTerm code/value?
+        ]
+      end
+    end
+  end
+
+  describe 'Unauthorized role term only' do
+    # Should this warn?
```
justinlittman, line 593: These seems crufty. Ticket or ask Arcadia?

spec/support/mods_mapping_spec_helper.rb
justinlittman: I don't understand what problem this is solving. Can you say more?

justinlittman (changes requested):

jmartin-sul:
> I am not finding which code is no longer covered by tests ...

according to the code climate link in the PR, it looks like the issue is in `contributor_writer.rb`, and comparing `main` against this branch using the code climate browser plugin, it looks like this is what lost coverage:
https://github.com/sul-dlss/dor-services-app/blob/mods-name-specs/app/services/cocina/to_fedora/descriptive/contributor_writer.rb#L136-L164

vs

https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/to_fedora/descriptive/contributor_writer.rb#L136-L164

(in this branch's version, it appears that much of the `write_note`, `write_identifier`, and `write_display_form` coverage was lost)

spec/support/mods_mapping_spec_helper.rb
ndushay: pull the branch and try running `rspec spec/services/cocina/mapping/descriptive/mods/name_spec.rb:537` - it will explain it better than I can.

Something about the deep_equal implementation cannot equate `nil` with `[{}]`, IIRC.

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
ndushay: oooOOOOOooooo

justinlittman (approved):

ndushay:
John - thanks for the research. @justinlittman - I'm inclined to ignore the coverage difference - what do you think?

PR #1933, opened 2021-01-27
add access_h2_spec for cocina mappings

ndushay:
## Why was this change made?

adding Arcadia's mappings for MODS h2 access specs.

it adds two "xit" specs ...

### DIFF
--- a/spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Cocina --> MODS mappings for access (H2 specific)' do
+  describe 'Contact email' do
+    xit 'FIXME: get MODS mapping for email (is it part of name ?)'
+
+    # this is clearly wrong
+    let(:mods) do
+      <<~XML
+        <location>
+          <physicalLocation type="email" displayLabel="Contact">me@stanford.edu</physicalLocation>
+        </location>
+      XML
+    end
+
+    let(:cocina) do
+      {
+        access: {
+          accessContact: [
+            {
+              value: 'me@stanford.edu',
+              type: 'email',
+              displayLabel: 'Contact'
+            }
+          ],
+          digitalRepository: [
+            {
+              value: 'Stanford Digital Repository'
+            }
+          ]
+        }
+      }
+    end
+  end
+
+  describe 'Contact email not provided' do
+    xit 'FIXME: cocina -> MODS ok, but no way to get back'
+
+    let(:mods) { '' }
+
+    let(:cocina) do
+      {
+        access: {
+          digitalRepository: [
+            {
+              value: 'Stanford Digital Repository'
+            }
+          ]
+        }
+      }
+    end
+  end
+end


spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+              value: 'Stanford Digital Repository'
+            }
+          ]
+        }
+      }
+    end
+  end
+
+  describe 'Contact email not provided' do
+    xit 'FIXME: cocina -> MODS ok, but no way to get back'
```
justinlittman, line 39: I don't understand. Is the example incomplete? In which case I would omit `let(:mods)` and have xit be something like "incomplete mapping".

spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Cocina --> MODS mappings for access (H2 specific)' do
+  describe 'Contact email' do
+    xit 'FIXME: get MODS mapping for email (is it part of name ?)'
```
justinlittman, line 7: I don't understand. Is this not implemented, in which case I would recommend "not implemented" as xit? Or is something else wrong?

justinlittman (changes requested):

spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
ndushay: it's more that h2 provides the "in the stanford repository" bit, which is good, but we don't put that in the MODS, so this particular h2 mapping is NEVER going to roundtrip properly.

spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
ndushay: Arcadia's spec itself doesn't have the MODS, but this will clearly map to MODS somewhere, so this seemed like a useful message.

You are going for very brief and standardized xits, and I am going for more specific ones as I write these examples and learn stuff.  :-P

justinlittman: I'm my mind, we have 3 types of incomplete tests:
1. Not implemented: the example is complete, but we haven't written the mapping code.
2. Broken: the example is complete and we wrote mapping code, but the mapping code doesn't work.
3. Incomplete mapping: the example is incomplete.

The challenge is that your xit comments make sense to you, but when I read them I can't make sense of them, or at least without digging into the test. Perhaps we want to some standardized prefixes and then additional details be optional, like `incomplete mapping: get MODS mapping for email (is it part of name ?)`.

justinlittman: Happy to discuss more.

ndushay: I propose to do a pass of all the cocina mapping xit comments as a separate PR when mine are merged.  I'll be sure to standardize on your prefixes.   Also, for h2 specs, I remembered we can have diff roundtrip_cocina but only after a bunch of PRs were already merged.

If you want me to address the xits and so on in this PR before merging, rather than bundling it in, I can do ...

justinlittman (approved):

PR #1935, opened 2021-01-27
Fix originInfo eventType differs from date type test.

### LINKED ISSUE #1932: Fix "originInfo eventType differs from date type" test
From `origin_info_spec.rb`.

justinlittman:
closes #1932

## Why was this change made?
Broken test bad.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/mods_normalizer.rb
+++ b/app/services/cocina/mods_normalizer.rb
@@ -184,43 +184,36 @@ module Cocina
     end
 
     # change original xml to have the event type that will be output
-    # rubocop:disable Metrics/CyclomaticComplexity
-    # rubocop:disable Metrics/PerceivedComplexity
     def normalize_origin_info_event_types
       ng_xml.root.xpath('//mods:originInfo', mods: MODS_NS).each do |origin_info_node|
-        date_issued_nodes = origin_info_node.xpath('mods:dateIssued', mods: MODS_NS)
-        add_event_type('publication', origin_info_node) && next if date_issued_nodes.present?
+        next if normalize_event_type(origin_info_node, 'dateIssued', 'publication', ->(oi_node) { oi_node['eventType'] != 'presentation' })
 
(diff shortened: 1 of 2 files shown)


justinlittman:
Before:
```
Status (n=100000):
  Success:   98117 (98.117%)
  Different: 1334 (1.334%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

After:
```
Status (n=100000):
  Success:   98153 (98.153%)
  Different: 1298 (1.298%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

app/services/cocina/mods_normalizer.rb
```diff
@@ -204,23 +208,24 @@ def normalize_origin_info_event_types
...
+        add_event_type('publication', origin_info_node) && next if frequency.present? && origin_info_node['eventType'].nil?
       end
     end
     # rubocop:enable Metrics/CyclomaticComplexity
     # rubocop:enable Metrics/PerceivedComplexity
+    # rubocop:enable Metrics/AbcSize
 
     def add_event_type(value, origin_info_node)
-      origin_info_node['eventType'] = value if origin_info_node[:eventType].blank?
+      origin_info_node['eventType'] = value
```
jcoyne, line 228: Since there's no longer a conditional here, would it make sense to just inline that in the method above?  I found this really hard to follow because you were testing `origin_info_node['eventType'].nil?` and then calling a separate method that was mutating that value.

jcoyne (changes requested):

app/services/cocina/mods_normalizer.rb
justinlittman: I tried inlining, but it just made things worse. So I took a different tact -- encapsulate more of the logic in the helper method to reduce the complexity of the main method.

jcoyne (approved):
Thanks, this is much easier for me to read.

PR #1942, opened 2021-01-28
Mods name specs tweak

justinlittman:
## Why was this change made?
So that contributor correctly return nil when empty rather than an empty structure.

Unit

NA


mjgiarlo (approved):

ndushay (approved):
thank you - this makes MUCH more sense now.

PR #1951, opened 2021-01-28
Remove cruft.

justinlittman:
## Why was this change made?
Remove cruft.


ndushay (approved):
approved but I think you have an extra space.  Merge when green.

PR #1954, opened 2021-01-28
H2 mappings - xit update

ndushay:
## Why was this change made?

Standardizing xit practices on mapping specs.  Also, I forgot about roundtrip_cocina so was able to add some specs to admin_metadata.

it's all spec code

### DIFF
--- a/spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
@@ -4,16 +4,7 @@ require 'rails_helper'
 
 RSpec.describe 'Cocina --> MODS mappings for access (H2 specific)' do
   describe 'Contact email' do
-    xit 'FIXME: get MODS mapping for email (is it part of name ?)'
-
-    # this is clearly wrong
-    let(:mods) do
-      <<~XML
-        <location>
-          <physicalLocation type="email" displayLabel="Contact">me@stanford.edu</physicalLocation>
-        </location>
-      XML
-    end
+    xit 'not mapped: email to MODS'
 
     let(:cocina) do
       {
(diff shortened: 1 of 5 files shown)


spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
```diff
@@ -33,12 +24,15 @@
         }
       }
     end
+
+    let(:mods) do
```
justinlittman, line 28: Why the empty `let(:mods)`? And why is it different than the empty mods let below?

spec/services/cocina/mapping/descriptive/h2/admin_metadata_h2_spec.rb
```diff
@@ -49,63 +49,102 @@
...
-    xit 'TODO: implement recordInfo/recordChangeDate as a note when mapping MODS -> cocina'
+    xit 'not implemented: recordInfo/recordChangeDate as a note when mapping MODS -> cocina'
 
     let(:create_date) { '2014-04-08' }
 
     let(:modification_date) { '2014-10-22' }
 
-    let(:druid) { 'jv545yc8727' }
+    it_behaves_like 'cocina MODS mapping' do
+      let(:druid) { 'jv545yc8727' }
```
justinlittman, line 59: This should be a comment, right? `# Adapted from druid:jv545yc8727`.

justinlittman (changes requested):

spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb
ndushay: I was trying to run these to see what happened.  I will remove both.

spec/services/cocina/mapping/descriptive/h2/admin_metadata_h2_spec.rb
ndushay: will fix.  when is druid "needed" ?

justinlittman (approved):

PR #1958, opened 2021-01-28
mods record_info mapping - fix broken spec

ndushay:
## Why was this change made?

fixes a broken mapping spec 

changed spec only.


justinlittman (approved):

ndushay:
this has been dealt with in another PR.

Issue #1936, opened 2021-01-27
add MODS mapping specs for value_dependencies examples


ndushay:
if there aren’t already specific examples for these https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_value_dependencies.txt, please add them to the `spec/services/cocina/mappings/descriptive/mods/xxx` spec files as appropriate (or make a new one).

PR #1964, opened 2021-01-29
Add external link mapping to table of contents spec

arcadiafalcone:
## Why was this change made?
Partially addresses #1936 

n/a

n/a

### DIFF
--- a/spec/services/cocina/mapping/descriptive/mods/table_of_contents_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/mods/table_of_contents_spec.rb
@@ -110,6 +110,27 @@ RSpec.describe 'MODS tableOfContents <--> cocina mappings' do
     end
   end
 
+  describe 'Link to external value only' do
+    xit 'not implemented'
+
+    let(:mods) do
+      <<~XML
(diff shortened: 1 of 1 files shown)


spec/services/cocina/mapping/descriptive/mods/table_of_contents_spec.rb
```diff
@@ -110,6 +110,27 @@
     end
   end
 
+  describe 'Link to external value only' do
+    it_behaves_like 'MODS cocina mapping' do
+      let(:mods) do
+        <<~XML
+          <tableOfContents xlink:href="http://contents.org/contents" />
+        XML
+
```
justinlittman, line 119: Missing an `end` here.

justinlittman (changes requested):
Does this work or should it be an `xit`?

spec/services/cocina/mapping/descriptive/mods/table_of_contents_spec.rb
arcadiafalcone: Fixed.

justinlittman (approved):

justinlittman:
#2027 is for implementation.

PR #1980, opened 2021-01-29
implement typeOfResource mapping when attribute without value

### LINKED ISSUE #1969: Implement mapping for attribute without value
See "Attribute without value" in `spec/services/cocina/mapping/descriptive/mods/type_of_resource_spec.rb`.

ndushay:
## Why was this change made?

Fixes #1969 

### This branch

I’m assuming that many of the mappings we’re now implementing occur rarely, and thus our roundtrip validation tests may show exactly the same numbers as main.   Roundtrip did show a difference before I adjusted mods_normalizer.

```
Status (n=50000):
  Success:   49078 (98.156%)
  Different: 595 (1.19%)
  To Cocina error:     16 (0.032%)
  To Fedora error:     0 (0.0%)
  Missing:     311 (0.622%)
```

### main branch

```
Status (n=50000):
  Success:   49078 (98.156%)
  Different: 595 (1.19%)
  To Cocina error:     16 (0.032%)
  To Fedora error:     0 (0.0%)
  Missing:     311 (0.622%)

```

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/form.rb
+++ b/app/services/cocina/from_fedora/descriptive/form.rb
@@ -118,14 +118,16 @@ module Cocina
 
         def add_types(forms)
           type_of_resource.each do |type|
-            forms << {
-              value: type.text,
-              type: 'resource type',
-              source: {
-                value: 'MODS resource types'
-              },
-              displayLabel: type[:displayLabel].presence
-            }.compact
+            if type.text.present?
+              forms << {
+                value: type.text,
+                type: 'resource type',
+                source: {
+                  value: 'MODS resource types'
+                },
+                displayLabel: type[:displayLabel].presence
+              }.compact
+            end
 
             if type[:manuscript] == 'yes'
               forms << {
--- a/app/services/cocina/to_fedora/descriptive/form.rb
+++ b/app/services/cocina/to_fedora/descriptive/form.rb
@@ -35,12 +35,16 @@ module Cocina
           is_manuscript = Array(forms).any? { |form| manuscript?(form) }
           is_collection = Array(forms).any? { |form| collection?(form) }
 
-          other_forms.each do |form|
-            if form.structuredValue
-              write_structured(form)
-            elsif form.value
-              write_basic(form, is_manuscript: is_manuscript, is_collection: is_collection)
+          if other_forms.present?
+            other_forms.each do |form|
+              if form.structuredValue
+                write_structured(form)
+              elsif form.value
+                write_basic(form, is_manuscript: is_manuscript, is_collection: is_collection)
+              end
             end
+          else
+            write_basic(nil, is_manuscript: is_manuscript, is_collection: is_collection)
           end
 
           physical_description_forms = Array(forms).select { |form| physical_description_member?(form) }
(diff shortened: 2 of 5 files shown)


app/services/cocina/to_fedora/descriptive/form.rb
```diff
@@ -86,6 +90,7 @@ def write_physical_description(forms)
         end
 
         def write_basic(form, is_manuscript: false, is_collection: false)
+          return write_attributes_only(is_manuscript, is_collection) if form.nil?
```
mjgiarlo, line 93: Was this line added to accommodate what's now on line 47 or are there other cases where `form` can be `nil`? If there are other cases, this seems right. If there aren't, line 47 could change to `write_attributes_only(is_manuscript, is_collection)` since we already known the form is `nil` there. Do not feel strongly about this at all.

mjgiarlo (approved):

app/services/cocina/from_fedora/descriptive/form.rb
```diff
@@ -118,14 +118,16 @@ def add_structured_genre(forms)
...
           type_of_resource.each do |type|
-            forms << {
-              value: type.text,
-              type: 'resource type',
-              source: {
-                value: 'MODS resource types'
-              },
-              displayLabel: type[:displayLabel].presence
-            }.compact
+            if type.text.present?
```
jcoyne, line 121: Why did you choose to use present? here? Is this to check for an empty string? Do we have a test for that case?

jcoyne (changes requested):

app/services/cocina/to_fedora/descriptive/form.rb
ndushay: See line 43, which is like line 47, but only one has form=nil.

app/services/cocina/from_fedora/descriptive/form.rb
ndushay: While developing, I wrote some extra tests for attribute `collection="yes"` and for <typeOfResource> without attributes or a value.  This code worked.   It's possible just `type.text` would also have worked.

I removed my extra tests since the mapping files are now really Arcadia's and I didn't want to clutter them up.

Does `.present?` offend because it might be unnecessary?  It seems like a small thing to quibble over.

ndushay: I also wonder if we do get an empty string from `<element></element>` - we've had to add `.present?` to a lot of mapping code over the last few months.

jcoyne: I'm concerned about this, because it's not clear to me as a reader of the code what cases you are actually looking for.  There are several cases handled by this method, but it's not clear that we need to handle all of these cases.  If we do, we should have tests for them.  That is, if I change the code to `if type.text` then one of the tests ought to break.   I feel like the code should be as simple as possible to handle the cases we need to.  And `type.text` is simpler than `type.text.present?`

ndushay: ah.  and I feel like our MODS is made in so many different ways, that it makes sense for our code to cope broadly with possible empty attributes and elements and am not sure if we want all of those tests in Arcadia's mapping specs or not, or if we want them all in to_fedora and from_fedora ... or not.   Since our practice has been inconsistent in terms of having specs for this in the past ... it seems weird that this is now a sticking point in my 2 PRs.

ndushay: added test.

jcoyne (approved):

PR #1984, opened 2021-01-30
Event encoding, qualifier cocina attribs at outermost level to which they apply

### LINKED ISSUE #1971: Implement event mapping for approximate creation date: approx. 1900
See "Approximate creation date: approx. 1900" in `spec/services/cocina/mapping/descriptive/h2/event_h2_spec.rb`.

### LINKED ISSUE #1972: Implement mapping for creation date range: 2020-01-01 to 2021-01-01
See "Creation date range: 2020-01-01 to 2021-01-01" in `spec/services/cocina/mapping/descriptive/h2/event_h2_spec.rb`.

ndushay:
## Why was this change made?

fixes event mapping for `encoding` and `qualifier` attributes -- Arcadia says 

> "the pattern is for properties to be at the highest [outermost] level to which they apply"

Also required fixing existing originInfo/event mapping specs to be consistent with this newly implemented rule.

Fixes #1971 
Fixes #1972

specs updated plus roundtrip validation:

### this branch

```
Status (n=50000):
  Success:   49078 (98.156%)
  Different: 595 (1.19%)
  To Cocina error:     16 (0.032%)
  To Fedora error:     0 (0.0%)
  Missing:     311 (0.622%)
```

### main branch

```
Status (n=50000):
  Success:   49078 (98.156%)
  Different: 595 (1.19%)
  To Cocina error:     16 (0.032%)
  To Fedora error:     0 (0.0%)
  Missing:     311 (0.622%)
```

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/event.rb
+++ b/app/services/cocina/from_fedora/descriptive/event.rb
@@ -372,8 +372,8 @@ module Cocina
 
           dates = node_set.map do |node|
             new_node = node.deep_dup
-            new_node.remove_attribute('encoding') if common_attribs[:encoding].present?
-            new_node.remove_attribute('qualifier') if common_attribs[:qualifier].present?
+            new_node.remove_attribute('encoding') if common_attribs[:encoding].present? || node[:encoding]&.size&.zero?
+            new_node.remove_attribute('qualifier') if common_attribs[:qualifier].present? || node[:qualifier]&.size&.zero?
             build_date(type, new_node)
           end
           { structuredValue: dates }.merge(common_attribs).compact
@@ -382,12 +382,12 @@ module Cocina
         def common_date_attributes(node_set)
           first_encoding = node_set.first['encoding']
           first_qualifier = node_set.first['qualifier']
-          encoding_is_common = node_set.all? { |node| node['encoding'] == first_encoding } if first_encoding.present?
-          qualifier_is_common = node_set.all? { |node| node['qualifier'] == first_qualifier } if first_qualifier.present?
-          {
-            qualifier: (first_qualifier if qualifier_is_common),
-            encoding: ({ code: first_encoding } if encoding_is_common)
-          }.compact
+          encoding_is_common = node_set.all? { |node| node['encoding'] == first_encoding }
+          qualifier_is_common = node_set.all? { |node| node['qualifier'] == first_qualifier }
+          attribs = {}
+          attribs[:qualifier] = first_qualifier if qualifier_is_common && first_qualifier.present?
+          attribs[:encoding] = { code: first_encoding } if encoding_is_common && first_encoding.present?
+          attribs
         end
 
         def build_date(event_type, node)
--- a/app/services/cocina/from_fedora/descriptive/event.rb
+++ b/app/services/cocina/from_fedora/descriptive/event.rb
@@ -368,8 +368,26 @@ module Cocina
         def build_structured_date(type, node_set)
           return nil if node_set.blank?
 
-          dates = node_set.map { |node| build_date(type, node) }
-          { structuredValue: dates }
+          common_attribs = common_date_attributes(node_set)
+
+          dates = node_set.map do |node|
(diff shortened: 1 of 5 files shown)


app/services/cocina/from_fedora/descriptive/event.rb
```diff
@@ -368,8 +368,26 @@ def build_event(type, node_set, language_script = nil)
         def build_structured_date(type, node_set)
           return nil if node_set.blank?
 
-          dates = node_set.map { |node| build_date(type, node) }
-          { structuredValue: dates }
+          common_attribs = common_date_attributes(node_set)
+
+          dates = node_set.map do |node|
+            new_node = node.deep_dup
+            new_node.remove_attribute('encoding') if common_attribs[:encoding].present?
```
jcoyne, line 375: Why do we need `present?` here? Can't we just check that the value is there or not?  Present is for checking things that could be arrays or scalars, but this could never be an array.

jcoyne (changes requested):

app/services/cocina/from_fedora/descriptive/event.rb
```diff
@@ -368,8 +368,26 @@ def build_event(type, node_set, language_script = nil)
...
+            new_node.remove_attribute('qualifier') if common_attribs[:qualifier].present?
+            build_date(type, new_node)
+          end
+          { structuredValue: dates }.merge(common_attribs).compact
+        end
+
+        def common_date_attributes(node_set)
+          first_encoding = node_set.first['encoding']
+          first_qualifier = node_set.first['qualifier']
+          encoding_is_common = node_set.all? { |node| node['encoding'] == first_encoding } if first_encoding.present?
```
jcoyne, line 385: Same with present here. These are never arrays, right?

jcoyne (changes requested):

arcadiafalcone (approved):

app/services/cocina/from_fedora/descriptive/event.rb
ndushay: an attribute can be present but empty.

app/services/cocina/from_fedora/descriptive/event.rb
ndushay: an attribute can be present but empty.  And sometimes is.

jcoyne: Do we have a test for this case?

app/services/cocina/from_fedora/descriptive/event.rb
jcoyne: Can you add a test for this?

ndushay: can you explain better what problem you are trying to have me solve here?   In implementing lots of mappings, we had to put "compact" and "present?" all over the place to deal with empty values.  I doubt we have tests for all of those spots.    Are you suggesting we should?  If so, maybe make a ticket for systematically doing that everywhere, and discuss w Arcadia if she wants those specs in the official "mapping" specs or in the "unit" test specs of to_fedora and from_fedora.

If you're not suggesting that we become consistent on this point ... then I wonder why we would want to add rspec tests for this in this particular case.

jcoyne: @ndushay in all of the places, if we remove the `present?` or `compact` the tests fail. I'm not sure any test fails if you remove this code here. Therefore, this is extra complexity.  Our code should be no more complex than we need it to be.  We should either make it simple or demonstrate through a test as to why we need this complexity.

ndushay: Ho ho - wrote tests;  they failed.  Fixed code and added the tests I wrote.

app/services/cocina/from_fedora/descriptive/event.rb
ndushay: yes, we do now.

jcoyne (approved):

PR #2003, opened 2021-02-01
Updated cocina-models

justinlittman:
## Why was this change made?
Keep cocina-models in sync.

Unit

openapi.yml

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1228,6 +1228,9 @@ components:
         order:
           description: Place of contributor in order of authorship for resource.
           type: integer
+        valueAt:
+          description: URL or other pointer to the location of the contributor information.
+          type: string
     Description:
       type: object
       additionalProperties: false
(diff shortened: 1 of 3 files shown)


openapi.yml
```diff
@@ -2015,9 +2027,9 @@ components:
         version:
           description: The version of the related resource.
           type: string
-        order:
-          description: The order of the related resource in a series of related resources.
-          type: integer
+        valueAt:
```
jcoyne, line 2030: Was `order` supposed to have disappeared?

jcoyne (changes requested):

openapi.yml
jcoyne: Ah, yes, https://github.com/sul-dlss/cocina-models/commit/2b1222a90f48753c7b4b41103978c0c19ad10870#diff-bcc7b22977da21470db68de156bbc9a8d019b43434505510d35b61f8cce04bc4 should have been removed in 0.44.0

jcoyne (approved):

PR #2016, opened 2021-02-01
Update subject/name mapping spec

arcadiafalcone:
## Why was this change made?
Add roundtrip mapping and partially address #1936 

n/a

n/a

Implementation tickets #2017 and #2018


justinlittman (commented):
This has multiple specs. Is that intended?

arcadiafalcone:
That was intentional. They're on separate implementation tickets.

justinlittman (approved):

Issue #2018, opened 2021-02-01
Implement mapping subject-name authority handling


arcadiafalcone:
See 'Single subject subelement with authority code same as subject, no URI' at https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/subject_name_spec.rb

PR #2020, opened 2021-02-01
Add MODS mapping for subject name with role

arcadiafalcone:
## Why was this change made?
Add a missing mapping

n/a

n/a

Implementation ticketed at #2018


justinlittman (commented):
Is this intended to have 3 specs in it?

arcadiafalcone:
It's just supposed to be the subject/role mapping.

justinlittman (approved):

PR #2030, opened 2021-02-02
Un-xited Single subject subelement with authority code same as subject, no URI spec

### LINKED ISSUE #2018: Implement mapping subject-name authority handling

justinlittman:
closes #2018

## Why was this change made?
The test passes.

Unit

NA


ndushay (approved):
wow - this must've taken hours!

PR #2045, opened 2021-02-03
Implement mapping for name with xlink.

### LINKED ISSUE #2007: Implement mapping for name with xlink
See 'Name with external link to value only' in https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/name_spec.rb

justinlittman:
closes #2007

## Why was this change made?
More mapping goodness.

Unit, sdr-deploy

NA

### DIFF
--- a/spec/services/cocina/mapping/descriptive/mods/name_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/mods/name_spec.rb
@@ -1350,26 +1350,27 @@ RSpec.describe 'MODS name <--> cocina mappings' do
   end
 
   describe 'Name with external link to value only' do
-    xit 'not implemented'
+    it_behaves_like 'MODS cocina mapping' do
+      # Note this handling of xlink:href is ONLY for when it is the only attribute and there are no children.
+      let(:mods) do
+        <<~XML
+          <name xlink:href="http://name.org/name" />
+        XML
+      end
 
-    let(:mods) do
-      <<~XML
-        <name xlink:href="http://name.org/name" />
-      XML
-    end
-
-    let(:cocina) do
-      {
-        contributor: [
-          {
-            name: [
-              {
-                valueAt: 'http://name.org/name'
-              }
-            ]
-          }
-        ]
-      }
+      let(:cocina) do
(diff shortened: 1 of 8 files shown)


spec/services/cocina/mapping/descriptive/mods/name_spec.rb
```diff
@@ -1350,22 +1350,28 @@
...
 
-    let(:cocina) do
-      {
-        contributor: [
-          {
-            uri: 'http://name.org/name'
-          }
-        ]
-      }
+      let(:cocina) do
```
justinlittman, line 1362: contributor does not have a URI attribute, so moved URI to name.

justinlittman:
Before:
```
Status (n=100000):
  Success:   98154 (98.154%)
  Different: 1297 (1.297%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

After:
```
Status (n=100000):
  Success:   98154 (98.154%)
  Different: 1297 (1.297%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
arcadiafalcone: Contributor should have a URI attribute - I think it didn't come up before because MODS is name-based rather than RWO-based.

arcadiafalcone: It should be in the contributor valueAt property, but a uri property should be added to contributor as well.

arcadiafalcone (changes requested):
Name xlink should map to the name valueAt property.

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
arcadiafalcone: I see I got this wrong in the mapping - will update that.

arcadiafalcone (approved):

PR #2052, opened 2021-02-03
Mapping multi primary

### LINKED ISSUE #1952: Write mapping examples for when multiple usage=primary

arcadiafalcone:
## Why was this change made?
Closes #1952 

n/a

n/a

Implementation ticket: #1952


ndushay:
This will likely need some developer effort to pass.

ndushay (approved):
LGTM;

two questions:
1.  what didn't pass about a bunch of xit specs added?
2. when changing the hash to a ruby hash in `subject_spec` mapping ... why not just do the entire object?

justinlittman (approved):

Issue #2054, opened 2021-02-03
Mass metadata refresh for Google Books and Project Phoenix


arcadiafalcone:
Many of the roundtrip/validation errors for Google Books are artifacts of the MARC2MODS XSLT in place at the time. The most recent local XSLT fixes at least some of these transformation errors, so refreshing the metadata from source will increase the validation rate.

PR #2084, opened 2021-02-05
Adds tools for generating druid lists from solr and refreshing metadata.

justinlittman:
refs #2054

## Why was this change made?
Produce cleaner data by refreshing metadata for none problem collections.

NA

README.


bin/generate-druid-list
```diff
@@ -5,26 +5,25 @@ require_relative '../config/environment'
...
-sample_size = Integer(ARGV[0])
-REQUEST_SIZE = 250
+query = ARGV[0]
 
-druids = Set.new
+druids = []
 
-druids.merge(File.read('druids.txt').split) if File.exist?('druids.txt')
+loop do
+  results = ActiveFedora::SolrService.query('*:*', fl: 'id', rows: 250, fq: query, start: druids.length, sort: 'id asc')
```
jcoyne, line 17: Rows 10000 is going to be much faster (and won't stress solr) than doing more calls to solr will

justinlittman: Done.

jcoyne (approved):

Issue #2021, opened 2021-02-01
Add 'recording' to valid eventType values


arcadiafalcone:
bc096kk4399

arcadiafalcone:
recording
performance
presentation

ndushay:
It would be great to have examples of all of these.

For event type `recording`, as in bc096kk4399;  should all of these, when in a cocina model, map to `dateCreated` for MODS?   (what about `dateIssued`)?

For event type `performance`, what flavor of MODS date should cocina map to?

For event type `presentation`, what flavor of MODS date should cocina map to?

arcadiafalcone:
It doesn't change the precedence - the event type should be the value of eventType if present, and the already established defaults if missing.

ndushay:
This is addressed with new mapping code (PR coming shortly)

PR #2086, opened 2021-02-05
Spec event type

arcadiafalcone:
## Why was this change made?
For implementation ticket #2021 

n/a

n/a


ndushay (approved):
This is awesome.  Love the splitting of files.  Love having my new specs written to un-xit!

PR #2098, opened 2021-02-05
Implement mapping for access (H2 specific). Changes approach to addin…

### LINKED ISSUE #2008: Implement mapping for access (H2 specific)
See spec/services/cocina/mapping/descriptive/h2/access_h2_spec.rb.

justinlittman:
…g digital repository.

closes #2008

## Why was this change made?
More mappings. Add digital repository whenever PURL present.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/access.rb
+++ b/app/services/cocina/from_fedora/descriptive/access.rb
@@ -16,7 +16,7 @@ module Cocina
         # @param [Nokogiri::XML::Element] resource_element mods or relatedItem element
         # @param [Cocina::FromFedora::Descriptive::DescriptiveBuilder] descriptive_builder
         # @return [Hash] a hash that can be mapped to a cocina model
-        def self.build(resource_element:, descriptive_builder:, add_sdr: true)
+        def self.build(resource_element:, descriptive_builder:)
           new(resource_element: resource_element, descriptive_builder: descriptive_builder).build
         end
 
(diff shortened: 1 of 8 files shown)


justinlittman:
Before:
```
Status (n=100000):
  Success:   98293 (98.293%)
  Different: 1158 (1.158%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

After:
```
Status (n=100000):
  Success:   98293 (98.293%)
  Different: 1158 (1.158%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

arcadiafalcone (approved):

justinlittman:
Needs infra dev review.

app/services/cocina/from_fedora/descriptive/access.rb
```diff
@@ -139,7 +132,7 @@ def descriptive_value_for(nodes)
                 uri: Authority.normalize_uri(node[:authorityURI])
               }.compact
               attrs[:source] = source unless source.empty?
-              attrs[:type] = node[:type]
+              attrs[:type] = type || node[:type]
```
ndushay, line 135: i believe got bitten by an expression like this -- won't it return a boolean value?   Do you have tests specifically for each side and both sides of this expression being true?

ndushay (changes requested):
I have one question / concern.

app/services/cocina/from_fedora/descriptive/access.rb
justinlittman:
```
irb(main):001:0> x = 'foo'
=> "foo"
irb(main):002:0> y = nil
=> nil
irb(main):003:0> x || y
=> "foo"
irb(main):004:0> y || x
=> "foo"
```

And if I take out either side of `type || node[:type]`, then tests fail.

ndushay: okay.  who knows what the heck I'm remembering.

ndushay (approved):

PR #2105, opened 2021-02-06
Implement descriptive standard with authority.

### LINKED ISSUE #2088: Implement descriptive standard with authority
See "Description standard with authority" in https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/record_info_spec.rb

justinlittman:
closes #2088

## Why was this change made?
To bring joy to the world.

Unit, sdr-deploy

NA

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/admin_metadata.rb
+++ b/app/services/cocina/from_fedora/descriptive/admin_metadata.rb
@@ -80,16 +80,22 @@ module Cocina
           return unless description_standards
 
           description_standards.map do |description_standard|
-            if description_standard['authority']
-              {
-                code: description_standard['authority'],
-                uri: ValueURI.sniff(description_standard['valueURI'], notifier),
-                value: description_standard.text.presence,
-                source: { uri: description_standard['authorityURI'] }
-              }.compact
-            else
-              { code: description_standard.text }
-            end
+            source = {
+              uri: Authority.normalize_uri(description_standard['authorityURI']),
+              code: description_standard['authority']
+            }.compact
+            {
+              uri: ValueURI.sniff(description_standard['valueURI'], notifier),
+              source: source.presence
+            }.tap do |attrs|
+              if description_standard.text.present?
+                if description_standard.text == description_standard.text.downcase
+                  attrs[:code] = description_standard.text
+                else
+                  attrs[:value] = description_standard.text
+                end
+              end
+            end.compact
           end.presence
         end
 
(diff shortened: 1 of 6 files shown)


arcadiafalcone (approved):

justinlittman:
Before:
```
Status (n=100000):
  Success:   98200 (98.2%)
  Different: 1269 (1.269%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

After:
```
Status (n=100000):
  Success:   98204 (98.204%)
  Different: 1265 (1.265%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
  Missing:     531 (0.531%)
```

justinlittman:
Needs an infra dev review.

app/services/cocina/from_fedora/descriptive/admin_metadata.rb
```diff
@@ -80,16 +80,22 @@ def build_standard
...
+            source = {
+              uri: Authority.normalize_uri(description_standard['authorityURI']),
+              code: description_standard['authority']
+            }.compact
+            {
+              uri: ValueURI.sniff(description_standard['valueURI'], notifier),
+              source: source.presence
+            }.tap do |attrs|
+              if description_standard.text.present?
+                if description_standard.text == description_standard.text.downcase
```
mjgiarlo, line 92: Huh, the codes are all lowercase and values are always not-all-lowercase?

app/services/cocina/from_fedora/descriptive/admin_metadata.rb
```diff
@@ -80,16 +80,22 @@ def build_standard
...
-            end
+            source = {
+              uri: Authority.normalize_uri(description_standard['authorityURI']),
+              code: description_standard['authority']
+            }.compact
+            {
+              uri: ValueURI.sniff(description_standard['valueURI'], notifier),
+              source: source.presence
+            }.tap do |attrs|
+              if description_standard.text.present?
```
mjgiarlo, line 91: I have a slight preference to use a guard clause here to make the code more vertically oriented than horizontally (`next if description_standard.text.blank?`) but I don't feel strongly about it.

mjgiarlo (approved):
One comment, one question, neither blocking

app/services/cocina/from_fedora/descriptive/admin_metadata.rb
justinlittman: The data is messy, so that is the way to differentiate.

app/services/cocina/from_fedora/descriptive/admin_metadata.rb
justinlittman: I would agree with you in almost every case, but that seems odd to me in a `tap`.

app/services/cocina/from_fedora/descriptive/admin_metadata.rb
mjgiarlo: If it works, it works!

PR #2124, opened 2021-02-09
migrate from_fedora contrib spec to mapping name spec, with additional normalizations

ndushay:
## Why was this change made?

We want to migrate the from_fedora and to_fedora mapping specs to a land where all the mapping specs are above average

spec/services/cocina/from_fedora/descriptive/contributor_spec.rb  contents were moved to spec/services/cocina/mappings/mods/name_spec.rb, or were found to be duplicates.

mods normalization changed and we got a 0.006% improvement in our success rate!   Go me!

### this branch

```
Status (n=50000):
  Success:   49117 (98.234%)
  Different: 571 (1.142%)
  To Cocina error:     1 (0.002%)
  To Fedora error:     0 (0.0%)
  Missing:     311 (0.622%)
```

### main branch

```
Status (n=50000):
  Success:   49114 (98.228%)
  Different: 574 (1.148%)
  To Cocina error:     1 (0.002%)
  To Fedora error:     0 (0.0%)
  Missing:     311 (0.622%)
```


app/services/cocina/mods_normalizers/name_normalizer.rb
```diff
@@ -56,6 +62,38 @@ def normalize_dupes_for(base_node)
...
+
+      def normalize_type
+        ng_xml.root.xpath('//mods:name[(@type)]', mods: ModsNormalizer::MODS_NS).each do |name_node_w_type|
+          raw_type = name_node_w_type['type']
+          next if FromFedora::Descriptive::Contributor::ROLES.keys.include?(raw_type)
+
+          if FromFedora::Descriptive::Contributor::ROLES.keys.include?(raw_type.downcase)
+            name_node_w_type['type'] = raw_type.downcase
+          else
+            name_node_w_type.remove_attribute('type')
```
justinlittman, line 74: Not a big deal, but elsewhere we've been using `.delete()`.

app/services/cocina/mods_normalizers/name_normalizer.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../from_fedora/descriptive/contributor'
```
justinlittman, line 3: I'm surprised that this is required.

app/services/cocina/mods_normalizers/name_normalizer.rb
```diff
@@ -56,6 +62,38 @@ def normalize_dupes_for(base_node)
           dupe_name_nodes[1..].each(&:remove)
         end
       end
+
+      def normalize_type
+        ng_xml.root.xpath('//mods:name[(@type)]', mods: ModsNormalizer::MODS_NS).each do |name_node_w_type|
```
justinlittman, line 67: I don't think the () are needed in "(@type)".

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
```diff
@@ -1919,4 +1919,480 @@
...
+              }
+            ]
+          }
+        end
+
+        let(:warnings) { [Notification.new(msg: 'Name type incorrectly capitalized', context: { type: 'Personal' })] }
+      end
+    end
+
+    context 'when unrecognized' do
```
justinlittman, line 1965: When what is unrecognized?

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
```diff
@@ -1919,4 +1919,480 @@
...
+        let(:warnings) do
+          [
+            Notification.new(msg: 'Name/namePart type attribute set to ""'),
+            Notification.new(msg: 'name/role/roleTerm missing value')
+          ]
+        end
+      end
+    end
+
+    context 'with a role that has no URI and has xlink uris from MODS 3.3' do
```
justinlittman, line 2219: Why not use the shared example? Might require some changes to allow setting the mods header.

justinlittman (changes requested):
Mostly suggestions.

app/services/cocina/mods_normalizers/name_normalizer.rb
ndushay: right you are.

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
ndushay: i felt the path of least resistance was to just port the MODS 3.3 example as it was.  If we want to make it possible to use different MODS versions, we should maybe write a ticket for that;  I don't think we have more than a handful of these tests.

app/services/cocina/mods_normalizers/name_normalizer.rb
ndushay: well, I'll be.  it isn't required now.  It was when I was developing the code;  must've had something weird going on for a spell.

app/services/cocina/mods_normalizers/name_normalizer.rb
ndushay: ja - docs imply they are aliases for one another.  Leaving this unless it really bugs you.

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
ndushay: amended.

spec/services/cocina/mapping/descriptive/mods/name_spec.rb
ndushay: Also wasn't sure if we had to have a specific xlink namespace declaration on the outer MODS element.

ndushay:
nice - elegant.  It just wasn't clicking for me.

justinlittman (approved):

PR #2158, opened 2021-02-11
run origin_info_logic_specs that pass

ndushay:
## Why was this change made?

Arcadia provided some new specs to go with her revised implementation of originInfo <--> event mappings.   This brings those specs into the fold.   It's a small PR!

added specs only


justinlittman (commented):
I don't understand these specs. Can you make them more explicit? (No 2 letter abbreviations.)

ndushay:
i am re-using variables so I can test roundtrip mappings in both directions.  I can’t use “mods” and “cocina” because they are taken.  What would you suggest?

justinlittman (approved):

PR #2197, opened 2021-02-16
validation: better label for 'Missing (no descMetadata)'

ndushay:
## Why was this change made?

Because most of us can't remember what "Missing" means in the validation stats.

n/a


justinlittman:
Do you also want to change in`validate-cocina-roundtrip` and `validate-to-cocina`?

ndushay:
@justinlittman excellent idea;  done.

justinlittman (approved):

PR #2206, opened 2021-02-17
Express date ranges as EDTF range in Cocina

jcoyne:
## Why was this change made?

Fixes sul-dlss/happy-heron#541

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/event.rb
+++ b/app/services/cocina/to_fedora/descriptive/event.rb
@@ -258,6 +258,7 @@ module Cocina
         def date_tag(date, cocina_event_type, structured_val_attribs = {})
           value = date.value
           tag = date_type_value_for(date) ? :dateOther : TAG_NAME.fetch(cocina_event_type, :dateOther)
+          return dates_for_range(date, cocina_event_type) if edtf_range?(date)
 
           attributes = {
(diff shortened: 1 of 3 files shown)


app/services/cocina/to_fedora/descriptive/event.rb
```diff
@@ -258,6 +258,8 @@ def write_basic_date(date, cocina_event_type)
         def date_tag(date, cocina_event_type, structured_val_attribs = {})
           value = date.value
           tag = date_type_value_for(date) ? :dateOther : TAG_NAME.fetch(cocina_event_type, :dateOther)
+          encoding = (date.encoding&.code || structured_val_attribs[:encoding]&.code)
```
mjgiarlo, line 261: What is this variable doing? Is it used below?

mjgiarlo (approved):
have a question about a potentially unused variable, but otherwise good to go

app/services/cocina/to_fedora/descriptive/event.rb
jcoyne: That may be cruft. I used it once, and then refactored.

jcoyne: Fixed that.

mjgiarlo: :beers:

PR #2211, opened 2021-02-18
revert part of PR #2130 to get more origin_info specs exercising our code

ndushay:
## Why was this change made?

To help us do origin info mappings and normalizations that don't break stuff ... before we do the big update of fixing originInfo mappings that roundtrip cleanly.

it uncomments some specs and they pass.


justinlittman (approved):

ndushay:
rebased ...

PR #2252, opened 2021-02-19
Add mapping for geographic subject with authority

### LINKED ISSUE #2232: Geographic subject mapping error
For druid:cv661zm2305:

```
Original node:
<subject authority="marcgac">
  <geographic>Africa</geographic>
</subject>
Best guess roundtripped node:
<subject>
  <geographic>Africa</geographic>
</subject>
```

arcadiafalcone:
## Why was this change made?

Closes #2232

n/a

n/a

Implementation ticket #2253

### DIFF
--- a/spec/services/cocina/mapping/descriptive/mods/subject_geographic_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/mods/subject_geographic_spec.rb
@@ -250,6 +250,34 @@ RSpec.describe 'MODS subject geographic <--> cocina mappings' do
     end
   end
 
+  describe 'Geographic subject with authority on subject element' do
+    xit 'unimplemented spec: geographic subject authority' do
+      let(:druid) { 'cv661zm2305' }
+
+      let(:mods) do
+        <<~XML
+          <subject authority="marcgac">
+            <geographic>Africa</geographic>
+          </subject>
+        XML
+      end
+
+      let(:cocina) do
+        {
+          subject: [
+            {
+              value: 'Africa',
+              type: 'place',
+              source: {
+                code: 'marcgac'
+              }
+            }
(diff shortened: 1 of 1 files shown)


spec/services/cocina/mapping/descriptive/mods/subject_geographic_spec.rb
```diff
@@ -250,6 +250,34 @@
...
   end
 
+  describe 'Geographic subject with authority on subject element' do
+    xit 'unimplemented spec: geographic subject authority' do
+      let(:druid) { 'cv661zm2305' }
+
+      let(:mods) do
+        <<~XML
+          <subject authority="marcgac">
+            <geographic>Africa</geographic>
```
justinlittman, line 260: We have some special handling around "marcgac" authority that keeps it on child instead of `<subject>`. Is that correct, or is it like every other authority?

arcadiafalcone: It should behave like other authorities. If `marcgac` is the authority there will only be `geographic` subelements and all of them will be from `marcgac`.

ndushay:
Roundtrip stats pls.

ndushay (approved):
Because you and Arcadia understand what you are talking about; and because this is adding an xit spec ... I'll let you merge it when you and Arcadia are in agreement.

spec/services/cocina/mapping/descriptive/mods/subject_geographic_spec.rb
arcadiafalcone: Roundtrip should be same as source.

justinlittman (approved):

PR #2393, opened 2021-03-01
Spec event type

ndushay:
## Why was this change made?

I fooched orig-info-refactor just now and this will fix it


ndushay:
fixed my git gaff another way.

PR #2395, opened 2021-03-01
Update publisher specs

arcadiafalcone:
## Why was this change made?
Mappings

n/a

n/a


ndushay (approved):

ndushay:
I will address failing tests in a follow on PR

PR #2402, opened 2021-03-02
[SUPERSEDED BUT DO NOT DELETE YET] Fix bug with rights determinations when metadata has multiple access nodes

mjgiarlo:
**SUPERSEDED** by #2716 

Connects to sul-dlss/argo#2382

## Why was this change made?

Also:
* Add the ability for `Cocina::FromFedora::DroStructural` to derive access information from the item rights when generating structural metadata.
  * Working implementation of file-specific rights that focuses only on world-readable/-downloadable files for now
* Prefer using `dor-rights-auth` to determine rights over duplicative (and wrong, see above) XPath checks

CI, to be tested in stage

None

### DIFF
--- a/app/services/cocina/from_fedora/access.rb
+++ b/app/services/cocina/from_fedora/access.rb
@@ -24,7 +24,7 @@ module Cocina
           copyright: copyright,
           useAndReproductionStatement: use_statement
         }.compact.tap do |h|
-          h[:controlledDigitalLending] = true if cdl?
+          h[:controlledDigitalLending] = true if controlled_digital_lending?
         end
       end
 
@@ -32,8 +32,10 @@ module Cocina
 
       attr_reader :rights_metadata_ds
 
+      delegate :controlled_digital_lending?, :dark?, to: :rights_object
+
       def rights_object
-        rights_metadata_ds.dra_object.obj_lvl
+        rights_metadata_ds.dra_object
       end
 
       def copyright
@@ -46,13 +48,13 @@ module Cocina
 
       # @return [Bool] true unless the rule="no-download" has been set or if the access is citation-only or dark
       def download?
-        return false if %w[citation-only dark].include? access_rights
+        return false if %w[citation-only dark].include?(access_rights)
 
-        !rights_object.world.rule && !rights_object.group.fetch(:stanford).rule
+        !rights_object.obj_lvl.world.rule && !rights_object.obj_lvl.group.fetch(:stanford).rule
       end
 
       def location
-        @location ||= rights_object.location.keys.first
+        @location ||= rights_object.obj_lvl.location.keys.first
       end
 
       # Map values from dor-services
@@ -60,9 +62,9 @@ module Cocina
       # to https://github.com/sul-dlss/cocina-models/blob/main/docs/maps/DRO.json#L102
       def access_rights
         @access_rights ||=
-          if world?
+          if rights_object.obj_lvl.world.value
             'world'
-          elsif stanford?
+          elsif rights_object.obj_lvl.group.fetch(:stanford).value
             'stanford'
           elsif dark?
             'dark'
@@ -71,26 +73,6 @@ module Cocina
           else
             'citation-only'
           end
-      end
-
-      def rights_xml
-        @rights_xml ||= rights_metadata_ds.ng_xml
-      end
-
-      def stanford?
-        rights_xml.search('//rightsMetadata/access[@type=\'read\']/machine/group').length == 1
-      end
-
-      def world?
-        rights_xml.search('//rightsMetadata/access[@type=\'read\']/machine/world').length == 1
-      end
-
-      def dark?
-        rights_xml.search('//rightsMetadata/access[@type=\'discover\']/machine/none').length == 1
-      end
-
-      def cdl?
-        rights_object.controlled_digital_lending
       end
     end
   end
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -46,8 +46,16 @@ module Cocina
 
       attr_reader :item, :type
 
+      delegate :stanford_only_downloadable_file?, :stanford_only_unrestricted_file?,
+               :world_downloadable_file?, :world_unrestricted_file?, to: :rights_object
+
+
       def project_phoenix?
         AdministrativeTags.for(pid: item.id).include?('Google Book : GBS VIEW_FULL')
+      end
+
+      def rights_object
+        item.rightsMetadata.dra_object
       end
 
       def build_has_member_orders
@@ -81,6 +89,76 @@ module Cocina
         content_metadata_ds.ng_xml.xpath('//resource/externalFile').map do |resource_node|
           "#{resource_node['resourceId']}/#{resource_node['fileId']}"
         end
+      end
+
+      def digests(node)
+        [].tap do |digests|
+          # The old google books use upcased versions. See https://argo.stanford.edu/view/druid:dd116zh0343
+          # Web archive crawls use SHA1
+          sha1 = node.xpath('checksum[@type="sha1" or @type="SHA1" or @type="SHA-1"]').text.presence
+          digests << { type: 'sha1', digest: sha1 } if sha1
+          md5 = node.xpath('checksum[@type="md5" or @type="MD5"]').text.presence
+          digests << { type: 'md5', digest: md5 } if md5
+        end
+      end
+
+      def build_files(file_nodes, version:, parent_id:)
+        file_nodes.map do |node|
+          height = node.xpath('imageData/@height').text.presence&.to_i
+          width = node.xpath('imageData/@width').text.presence&.to_i
+          use = node.xpath('@role').text.presence
+          {
+            externalIdentifier: "#{parent_id}/#{node['id']}",
+            type: Cocina::Models::Vocab.file,
+            label: node['id'],
+            filename: node['id'],
+            size: node['size'].to_i,
+            version: version,
+            hasMessageDigests: digests(node),
+            access: access(node),
+            administrative: {
+              publish: node['publish'] == 'yes',
+              sdrPreserve: node['preserve'] == 'yes',
+              shelve: node['shelve'] == 'yes'
+            }
+          }.tap do |attrs|
+            # Files from Goobi and Hydrus don't have mimetype until they hit exif-collect in the assemblyWF
+            attrs[:hasMimeType] = node['mimetype'] if node['mimetype']
+            attrs[:presentation] = { height: height, width: width } if height && width
+            attrs[:use] = use if use
+          end
+        end
+      end
+
+      def access(node)
+        # file_specific_rights = file_rights_for(node['id'])
+        # item_rights_defaults.merge(file_specific_rights)
+        if (file_specific_rights = file_rights_for(node['id']))
+          byebug
+          file_specific_rights
+        else
+          item_rights_defaults
+        end
+      end
+
+      def file_rights_for(file_name)
+        {}.tap do |file_rights|
+          if world_unrestricted_file?(file_name)
+            file_rights[:access] = 'world'
+          elsif stanford_only_unrestricted_file?(file_name)
+            file_rights[:access] = 'stanford'
+          end
+
+          if world_downloadable_file?(file_name)
+            file_rights[:download] = 'world'
+          elsif stanford_only_downloadable_file?(file_name)
+            file_rights[:download] = 'stanford'
+          end
+        end.presence # drop?
+      end
+
+      def item_rights_defaults
+        Access.props(item)
       end
     end
   end
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -88,7 +88,7 @@ module Cocina
         add_dro_tags(pid, obj)
 
         apply_default_access(item)
-        Cocina::ToFedora::DROAccess.apply(item, obj.access) if obj.access
+        Cocina::ToFedora::DROAccess.apply(item, obj) if obj.access
 
         item.contentMetadata.content = Cocina::ToFedora::ContentMetadataGenerator.generate(druid: pid, object: obj)
         Cocina::ToFedora::Identity.apply(item, label: obj.label, agreement_id: obj.structural&.hasAgreement)
@@ -109,7 +109,7 @@ module Cocina
         add_description(item, obj)
         add_collection_tags(pid, obj)
         apply_default_access(item)
-        Cocina::ToFedora::Access.apply(item, obj.access) if obj.access
+        Cocina::ToFedora::Access.apply(item, obj) if obj.access
         Cocina::ToFedora::Identity.apply(item, label: obj.label)
       end
     end
--- a/app/services/cocina/to_fedora/access.rb
+++ b/app/services/cocina/to_fedora/access.rb
@@ -19,12 +19,7 @@ module Cocina
       end
 
       def apply
-        # See https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb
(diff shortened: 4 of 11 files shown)


mjgiarlo:
@jcoyne I'd welcome some pairing on this tomorrow to get your input on how to take this forward. (Not only getting the failing tests passing but determining whether the changes are 1) sufficiently well tested otherwise, and 2) likely to cause unintended side effects.)

app/services/cocina/from_fedora/access.rb
```diff
@@ -71,15 +71,15 @@ def rights_xml
...
 
       def world?
-        rights_xml.search('//rightsMetadata/access[@type=\'read\']/machine/world').length == 1
+        rights_xml.search('//rightsMetadata/access[@type=\'read\'][not(file)]/machine/world').length == 1
       end
 
       def dark?
-        rights_xml.search('//rightsMetadata/access[@type=\'discover\']/machine/none').length == 1
+        rights_xml.search('//rightsMetadata/access[@type=\'discover\'][not(file)]/machine/none').length == 1
       end
```
mjgiarlo, line 83: This corrects a bug when making rights determinations about objects with rights XML that includes file-specific rights declarations, such as the following:

```xml
  <access type="read">
    <machine>
      <world rule="no-download"/>
    </machine>
  </access>
  <access type="read">
    <file>gs491bt1345_sample_md.pdf</file>
    <machine>
      <world/>
    </machine>
  </access>
```

Before this change, such objects would be marked as citation-only because the matching XPath returns **2** hits instead of **1**. Exclude file-specific rights subelements from the XPaths to fix the bug.

.rubocop.yml
```diff
@@ -31,19 +31,20 @@ Metrics/BlockLength:
...
+    - 'app/services/cocina/mods_normalizers/subject_normalizer.rb'
     - 'app/services/cocina/to_fedora/descriptive/contributor_writer.rb'
+    - 'app/services/cocina/to_fedora/descriptive/event.rb'
     - 'app/services/cocina/to_fedora/descriptive/form.rb'
-    - 'app/services/cocina/mods_normalizers/subject_normalizer.rb'
-    - 'app/services/cocina/mods_normalizers/origin_info_normalizer.rb'
-    - 'app/services/cocina/from_fedora/descriptive/access.rb'
     - 'app/services/cocina/to_fedora/descriptive/related_resource.rb'
+    - 'app/services/cocina/to_fedora/descriptive/title.rb'
+    - 'app/services/mods_equivalent_service.rb'
```
mjgiarlo, line 47: Alpha sort.

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -137,11 +137,68 @@ def build_files(file_nodes, version:, parent_id:)
 
       def access(node)
         if node['publish'] == 'yes'
-          { access: 'world', download: 'world' }
+          derive_access_from_rights_metadata(node)
         else
           { access: 'dark', download: 'none' }
         end
       end
```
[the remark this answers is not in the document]
andrewjbtw: I want to say yes, but this is tricky because what ultimately determines access is what's published on the Purl XML. The access system won't know anything about what Cocina says on a file-by-file basis as long as what we continue to publish is the rightsMetadata block. If we publish anything other than that, then they will need to update their code accordingly.

That said, I think this is correct:

>If the item's rights metadata has a file-specific section for the file in question, use the rights declared in that section. Else, use the overall object's rights.

But when `publish` is `no` I think using the overall object's rights is also consistent with current behavior, rather than setting that file to dark. We have files on current purls that are `publish="no"` that are downloadable (if someone could get the Stacks URL). ETDs do this, for example.

app/services/cocina/from_fedora/access.rb
```diff
@@ -25,37 +25,39 @@ def props
...
 
       private
 
       attr_reader :item
 
+      delegate :controlled_digital_lending?, :dark?, to: :rights_object
+
       def rights_object
-        item.rightsMetadata.dra_object.obj_lvl
+        item.rightsMetadata.dra_object
```
mjgiarlo, line 39: Part of the changes in this file are inspired by wanting to replace custom XPaths, some of which were mistakenly crafted, with using what we already have in `dor-rights-auth` (which uses the correct XPaths). So not only more correctness, but more reuse.

app/services/cocina/from_fedora/access.rb
```diff
@@ -65,26 +67,6 @@ def access_rights
...
-        rights_xml.search('//rightsMetadata/access[@type=\'read\']/machine/group').length == 1
-      end
-
-      def world?
-        rights_xml.search('//rightsMetadata/access[@type=\'read\']/machine/world').length == 1
-      end
-
-      def dark?
-        rights_xml.search('//rightsMetadata/access[@type=\'discover\']/machine/none').length == 1
-      end
```
mjgiarlo, line 83: These were buggy XPath statements in that they failed to acknowledge that we can in fact have **multiple** `access` nodes in `rightsMetadata`, which is part of the bug in sul-dlss/argo#2382, which manifests in objects that have file-specific rights metadata (in which case the equality checks above return `false` because the `length` is `2`! So world-readable objects, e.g., were showing as citation-only. The better XPath statements are already in `dor-rights-auth`, so use those instead.

app/services/cocina/from_fedora/access.rb
```diff
@@ -65,26 +67,6 @@ def access_rights
...
-
-      def cdl?
-        rights_object.controlled_digital_lending
-      end
```
mjgiarlo, line 87: Handling this through delegation now.

app/services/cocina/from_fedora/descriptive/event.rb
```diff
@@ -70,7 +70,6 @@ def merge_parallel_grouped_event(grouped_event, base_event_group, base_event_gro
           end
         end
 
-        # rubocop:disable Metrics/AbcSize
```
mjgiarlo, line 73: Sorry, DSA's rubocop config really threw a monkeywrench into this PR, which is partly why I regenerated the TODOs above.

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -46,6 +46,13 @@ def props
 
       attr_reader :item, :type
 
+      delegate :stanford_only_downloadable_file?, :stanford_only_unrestricted_file?,
+               :world_downloadable_file?, :world_unrestricted_file?, to: :rights_object
+
+      def rights_object
+        item.rightsMetadata.dra_object
+      end
```
mjgiarlo, line 54: Again, leave the rightsMD heavy lifting to `dor-rights-auth`.

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -137,11 +144,39 @@ def build_files(file_nodes, version:, parent_id:)
 
       def access(node)
         if node['publish'] == 'yes'
-          { access: 'world', download: 'world' }
+          derive_access_from_rights_metadata(node)
         else
           { access: 'dark', download: 'none' }
         end
       end
```
mjgiarlo, line 151: This is the second big change in this PR: deriving access to files from the rights metadata when generating structural metadata. Andrew has confirmed that this is how we ought to handle things for now.

app/services/cocina/object_creator.rb
```diff
@@ -81,7 +81,7 @@ def create_dro(obj)
         add_dro_tags(pid, obj)
 
         apply_default_access(item)
-        Cocina::ToFedora::DROAccess.apply(item, obj.access) if obj.access
+        Cocina::ToFedora::DROAccess.apply(item, obj) if obj.access
```
mjgiarlo, line 84: The `DROAccess` class now needs the entire Cocina object to generate access since both object-level access and file-level access (down in the structural MD) must be considered for fidelity on the ToFedora side, else the file-specific rights are omitted from the generated rightsMD.

app/services/cocina/object_creator.rb
```diff
@@ -100,7 +100,7 @@ def create_collection(obj)
         add_description(item, obj)
         add_collection_tags(pid, obj)
         apply_default_access(item)
-        Cocina::ToFedora::Access.apply(item, obj.access) if obj.access
+        Cocina::ToFedora::Access.apply(item, obj) if obj.access
```
mjgiarlo, line 103: Same as above re: `DROAccess`.

app/services/cocina/to_fedora/dro_access.rb
```diff
@@ -21,12 +21,6 @@ def create_embargo(embargo)
                               access: embargo.access,
                               use_and_reproduction_statement: embargo.useAndReproductionStatement)
       end
-
-      def rights_type
-        return 'cdl-stanford-nd' if access.controlledDigitalLending
-
-        super
-      end
```
mjgiarlo, line 29: This moved to the new `RightsMetadataGenerator` class.

app/services/cocina/to_fedora/access.rb
```diff
@@ -8,38 +8,25 @@ class Access
...
-        when 'location-based'
-          "loc:#{access.readLocation}"
-        when 'citation-only'
-          'none'
-        when 'dark'
-          'dark'
-        else
-          access.respond_to?(:download) && access.download == 'none' ? "#{access.access}-nd" : access.access
-        end
-      end
```
mjgiarlo, line 42: Moved to the new `RightsMetadataGenerator` class.

app/services/cocina/to_fedora/rights_metadata_generator.rb
```diff
@@ -0,0 +1,160 @@
...
+        new(item: item, object: object).generate
+      end
+
+      def initialize(item:, object:)
+        @item = item
+        @object = object
+      end
+
+      # Adapted copypasta from https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb
+      def generate
```
mjgiarlo, line 19: This method is essentially inlined from `dor-services` with one addition: file-specific rights handling. Somehow, we never had this in `dor-services`. So instead of mucking about down in that :rat: nest, make it work better up here in DSA.

app/services/cocina/to_fedora/rights_metadata_generator.rb
```diff
@@ -0,0 +1,160 @@
...
+          elsif rights_type.start_with?('cdl')
+            cdl_node = Nokogiri::XML::Node.new('cdl', rights_xml)
+            group_node = Nokogiri::XML::Node.new('group', cdl_node)
+            group_node.content = 'stanford'
+            group_node.set_attribute('rule', 'no-download')
+            cdl_node.add_child(group_node)
+            machine_node.add_child(cdl_node)
+          else # we know it is none or dark by the argument filter (first line)
+            machine_node.add_child Nokogiri::XML::Node.new('none', rights_xml)
+          end
```
mjgiarlo, line 56: None of this changed. This is what was inlined without modifications.

app/services/cocina/to_fedora/rights_metadata_generator.rb
```diff
@@ -0,0 +1,160 @@
...
+
+          # Append file-specific rights to the first `access@type='read'` node
+          # if the object is a DRO; collections and APOs don't directly contain files
+          if object.dro? && index.zero?
+            file_specific_rights = generate_file_specific_rights
+            node.add_next_sibling(file_specific_rights) unless file_specific_rights.empty?
+          end
+        end
```
mjgiarlo, line 64: This is new, and it's how we maintain file-specific rights fidelity.

app/services/cocina/to_fedora/rights_metadata_generator.rb
```diff
@@ -0,0 +1,160 @@
...
+
+              rights_to_add = rights_to_add_for(file_specific_rights)
+
+              # Patching this in to handle objects like those in
+              # sul-dlss/argo#2382 without committing to a major overhaul
+              if rights_to_add.values.uniq != ['world']
+                Honeybadger.notify('Found file-specific rights that cannot be handled yet',
+                                   context: { rights: rights_to_add, druid: item.pid })
+                next
+              end
```
mjgiarlo, line 120: Only handling the one case now, so as 1) not to bite off more than I can chew, and to 2) remediate the specific sul-dlss/argo#2382 case in the short-term. I added the HB experiment to gather data over time about how this method needs to evolve to get even gooderer.

spec/requests/create_collection_spec.rb
```diff
@@ -112,7 +112,7 @@
         post '/v1/objects',
              params: data,
              headers: { 'Authorization' => "Bearer #{jwt}", 'Content-Type' => 'application/json' }
-        expect(response.body).to eq expected.to_json
+        expect(response.body).to equal_cocina_model(expected)
```
mjgiarlo, line 115: This is a new matcher I added to make it easier to compare two cocina structures. Comparing two giant JSON strings is a PITA. Instead, RSpec will compare both sides as JSON strings and if they do not match, will coerce both sides to hashes which lets `super_diff` makes our lives much easier when trying to make sense of failing tests.

spec/services/cocina/from_fedora/access_spec.rb
```diff
@@ -76,6 +76,12 @@
               <world rule="no-download"/>
             </machine>
           </access>
+          <access type="read">
+            <file>foo_bar.pdf</file>
+            <machine>
+              <world/>
+            </machine>
+          </access>
```
mjgiarlo, line 84: Added to make sure we don't get a regression w/r/t rightsMD with multiple access@type=read nodes.

app/services/cocina/to_fedora/rights_metadata_generator.rb
```diff
@@ -0,0 +1,164 @@
...
+          # Append file-specific rights to the first `access@type='read'` node
+          # if the object is a DRO; collections and APOs don't directly contain files
+          if object.dro? && index.zero?
+            # Remove file-specific rights from XML (because we will be
+            # regenerating them from Cocina)
+            rights_xml.search("//rightsMetadata/access[@type='read' and file]").map(&:remove)
+
+            file_specific_rights = generate_file_specific_rights
+            node.add_next_sibling(file_specific_rights) unless file_specific_rights.empty?
+          end
```
mjgiarlo, line 67: This is the new code I added (not copypasta).

app/services/cocina/to_fedora/rights_metadata_generator.rb
```diff
@@ -0,0 +1,164 @@
...
+        when 'citation-only'
+          'none'
+        when 'dark'
+          'dark'
+        else
+          return "#{access.access}-nd" if access.respond_to?(:download) && access.download == 'none'
+
+          access.access
+        end
+      end
```
mjgiarlo, line 99: This code was just moved.

spec/services/cocina/from_fedora/dro_structural_spec.rb
```diff
@@ -187,7 +187,7 @@
       expect(file1[:hasMessageDigests].first[:type]).to eq 'sha1'
       expect(file1[:administrative][:shelve]).to eq true
       expect(file1[:administrative][:sdrPreserve]).to eq false
-      expect(file1[:access][:access]).to eq('world')
+      expect(file1[:access][:access]).to eq('dark')
```
mjgiarlo, line 190: Some existing specs needed changes since we now derive rights from the item.

justinlittman (approved):
Thanks for the annotations.

mjgiarlo:
# Changes to make

1. When generating cocina structural metadata, do not generate file access from publish setting
2. Add publish attribute to file_administrative cocina model (because publish and rights do not have a 1-to-1 mapping)
3.

ndushay:
Noting here that #510, about indexing cocina access, is dependent on
-  understanding the representation of file level rights in cocina access

And I think we would be well served to come up with a way to test:  "does THIS rightsMetadata.xml end up getting mapped to cocina and indexed from cocina with the same solr values it got when it was being indexed from rightsMetadata.xml"

ndushay:
feature tests for, say:  "when the user selects xx in the form, then the bulk action gets the following args" -- not for the background job part.

PR #2427, opened 2021-03-05
Support partial object update.

### LINKED ISSUE #2410: Spike: Partial object update in api - allow specifying keys to update
When provided, only the values referred to by the keys are updated.

I.e. just update the identifiers

Purpose/goal: avoid introducing roundtripping errors for unchanged parts.

justinlittman:
closes #2410

## Why was this change made?
Reduce the risk of unrelated mapping errors when performing an object update by only update Fedora from the parts of the cocina object that have changed.

In particular, it avoids changes being accidentally made to descriptive metadata.

Not yet.

NA

### DIFF
--- a/app/services/cocina/object_updater.rb
+++ b/app/services/cocina/object_updater.rb
@@ -17,6 +17,7 @@ module Cocina
       @params = params
       @obj = obj
       @item = item
+      @orig_obj = Mapper.build(item)
     end
 
     def run(event_factory:)
@@ -34,7 +35,7 @@ module Cocina
         raise "unsupported type #{obj.type}"
       end
 
-      update_descriptive if update_descriptive?
+      update_descriptive if has_changed?(:description) && update_descriptive?
 
       item.save!
 
@@ -51,42 +52,65 @@ module Cocina
 
     private
 
-    attr_reader :obj, :item, :params
+    attr_reader :obj, :item, :params, :orig_obj
 
-    def update_apo
-      item.admin_policy_object_id = obj.administrative.hasAdminPolicy
-      # item.source_id = obj.identification.sourceId
-      item.label = truncate_label(obj.label)
-
-      Cocina::ToFedora::ApoRights.write(item.administrativeMetadata, obj.administrative)
-      Cocina::ToFedora::Roles.write(item, Array(obj.administrative.roles))
-      Cocina::ToFedora::Identity.apply(obj, item, object_type: 'adminPolicy')
+    def has_changed?(key)
+      # Update only if changed.
+      obj.public_send(key) != orig_obj.public_send(key)
     end
 
-    def update_collection
-      item.admin_policy_object_id = obj.administrative.hasAdminPolicy
-      item.catkey = catkey_for(obj)
-      item.label = truncate_label(obj.label)
+    # rubocop:disable Style/GuardClause
+    def update_apo
+      # item.source_id = obj.identification.sourceId
+      if has_changed?(:label)
+        Cocina::ToFedora::Identity.apply(item, label: obj.label, object_type: 'adminPolicy')
+        item.label = truncate_label(obj.label)
+      end
 
-      Cocina::ToFedora::Access.apply(item, obj.access)
-      Cocina::ToFedora::Identity.apply(obj, item, object_type: 'collection')
+      if has_changed?(:administrative)
+        item.admin_policy_object_id = obj.administrative.hasAdminPolicy
+        Cocina::ToFedora::ApoRights.write(item.administrativeMetadata, obj.administrative)
+        Cocina::ToFedora::Roles.write(item, Array(obj.administrative.roles))
+      end
+    end
+    # rubocop:enable Style/GuardClause
+
+    def update_collection
+      if has_changed?(:label)
+        item.label = truncate_label(obj.label)
+        Cocina::ToFedora::Identity.apply(item, label: obj.label, object_type: 'collection')
+      end
+      item.admin_policy_object_id = obj.administrative.hasAdminPolicy if has_changed?(:administrative)
+
+      Cocina::ToFedora::Access.apply(item, obj.access) if has_changed?(:access)
     end
 
     # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/PerceivedComplexity
     def update_dro
-      item.admin_policy_object_id = obj.administrative.hasAdminPolicy
-      item.source_id = obj.identification.sourceId
-      item.collection_ids = Array.wrap(obj.structural&.isMemberOf).compact
-      item.catkey = catkey_for(obj)
-      item.label = truncate_label(obj.label)
+      item.admin_policy_object_id = obj.administrative.hasAdminPolicy if has_changed?(:administrative)
+      item.collection_ids = Array.wrap(obj.structural&.isMemberOf).compact if has_changed?(:structural)
+      item.label = truncate_label(obj.label) if has_changed?(:label)
 
-      add_tags(item.id, obj)
+      if has_changed?(:identification)
+        item.source_id = obj.identification.sourceId
+        item.catkey = catkey_for(obj)
+      end
 
-      Cocina::ToFedora::DROAccess.apply(item, obj.access)
-      update_content_metadata(item, obj)
-      Cocina::ToFedora::Identity.apply(obj, item, object_type: 'item', agreement_id: obj.structural&.hasAgreement)
+      if has_changed?(:label) || has_changed?(:structural)
+        label = has_changed?(:label) ? obj.label : item.label
+        agreement_id = has_changed?(:structural) ? obj.structural&.hasAgreement : nil
+        Cocina::ToFedora::Identity.apply(item, label: label, object_type: 'item', agreement_id: agreement_id)
+      end
+      Cocina::ToFedora::DROAccess.apply(item, obj.access) if has_changed?(:access)
+      update_content_metadata(item, obj) if has_changed?(:structural) || has_changed?(:type)
+
+      add_tags(item.pid, obj)
     end
     # rubocop:enable Metrics/AbcSize
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
 
     def update_content_metadata(item, obj)
       # We don't want to overwrite contentMetadata unless they provided structural.contains
@@ -108,7 +132,7 @@ module Cocina
 
       raise ValidationError, "Identifier on the query and in the body don't match" if item.pid != obj.externalIdentifier
 
-      validator = Cocina::ApoExistenceValidator.new(obj)
+      validator = ApoExistenceValidator.new(obj)
       raise ValidationError, validator.error unless validator.valid?
 
       raise NotImplemented, 'Updating descriptive metadata not supported' if !update_descriptive? && client_attempted_metadata_update?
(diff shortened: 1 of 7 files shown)


app/services/cocina/to_fedora/access.rb
```diff
@@ -21,6 +21,8 @@ def initialize(item, access)
       def apply
         # See https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb
         Dor::RightsMetadataDS.upd_rights_xml_for_rights_type(item.rightsMetadata.ng_xml, rights_type)
+        # This invalidates the dra_object, which is necessary if re-mapping.
+        item.rightsMetadata.content = item.rightsMetadata.ng_xml.to_s
```
mjgiarlo, line 25: This change may conflict with #2402 but it shouldn't be difficult to work it back in.

justinlittman: Yeah, looking at #2402, I think there will be several but no problem.

jcoyne:
> Reduce the risk of unrelated mapping errors when performing an object update.

how?

spec/services/cocina/object_updater_spec.rb
```diff
@@ -0,0 +1,482 @@
...
+          update
+          expect(desc_metadata).to have_received(:content=)
+          expect(desc_metadata).to have_received(:content_will_change!)
+          expect(Cocina::ToFedora::Descriptive).to have_received(:transform)
+        end
+      end
+
+      context 'when description has not changed' do
+        it 'does not update descriptive' do
+          update
```
jcoyne, line 101: is there some expectation here?

spec/services/cocina/object_updater_spec.rb
```diff
@@ -0,0 +1,482 @@
...
+          expect(item).to have_received(:admin_policy_object_id=).with('druid:dd999df4567')
+          expect(Cocina::ToFedora::ApoRights).to have_received(:write)
+          expect(Cocina::ToFedora::Roles).to have_received(:write)
+        end
+      end
+
+      context 'when administrative has not changed' do
+        it 'does not update administrative' do
+          update
```
jcoyne, line 130: Is there an expectation here?

spec/services/cocina/object_updater_spec.rb
```diff
@@ -0,0 +1,482 @@
...
+        it 'updates label' do
+          update
+          expect(item).to have_received(:label=).with('new label')
+          expect(Cocina::ToFedora::Identity).to have_received(:apply)
+        end
+      end
+
+      context 'when label has not changed' do
+        it 'does not update label' do
+          update
```
jcoyne, line 177: Is there an expectation here?

app/services/cocina/object_updater.rb
```diff
@@ -7,108 +7,132 @@ class ObjectUpdater
...
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
 
-    def update_content_metadata(item, obj)
+    def update_content_metadata(item, cocina_object)
       # We don't want to overwrite contentMetadata unless they provided structural.contains
-      if obj.structural&.contains
-        item.contentMetadata.content = Cocina::ToFedora::ContentMetadataGenerator.generate(druid: item.pid, object: obj)
+      if cocina_object.structural&.contains
+        item.contentMetadata.content = Cocina::ToFedora::ContentMetadataGenerator.generate(druid: item.pid, object: cocina_object)
```
jcoyne, line 118: what about `hasMemberOrders`?

app/services/cocina/to_fedora/identity.rb
```diff
@@ -6,11 +6,11 @@ module ToFedora
     # Fedora 3 data model identityMetadata
     class Identity
       # @param [String] agreement_id (nil) the identifier for the agreement. Note that only items have an agreement.
-      def self.apply(obj, item, object_type:, agreement_id: nil)
+      def self.apply(item, label:, object_type:, agreement_id: nil)
```
jcoyne, line 9: I'm curious as to why youl've extracted `label` since this already has a dependency on the object where the label comes from.

app/services/cocina/object_updater.rb
```diff
@@ -7,108 +7,132 @@ class ObjectUpdater
...
+      event_factory.create(druid: item.pid, event_type: 'update', data: { success: false, error: e.message, request: cocina_object.to_h })
       raise
     end
 
     private
 
-    attr_reader :obj, :item, :params
+    attr_reader :cocina_object, :item, :params, :orig_cocina_object
 
+    def update?(key)
```
jcoyne, line 57: can we call this `has_changed?(key)`

app/services/cocina/object_updater.rb
```diff
@@ -7,108 +7,132 @@ class ObjectUpdater
...
+      # Update only if changed.
+      cocina_object.public_send(key) != orig_cocina_object.public_send(key)
```
jcoyne, line 59: Is this basically doing the same as we're doing at the rubydora level? https://github.com/samvera/rubydora/blob/2cfedc707daa7203bce8c34a1a86ed72b6507e06/lib/rubydora/datastream.rb#L180

jcoyne: And https://github.com/samvera/rubydora/blob/2cfedc707daa7203bce8c34a1a86ed72b6507e06/lib/rubydora/datastream.rb#L22 for the label (i.e. `dsLabel`)

app/services/cocina/object_updater.rb
justinlittman: Yup.

app/services/cocina/object_updater.rb
justinlittman: No idea. I'm not changing this code.

app/services/cocina/to_fedora/identity.rb
justinlittman: It avoids passing the cocina object. Passing the cocina object might allow a future developer to accidentally bypass the has_changed? checking.

app/services/cocina/object_updater.rb
justinlittman: Kinda looks like it.

spec/services/cocina/object_updater_spec.rb
justinlittman: It's implicit. If an attempt was made to update descriptive, then an error would have been thrown, something like:

```
 Failure/Error: item.descMetadata.content = Cocina::ToFedora::Descriptive.transform(obj.description, item.pid).to_xml
       #<Double (anonymous)> received unexpected message :content= with ("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<mods xmlns=\"http://www.loc.gov/mods/v3\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:rdf=\"http://www.w3.org/1999/02/22-rdf-syntax-ns#\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" version=\"3.7\" xsi:schemaLocation=\"http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-7.xsd\">\n  <titleInfo>\n    <title>new title</title>\n  </titleInfo>\n</mods>\n")
```

app/services/cocina/object_updater.rb
jcoyne: what if the user provides `cocina_object.structural.hasMemberOrders` and not cocina_object.structural.contains`? Then their change would never be saved.

app/services/cocina/object_updater.rb
```diff
@@ -51,42 +52,65 @@ def run(event_factory:)
...
+      item.collection_ids = Array.wrap(obj.structural&.isMemberOf).compact if has_changed?(:structural)
+      item.label = truncate_label(obj.label) if has_changed?(:label)
+
+      if has_changed?(:identification)
+        item.source_id = obj.identification.sourceId
+        item.catkey = catkey_for(obj)
+      end
 
-      add_tags(item.id, obj)
+      if has_changed?(:label) || has_changed?(:structural)
```
jcoyne, line 101: Does this handle if the agreement_id is changed?

jcoyne (changes requested):

app/services/cocina/object_updater.rb
justinlittman: I believe so. Agreement is `obj.structural.hasAgreement`, so it would pass this if statement.

app/services/cocina/object_updater.rb
justinlittman: I'm not following.

spec/services/cocina/object_updater_spec.rb
jcoyne: It might be more clear to say `expect(foo).not_to have_received(:bar)`

app/services/cocina/object_updater.rb
jcoyne: It is possilbe for a cocina object to have nothing in the `contains` property, but to have `hasMemberOrders`.  If you are checking only for the presence of `contains`, then my changes to `hasMemberOrders` will not be preserved.

jcoyne:
I'm sorry to be really critical of this PR, but this adds a whole bunch of complexity and I'm not exactly sure what problem it solves.  I feel like these conditionals may end up introducing several hard to diagnose bugs.  I'd like some other team members to look at this very carefully and see what they think.

app/services/cocina/object_updater.rb
justinlittman: The comment in the line above says ` # We don't want to overwrite contentMetadata unless they provided structural.contains`, which would suggest that this behavior is deliberate. Maybe?

However, this PR isn't changing this code so if you think it is incorrect, I'd suggest another ticket.

spec/services/cocina/object_updater_spec.rb
justinlittman: Added `not_to`s

justinlittman:
I agree that the complexity is undesirable. However, until our mappings are rock solid, it reduces the risk of changes to one part of an item causing changes in an unrelated part of the item. For example, if a change is made to the access only the rightsMetadata datastream should change, the descMetadata datastream should not change. At worst, such changes might include errors; at best, they would be disconcerting to a user.

jcoyne:
I'm not sure this PR does that though. The cocina schemas don't map 1-1 with datastreams.   Many schemas are an amalgamation of several datastreams. For example changing the "type" property affects the "contentMetadata" datastream.  Changing the collection something belongs to is in the structural metadata, but it's written into the RELS-EXT datastream. The adminPolicy is in the administrative schema, but that also is in the RELS-EXT datastream.

justinlittman:
It doesn't matter if it is a 1-1 mapping. All that it matters is that a cocina change does not result in an _unrelated_ fedora change.

But let's step back. The motivation for this line of work was to enable H2 to use the updates endpoint, including for description changes. The risk that was identified was that for non-H2 uses of the updates endpoint, non-description updates would result in changes in the descMetadata datastream.

The changes might be of two sorts:
1. Roundtripping differences: Because there is not a 1-1 mapping of MODS to Cocina, if MODS is converted to Cocina, even if the Cocina is not changed, and then converted back to MODS, the MODS will have changed. While the MODS should be logically the same, it would be disconcerting for users to have the MODS unexpectedly changed.
2. Roundtripping errors: Because we continue to identify bugs in our mapping, forcing an unnecessary roundtripping and saving the resulting MODS will codify those errors. As we discover the errors, we will probably need to remediate, all for changes that weren't necessary.

While this was primarily motivated by and is clearest for descriptive metadata, it seems like it would apply in other areas, e.g., access rights, where we continue to discover wrinkles in our mappings / logic.

The goal here is not a perfect solution, but just to reduce the risk of unintended changes from the cocina update endpoint.

And to be clear, there are multiple ways of accomplishing this; I'm not wedded to this approach.

jcoyne:
I'm all for the descMetadata part of this. I understand how it's helpful there. I'd like an example case for an error in accessRights for instance.

justinlittman:
See https://github.com/sul-dlss/argo/issues/2382

jcoyne:
@justinlittman that is exactly what i'm thinking of.  While this could prevent the source_id from messing up the rights, it still will mess up the rights when the structural metadata gets changed.  And that sort of complexity makes it really hard to know we have a bug that only occurs in field Y when you change field X since it doesn't occur when you change field Z.  I think it is easier to find and fix a bug that occurs consistently whenever a change is made.

justinlittman:
I'm going to disagree -- it is more difficult to track down bugs that occur in totally unrelated places. For example, making the connection between a change in descriptive metadata breaking something in rights.

I'd suggest that we bring @andrewjbtw into the discussion, because it comes down what strategy and priority (if any) he wants to pursue for reducing risk of unintended changes.

jcoyne:
Sure absolutely. Discussion during standup?

jcoyne (approved):

PR #2464, opened 2021-03-10
Remove empty elements from originInfo.

### LINKED ISSUE #2445: originInfo -> event roundtripping errors introduced with PR #2440
These four records errored with #2440 but not in main prior to that:

gy852bt1470
ks031qj5177
vj932ns8042
zn500ww2119

They all _appear_ to be complaining about empty elements within `<originInfo>` not being removed, but normalization does remove them.

vj932ns8042 and zn500ww2119 also have a correctly XML encoded ampersand -- I thought that was the problem for a while.  At least some of these four records are MODS 3.6.  Not sure if that's the problem either.

In all cases, I was unable to replicate the error in spec code.   The last 3 specs in `origin_info_normalizer_spec` were all attempts to figure this out, and might just be noise that can be removed.  These are repeated in the mapping  origin_info_spec with the same context descriptions (currently lines 1709 - 1822)

When I did preliminary roundtrip stats for the event mapping redo, there was a different batch of 25 errors that contained some similar records for which I couldn't figure out what was wrong.  I no longer have those druids handy (perhaps the druid list was shufffled since then, or some descMetadata has been refreshed/remediated), but two of them were vj932ns8042 and zn500ww2119 which are already included above.

justinlittman:
closes #2445

## Why was this change made?
Now ready to clean up originInfo nodes.

Unit, sdr-deploy

NA


justinlittman:
Before:
```
  Success:   99381 (99.904%)
  Different: 95 (0.096%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
  Missing (no descMetadata):     524 (0.524%)
```

After:
```
  Success:   99384 (99.908%)
  Different: 92 (0.092%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
  Missing (no descMetadata):     524 (0.524%)
```

app/services/cocina/mods_normalizer.rb
```diff
@@ -302,10 +302,6 @@ def normalize_usage_primary_for(base_node)
     def remove_empty_elements(start_node)
       return unless start_node
 
-      # Temporarily ignoring <originInfo> pending https://github.com/sul-dlss/dor-services-app/issues/2128
-      # See also xited specs in mods_normalizer_spec.rb and origin_info_normalizer_spec.rb.
-      return if start_node.ancestors('originInfo').present?
-
```
ndushay, line 308: d'oh!

jcoyne (approved):

ndushay (approved):
thank you for solving this!   And thanks for removing cruft from my flailings.

PR #2477, opened 2021-03-11
New Delete Object Endpont

### LINKED ISSUE #2385: Expose a delete endpoint for the DeleteService
This will remove the argo dependency on Fedora 3 for deleting objects.

https://github.com/sul-dlss/dor-services-app/blob/399959cb9b2da96fcb37aeaee5f63808e787520f/app/services/delete_service.rb

jermnelson:
## Why was this change made?
For #2385, still need to update DSC and Argo to support this Delete Object Endpoint. 

New spec for delete request, test suite.

n/a

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ObjectsController < ApplicationController
+class ObjectsController < ApplicationController # rubocop:disable Metrics/ClassLength
   before_action :load_item, except: [:create]
 
   rescue_from(Cocina::ObjectUpdater::NotImplemented) do |e|
@@ -107,6 +107,11 @@ class ObjectsController < ApplicationController
 
   def update_marc_record
     Dor::UpdateMarcRecordService.new(@item).update
+    head :created
+  end
+
+  def destroy
+    DeleteService.destroy(@item.pid)
     head :created
   end
 
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -112,7 +112,11 @@ class ObjectsController < ApplicationController # rubocop:disable Metrics/ClassL
 
   def destroy
     DeleteService.destroy(@item.pid)
-    head :created
+    head :no_content
+  rescue StandardError => e
+    json_api_error(status: :internal_server_error,
+                   title: "Internal server error destroying #{@item.pid}",
+                   message: e.message)
   end
 
   # This endpoint is called by the goobi-notify process in the goobiWF
--- a/app/services/delete_service.rb
+++ b/app/services/delete_service.rb
@@ -2,7 +2,7 @@
 
 # Remove all traces of the object's data files from the workspace and export areas
 class DeleteService
-  # Tries to remove any exsitence of the object in our systems
+  # Tries to remove any existence of the object in our systems
   #   Does the following:
   #   - Removes item from Fedora/Solr
   #   - Removes content from dor workspace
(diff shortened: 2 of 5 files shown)


app/services/delete_service.rb
```diff
@@ -2,7 +2,7 @@
 
 # Remove all traces of the object's data files from the workspace and export areas
 class DeleteService
-  # Tries to remove any exsitence of the object in our systems
+  # Tries to remove any existence of the object in our systems
```
mjgiarlo, line 5: :sunglasses:

openapi.yml
```diff
@@ -1043,6 +1043,26 @@ paths:
...
           schema:
             $ref: '#/components/schemas/Druid'
+    delete:
+      tags:
+        - objects
+      summary: Destroys the object
+      description: ''
+      operationId: 'objects#destroy'
+      responses:
+        '201':
```
mjgiarlo, line 1053: What would you think about using 204 instead of 201 here? 201 implies the creation of a new resource.

openapi.yml
```diff
@@ -1043,6 +1043,26 @@ paths:
...
+          description: Deleted
+        '400':
+          description: Bad request
+        '404':
```
mjgiarlo, line 1057: Should this also include a 500 response if any of the delete service methods raise an exception?

spec/requests/destroy_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  before do
+    allow(Dor).to receive(:find).and_return(object)
+    allow(DeleteService).to receive(:destroy).and_return(nil)
+  end
+
+  context 'when the request is successful' do
+    it 'returns a 201 response' do
+      delete "/v1/objects/#{druid}", headers: { 'Authorization' => "Bearer #{jwt}" }
+      expect(response.status).to eq(201)
+    end
```
mjgiarlo, line 18: See above re: replacing 201 with 204.

I also wonder if we should add a test to confirm a 500 is raised with an informative message if any of the methods in the delete service raise an exception.

openapi.yml
jcoyne: I'm not sure we need to expect that it will raise a 500, because it should never. 🤷

openapi.yml
jermnelson: Done, fixed in latest commit

app/controllers/objects_controller.rb
```diff
@@ -110,6 +110,15 @@ def update_marc_record
     head :created
   end
 
+  def destroy
+    DeleteService.destroy(@item.pid)
+    head :no_content
+  rescue => e
```
jermnelson, line 116: @mjgiarlo, is doing a rescue for any exceptions raised by the `DeleteService` the best pattern here?

spec/requests/destroy_spec.rb
jermnelson: Added new test to test for 500, changed to 204

app/controllers/objects_controller.rb
```diff
@@ -110,6 +110,15 @@ def update_marc_record
     head :created
   end
 
+  def destroy
+    DeleteService.destroy(@item.pid)
+    head :no_content
+  rescue StandardError => e
+    json_api_error(status: :internal_server_error,
+                   title: "Internal server error destroying #{@item.pid}",
+                   message: e.message)
```
mjgiarlo, line 119: LGTM, @jermnelson! Thanks for being open to this.

mjgiarlo (approved):

PR #2479, opened 2021-03-11
Update cocina models to 0.53.1

jcoyne:
## Why was this change made?

Accept the full list of apo roles


justinlittman:
No openapi changes?

justinlittman (approved):

PR #2487, opened 2021-03-11
add metadata/mods endpoint to retrieve source mods

### LINKED ISSUE #1885: Create an endpoint to retrieve the source mods
Currently we have `Metadata#descriptive` to get the public mods, but we don't have a way to get the source mods.  There should be an endpoint to get this representation.  This would allow Argo to present the mods for editing without needing to hit Fedora. It also allows Argo's background job to see if the mods from modsulator is different from what is currently there.

ndushay:
## Why was this change made?

Fixes #1885

Adds an endpoint to retrieve source MODS for a DOR object.

test suite

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -14,6 +14,10 @@ class MetadataController < ApplicationController
     render xml: service
   end
 
+  def mods
+    render xml: @item.descMetadata.content
+  end
+
   # This supports the Legacy Fedora 3 data model. This is used by the accessionWF.
   def update_legacy_metadata
     datastream_names = {
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -59,6 +59,7 @@ Rails.application.routes.draw do
           patch 'legacy', action: :update_legacy_metadata
           get 'dublin_core'
           get 'descriptive'
+          get 'mods'
         end
       end
 
--- a/openapi.yml
+++ b/openapi.yml
@@ -782,6 +782,23 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+  '/v1/objects/{object_id}/metadata/mods':
+    get:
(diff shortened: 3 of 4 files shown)


app/services/publish/mods_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    # @return [String] the source MODS XML
+    delegate :to_xml, to: :ng_xml
+
+    private
+
+    NOKOGIRI_DEEP_COPY = 1
+
+    # @return [Nokogiri::XML::Document] A copy of the descriptiveMetadata of the DOR object
+    def doc
+      @doc ||= dor_object.descMetadata.ng_xml.dup(NOKOGIRI_DEEP_COPY)
```
jcoyne, line 31: can this just be `dor_object.descMetadata.content`?

jcoyne (changes requested):

app/services/publish/mods_service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Publish
```
jcoyne, line 3: This namespace seems confusing to me.  This might be misleading as this is not related to the `publish` endpoint.

ndushay: happy to move it -- what do you suggest?

app/services/publish/mods_service.rb
ndushay: ah - good point - in this context, yes.

app/services/publish/mods_service.rb
jcoyne: Do we need this service at all? Can we just return `item.descMetadata.content`?

ndushay: another excellent point.  Done.

jcoyne (approved):

PR #2494, opened 2021-03-11
Check round trip before doing cocina update or create

justinlittman:
## Why was this change made?
To make sure that only changes that can be roundtripped are accepted for create or update. This ensures the integrity of data that is submitted via cocina update or create endpoints.

Unit

NA

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -28,6 +28,9 @@ class ObjectsController < ApplicationController
     render status: :created, location: object_path(cocina_object.externalIdentifier), json: cocina_object
   rescue SymphonyReader::ResponseError
     json_api_error(status: :bad_gateway, title: 'Catalog connection error', message: 'Unable to read descriptive metadata from the catalog')
+  rescue Cocina::RoundtripValidationError => e
+    Honeybadger.notify(e)
+    json_api_error(status: e.status, message: e.message)
   rescue Cocina::ValidationError => e
     json_api_error(status: e.status, message: e.message)
   end
@@ -38,6 +41,9 @@ class ObjectsController < ApplicationController
     cocina_object = Cocina::ObjectUpdater.run(obj, update_request)
 
     render json: cocina_object
+  rescue Cocina::RoundtripValidationError => e
+    Honeybadger.notify(e)
+    json_api_error(status: e.status, message: e.message)
   rescue Cocina::ValidationError => e
     json_api_error(status: e.status, message: e.message)
   end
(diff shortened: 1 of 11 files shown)


app/services/cocina/object_creator.rb
```diff
@@ -155,5 +154,15 @@ def apply_default_access(item)
     def truncate_label(label)
       label.length > 254 ? label[0, 254] : label
     end
+
+    def validate(obj)
+      if Settings.enabled_features.validate_descriptive_roundtrip.create
+        validator = DescriptionRoundtripValidator.new(obj)
+        raise ValidationError.new(validator.error, status: :internal_server_error) unless validator.valid?
```
jcoyne, line 161: I'm not sure that 500 is the right code here. Would you consider 409?

app/controllers/objects_controller.rb
```diff
@@ -39,6 +42,9 @@ def update
 
     render json: cocina_object
   rescue Cocina::ValidationError => e
+    # Notify if this is a roundtripping error.
+    Honeybadger.notify(e) if e.status == :internal_server_error
```
jcoyne, line 46: Can this be a subclass of `Cocina::ValidationError` so we don't have to test on properties?

app/services/cocina/object_updater.rb
```diff
@@ -144,7 +145,13 @@ def validate
       raise ValidationError, validator.error unless validator.valid?
 
       raise NotImplemented, 'Updating descriptive metadata not supported' if !update_descriptive? && client_attempted_metadata_update?
+
+      if has_changed?(:description) && update_descriptive? && Settings.enabled_features.validate_descriptive_roundtrip.update
```
jcoyne, line 149: why not do a `return unless ...` here?

app/validators/cocina/description_roundtrip_validator.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      true
+    end
+
+    private
+
+    attr_reader :cocina_object
+
+    def druid
+      # Requests do not have druids.
```
jcoyne, line 35: 👍

jcoyne (changes requested):

app/services/cocina/object_creator.rb
justinlittman: Done.

app/controllers/objects_controller.rb
justinlittman: Done.

app/services/cocina/object_updater.rb
justinlittman: While I generally support guard statements, I find this sort of a use of a guard statement confusing and requires additional effort to understand the code. In particular, I think guard statements should be used to filter the unhappy/atypical path, leaving the normal code path as the happy/typical path. In this case, if filters the happy/typical path and leaves the normal code path the unhappy/atypical path.

justinlittman: But this is a pet peeve of mine.

jcoyne (approved):

Issue #2411, opened 2021-03-04
Update legacy api endpoint to round trip and provide xml schema validation for mods data


aaron-collier:
Whenever a user is providing or updating mods, we will check that the roundtripped mods matches, if different the user will be informed and possibly prevented from submitting it.

Under a feature flag, When in the UI: raise an error so that data is not submitted to avoid complicated remediation workflows.

How/Where can we introduce MODS validation? (mods validation service)?

PR #2504, opened 2021-03-12
Adds MODS validation to legacy metadata endpoint.

justinlittman:
refs #2411

## Why was this change made?
Keep out bad data.

Unit

openapi


justinlittman:
See https://github.com/sul-dlss/argo/pull/2480 for the Argo side.

jcoyne (approved):

PR #2507, opened 2021-03-15
Adds script for roundtripping full cocina objects.

justinlittman:
## Why was this change made?
So we can find mapping errors before our users do.

Local

README

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,9 @@ gem 'rails', '~> 5.2.0'
 gem 'puma', '~> 3.12'
 
 group :development do
+  gem 'dor-services-client' # used by bin/validate-cocina-roundtrip
   gem 'listen', '~> 3.0.5'
+  gem 'rubyzip', '>= 1.0.0'
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
--- a/app/services/cocina/object_updater.rb
+++ b/app/services/cocina/object_updater.rb
@@ -9,20 +9,25 @@ module Cocina
     # @param [ActiveFedora::Base] item the object to update
     # @param [Cocina::Models::RequestDRO,Cocina::Models::RequestCollection,Cocina::Models::RequestAdminPolicy] obj the cocina model provided by the client
     # @param [#create] event_factory creates events
-    def self.run(item, obj, event_factory: EventFactory)
-      new(item, obj).run(event_factory: event_factory)
+    # @param [boolean] trial do not persist or event; run all mappings regardless of changes
+    # @param [Cocina::FromFedora::DataErrorNotifier] notifier
+    def self.run(item, obj, event_factory: EventFactory, trial: false, notifier: nil)
+      new(item, obj, trial: trial).run(event_factory: event_factory, notifier: notifier)
     end
 
-    def initialize(item, obj)
+    def initialize(item, obj, trial: false)
       @params = params
       @obj = obj
       @item = item
-      @orig_obj = Mapper.build(item)
+      @trial = trial
     end
 
-    def run(event_factory:)
+    # rubocop:disable Metrics/AbcSize
+    def run(event_factory:, notifier: nil)
+      @orig_obj = Mapper.build(item, notifier: notifier)
+
       # Validate will raise an error if not valid.
-      validate
+      validate unless trial
 
       case obj
       when Cocina::Models::AdminPolicy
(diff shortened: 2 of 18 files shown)


bin/validate-cocina-roundtrip
```diff
@@ -25,151 +26,147 @@ parser.parse!(into: options)
...
+
+# Monkeypatch to retrieve tags using dor-services-client.
+class AdministrativeTags
+  def self.project(pid:)
+    [Dor::Services::Client.object(pid)
+                          .administrative_tags
+                          .list
+                          .find { |tag| tag.start_with?('Project :') }
+                          .split(' : ', 2)
+                          .last]
```
mjgiarlo, line 74: Can you replace this code with:

```ruby
AdministrativeTags.for(pid: pid)
```

Since that DSC call ultimately is an HTTP wrapper around a service object call that already lives in this codebase? If so, we can get around having to add DSC as a dependency of DSA.

bin/validate-cocina-roundtrip
```diff
@@ -25,151 +26,147 @@ parser.parse!(into: options)
...
+  end
 
-  result.failure? ? result.failure : 'Valid MODS'
+  def self.content_type(pid:)
+    [Dor::Services::Client.object(pid)
+                          .administrative_tags
+                          .list
```
mjgiarlo, line 80: Same as above.

bin/validate-cocina-roundtrip
justinlittman: The reason for this whole admin tags kludge is because this isn't run on the dor-services-app server. It may be run locally or on sdr-deploy.

mjgiarlo: @justinlittman OK, now I get it.

app/services/cocina/object_updater.rb
```diff
@@ -9,20 +9,25 @@ class NotImplemented < StandardError; end
...
-    def self.run(item, obj, event_factory: EventFactory)
-      new(item, obj).run(event_factory: event_factory)
+    # @param [boolean] trial do not persist or event; run all mappings regardless of changes
+    # @param [Cocina::FromFedora::DataErrorNotifier] notifier
+    def self.run(item, obj, event_factory: EventFactory, trial: false, notifier: nil)
+      new(item, obj, trial: trial).run(event_factory: event_factory, notifier: notifier)
     end
 
-    def initialize(item, obj)
+    def initialize(item, obj, trial: false)
```
justinlittman, line 18: The crux of this change is adding the `trial` flag.

justinlittman:
I don't understand why code climate is so unhappy. The only red bars are in code that is supposed to be excluded from code climate.

Gemfile
```diff
@@ -38,6 +38,7 @@ gem 'pg'
 gem 'progressbar' # for the cleaner rake task
 gem 'retries' # for ReleaseTags::PurlClient and Goobi
 gem 'ruby-cache', '~> 0.3.0'
+gem 'rubyzip', '>= 1.0.0'
```
mjgiarlo, line 41: OK, I now get that we need these dependencies to run the scripts on our laptops (or sdr-deploy or anywhere). Does that mean we can safely move the two new deps to the `:development` bundler group to keep them out of deployed instances? That's my only request/question, and then I'm happy to :white_check_mark:

mjgiarlo:
@justinlittman :speech_balloon: 
> I don't understand why code climate is so unhappy. The only red bars are in code that is supposed to be excluded from code climate.

It seems really unhappy about the FedoraCache object, which is only used by our migration/roundtripping scripts, right? I think we can ignore CodeClimate in terms of judging this PR if so. :man_shrugging:

justinlittman:
@mjgiarlo code climate is configured to ignore `lib/` so I wish it would keep its opinion to itself.

image

Gemfile
justinlittman: Thanks for the suggestion. Fixed.

Gemfile
```diff
@@ -52,6 +52,11 @@ gem 'marc'
 gem 'moab-versioning', '~> 4.0', require: 'moab/stanford'
 gem 'preservation-client', '>= 3.0' # 3.x or greater is needed for token auth
 
+group :development do
+  gem 'dor-services-client' # used by bin/validate-cocina-roundtrip
+  gem 'rubyzip', '>= 1.0.0'
```
mjgiarlo, line 57: Could you shift these into the other development group up around lines 11-16? Approving pending consideration!

mjgiarlo (approved):

mjgiarlo:
> @mjgiarlo code climate is configured to ignore `lib/` so I wish it would keep its opinion to itself.
> 
> image

I can only conclude from this that CodeClimate may not be applying exclude patterns to coverage (only applying to maintainability analysis?). I have no idea. But I'm happy to disregard it when it's being unreasonable.

Gemfile
justinlittman: Thanks for the catch.

mjgiarlo (approved):

PR #2514, opened 2021-03-16
Eventual unique identifiers for files and file sets.

### LINKED ISSUE #2513: New strategy for ids for files and file sets
Our approach to assigned identifiers for files and file sets is inconsistent. In addition, there the id for file is not stored in Fedora. The goal is to allow files and file identifiers to be individually addressable, which requires a unique identifier.

Content metadata generator does not preserve existing fileset ids https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/to_fedora/content_metadata_generator.rb#L97

whereas `FromFedora::FileSets` does: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/from_fedora/file_sets.rb#L13

justinlittman:
closes #2513

## Why was this change made?
Allow for unique and addressable identifiers for files and filesets.

Unit

NA

### DIFF
--- a/app/services/cocina/id_generator.rb
+++ b/app/services/cocina/id_generator.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module Cocina
+  # Generates identifiers for use within Cocina model.
+  class IdGenerator
+    ID_NAMESPACE = 'http://cocina.sul.stanford.edu/'
+
+    def self.generate_fileset_id
+      generate_id('fileSet')
+    end
+
+    def self.generate_or_existing_fileset_id(existing_id)
+      cocina_id?(existing_id) ? existing_id : generate_fileset_id
+    end
+
+    def self.generate_file_id
+      generate_id('file')
+    end
+
+    def self.generate_or_existing_file_id(existing_id)
+      cocina_id?(existing_id) ? existing_id : generate_file_id
+    end
+
+    def self.cocina_id?(id)
+      id.present? && id.start_with?(ID_NAMESPACE)
+    end
+
+    def self.generate_id(type)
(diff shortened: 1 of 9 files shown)


app/services/cocina/id_generator.rb
```diff
@@ -0,0 +1,33 @@
...
+  # Generates identifiers for use within Cocina model.
+  class IdGenerator
+    ID_NAMESPACE = 'http://cocina.sul.stanford.edu/'
+
+    def self.generate_fileset_id
+      generate_id('fileSet')
+    end
+
+    def self.generate_or_existing_fileset_id(existing_id)
+      cocina_id?(existing_id) ? existing_id : generate_id('fileSet')
```
jcoyne, line 13: can this have `generate_id('fileSet')` as the last term?

app/services/cocina/id_generator.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def self.generate_file_id
+      generate_id('file')
+    end
+
+    def self.generate_or_existing_file_id(existing_id)
+      cocina_id?(existing_id) ? existing_id : generate_id('file')
```
jcoyne, line 21: can this have `generate_file_id` as the last term?

jcoyne (changes requested):

app/services/cocina/id_generator.rb
justinlittman: Good idea. Fixed.

app/services/cocina/id_generator.rb
justinlittman: Good idea. Fixed.

jcoyne (approved):

PR #2525, opened 2021-03-17
Create an endpoint for applying admin policy defaults

jcoyne:
## Why was this change made?
This will be used so that we can decouple Argo from Fedora.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -540,6 +540,27 @@ paths:
               properties:
                 administrative_tag:
                   $ref: '#/components/schemas/AdministrativeTag'
+  '/v1/objects/{id}/apply_admin_policy_defaults':
+    post:
+      tags:
+        - objects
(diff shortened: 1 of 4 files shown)


openapi.yml
```diff
@@ -540,6 +540,27 @@ paths:
               properties:
                 administrative_tag:
                   $ref: '#/components/schemas/AdministrativeTag'
+  '/v1/objects/{id}/apply_admin_policy_defaults':
+    post:
+      tags:
+        - objects
+      summary: Applies the default values (copyright, use_statement, rights) from the admin policy
+      description: ''
```
justinlittman, line 548: Cruft?

justinlittman (approved):

openapi.yml
jcoyne: Not really.  The description is valid and we're just being explicit that the description is empty.  This is used elsewhere in the file too.

PR #2536, opened 2021-03-17
Create an API endpoint that returns the version log

jcoyne:
## Why was this change made?

To decouple Argo from Fedora 3 and to create APIs.


mjgiarlo (changes requested):
Should we add a new request spec for the new controller action?

jcoyne:
Sorry. I thought I did. I just didn't add it to the commit.

mjgiarlo (approved):

PR #2542, opened 2021-03-18
Provide an API to provide datastream list and content for the Argo UX

jcoyne:
## Why was this change made?

This will allow us to maintain the existing Argo UI while decoupling it from Fedora.

### DIFF
--- a/app/controllers/datastreams_controller.rb
+++ b/app/controllers/datastreams_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# A controller to display datastreams for an object.
+# This controller is intended to be temporary until we can decouple Argo's UX
+# from the datastream abstraction.
+class DatastreamsController < ApplicationController
+  before_action :load_item
+
+  def index
+    result = @item.datastreams
+                  .reject { |name, instance| instance.new? || name == 'workflows' }
+                  .values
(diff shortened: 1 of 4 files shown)


justinlittman (commented):
Should this have the `/v1/objects/{object_id}/metadata/legacy` endpoint?

jcoyne:
I'm not understanding the question. Should what have that?

justinlittman:
This PR adds the `/v1/objects/{object_id}/metadata/datastreams` endpoint. Should it use the existing `/v1/objects/{object_id}/metadata/legacy` endpoint?

jcoyne:
I considered that, but ultimately decided against it for the following reasons:

1. This supports all datastreams, where the legacy API only supports a subset (https://github.com/sul-dlss/dor-services-app/blob/524fbc1579f725e5f896f9cecd280ef4c478cbae/app/controllers/metadata_controller.rb#L23-L33)
1. This uses the datastream name, where the legacy API has modifed the name
1. This API has a different lifecycle, it is intended to go away as we replace certain UX components in Argo, and certainly before we decouple dor-services-app from Fedora.  The legacy API may need to be supported over a longer timeframe.

mjgiarlo (approved):

justinlittman:
Thanks for the explanation @jcoyne. Makes sense to me.

mjgiarlo:
I missed the outstanding question when I merged the PR. Sorry about that!

justinlittman:
No worries. If it was a deal breaker, I would have marked as changes required.

Issue #2499, opened 2021-03-11
Implement additional part specs


arcadiafalcone:
See #2498

PR #2552, opened 2021-03-19
Implement top level part mapping.

justinlittman:
refs #2499

## Why was this change made?
Infinite MODS.

Unit, sdr-deploy

NA

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,11 +46,23 @@ Metrics/ClassLength:
     - 'app/services/cocina/to_fedora/descriptive/related_resource.rb'
     - 'app/services/cocina/object_updater.rb'
 
-Rails/Exit:
+Performance/CollectionLiteralInLoop:
+  Enabled: true
+  Exclude:
+    - 'app/models/dor/update_marc_record_service.rb'
+    - 'app/models/symphony_reader.rb'
+    - 'app/services/cocina/from_fedora/descriptive/access.rb'
+    - 'app/services/cocina/from_fedora/descriptive/form.rb'
+    - 'app/services/cocina/from_fedora/descriptive/title_builder.rb'
+    - 'app/services/cocina/to_fedora/descriptive/access.rb'
+    - 'app/services/cocina/to_fedora/descriptive/related_resource.rb'
+    - 'app/services/cocina/to_fedora/descriptive/part_writer.rb'
+
+Rails/Output:
   Exclude:
     - 'lib/report.rb'
 
-Rails/Output:
+Rails/Exit:
   Exclude:
     - 'lib/report.rb'
 
@@ -333,8 +345,6 @@ Performance/AncestorsInclude: # (new in 1.7)
 Performance/BigDecimalWithNumericArgument: # (new in 1.7)
   Enabled: true
 Performance/BlockGivenWithExplicitBlock: # (new in 1.9)
-  Enabled: true
-Performance/CollectionLiteralInLoop: # (new in 1.8)
   Enabled: true
 Performance/ConstantRegexp: # (new in 1.9)
   Enabled: true
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -89,18 +89,6 @@ Naming/VariableNumber:
     - 'spec/services/cleanup_service_spec.rb'
     - 'spec/support/datastreams.rb'
 
-# Offense count: 9
-# Configuration parameters: MinSize.
-Performance/CollectionLiteralInLoop:
-  Exclude:
-    - 'app/models/dor/update_marc_record_service.rb'
-    - 'app/models/symphony_reader.rb'
-    - 'app/services/cocina/from_fedora/descriptive/access.rb'
-    - 'app/services/cocina/from_fedora/descriptive/form.rb'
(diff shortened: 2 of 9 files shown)


justinlittman:
Before:
```
  Success:   99499 (99.499%)
  Different: 49 (0.049%)
  To Cocina error:     452 (0.452%)
  To Fedora error:     0 (0.0%)
```

After:
```
  Success:   99500 (99.5%)
  Different: 48 (0.048%)
  To Cocina error:     452 (0.452%)
  To Fedora error:     0 (0.0%)
```

.rubocop.yml
```diff
@@ -327,7 +327,7 @@ Performance/BigDecimalWithNumericArgument: # (new in 1.7)
 Performance/BlockGivenWithExplicitBlock: # (new in 1.9)
   Enabled: true
 Performance/CollectionLiteralInLoop: # (new in 1.8)
-  Enabled: true
+  Enabled: false
```
jcoyne, line 330: Is there a reason we are turning this cop off and not adding a TODO?

jcoyne (changes requested):

.rubocop.yml
justinlittman: The changes required make the code less readable and that sort of performance optimization isn't a consideration.

jcoyne: That sounds like a local consideration for the method you were working on rather than something we want to change globally.  Can you wrap your method with an exception?

justinlittman: Adding:
```
Performance/CollectionLiteralInLoop: # (new in 1.8)
  Enabled: true
  Exclude:
    - 'app/services/cocina/to_fedora/descriptive/part_writer.rb'
```

resulted in:
```
app/models/dor/update_marc_record_service.rb:164:9: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
        %w(partName partNumber).include?(child.name)
        ^^^^^^^^^^^^^^^^^^^^^^^
app/models/symphony_reader.rb:34:42: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
      record << marc_field(field) unless %w[001 003].include? field['tag'] # explicitly remove all 001 and 003 fields from the record
                                         ^^^^^^^^^^^
app/services/cocina/from_fedora/descriptive/access.rb:148:18: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
              if %w[marcorg oclcorg].include?(node[:authority])
                 ^^^^^^^^^^^^^^^^^^^
app/services/cocina/from_fedora/descriptive/form.rb:96:47: C: Performance/CollectionLiteralInLoop: Avoid immutable Hash literals in loops. It is better to extract it into a local variable or a constant.
          other_genre.each { |genre| forms << { type: 'genre' }.merge(build_genre(genre)) }
                                              ^^^^^^^^^^^^^^^^^
app/services/cocina/from_fedora/descriptive/title_builder.rb:97:19: C: Performance/CollectionLiteralInLoop: Avoid immutable Array literals in loops. It is better to extract it into a local variable or a constant.
            add = ['-', "'", ' '].include?(last_character) ? 0 : 1
                  ^^^^^^^^^^^^^^^
app/services/cocina/to_fedora/descriptive/access.rb:65:83: C: Performance/CollectionLiteralInLoop: Avoid immutable Hash literals in loops. It is better to extract it into a local variable or a constant.
                xml.physicalLocation access_contact.value || access_contact.code, { type: 'repository' }.merge(descriptive_attrs(access_contact))
```

which had nothing to do with my changes (or had performance considerations). And so I turned it off.

jcoyne: It's because the settings in `.rubocop.yml` are overwriting what you already have in https://github.com/sul-dlss/dor-services-app/blob/8461592524975be180ccd0d688033715c4fa2090/.rubocop_todo.yml#L94

justinlittman: Ah ... since we're ignoring it in all cases, then we should just turn it off in `.rubocop.yml`. It's confusing to split our ignores across the two rubo files, especially when we never intend to fix the TODOs.

However, at the end of the day, I don't care that much.

jcoyne: I don't think we are ignoring it in all cases.  I've hit it before and said "Ah-ha! I should fix this."  We are ignoring it in 7 specific places and five of them are in the descriptive metadata mapping.

justinlittman: OK, then I moved them from `.rubocop_todo.yml` to `.rubocop.yml` since they're not TODOs.

jcoyne: I rather think the ones in app/models were TODOs, as they were detected and never fixed, but I don't want to spend more time on this as it's really not worth it.

jcoyne (approved):

.rubocop.yml
justinlittman: Agree. I conceptually understand the reason for TODOs, but I think in practice they do more harm then good.

justinlittman: Hah, and to add insult to injury, now there is a conflict on rubocop.yml.

PR #2558, opened 2021-03-20
Revert "Bootstrap the Ur-APO"

jcoyne:
This reverts commit ea6ce21165c38bbd180357dc830ee9a839c7614b.

## Why was this change made?

This was preventing the argo tests from passing.  While this is a nice to have, it's not critical. We can do this with a `docker-compose web exec` instead.

### DIFF
--- a/docker/invoke.sh
+++ b/docker/invoke.sh
@@ -22,9 +22,5 @@ set -e
 echo "Migrating db"
 bin/rails db:migrate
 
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

justinlittman:
How is `docker-compose web exec` going to solve the problem in ci? Won't we still have the same problem of having to wait for Solr to be up and ready?

PR #2561, opened 2021-03-22
Provide a reports on otherId names from identity datastream. Refactor…

justinlittman:
… reports to be single exe per report, use a common report class, and get all data from cache.

## Why was this change made?
Provide additional report for analyzing identity metadata.
Improve maintainability and performance of reports.

Locally

README


justinlittman:
`lib/` is supposed to be omitted from code climate, but it is angry nonetheless.

jcoyne (approved):

Issue #2598, opened 2021-03-24
Unable to create a new collection


andrewjbtw:
**Describe the bug**
I get a 500 error trying to create a collection from the "Create collection" button on an APO: https://argo-stage.stanford.edu/view/druid:bb973ms0268

Argo error: https://app.honeybadger.io/projects/49894/faults/78554731
Dor services app error: https://app.honeybadger.io/projects/50568/faults/78554730

**User Impact**
It's not possible to create new collections.

**Expected behavior**
A collection should have been created.

jcoyne:
The problem is occurring when it is updating the APO to add the new collection to the APO.

PR #2600, opened 2021-03-25
"obj" var name disambiguation

jmartin-sul:
## Why was this change made?

I found these var names a bit confusing when i was doing some initial debugging on https://github.com/sul-dlss/dor-services-app/issues/2598.  In particular, the use of `obj` as the second param to `ObjectUpdater.run` (and corresponding ivar name in the class) for a cocina model instance, but in the caller (`ObjectsController#update`) as a Fedora object passed as the first param to `ObjectUpdater.run` made it cumbersome to trace through what was happening by copying and running parts of the code individually in Rails console.

unit tests

n/a, though hopefully this hews closer to the convention suggested by https://github.com/sul-dlss/cocina-models/#usage-conventions

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -36,11 +36,11 @@ class ObjectsController < ApplicationController
   end
 
   def update
-    obj = Dor.find(params[:id])
+    fedora_object = Dor.find(params[:id])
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

jcoyne (approved):
Thanks!

jmartin-sul:
> Thanks!

np!

PR #2605, opened 2021-03-25
update object updater spec

peetucket:
## Why was this change made?

Related to sul-dlss/argo#2562.  Ideally, this should be merged and deployed first, but not super critical (it will just leave some extraneous XML in contentMetadata in certain circumstances if the argo PR is deployed first).

The problem occurs if you change the content type from a book type to a non-book type, as this will currently leave behind the unneeded `<bookData>` node. As noted in a comment in the code, everything in DSA works fine when you switch from a non-book type to a book type, because in this case the changes in sul-dlss/argo#2562 will generate  a completely updated structural part of the cocina-model, which will then rebuild and replace the contentMetadata, thus adding the new `<bookData>` node.  

This PR:
- removes these extra nodes when you are switching content types in a way that does not generate new structural (i.e. not going to a book type).
- adds a new test block
- updates the explanation of some existing tests

Updated/new tests

### DIFF
--- a/spec/services/cocina/object_updater_spec.rb
+++ b/spec/services/cocina/object_updater_spec.rb
@@ -297,6 +297,10 @@ RSpec.describe Cocina::ObjectUpdater do
 
     let(:content_metadata) { double }
 
+    let(:content_metadata_ng_xml) { double }
+
+    let(:book_data_node) { double }
+
     let(:desc_metadata) { double }
 
     let(:identity_metadata) { double }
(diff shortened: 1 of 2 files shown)


spec/services/cocina/object_updater_spec.rb
```diff
@@ -470,6 +476,35 @@
       end
     end
 
+    context 'when updating structural with a contains block' do
```
peetucket, line 479: this is a new context block testing a code path that didn't seem to be hit before

mjgiarlo (approved):

PR #2606, opened 2021-03-25
Update H2 approximate date mappings

### LINKED ISSUE #2603: H2 mapping: Provide H2 mapping for approximate date range
image

Note that for a range, each part can be approximate. However, the current examples only show how to indicate that a single date is approximate.

arcadiafalcone:
## Why was this change made?
Closes #2603

CircleCI

n/a

Implementation ticket #2607 and https://github.com/sul-dlss/happy-heron/issues/1306


ndushay (approved):
I'm hesitant to merge this as it implies we need a ticket to implement the fix, and I don't know if Arcadia will create that and prioritize in on our current zenhub board.   

But I do approve!

PR #2609, opened 2021-03-26
Fix software-code H2 example.

justinlittman:
## Why was this change made?
So that it is in sync with H2.

Unit

NA


arcadiafalcone (changes requested):
Retain technical manuals - I'll add it to the mapping spreadsheet.

justinlittman:
See https://github.com/sul-dlss/happy-heron/pull/1311

arcadiafalcone (approved):

mjgiarlo (approved):

PR #2614, opened 2021-03-26
Filter hydrus objects out of HB alerts

### LINKED ISSUE #2478: Filter out Hydrus from update_legacy_metadata date experiment
See https://github.com/sul-dlss/dor-services-app/blob/c450ab7ff178e6b1687d194cba48ec55345b408f/app/services/legacy_metadata_service.rb#L17

Do not send an HB alert if this is a Hydrus object (based on the project tag)

peetucket:
## Why was this change made?

Fixes #2478 - filter hydrus objects out when sending alerts to HB of date mismatches

Existing unit tests

### DIFF
--- a/app/services/legacy_metadata_service.rb
+++ b/app/services/legacy_metadata_service.rb
@@ -27,6 +27,7 @@ class LegacyMetadataService
     @event_factory = event_factory
   end
 
+  # rubocop:disable Metrics/AbcSize
   def update_datastream_if_newer
     if !datastream.createDate || updated > datastream.createDate
       datastream.content = content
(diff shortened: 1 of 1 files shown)


app/services/legacy_metadata_service.rb
```diff
@@ -38,10 +39,14 @@ def update_datastream_if_newer
 
     return if !datastream.createDate || updated > datastream.createDate
 
+    # do not send alerts to HB for Hydrus items (see https://github.com/sul-dlss/dor-services-app/issues/2478)
+    return if AdministrativeTags.for(pid: @item.id).include?('Project : Hydrus')
```
mjgiarlo, line 43: Should we also consider `AdministrativeTags.project(pid: @item.id).include?('Hydrus')` ?

mjgiarlo: Related: why does this method assume an item only belongs to a single project? Tagging @andrewjbtw on this so we can determine if this is desired behavior or problematic behavior.

peetucket: I didn't realize there was a `project` method ... that makes sense.

peetucket: > Related: why does this method assume an item only belongs to a single project? Tagging @andrewjbtw on this so we can determine if this is desired behavior or problematic behavior.

Can you say more about this?  I'm not sure I follow.

mjgiarlo: The method I recommended you might use instead has a bug: it only returns the FIRST project, not every project. I will take a look after lunch. If you want to pair, lmk

mjgiarlo: but I don't think that work needs to block this PR!

mjgiarlo (approved):

app/services/legacy_metadata_service.rb
peetucket: Cool, happy to pair after lunch.  Ping me.

PR #2616, opened 2021-03-27
Bootstrap the Ur-AdminPolicy

jcoyne:
## Why was this change made?

There is no way to create new objects in the system until the Ur-AdminPolicy exists.  Presently Argo has a factory that writes directly into Fedora to accomplish this, but we want to decouple Argo from Fedora.  Now, we can turn on a feature flag when this is run in a test container which will automatically bootstrap the Ur-AdminPolicy.


jmartin-sul (changes requested):
small suggestion and i think i noticed a minor settings inconsistency, but otherwise lgtm

jcoyne:
Thanks @jmartin-sul I made those changes.

jmartin-sul (approved):

PR #2626, opened 2021-03-30
do not send HB alerts for project phoenix books with invalid resource types

peetucket:
## Why was this change made?

So we don't spam HB

### DIFF
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -27,7 +27,7 @@ module Cocina
           has_member_orders = build_has_member_orders
           structural[:hasMemberOrders] = has_member_orders if has_member_orders.present?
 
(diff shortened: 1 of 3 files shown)


ndushay (approved):

jcoyne:
I think we should revert this, because this notice is saying "fix this data, because we can't migrate it to the next generation repository system.

andrewjbtw:
We already know Project Phoenix is bad. Excluding it from this alert helps us find what else is bad.

peetucket:
This was a pretty small change (took about 30 minutes) with a goal of not spamming HB in the near term.  Remediating the data wouldn't be hard either, though actually touching and versioning ~187k objects would probably take a while to run (maybe a couple days?).  We could do that too if we were actually going to migrate these books, which I believe from the initial google books workcycle we had decided would not be necessary, given the fact that our current GB pipleine is likely going to get the vast majority of these books again anyway.

Issue #2623, opened 2021-03-30
Map collection copyright property


mjgiarlo:
`rightsMetadata/copyright` :left_right_arrow: `access/copyright`

Blocked by release of cocina-models 0.57.0

jcoyne:
@andrewjbtw can you confirm that collections should be allowed to bear copyright statements.  Since a collection itself is not a creative work, this seems suspect to me.

andrewjbtw:
It's an odd case, but I think we should still map copyright here because:

1. The metadata itself is copyrightable. Probably if people are really thinking of it this way, they would choose a different copyright for the metadata than for the objects but if they do they need the field.
2. From the public access standpoint the idea of a "collection object" is not clear. The collection display can stand in for a description of the collection overall, in which case having a copyright statement that is consistent with what's on the items themselves makes sense. In most cases, the collection will have the same statement as items because they all start from the same defaults.
3. We currently show copyright on collection purls. While "that's how we've always done it" doesn't necessarily mean we should continue to do that, in this case removing the copyright statement from the collection object would be a visible change (for new or republished purls) and I don't think we can make the call to change this right now.

mjgiarlo:
Closed by #2632.

PR #2632, opened 2021-03-31
Consistently map licenses, copyright statements, and use statements for items and collections

mjgiarlo:
Fixes sul-dlss/argo#2612

## Why was this change made?

Data loss is bad.

CI. Should prolly be tested in QA by @andrewjbtw before we go forward.

None

### DIFF
--- a/app/services/cocina/from_fedora/access.rb
+++ b/app/services/cocina/from_fedora/access.rb
@@ -4,6 +4,8 @@ module Cocina
   module FromFedora
     # builds the Access subschema for Collections
     class Access
+      NONE_LICENSE_URI = 'http://cocina.sul.stanford.edu/licenses/none'
+
       def self.collection_props(rights_metadata_ds)
         props = new(rights_metadata_ds).props
         # Collection access does not have download
@@ -20,7 +22,9 @@ module Cocina
           access: access_rights,
           download: download? ? access_rights : 'none',
           readLocation: location,
-          license: license_uri
+          license: license_uri,
+          copyright: copyright,
+          useAndReproductionStatement: use_statement
         }.compact.tap do |h|
           h[:controlledDigitalLending] = true if cdl?
         end
@@ -31,6 +35,8 @@ module Cocina
       attr_reader :rights_metadata_ds
 
       def license_uri
+        return NONE_LICENSE_URI if rights_metadata_ds.use_license.first == 'none'
+
         if rights_metadata_ds.open_data_commons.first.present?
           Dor::OpenDataLicenseService.property(rights_metadata_ds.open_data_commons.first).uri
         elsif rights_metadata_ds.creative_commons.first.present?
@@ -40,6 +46,14 @@ module Cocina
 
       def rights_object
         rights_metadata_ds.dra_object.obj_lvl
+      end
+
+      def copyright
+        rights_metadata_ds.copyright.first.presence
+      end
+
+      def use_statement
+        rights_metadata_ds.use_statement.first.presence
       end
 
       # @return [Bool] true unless the rule="no-download" has been set or if the access is citation-only or dark
--- a/app/services/cocina/from_fedora/dro_access.rb
+++ b/app/services/cocina/from_fedora/dro_access.rb
@@ -16,8 +16,9 @@ module Cocina
       def props
         super.tap do |access|
           access[:embargo] = embargo unless embargo.empty?
-          access[:useAndReproductionStatement] = rights_metadata_ds.use_statement.first if rights_metadata_ds.use_statement.first.present?
-          access[:copyright] = rights_metadata_ds.copyright.first if rights_metadata_ds.copyright.first.present?
+          access[:useAndReproductionStatement] = use_statement if use_statement.present?
+          access[:copyright] = copyright if copyright.present?
+          access[:license] = license_uri if license_uri.present?
         end
       end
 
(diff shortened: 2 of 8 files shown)


app/services/cocina/from_fedora/access.rb
```diff
@@ -4,6 +4,8 @@ module Cocina
   module FromFedora
     # builds the Access subschema for Collections
     class Access
+      NONE_LICENSE_URI = 'http://cocina.sul.stanford.edu/licenses/none'
```
mjgiarlo, line 7: I made this up, obviously.

app/services/cocina/from_fedora/access.rb
```diff
@@ -20,7 +22,9 @@ def props
           access: access_rights,
           download: download? ? access_rights : 'none',
           readLocation: location,
-          license: license_uri
+          license: license_uri,
+          copyright: copyright,
+          useAndReproductionStatement: use_statement
```
mjgiarlo, line 27: These previously were not mapped.

app/services/cocina/from_fedora/dro_access.rb
```diff
@@ -16,8 +16,9 @@ def initialize(rights_metadata_ds, embargo:)
       def props
         super.tap do |access|
           access[:embargo] = embargo unless embargo.empty?
-          access[:useAndReproductionStatement] = rights_metadata_ds.use_statement.first if rights_metadata_ds.use_statement.first.present?
-          access[:copyright] = rights_metadata_ds.copyright.first if rights_metadata_ds.copyright.first.present?
+          access[:useAndReproductionStatement] = use_statement if use_statement.present?
+          access[:copyright] = copyright if copyright.present?
+          access[:license] = license_uri if license_uri.present?
```
mjgiarlo, line 21: This was not previously mapped.

app/services/cocina/to_fedora/access.rb
```diff
@@ -23,12 +23,42 @@ def apply
...
+          item.rightsMetadata.creative_commons_human = Dor::CreativeCommonsLicenseService.property(license_code).label
+        elsif Dor::OpenDataLicenseService.key?(license_code)
+          item.rightsMetadata.open_data_commons = license_code
+          item.rightsMetadata.open_data_commons.uri = Dor::OpenDataLicenseService.property(license_code).uri
+          item.rightsMetadata.open_data_commons_human = Dor::OpenDataLicenseService.property(license_code).label
+        elsif license_code == 'none'
+          item.rightsMetadata.creative_commons = license_code
+          item.rightsMetadata.creative_commons_human = 'no Creative Commons (CC) license'
+        else
+          raise ArgumentError, "'#{license_code}' is not a valid license code"
```
mjgiarlo, line 49: This logic is essentially ported from dor-services gem (DefaultObjectRights class) and adapted to work here, with the `'none'` check added.

app/services/cocina/to_fedora/access.rb
```diff
@@ -23,12 +23,42 @@ def apply
...
+      end
+
+      def license_code
+        license_codes.fetch(access.license)
+      end
+
+      def license_codes
+        DefaultRights::LICENSE_CODES.merge(
+          Cocina::FromFedora::Access::NONE_LICENSE_URI => 'none'
+        )
```
mjgiarlo, line 60: Here's where we patch in the ability to map nones across.

app/services/cocina/from_fedora/access.rb
jcoyne: @andrewjbtw I find this confusing. Is this saying "no creativecommons license selected", "in copyright" or "undetermined"? Can we get more specific?

app/services/cocina/from_fedora/access.rb
```diff
@@ -20,7 +22,9 @@ def props
           access: access_rights,
           download: download? ? access_rights : 'none',
           readLocation: location,
-          license: license_uri
+          license: license_uri,
+          copyright: copyright,
```
jcoyne, line 26: @andrewjbtw please confirm that we want to allow collections to have copyright statements.  I'm not sure collections qualify as creative works that can be copyrighted.

mjgiarlo: @jcoyne I believe he confirmed this elsewhere.

app/services/cocina/to_fedora/access.rb
```diff
@@ -23,12 +23,42 @@ def apply
...
+        lookup_and_assign_license! if access.license
         item.rightsMetadata.ng_xml_will_change!
       end
 
       private
 
       attr_reader :item, :access
+
+      def lookup_and_assign_license! # rubocop:disable Metrics/AbcSize
+        if Dor::CreativeCommonsLicenseService.key?(license_code)
```
jcoyne, line 37: why bother with `license_code` in here and not lookup by the URI in`access.license`.  If we do this, then we can add the CC 4.0 licenses as something that can be round tripped.

jcoyne (changes requested):

app/services/cocina/to_fedora/access.rb
mjgiarlo: @jcoyne Currently we do look up the URI in `access.license` in `DefaultRights::LICENSE_CODES`, and then we use the license code to grab the label (and the URI but we can prolly drop that), and we also drop the license code into the Fedora XML.

I think the code as currently written on the ToFedora side has symmetry with the FromFedora side (both use the underlying Dor license services and associated lookup hashes), so I'm currently meh on making this change in this PR, anyway.

app/services/cocina/from_fedora/access.rb
jcoyne: Can you track that down?

mjgiarlo: @jcoyne no problem: https://github.com/sul-dlss/dor-services-app/issues/2623#issuecomment-811203233

PR #2634, opened 2021-03-31
allow objects that are currently opened to be accessioned properly

peetucket:
## Why was this change made?

~~HOLD:~~ Currently causing an integration tests to fail in QA (UPDATE: These tests all fail even with main branch of DSA deployed to QA, so likely unrelated to the work in this PR)

```
SDR_ENV=qa rspec ./spec/features/create_apo_spec.rb:19 # Use Argo to create an APO and verify new objects inherit it's rights is expected to have text "The following defaults will apply to all newly registered objects."
```
~~The pre-assembly spec is causing this exception with the legacy_metadata service: https://app.honeybadger.io/projects/50568/faults/78762991   .... assume because it is closing a version now via the new codepath?  But not sure why this is a problem.~~

-------
Came out of discussions on some work being done by John here: https://github.com/sul-dlss/sdr-api/pull/285 

The method we call to start accessioning (currently in pre-assembly and goobi, but soon to be sdr-api as well) handles the following two scenarios:

* A newly registered object that has not been accessioned
* An object that has been accessioned and is not yet opened

This change handles a potential third scenario:

* An object that has been accessioned and is in an open state

If not handled as in this PR, I believe objects in the last state will currently result in a version mismatch error, as we will start the accessionWF without closing them, which will end up trying to send an object to Preservation core with the same version as the last one sent.

NEWLY ADDED:

* if an object is currently being accessioned, return a 406 and do nothing

Added a new test

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -80,9 +80,12 @@ class ObjectsController < ApplicationController
   def accession
     workflow = params[:workflow] || default_start_accession_workflow
 
-    # if this is an existing versionable object, open and close it without starting accessioning
+    # if this is an existing versionable object, open and close it without starting accessionWF
     if VersionService.can_open?(@item, params)
       VersionService.open(@item, params, event_factory: EventFactory)
+      VersionService.close(@item, params.merge(start_accession: false), event_factory: EventFactory)
+    # if this is an existing accessioned object that is currently open, just close it without starting accessionWF
+    elsif VersionService.open_for_versioning?(@item, params)
       VersionService.close(@item, params.merge(start_accession: false), event_factory: EventFactory)
     end
 
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -91,7 +91,7 @@ class ObjectsController < ApplicationController
       VersionService.open(@item, params, event_factory: EventFactory)
       VersionService.close(@item, params.merge(start_accession: false), event_factory: EventFactory)
     # if this is an existing accessioned object that is currently open, just close it without starting accessionWF
-    elsif VersionService.open_for_versioning?(@item, params)
+    elsif VersionService.open?(@item)
       VersionService.close(@item, params.merge(start_accession: false), event_factory: EventFactory)
     end
 
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -80,6 +80,12 @@ class ObjectsController < ApplicationController
   def accession
     workflow = params[:workflow] || default_start_accession_workflow
 
+    # if this object is currently already in accessioning, we cannot start it again
+    if VersionService.in_accessioning?(@item)
+      head :not_acceptable
+      return
+    end
+
     # if this is an existing versionable object, open and close it without starting accessionWF
     if VersionService.can_open?(@item, params)
       VersionService.open(@item, params, event_factory: EventFactory)
--- a/spec/requests/start_accession_spec.rb
+++ b/spec/requests/start_accession_spec.rb
@@ -18,6 +18,7 @@ RSpec.describe 'Start Accession or Re-accession an object (with versioning)' do
 
   context 'when newly registered object that has not been accessioned yet' do
     before do
+      allow(VersionService).to receive(:in_accessioning?).and_return(false)
       allow(VersionService).to receive(:can_open?).and_return(false)
       allow(VersionService).to receive(:open_for_versioning?).and_return(false)
     end
@@ -47,6 +48,7 @@ RSpec.describe 'Start Accession or Re-accession an object (with versioning)' do
     let(:base_params) { { 'controller' => 'objects', 'action' => 'accession', 'id' => druid } }
 
     before do
+      allow(VersionService).to receive(:in_accessioning?).and_return(false)
       allow(VersionService).to receive(:can_open?).and_return(true)
     end
 
@@ -73,6 +75,7 @@ RSpec.describe 'Start Accession or Re-accession an object (with versioning)' do
     let(:base_params) { { 'controller' => 'objects', 'action' => 'accession', 'id' => druid } }
 
     before do
+      allow(VersionService).to receive(:in_accessioning?).and_return(false)
       allow(VersionService).to receive(:can_open?).and_return(false)
       allow(VersionService).to receive(:open_for_versioning?).and_return(true)
     end
@@ -83,6 +86,23 @@ RSpec.describe 'Start Accession or Re-accession an object (with versioning)' do
       expect(workflow_client).to have_received(:create_workflow_by_name).with(object.pid, default_start_accession_workflow, version: '1')
       expect(VersionService).not_to have_received(:open).with(object, base_params, event_factory)
       expect(VersionService).to have_received(:close).with(object, base_params.merge('start_accession' => false), event_factory)
+    end
+  end
+
+  context 'when object currently in accessioning' do
+    let(:base_params) { { 'controller' => 'objects', 'action' => 'accession', 'id' => druid } }
+
+    before do
+      allow(VersionService).to receive(:in_accessioning?).and_return(true)
+    end
+
+    it 'returns an unacceptable response and does not start any workflows' do
+      post "/v1/objects/#{druid}/accession",
+           headers: { 'Authorization' => "Bearer #{jwt}" }
+      expect(workflow_client).not_to have_received(:create_workflow_by_name)
+      expect(VersionService).not_to have_received(:open)
+      expect(VersionService).not_to have_received(:close)
+      expect(response).to have_http_status(:not_acceptable)
     end
   end
 end
--- a/spec/requests/start_accession_spec.rb
+++ b/spec/requests/start_accession_spec.rb
@@ -84,7 +84,7 @@ RSpec.describe 'Start Accession or Re-accession an object (with versioning)' do
       post "/v1/objects/#{druid}/accession",
            headers: { 'Authorization' => "Bearer #{jwt}" }
       expect(workflow_client).to have_received(:create_workflow_by_name).with(object.pid, default_start_accession_workflow, version: '1')
-      expect(VersionService).not_to have_received(:open).with(object, base_params, event_factory)
+      expect(VersionService).not_to have_received(:open)
       expect(VersionService).to have_received(:close).with(object, base_params.merge('start_accession' => false), event_factory)
     end
   end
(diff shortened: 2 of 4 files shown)


ndushay:
seems like functionality / concepts Andrew should weigh in on ...

ndushay (approved):
Aside from my concern that Andrew think through whether there are undesired side effects, the code looks fine.

jmartin-sul (approved):
LGTM also, agree andrew's sign-off would be good to have

app/controllers/objects_controller.rb
```diff
@@ -79,10 +79,13 @@ def show
...
 
-    # if this is an existing versionable object, open and close it without starting accessioning
+    # if this is an existing versionable object, open and close it without starting accessionWF
     if VersionService.can_open?(@item, params)
       VersionService.open(@item, params, event_factory: EventFactory)
       VersionService.close(@item, params.merge(start_accession: false), event_factory: EventFactory)
+    # if this is an existing accessioned object that is currently open, just close it without starting accessionWF
+    elsif VersionService.open_for_versioning?(@item, params)
+      VersionService.close(@item, params.merge(start_accession: false), event_factory: EventFactory)
     end
```
jmartin-sul, line 89: actually, question:  should there be a check that throws an error if it's not open, not openable, and not registered/v1?  that is, should it be possible to kick off accessioning on something that's not the initial v1 registration, and which is also already closed by some other process?  or am i misunderstanding the object state transitions?  how would the accessioned-to-an-existing-version content get preserved?

jmartin-sul (commented):
ok, actually, a question about whether there's another opportunity for preventing version mismatch errors from pres cat, which is possibly just my lack of understanding of SDR object states and versioning.

app/controllers/objects_controller.rb
jmartin-sul: i wouldn't hold up this PR over my question, since i think the check that was added makes sense, and stays the same regardless of the answer to my question (which i think results in no change if i'm off-base, and a check that throws an error if i did notice an opportunity for a preservation version mismatch problem).

peetucket: So let's think through the possibly object states when we reach this method:

- a newly registered object, never accessioned: `VersionService.can_open?` is false and `VersionService.open_for_versioning?` is also false so we just kick off accessioning.  This seems fine and should not generate any errors
- an object that is currently closed and ready to be opened: ``VersionService.can_open?` is true, so we will open/close the version and kick off accessioning
- an object that is currently opened (and therefore by definition has already been accessioned once): `VersionService.can_open?` is false and ` VersionService.open_for_versioning?` is true, so we close and then start accessioning (this is the work in this PR)

I don't know if its possible for something to be in a state of `not open, not openable, and not registered/v1` ... I think it may be possible for something to be openable but for the version number to not be correct for whatever reason.

andrewjbtw:
I'm pretty sure that Preassembly currently skips objects that aren't open and writes a message to the log that says something like "object skipped when object is open and version is greater than 1". Would this mean that Preassembly will start accessioning objects in this state? I believe preassembly is avoiding running on open objects because they may be open for another reason and running preassembly would then close them.

app/controllers/objects_controller.rb
andrewjbtw: `not open, not openable, and not registered/v1`

I believe this describes "in accessioning".

peetucket: Ahh, that's a good point and we aren't checking for that.  I'm going to propose we add another check and return a 406 unacceptable if an item is currently in accessioning.  This would prevent us from adding a new workflow if there is one that is in process (or stuck in an error state).  Thoughts?

peetucket:
Ahh yes, that is happening here: https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/digital_object.rb#L70

Basically, it won't bother calling this start accession method if we can't open it and its not an initial version (basically "in accessioning").  The same check I'd propose adding here, though the caller would have to handle the 406 method.  

There is no harm in leaving pre-assembly the way it is, it would just an extra check to try and prevent creating bad data by kicking off accessioning when it shouldn't be.



> I'm pretty sure that Preassembly currently skips objects that aren't open and writes a message to the log that says something like "object skipped when object is open and version is greater than 1". Would this mean that Preassembly will start accessioning objects in this state? I believe preassembly is avoiding running on open objects because they may be open for another reason and running preassembly would then close them.

peetucket:
Note that the current consumers of start accession API endpoint modified here are:

* pre-assembly (does it's own check to try and prevent starting accessioning on something that is in process)
* goobi 
* sdr-api (this is the recent work done by John)

I think it is acceptable for any of these apps to get a 406 if they try to start accessioning on an object that is already currently in accessioning.  But open to whether the extra check I just added is a good idea or not.

app/controllers/objects_controller.rb
jmartin-sul: @peetucket 👍 from me to that `406` idea.

i think i was vaguely imagining the version mismatch scenario ("it may be possible for something to be openable but for the version number to not be correct for whatever reason").  but @andrewjbtw's point about "in accessioning" is a good one, and makes for a more concrete/testable scenario.  that does seem like a possible race condition that could introduce versioning issues?

peetucket: It certainly seems plausible that a race condition could exist if you start accessioning on something already in progress, though the more likely situation is just an extra accessionWF on something that stopped mid-way due to an error (and will then error out again).  So I think this new check for "in accessioning" is not likely to prevent version mismatch errors, but seems reasonable to guard against anyway.

peetucket: new check for in accessioning added, along with a new test

jmartin-sul:
fwiw, it appears that the openability check does check to see whether something is in accessioning:

https://github.com/sul-dlss/dor-services-app/blob/5500c31d1c5ba4821559d7e4c2a7712c4a229b8a/app/services/version_service.rb#L107-L124

that said, i do like the specific error code for that situation, so no objection to @peetucket's latest proposed modification.

peetucket:
> fwiw, it appears that the openability check does check to see whether something is in accessioning:
> 
> https://github.com/sul-dlss/dor-services-app/blob/5500c31d1c5ba4821559d7e4c2a7712c4a229b8a/app/services/version_service.rb#L107-L124
> 
> that said, i do like the specific error code for that situation, so no objection to @peetucket's latest proposed modification.

Yes, I realized that too.  The problem is that if something is in accessioning and thus fail the openability check, that just means it won't be versioned... it will still have accessioning started.  So the new check is the only one that stops accessioning from being started at all.

jmartin-sul:
> > fwiw, it appears that the openability check does check to see whether something is in accessioning:
> > that said, i do like the specific error code for that situation, so no objection to @peetucket's latest proposed modification.
> 
> Yes, I realized that too. The problem is that if something is in accessioning and thus fails the openability check, that just means it won't be versioned... it will still have accessioning started. So the new check is the only one that stops accessioning from being started at all.

ah, yeah, good point.  so then i'm definitely in favor of the additional check.

spec/requests/start_accession_spec.rb
```diff
@@ -67,4 +70,39 @@
...
+      allow(VersionService).to receive(:in_accessioning?).and_return(false)
+      allow(VersionService).to receive(:can_open?).and_return(false)
+      allow(VersionService).to receive(:open_for_versioning?).and_return(true)
+    end
+
+    it 'closes a version and starts default workflow' do
+      post "/v1/objects/#{druid}/accession",
+           headers: { 'Authorization' => "Bearer #{jwt}" }
+      expect(workflow_client).to have_received(:create_workflow_by_name).with(object.pid, default_start_accession_workflow, version: '1')
+      expect(VersionService).not_to have_received(:open).with(object, base_params, event_factory)
```
jmartin-sul, line 87: i've seen rubocop warn about this before, but if you don't expect `open` to be called at all in this scenario, you might want to remove the `with(...)` qualifier from this expectation

jmartin-sul (changes requested):
one small testing quibble, otherwise looks good to me!

spec/requests/start_accession_spec.rb
peetucket: good call...fixed

jmartin-sul: thanks!

jmartin-sul (approved):

jmartin-sul:
@peetucket i wasn't sure if you wanted to deploy the branch to -qa or -stage to run integration tests (or have @andrewjbtw test) before merge to `main`, so i held off on merging myself.

peetucket:
> @peetucket i wasn't sure if you wanted to deploy the branch to -qa or -stage to run integration tests (or have @andrewjbtw test) before merge to `main`, so i held off on merging myself.

Yup, good idea.  I'll deploy this branch to qa and then run the integration tests there

andrewjbtw:
I'm not sure how to tell if there are going to be any side effects without actually deploying this, so the QA, test, merge route makes sense to me.

app/controllers/objects_controller.rb
```diff
@@ -80,10 +80,19 @@ def show
...
+      head :not_acceptable
+      return
+    end
+
+    # if this is an existing versionable object, open and close it without starting accessionWF
     if VersionService.can_open?(@item, params)
       VersionService.open(@item, params, event_factory: EventFactory)
       VersionService.close(@item, params.merge(start_accession: false), event_factory: EventFactory)
+    # if this is an existing accessioned object that is currently open, just close it without starting accessionWF
+    elsif VersionService.open_for_versioning?(@item, params)
```
mjgiarlo, line 94: I believe this line causes an exception: https://app.honeybadger.io/projects/50568/faults/78761944 

Should be `open?` instead of `open_for_versioning?`. Would be great for this `NoMethodError` to be raised in a test.

spec/requests/start_accession_spec.rb
```diff
@@ -16,9 +16,11 @@
     allow(WorkflowClientFactory).to receive(:build).and_return(workflow_client)
   end
 
-  context 'when new object' do
+  context 'when newly registered object that has not been accessioned yet' do
     before do
+      allow(VersionService).to receive(:in_accessioning?).and_return(false)
       allow(VersionService).to receive(:can_open?).and_return(false)
+      allow(VersionService).to receive(:open_for_versioning?).and_return(false)
```
mjgiarlo, line 23: This line is why we're not seeing the `NoMethodError` get raised.

app/controllers/objects_controller.rb
peetucket: i think in order to actually hit the `VersionService` in this test instead of just mocking the calls, we'd need to mock all of the actual underlying calls to the workflow service...but this was an unfortunate case of the actual bug slipping through the test

peetucket:
Releasing hold and I think it's ok to go.  It works fine with integration tests in QA (after a couple legit bug fixes pointed out by Mike).  Only one integration test is failing in QA, but it fails on stage and without my changes here as well in the same way.

PR #2635, opened 2021-03-31
New Unpublish Endpont

### LINKED ISSUE #2538: Add an endpoint to unpublish an object.
We have a "publish" action in Argo but not an unpublish action. The only way to unpublish seems to be to set rights to "dark" and then run the publish action, which is both counterintuitive and potentially risky because "dark" rights carry additional settings changes.

This is a problem when things are published accidentally as it's difficult to undo that action.

We could mitigate this problem by preventing publish in circumstances where it clearly should be prohibited, such as when an item is registered (see #2384).

jermnelson:
## Why was this change made?
Fixes #2538 

New tests for `UnpublishJob`, `UnpublishService`, and a new request spec.

n/a

### DIFF
--- a/app/services/unpublish_service.rb
+++ b/app/services/unpublish_service.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+# Unpublishes Druid from PURL
+class UnpublishService
+  def unpublish(druid:)
+    raise 'You have not configured purl-fetcher (Settings.purl_services_url).' unless Settings.purl_services_url
+
+    id = druid.gsub(/^druid:/, '')
+    Faraday.delete("#{Settings.purl_services_url}/purls/#{id}")
+  end
+end
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
 class ObjectsController < ApplicationController
   before_action :load_item, except: [:create]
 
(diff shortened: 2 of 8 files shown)


app/services/unpublish_service.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Unpublishes Druid from PURL
+class UnpublishService
+  def unpublish(druid:)
+    raise 'You have not configured perl-fetcher (Settings.purl_services_url).' unless Settings.purl_services_url
+
+    purl_druid = DruidTools::PurlDruid.new druid
+    PruneService.new(druid: purl_druid).prune!
+    id = druid.gsub(/^druid:/, '')
```
peetucket, line 10: Did we determine if this is actually an "unshelve", i.e. removing files from the stacks as opposed to the PURL page, and if so, is this desired via an Unpublish?  This would effectively make Unpublish have the same result as setting the rights to dark and re-accessioning.

peetucket:
Also wondering if we need to add a blue unpublish button for this action too on the item page.

app/services/unpublish_service.rb
jermnelson: Removed call to `PruneService` as per Andrew's comments during stand-up.

jermnelson:
> Also wondering if we need to add a blue unpublish button for this action too on the item page.

This will wait until the new Argo UI re-design.  If needed Andrew will be able to unpublish from the command-line.

peetucket (approved):

PR #2636, opened 2021-03-31
Ensure use and license fields are initialized before setting them

mjgiarlo:

## Why was this change made?

I honestly am not sure if we would ever see this in production but I have data that looks like this in Argo locally and I figured it was worth writing a new test and patching it in. Not sure what the harm is, despite the small cost of added code.

CI, local testing

None

### DIFF
--- a/app/services/cocina/to_fedora/access.rb
+++ b/app/services/cocina/to_fedora/access.rb
@@ -34,6 +34,8 @@ module Cocina
       attr_reader :item, :access
 
       def lookup_and_assign_license! # rubocop:disable Metrics/AbcSize
+        initialize_license_fields!
+
         if Dor::CreativeCommonsLicenseService.key?(license_code)
           item.rightsMetadata.creative_commons = license_code
           item.rightsMetadata.creative_commons.uri = access.license
@@ -58,6 +60,20 @@ module Cocina
         DefaultRights::LICENSE_CODES.merge(
           Cocina::FromFedora::Access::NONE_LICENSE_URI => 'none'
         )
+      end
+
+      def use_field
+        item.rightsMetadata.find_by_terms(:use).first # rubocop:disable Rails/DynamicFindBy
+      end
+
+      def initialize_field!(field_name, root_term = item.rightsMetadata.ng_xml.root)
+        item.rightsMetadata.add_child_node(root_term, field_name)
+      end
+
+      def initialize_license_fields!
+        initialize_field!(:use) if use_field.blank?
+        initialize_field!(:creative_commons, use_field) if item.rightsMetadata.creative_commons.blank?
+        initialize_field!(:open_data_commons, use_field) if item.rightsMetadata.open_data_commons.blank?
       end
     end
   end
--- a/app/services/cocina/to_fedora/dro_access.rb
+++ b/app/services/cocina/to_fedora/dro_access.rb
@@ -7,9 +7,6 @@ module Cocina
     class DROAccess < Access
       def apply
         create_embargo(access.embargo) if access.embargo
-        item.rightsMetadata.copyright = access.copyright if access.copyright
-        item.rightsMetadata.use_statement = access.useAndReproductionStatement if access.useAndReproductionStatement
-        lookup_and_assign_license! if access.license
 
         super
       end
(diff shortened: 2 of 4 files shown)


app/services/cocina/to_fedora/access.rb
```diff
@@ -59,6 +61,20 @@ def license_codes
...
+      end
+
+      def initialize_field!(field_name, root_term = item.rightsMetadata.ng_xml.root)
+        item.rightsMetadata.add_child_node(root_term, field_name)
+      end
+
+      def initialize_license_fields!
+        initialize_field!(:use) if use_field.blank?
+        initialize_field!(:creative_commons, use_field) if item.rightsMetadata.creative_commons.blank?
+        initialize_field!(:open_data_commons, use_field) if item.rightsMetadata.open_data_commons.blank?
```
mjgiarlo, line 76: These are the lines needed to make sure the rights metadata XML is sparse such as the example in the test below: https://github.com/sul-dlss/dor-services-app/pull/2636/files#diff-6a34c5e09cfb46d3721355fa41db34f7bb901040af862082fd5b78a7e4e38e6bR26-R38

app/services/cocina/to_fedora/access.rb
```diff
@@ -59,6 +61,20 @@ def license_codes
           Cocina::FromFedora::Access::NONE_LICENSE_URI => 'none'
         )
       end
+
+      def use_field
+        item.rightsMetadata.find_by_terms(:use).first # rubocop:disable Rails/DynamicFindBy
```
mjgiarlo, line 66: This is a Rubocop false positive.

app/services/cocina/to_fedora/dro_access.rb
```diff
@@ -7,9 +7,6 @@ module ToFedora
     class DROAccess < Access
       def apply
         create_embargo(access.embargo) if access.embargo
-        item.rightsMetadata.copyright = access.copyright if access.copyright
-        item.rightsMetadata.use_statement = access.useAndReproductionStatement if access.useAndReproductionStatement
-        lookup_and_assign_license! if access.license
```
mjgiarlo, line 12: These same lines are called by the `ToFedora::Access` class when `super` is called a couple lines down, so remove the duplicate code.

config/initializers/rights_metadata_monkeypatch.rb
```diff
@@ -0,0 +1,20 @@
...
+      xml.human(type: 'creativeCommons')
+      xml.machine(type: 'creativeCommons', uri: '')
+    end
+
+    define_template :open_data_commons do |xml|
+      xml.human(type: 'openDataCommons')
+      xml.machine(type: 'openDataCommons', uri: '')
+    end
+
+    define_template(:use, &:use)
```
mjgiarlo, line 18: Since we endeavor to replace the dor-services gem, ultimately, I thought it defensible to patch this into DSA.

spec/services/cocina/to_fedora/dro_access_spec.rb
```diff
@@ -9,7 +9,67 @@
...
+            <machine>
+              <none/>
+            </machine>
+          </access>
+          <access type="read">
+            <machine>
+              <none/>
+            </machine>
+          </access>
+        </rightsMetadata>
```
mjgiarlo, line 38: Here's the sparse rights MD case I'm seeing locally.

jcoyne: I'm struggling to see why this is necessary, isn't this provided by https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb#L55-L61? You can see it if you do: 

```ruby
puts Dor::AdminPolicyObject.new.rightsMetadata.ng_xml.to_s
```

mjgiarlo: @jcoyne I don't know, but I am seeing it locally.

mjgiarlo: @jcoyne Happy to pair to show you what I'm seeing in a bit

mjgiarlo: @jcoyne ready for another look

jcoyne (approved):

PR #2639, opened 2021-04-01
When setting the license, make sure to blank existing licenses

mjgiarlo:

## Why was this change made?

Without this change, if I have (e.g.) an ODC license set and then set a CC license, the item or collection will now have multiple licenses.

CI

None

### DIFF
--- a/app/services/cocina/to_fedora/access.rb
+++ b/app/services/cocina/to_fedora/access.rb
@@ -38,13 +38,22 @@ module Cocina
           item.rightsMetadata.creative_commons = license_code
           item.rightsMetadata.creative_commons.uri = access.license
           item.rightsMetadata.creative_commons_human = Dor::CreativeCommonsLicenseService.property(license_code).label
+          item.rightsMetadata.open_data_commons = ''
+          item.rightsMetadata.open_data_commons.uri = ''
+          item.rightsMetadata.open_data_commons_human = ''
         elsif Dor::OpenDataLicenseService.key?(license_code)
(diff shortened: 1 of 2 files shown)


app/services/cocina/to_fedora/access.rb
```diff
@@ -38,13 +38,22 @@ def lookup_and_assign_license! # rubocop:disable Metrics/AbcSize
...
           item.rightsMetadata.open_data_commons_human = Dor::OpenDataLicenseService.property(license_code).label
+          item.rightsMetadata.creative_commons = ''
+          item.rightsMetadata.creative_commons.uri = ''
+          item.rightsMetadata.creative_commons_human = ''
         elsif license_code == 'none'
           item.rightsMetadata.creative_commons = license_code
           item.rightsMetadata.creative_commons_human = 'no Creative Commons (CC) license'
+          item.rightsMetadata.open_data_commons = ''
+          item.rightsMetadata.open_data_commons.uri = ''
+          item.rightsMetadata.open_data_commons_human = ''
```
ndushay, line 56: my, isn't this crufty.  Perhaps a case statement?  Nothing to hold up the PR.

ndushay (approved):

app/services/cocina/to_fedora/access.rb
mjgiarlo: In this case since we have at least a different comparisons for the conditions, I think a `case` statement might even make this worse. But, yes, rights metadata is a `~ m e s s ~`.

PR #2646, opened 2021-04-02
do not type type="parallel" in multiscript titles

### LINKED ISSUE #2528: Implement mappings for multiscript titles converted from MARC
See #2527 (PR that added xit specs)  (and also #2468 - issue for parallel titles)

To spin up on mapping tickets, Naomi or JLitt. can help w context that might be necessary.

Arcadia says:

> 2528 is when a record has multiple titleInfo elements with the same altRepGroup and different/no types (specifically the output generated when mapping from marc), and in theory represents the situation where the record includes the title in original script transcribed from the resource and in translation/transliteration created by the cataloger.
>
> 2468 is when all titleInfo elements with the same altRepGroup have type="translated". it represents the situation where multiple titles in different scripts/languages are transcribed directly from the resource by the cataloger - i.e., the resource itself has titles in multiple scripts/languages on itself and all are considered of equal importance.
>
> some (but not all) of the titles included in 2528 should actually be 2468 titles, but it's nontrivial to determine that from the marc via the xslt so the mods doesn't treat them that way. so they're linked intellectually but not necessarily in the way they're implemented.

peetucket:
## Why was this change made?

Fixes #2528 - extraneous type="parallel" node in some mappings

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/titles.rb
+++ b/app/services/cocina/from_fedora/descriptive/titles.rb
@@ -56,7 +56,7 @@ module Cocina
             parallelValue: simple_or_structured(node_set, display_types: display_types?(node_set))
           }.tap do |result|
             type = parallel_type(node_set)
-            result[:type] = type if type
+            result[:type] = type if type && type != 'parallel'
           end
         end
 
--- a/spec/services/cocina/mapping/descriptive/mods/subject_title_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/mods/subject_title_spec.rb
@@ -153,8 +153,9 @@ RSpec.describe 'MODS subject topic <--> cocina mappings' do
   end
 
   describe 'altRepGroup for alternative title (880-246)' do
-    # based on druid:cp165bv2167
-    xit 'not implemented' do
+    it_behaves_like 'MODS cocina mapping' do
+      # based on druid:cp165bv2167
+
       let(:mods) do
         <<~XML
           <titleInfo usage="primary">
@@ -166,10 +167,10 @@ RSpec.describe 'MODS subject topic <--> cocina mappings' do
           <titleInfo type="translated">
              <title>Contemporary Chinese literature from Taiwan, &lt;2005-&gt;</title>
           </titleInfo>
-          <titleInfo type="translated" altRepGroup="01">
+          <titleInfo type="translated" altRepGroup="1">
              <title>Dang dai Taiwan wen xue xuan yi &lt;2002-&gt;</title>
           </titleInfo>
-          <titleInfo type="alternative" altRepGroup="01">
+          <titleInfo type="alternative" altRepGroup="1">
              <title>當代台灣文學選譯 &lt;2002-&gt;</title>
           </titleInfo>
         XML


spec/services/cocina/mapping/descriptive/mods/subject_title_spec.rb
```diff
@@ -167,10 +167,10 @@
           <titleInfo type="translated">
              <title>Contemporary Chinese literature from Taiwan, &lt;2005-&gt;</title>
           </titleInfo>
-          <titleInfo type="translated" altRepGroup="01">
```
peetucket, line 170: It looks like the "01" gets round tripped as "1" causing the spec to fail unless you start it out as "1"?   Or am I misunderstanding how the specs work?  Is that a problem?

spec/services/cocina/mapping/descriptive/mods/subject_title_spec.rb
```diff
@@ -153,8 +153,8 @@
   end
 
   describe 'altRepGroup for alternative title (880-246)' do
-    xit 'not implemented' do
-      let(:druid) { 'druid:cp165bv2167' }
+    it_behaves_like 'MODS cocina mapping' do
+      # based on oo000oo0000
```
jcoyne, line 157: Really?

spec/services/cocina/mapping/descriptive/mods/subject_title_spec.rb
jcoyne: I believe the important part is that the values for altRepGroup are internally consistent. It doesn't matter that they vary in the round trip. I believe there was part of the normalizer that deals with that.

spec/services/cocina/mapping/descriptive/mods/subject_title_spec.rb
peetucket: That's what @ndushay suggested I add as a comment. I didn't investigate why. Am happy to defer to Naomi and Justin L as to utility.

peetucket: Never mind, I presume I was supposed to put the actual druid that arcadia user in the comment, which I just did

spec/services/cocina/mapping/descriptive/mods/subject_title_spec.rb
ndushay: I have been encouraged by JLitt to change the examples from Arcadia to use "1" instead of "01" for altRepGroup as it will all be handled appropriately in production.

app/services/cocina/from_fedora/descriptive/titles.rb
```diff
@@ -56,7 +56,7 @@ def parallel(node_set)
             parallelValue: simple_or_structured(node_set, display_types: display_types?(node_set))
           }.tap do |result|
             type = parallel_type(node_set)
-            result[:type] = type if type
+            result[:type] = type if type && type != 'parallel'
```
ndushay, line 59: `if type.present?` -- didn't look at the parallel_type code to see if it's needed.

ndushay (approved):
approving, modulo a question.

app/services/cocina/from_fedora/descriptive/titles.rb
peetucket: Not sure I understand...

ndushay: just wondering if type can be an empty string.

peetucket: oh i see -- it appears in this particular case that `type` will never be an empty string: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/from_fedora/descriptive/titles.rb#L70-L77

PR #2648, opened 2021-04-05
Create a license node for storing the license URI

### LINKED ISSUE #2592: Read and write licenses from new XML node
In order to support licenses that are neither creative commons or open data. We will be adding a new node to the rightsMetadata xml:

```xml
<rightsMetadata>
  <use>
    <license>https://creativecommons.org/licenses/by-nc/3.0/</license>
  </use>
 </rightsMetadata>
```

While we are making this the canonical record, we should fall back to reading and writing from the old `<machine>` nodes if possible. 

Note that some items have `<use><machine uri="https://creativecommons.org/licenses/by-nc/4.0/">by-nc</machine></use>` so we should look for the `uri` attribute rather than the text content.

### LINKED ISSUE #2628: Map existing license data from the URI if there is one, not from the text abbreviation
SDR has content where people have entered specific URIs in the rightsMetadata in order to give content a license that has not been implemented in the license menus on the APO. When mapping this data to Cocina, we should respect the URI and map it as entered, provided it's a valid URI.

Example: https://argo.stanford.edu/view/druid:bc677qs7597 has a 4.0 Creative Commons license

Fedora rightsMetadata:
```xml
<use>
    <human type="creativeCommons">CC-BY SA 4.0</human>
    <machine type="creativeCommons" uri="https://creativecommons.org/licenses/by-sa/4.0/">by-sa</machine>
    <human type="useAndReproduction">This material may be shared or adapted provided appropriate credit is given to the originator, and the same license is used.</human>
  </use>
```
Currently this is mapped to Cocina as a 3.0 CC license, which it clearly should not have:

```json
{"access": {
        "access": "world",
        "license": "https://creativecommons.org/licenses/by-sa/3.0/"
    }}
```

I'm guessing this mapping is based on the text "by-sa" in the machine tag, which does not express a version. Since our data may or may not have a URI along with the text abbreviation for the license, our mapping should do the following:

1. If there is a URI, follow the URI
2. If there is no URI, go with what is in the Argo license menu on the assumption that this is what the user would have chosen.

jcoyne:
## Why was this change made?

This allows us to accept more licenses including the CC 4.0 licenses.  This still preserves how we publish the data to PURL, so although we can now accept all the licenses that H2 accepts, PURL will not yet know how to display them all.

Fixes #2628
Fixes #2592

Once this is merged and deployed we should be able to remediate: https://argo.stanford.edu/view/druid:qk286xn8411

### DIFF
--- a/app/services/cocina/from_fedora/access.rb
+++ b/app/services/cocina/from_fedora/access.rb
@@ -4,8 +4,6 @@ module Cocina
   module FromFedora
     # builds the Access subschema for Collections
     class Access
-      NONE_LICENSE_URI = 'http://cocina.sul.stanford.edu/licenses/none'
-
       def self.collection_props(rights_metadata_ds)
         props = new(rights_metadata_ds).props
         # Collection access does not have download
@@ -22,7 +20,7 @@ module Cocina
           access: access_rights,
           download: download? ? access_rights : 'none',
           readLocation: location,
-          license: license_uri,
+          license: License.find(rights_metadata_ds),
           copyright: copyright,
           useAndReproductionStatement: use_statement
         }.compact.tap do |h|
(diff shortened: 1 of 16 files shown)


mjgiarlo:
Do you think this will affect/snag https://github.com/sul-dlss/argo/pull/2629 ?

jcoyne:
@mjgiarlo it shouldn't affect anything that is using the API.

justinlittman (approved):

PR #2653, opened 2021-04-06
Do not roundtrip parts.

justinlittman:
## Why was this change made?
They don't roundtrip, so should be excluded from validation.

sdr-deploy

NA


ndushay (approved):
would this benefit from a test?

justinlittman:
There aren't currently any tests for the tooling. We can discuss as a separate ticket, but not sure it is worth the work.

PR #2654, opened 2021-04-06
Remove blanks when validating descriptive roundtripping.

### LINKED ISSUE #2621: normalization to address Cocina::RoundtripValidationError in objects_creator for H2
See https://app.honeybadger.io/projects/50568/faults/78609501

This is from turning roundtrip validation on in the qa environment for H2 and running the h2 integration spec.  

The error is "Cocina::RoundtripValidationError: Roundtripping of descriptive metadata unsuccessful. Expected {:title=>[{:value=>"flory"}], :relatedResource=>[]} but received {:title=>[{:value=>"flory"}]}"

Not sure if this is an H2 thing or a dor-services-app thing.

justinlittman:
closes #2621

## Why was this change made?
Cocina allows empty arrays, which can confuse roundtripping.

Unit

NA

### DIFF
--- a/app/validators/cocina/description_roundtrip_validator.rb
+++ b/app/validators/cocina/description_roundtrip_validator.rb
@@ -21,8 +21,11 @@ module Cocina
       title_builder = FromFedora::Descriptive::TitleBuilderStrategy.find(label: cocina_object.label)
       roundtrip_description = FromFedora::Descriptive.props(title_builder: title_builder, mods: descriptive_ng_xml, druid: druid)
 
+      # With Rails 6.1, .delete_if can be replace by compact_blank (for deleting empty array).
+      orig_description = cocina_object.description.to_h.delete_if { |_k, v| v.blank? }
+
       # Compare original description against roundtripped Cocina.
-      unless DeepEqual.match?(roundtrip_description, cocina_object.description.to_h)
+      unless DeepEqual.match?(roundtrip_description, orig_description)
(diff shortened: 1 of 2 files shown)


app/validators/cocina/description_roundtrip_validator.rb
```diff
@@ -21,8 +21,10 @@ def self.valid_from_cocina?(cocina_object)
       title_builder = FromFedora::Descriptive::TitleBuilderStrategy.find(label: cocina_object.label)
       roundtrip_description = FromFedora::Descriptive.props(title_builder: title_builder, mods: descriptive_ng_xml, druid: druid)
 
+      orig_description = cocina_object.description.to_h.delete_if { |_k, v| v.blank? }
```
jcoyne, line 24: Coul this use `cocina_object.description.to_h.compact`?

jcoyne (changes requested):

app/validators/cocina/description_roundtrip_validator.rb
justinlittman: `compact` removes `nil`. This needs to remove `[]`. `compact_blank` would work if we were running newer rails.

jcoyne: Would you mind dropping a comment about that and then it's good to merge.

jcoyne (approved):

app/validators/cocina/description_roundtrip_validator.rb
justinlittman: Done.

Issue #2659, opened 2021-04-07
map the APO referencesAgreement to Cocina


andrewjbtw:
In https://github.com/sul-dlss/cocina-models/pull/234 we added `referencesAgreement` to Cocina to store the druid for the agreement that each APO references (in Fedora this is in the APO RELS-EXT). But we don't seem to be mapping this data to Cocina. We need to map this data so we don't leave it behind.
 
This APO shows no agreement in Cocina: https://argo.stanford.edu/view/druid:bz845pv2292

Issue #2644, opened 2021-04-02
Repeated edits to an APO create a lot of XML nodes


andrewjbtw:
**Describe the bug**
Repeatedly updating an APO (using https://github.com/sul-dlss/argo/pull/2529) is adding a lot of XML nodes to the administrativeMetadata. 

The administrativeMetadata for https://argo-qa.stanford.edu/view/druid:vr838wn4827 looks like this after many rounds of changed:

```xml
<?xml version="1.0"?>
<administrativeMetadata>
  <dissemination>
    
  </dissemination>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination/>
  <registration/>
  <dissemination>
    <workflow id=""/>
  </dissemination>
  <registration>
    <workflow id="accessionWF"/>
    <workflow id="goobiWF"/>
    <workflow id="registrationWF"/>
    <collection id="druid:ns375cy3379"/>
  </registration>
</administrativeMetadata>
```

**User Impact**
It's kind of a mess but doesn't seem to hurt functionality.

jcoyne:
This involves fixing this todo: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/to_fedora/apo_rights.rb#L15

ndushay:
See https://github.com/sul-dlss/dor-services-app/pull/2677/files#r612116570 -- it may be there's a more appropriate solution to ensure that there is always exactly one dissemination node, which may be empty ...    just not MORE than one.  Diddot for registration nodes

PR #2661, opened 2021-04-08
notify if names have same altRepGroup but differ in type

jmartin-sul:
## Why was this change made?

fixes #2456

## How was this change tested?

unit tests, roundtrip validation on sdr-deploy (as expected, numbers remain unchanged, since this only adds notification for bad data):
```
[deploy@sdr-deploy dor-services-app]$ bin/validate-desc-cocina-roundtrip -s 100000 
Testing |Time: 00:10:27 | ================================================================================== | Time: 00:10:27
Status (n=100000; not using Missing for success/different/error stats):
  Success:   99346 (99.776%)
  Different: 223 (0.224%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
  MODS normalizer error:     0 (0.0%)
  Missing (no descMetadata):     431 (0.431%)
[deploy@sdr-deploy dor-services-app]$ 
[deploy@sdr-deploy dor-services-app]$ git checkout 2456-notify-on-names-same-altRepGroup-diff-type
Branch '2456-notify-on-names-same-altRepGroup-diff-type' set up to track remote branch '2456-notify-on-names-same-altRepGroup-diff-type' from 'origin'.
Switched to a new branch '2456-notify-on-names-same-altRepGroup-diff-type'
[deploy@sdr-deploy dor-services-app]$ 
[deploy@sdr-deploy dor-services-app]$ bin/validate-desc-cocina-roundtrip -s 100000 
Testing |Time: 00:05:00 | ================================================================================== | Time: 00:05:00
Status (n=100000; not using Missing for success/different/error stats):
  Success:   99346 (99.776%)
  Different: 223 (0.224%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
  MODS normalizer error:     0 (0.0%)
  Missing (no descMetadata):     431 (0.431%)
[deploy@sdr-deploy dor-services-app]$ 
```

## Which documentation and/or configurations were updated?

n/a

Issue #2456, opened 2021-03-09
notify on inconsistent types for names in same altRepGroup [was: Don't normalize name type="corporate" to name type="conference"]


arcadiafalcone:
mx303pc7527

jmartin-sul:
@arcadiafalcone, sorry this got tied up in progress without actually moving forward -- i picked this up, got pulled back to working on an H2 feature, had a busy first responder week, and then got pulled back to finishing that H2 feature.

I started back in on this on Friday, and after getting a bit stuck, paired with @ndushay this afternoon.  We were both a bit unsure how to proceed since we weren't sure what exactly the desired Cocina output is in this case, and since the druid mentioned in the issue description has been remediated to no longer have this issue (maybe we think).

I did have a copy of the old problematic desc metadata fragments that were causing trouble, and it seemed to me that the issue only occurs when there are two different `name` elements, one with `type="conference"` and one with `type="corporate"`, e.g.:

```xml
<name type="conference" altRepGroup="01" script="Latn">
  <namePart>Zhonghua gong nong bing Suwei'ai quan guo dai biao da hui 1931 : Ruijin)</namePart>
</name>
<name type="corporate" altRepGroup="01" script="Latn">
<!-- snip some other desc metadata in between -->
<namePart>&#x4E2D;&#x83EF;&#x5DE5;&#x8FB2;&#x5175;&#x8607;&#x7DAD;&#x57C3;&#x5168;&#x570B;&#x4EE3;&#x8868;&#x5927;&#x6703;</namePart>
  <namePart>(1st : 1931 : Ruijin)</namePart>
</name>
```

If there is only a `name type="corporate"`, that seems to work fine, assuming the Cocina `type` in that case should be `'organization'`.

So, questions:

1. Have you seen cases where the Cocina `type` is incorrect when only one `name` element is in play?
2. Is there other data like this in production, now that mx303pc7527 seems to have been remediated?  Is mx303pc7527 still producing incorrect contributor Cocina?
3. I'm pretty unfamiliar with `altRepGroup` usage (Naomi gave me a short primer this afternoon), but we were wondering whether the `name` elements in this example's original desc metadata were meant to have a structured relationship?
4. If there is still problematic data like this in production, would it be acceptable to alert and to keep producing the same Cocina output we do now?
5. If no, and we should fix the Cocina output, what's the desired JSON look like?  Before pairing this afternoon, I'd had something like:

```ruby
      let(:cocina) do
        {
[...]

PR #2677, opened 2021-04-13
apo adminMetadata mapping: roundtrip spec AND only create disseminationWF if there is one

ndushay:
## Why was this change made?

part of #2704

- Adds roundtripping apo adminstrativeMetadata spec before working on ticket #2659
- doesn't create empty dissemination elements in APO administrativeMetaadata (part of #2644)

- it's a spec, and a fix added because the spec illustrated it.

### DIFF
--- a/spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
+++ b/spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
@@ -1,0 +1,561 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.shared_examples 'APO Fedora Cocina mapping' do
+  # Required: admin_metadata_xml, default_object_rights_xml, role_metadata_xml, rels_ext, cocina
+  # Optional: roundtrip_admin_metadata_xml, roundtrip_default_object_rights_xml
+
+  let(:apo_druid) { 'apo_druid' }
+  let(:apo_label) { 'apo_label' }
+  let(:ur_apo_druid) { 'druid:hv992ry2431' }
+  let(:orig_fedora_apo_mock) do
+    # need to mock to avoid call to Solr
+    instance_double(
+      Dor::AdminPolicyObject,
+      pid: apo_druid,
+      label: apo_label,
+      current_version: '1',
+      admin_policy_object_id: ur_apo_druid,
+      administrativeMetadata: Dor::AdministrativeMetadataDS.from_xml(admin_metadata_xml),
+      descMetadata: Dor::DescMetadataDS.from_xml('<mods/>'),
+      # NOTE: DefaultObjectRights doesn't work with :from_xml
+      defaultObjectRights: instance_double(Dor::DefaultObjectRightsDS, content: default_object_rights_xml),
+      roleMetadata: Dor::RoleMetadataDS.from_xml(role_metadata_xml)
+    )
+  end
+  let(:actual_cocina_props) { Cocina::FromFedora::APO.props(orig_fedora_apo_mock) }
+  let(:expected_cocina_props) do
+    {
+      administrative: cocina.deep_dup,
+      externalIdentifier: apo_druid,
+      type: Cocina::Models::Vocab.admin_policy,
+      label: apo_label,
+      version: 1,
+      description: {
+        purl: "#{Settings.release.purl_base_url}/apo_druid",
+        access: {
+          digitalRepository: [
+            {
+              value: 'Stanford Digital Repository'
+            }
+          ]
+        }
+      }
+    }
+  end
+
+  context 'when mapping from Fedora to Cocina' do
+    it 'cocina hash produces valid Cocina AdminPolicyAdministrative' do
+      expect { Cocina::Models::AdminPolicyAdministrative.new(cocina) }.not_to raise_error
+    end
+
+    it 'Fedora maps to expected Cocina object props' do
+      expect(actual_cocina_props).to be_deep_equal(expected_cocina_props)
+    end
+  end
+
+  context 'when mapping from Cocina to Fedora' do
+    let(:actual_cocina_apo_admin) { Cocina::Models::AdminPolicyAdministrative.new(cocina) }
+    let(:roundtrip_fedora_apo) do
+      Dor::AdminPolicyObject.new(pid: apo_druid,
+                                 admin_policy_object_id: actual_cocina_apo_admin.hasAdminPolicy,
+                                 label: apo_label)
+    end
+
+    describe 'Cocina::ToFedora::AdministrativeMetadata' do
+      let(:actual_admin_metadata_xml) do
+        Cocina::ToFedora::AdministrativeMetadata.write(roundtrip_fedora_apo.administrativeMetadata, actual_cocina_apo_admin)
+        roundtrip_fedora_apo.administrativeMetadata.to_xml
+      end
+
+      it 'AdminPolicyAdministrative cocina model roundtrips to original administrativeMetadata.xml' do
+        expect(actual_admin_metadata_xml).to be_equivalent_to(admin_metadata_xml)
+      end
+    end
+
+    describe 'Cocina::ToFedora::DefaultRights' do
+      let(:actual_default_rights_xml) do
+        Cocina::ToFedora::DefaultRights.write(roundtrip_fedora_apo.defaultObjectRights, actual_cocina_apo_admin.defaultAccess) if actual_cocina_props[:administrative][:defaultAccess]
+        roundtrip_fedora_apo.defaultObjectRights.to_xml
+      end
+
+      it 'AdminPolicyAdministrative cocina model roundtrips to original defaultObjectRights.xml' do
+        expect(actual_default_rights_xml).to be_equivalent_to(default_object_rights_xml)
+      end
+    end
+
+    describe 'Cocina::ToFedora::Roles' do
+      let(:actual_role_xml) do
+        Cocina::ToFedora::Roles.write(roundtrip_fedora_apo, actual_cocina_apo_admin.roles)
+        roundtrip_fedora_apo.roleMetadata.to_xml
+      end
+
+      it 'AdminPolicyAdministrative cocina model roundtrips to original roleMetadata.xml' do
+        expect(actual_role_xml).to be_equivalent_to(role_metadata_xml)
+      end
+    end
+  end
+
+  context 'when mapping from roundtrip Fedora to Cocina' do
+    let(:my_roundtrip_admin_metadata_xml) do
+      defined?(roundtrip_admin_metadata_xml) ? roundtrip_admin_metadata_xml : admin_metadata_xml
+    end
+    let(:my_roundtrip_default_object_rights_xml) do
+      defined?(roundtrip_default_object_rights_xml) ? roundtrip_default_object_rights_xml : default_object_rights_xml
+    end
+    let(:roundtrip_fedora_apo_mock) do
+      # need to mock to avoid call to Solr
+      instance_double(
+        Dor::AdminPolicyObject,
+        pid: apo_druid,
+        label: apo_label,
+        current_version: '1',
+        admin_policy_object_id: ur_apo_druid,
+        administrativeMetadata: Dor::AdministrativeMetadataDS.from_xml(my_roundtrip_admin_metadata_xml),
+        descMetadata: Dor::DescMetadataDS.from_xml('<mods/>'),
+        # NOTE: DefaultObjectRights doesn't work with :from_xml
+        defaultObjectRights: instance_double(Dor::DefaultObjectRightsDS, content: my_roundtrip_default_object_rights_xml),
+        roleMetadata: Dor::RoleMetadataDS.from_xml(role_metadata_xml)
+      )
+    end
+    let(:roundtrip_cocina_props) { Cocina::FromFedora::APO.props(roundtrip_fedora_apo_mock) }
+
+    it 'roundtrip Fedora maps to expected Cocina object props' do
+      expect(roundtrip_cocina_props).to be_deep_equal(expected_cocina_props)
+    end
+  end
+end
+
+RSpec.describe 'Fedora APO objects <--> cocina administrative mappings' do
+  # NOTE:  some APO have descMetadata, contact, accessioning (WF) in the administrativeMetadata.  Ignoring for now
+  describe 'world access, registration workflows, collection, single role' do
+    # from bz845pv2292
+    it_behaves_like 'APO Fedora Cocina mapping' do
+      let(:admin_metadata_xml) do
+        <<~XML
+          <administrativeMetadata>
+            <registration>
+              <workflow id="goobiWF"/>
+              <workflow id="registrationWF"/>
+              <collection id="druid:ny719df8518"/>
+            </registration>
+          </administrativeMetadata>
+        XML
+      end
+      let(:default_object_rights_xml) do
+        <<~XML
+          <rightsMetadata>
+            <access type="discover">
+              <machine>
+                <world/>
+              </machine>
+            </access>
+            <access type="read">
+              <machine>
+                <world/>
+              </machine>
+            </access>
+            <use>
+              <human type="useAndReproduction">Blah blah</human>
+            </use>
+          </rightsMetadata>
+        XML
+      end
+      let(:role_metadata_xml) do
+        <<~XML
+          <roleMetadata>
+            <role type="dor-apo-manager">
+              <group>
+                <identifier type="workgroup">sdr:developer</identifier>
+              </group>
+            </role>
+          </roleMetadata>
+        XML
+      end
+      let(:rels_ext) do
+        <<~XML
+          <rdf:RDF xmlns:fedora-model="info:fedora/fedora-system:def/model#" xmlns:hydra="http://projecthydra.org/ns/relations#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
+            <rdf:Description rdf:about="info:fedora/druid:bz845pv2292">
+              <hydra:isGovernedBy rdf:resource="info:fedora/druid:hv992ry2431"/>
+              <hydra:referencesAgreement rdf:resource="info:fedora/druid:yr951qr4199"/>
+              <fedora-model:hasModel rdf:resource="info:fedora/afmodel:Dor_AdminPolicyObject"/>
+            </rdf:Description>
+          </rdf:RDF>
+        XML
+      end
+      let(:cocina) do
+        {
+          defaultObjectRights: default_object_rights_xml,
+          defaultAccess: {
+            access: 'world',
+            download: 'world',
+            useAndReproductionStatement: 'Blah blah'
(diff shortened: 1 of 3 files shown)


app/services/cocina/to_fedora/apo_rights.rb
```diff
@@ -11,8 +11,7 @@ def self.write(administrative_metadata, administrative)
         # Clear out the nodes we are updating
         admin_node.xpath('registration/workflow|registration/collection|dissemination/workflow').each(&:remove)
 
-        # TODO: need to see if this node already exists
-        admin_node.add_child "<dissemination><workflow id=\"#{administrative.disseminationWorkflow}\" /></dissemination>"
+        admin_node.add_child "<dissemination><workflow id=\"#{administrative.disseminationWorkflow}\" /></dissemination>" if administrative.disseminationWorkflow
```
ndushay, line 14: this is a partial fix for #2644, and was discovered with the new specs.   whee!

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+          },
+          registrationWorkflow: [
+            'goobiWF',
+            'registrationWF'
+          ],
+          collectionsForRegistration: [
+            'druid:ny719df8518'
+          ],
+          hasAdminPolicy: ur_apo_druid,
+          # referencesAgreement: 'druid:yr951qr4199',
```
ndushay, line 171: this is issue #2659, which I will work on as soon as this PR is merged.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+    # <descMetadata>
+    #   <format>MODS</format>
+    #   <source>Symphony</source>
+    # </descMetadata>
+    # or empty <registration/> ...
+
+    # TODO: defaultAccess:  readLocation; controlleddigitallending
+    # TODO: test without registrationWF ?
+    # TODO: test without admin policy (? - should throw error?)
+    # TODO: test without roles ?
```
ndushay, line 527: @aaron-collier, @justinlittman some tests/normalization to work on for clean APO roundtripping.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+          ]
+        }
+      end
+    end
+  end
+
+  describe 'no collections, license, roles with type sunetid' do
+    # based on kt538yv1733 combined with default_obj_rights and roles from qv549bf9093
+    # it_behaves_like 'APO from Fedora to Cocina mapping' do
+    xit 'problems with empty name elements in role/person, with sunetid members, with license (roundtrips as license url)' do
```
ndushay, line 389: @aaron-collier, @justinlittman a test to work on for clean APO roundtripping.

app/services/cocina/to_fedora/apo_rights.rb
jcoyne: If we do the full fix for for #2644, will this check be necessary?  I rather like the explicitness of having a `<dissemination>` node for every APO.  This shows that this is a configuration option that belongs to an APO and that in this case it has zero workflows in it.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.shared_examples 'APO from Fedora to Cocina mapping' do
```
jcoyne, line 5: Isn't this spec duplicating what we already have in `spec/requests/create_object_spec.rb` and `spec/requests/update_object_spec.rb` for full integration tests as well as `spec/services/cocina/to_fedora/*_spec.rb` for unit tests?   I don't think having specs in yet another location makes it easy to find the tests we are looking for. I'd rather we address any deficiencies in these previous locations rather than add yet another set of specs.

jcoyne (changes requested):

app/services/cocina/to_fedora/apo_rights.rb
ndushay: noted;  will note in #2644 and I have taken that ticket to work on shortly.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
ndushay: The roundtripping is crucial to get right before we cutover and the way the unit tests are written, it is not addressed.   We can certainly move some of the tests in those other place to this location, especially the  to_fedora  tests that are relevant.   Testing this way has already surfaced a bunch of stuff that won't roundtrip cleanly.

ndushay: Also - we discussed this approach in mid-sprint planning yesterday and many folks chimed in that doing roundtrip testing of mappings beyond MODS (e.g. access, identity, structural) will likely be very useful for the same reasons -- we need to be sure we can do clean roundtrips and the current tests don't cover that.

jcoyne: The round tripping is fine, is there a reason that can't be done in the request specs or the unit specs I've pointed out?  I don't want to introduce yet another place that one should look for tests.  Also, I'm skeptical that "round-tripping" is the goal. I know that we don't want to loose important data, but sometimes we are writing things in a way that was better (more explicit) than it would otherwise be.

ndushay: this is following the pattern used for MODS.  unit specs ... are only for the direction of the code -- one way -- so they are not a good choice.   We found this approach worked really really well for MODS roundtripping.   It does not replace the need for unit tests or for request specs ... it augments.   I think maybe you'll need to let this go through and we can discuss it further when you and I are both available, and possibly as a team discussion.  I think you might be missing some context.

ndushay: "Request specs provide a thin wrapper around Rails' integration tests, and are
designed to drive behavior through the full stack, including routing
(provided by Rails) and without stubbing (that's up to you)." -- this is useful, but it's not about roundtripping metadata in an API app that would only ever get one direction from a single API call.

ndushay: - where should roundtrip specs live?  In the code base where the mappings are done.
- the app where the mappings are done is an API app
- each direction of mapping is used for separate API calls;  testing the individual API calls works well as a request spec

I am more than happy to discuss consolidating tests or moving test code to a different location when you and I are both working and can talk in realtime.;  I think this is a useful PR as it stands, and the "cleanup" or "consolidation" can happen separately

jcoyne: Round tripping is not important for this data. It is very important for MODS as this is something we plan to do for a long time.  So because we have different requirements for those, let's not try to use the same approach.  Any update request is already doing a round trip because it writes to fedora and then builds the cocina model anew from Fedora: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_updater.rb#L51-L54  So any test in tthttps://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_updater.rb#L51-L54 is already doing this.

justinlittman: One of the lessons of our MODS work is that one-direction testing inevitably produced gaps in the testing. An approach that made roundtrip testing the default definitely reduced mapping errors.

That being said, if tests are moved here, we want to make sure redundant tests are removed from other areas.

jcoyne: @ndushay the conference, aside from keynotes is completely asynchronous and I have availability to have real time discussion.

ndushay: FYI:

There are no unit tests for the mapping classes:  Cocina::FromFedora::APO and Cocina::ToFedora::AdministativeMetadata (was apo_rights).    

The unit tests for Cocina::ObjectUpdater, relevant to an APO, are checking that when a particular part of the cocina model is updated, the relevant fedora datastream is updated -- an excellent unit test, but not on point for the roundtrip testing.  There is no unit test for an APO in Cocina::ObjectCreator.

The request specs of relevance appear to be in 
- create_object_spec.rb -- there *may* be some duplicative testing there, but I would argue that it is testing that "given this api call, the right thing happens", and that   I would be happy to make a ticket to indicate this needs revisiting.

app/services/cocina/to_fedora/apo_rights.rb
ndushay: noted in ticket #2644

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
jcoyne: > There are no unit tests for the mapping classes: Cocina::FromFedora::APO and Cocina::ToFedora::AdministativeMetadata (was apo_rights).

The relevant steps are likely covered in the integration tests and higher level unit tests.  If you see a gap, is it possible to improve it there rather than creating a new paradigm?

> The unit tests for Cocina::ObjectUpdater, relevant to an APO, are checking that when a particular part of the cocina model is updated, the relevant fedora datastream is updated -- an excellent unit test, but not on point for the roundtrip testing. There is no unit test for an APO in Cocina::ObjectCreator.

Why is it "not on point"?  What is preventing you from  you making it so? I'm absolutely certain we are missing test cases here (as evidenced by the tickets being open), but why can't we add them here.

I don't think that I'm being clear about why I think that round tripping for administrativeMetadata is not important.
AdministrativeMetadata is the serialization format exclusively for DSA it should be completely opaque from the outside. There are no other systems that need to know how to parse this XML.  We could convert it to binary data and nobody would be the wiser because our API encapsulates that implementation detail.  We intend to do this in the near future when we move to database for persistence.  This is why using request tests for this scenario is the best way to ensure the data we send in (via cocina) is the data we get out (also via cocina). 

On the other hand, MODS is produced externally and consumed externally.  Thus we need to go bi-directionally from MODS to Cocina.  This is not the use case for administrativeMetadata.   I'm concerned that tests written they way you have now in this PR will likely be deleted when we go to a new persistence layer.

jcoyne: Here's how I would expand the existing test that does round tripping: #2681

justinlittman: I'm not understanding why roundtripping is important for descriptive/MODS, but not other types of data. Since we will be creating/updating objects in the immediate future and mapping all objects permanently from Fedora to Cocina in the near future, it seems important regardless.

Having worked on the request specs, they are a cumbersome way to test mappings. More direct unit tests such as these are cleaner and easier to work on.

jcoyne: @justinlittman  I think the 2 line change is #2681 is a much simpler change than the ~530 line change proposed here. 

MODS roundtrip tests were important because Arcadia was writing them, because MODS is complex, because the timeframe to keep this is longer, because the app needs to accept both MODS and Cocina and because there was no underlying modeling approach (ActiveFedora/OM models).  All of these circumstances are different here.

justinlittman: I agree with your logic -- we don't want to test the implementation detail of persisting as Fedora. However, that doesn't seem to correspond with request specs. If you look at the request specs, e.g., for object_updater, most of what is being tested is mapping / the implementation detail of persisting as Fedora. When we take Fedora out of the equation, we'll probably discard/simplify most of the request specs.

ndushay: @jcoyne the issue here isn't whether or not I can address #2659 for the referencesAgreement with an easy change.  **The issue is that roundtrip testing is desired by Andrew and this is a way to do it.**  Is it perfect?  Of course not.  But it is a way.   Will we need this roundtripping test code forever?  Nope -- it's all about cocina to fedora and back and vice versa.  

Given that we at least temporarily have been asked to provide this roundtrip testing, it makes sense to alter relevant mappings AFTER we have this testing in place and have this way to ensure that the new mappings roundtrip cleanly.

jcoyne: I think I've been missing context.

> The issue is that roundtrip testing is desired by Andrew and this is a way to do it

> Given that we at least temporarily have been asked to provide this roundtrip testing

That was not clear in the description of this ticket an I didn't see any issues like that on the board.

> it's all about cocina to fedora and back and vice versa.

We already have the "cocina to fedora and back part" covered by existing tests.    We do not yet have "vice-versa" covered an as I've already explained, I'm not sure why this latter part is necessary.  What user case are we addressing there?

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+      expect(actual_cocina_props).to be_deep_equal(expected_cocina_props)
+    end
+  end
+
+  context 'when mapping from Cocina to Fedora' do
+    let(:actual_cocina_apo_admin) { Cocina::Models::AdminPolicyAdministrative.new(cocina) }
+    let(:fedora_apo) do
+      Dor::AdminPolicyObject.new(pid: apo_druid,
+                                 admin_policy_object_id: actual_cocina_apo_admin.hasAdminPolicy,
+                                 # source_id: cocina_admin_policy.identification.sourceId,
```
justinlittman, line 59: Cruft?

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+      end
+
+      it 'AdminPolicyAdministrative cocina model roundtrips to original defaultObjectRights.xml' do
+        expect(actual_default_rights_xml).to be_equivalent_to(default_object_rights_xml)
+      end
+    end
+
+    describe 'Cocina::ToFedora::Roles' do
+      let(:actual_role_xml) do
+        Cocina::ToFedora::Roles.write(fedora_apo, Array(actual_cocina_apo_admin.roles))
```
justinlittman, line 87: Does the Array() belong here or does this indicate a condition that should be handled by `Cocina::ToFedora::Roles.write()`?

justinlittman (approved):

Are the various xit's/TODO's ticketed?

jmartin-sul (approved):
a couple small questions, but this looks good to me.

restating some things to make sure i understand them:
* we have a lot of data with a very complex data serialization format in principle.  in practice we've collected millions of objects in a non-transactional data store, with little pre-save validation for conformance to the expected structure, and much hand editing.  there's a lot of unknown and irregularity we have to tolerate.
* we essentially don't want to lose any (?) of it in translation/transfer to the successor serialization format/data store, except where we explicitly decide it's ok.
* fedora -> cocina -> fedora roundtrip testing seems to have caught mapping bugs we'd otherwise have missed (in PRs like this / #2701 / #2703 / #2704, and for real data as with `bin/validate-desc-cocina-roundtrip`)

and if those things are true, my $0.02 to the larger discussion in this PR and the associated meeting:
* i'm in favor of using and expanding (e.g #2700) on the existing roundtrip validation of real fedora datastream xml to look for regressions from any mapping changes, and to make sure changes improve things when expected
  *  and when the time comes to actually migrate, to maybe do a last check to make sure that the final fedora representation roundtrips cleanly before being serialized in cocina for storage using the successor data store as the source of truth, and alerting for manual remediation if the check fails
* but i also find that roundtrip unit tests on basic/represenatative/tricky examples, as with this PR, provide good quick guard rails for development.  e.g. breaking a bunch of existing roundtrip unit tests in the style of this PR prevented me from inadvertently going down the wrong path on #2456 / #2661.
* personally, instead of framing this sort of test code as specs for dor-services app that we know we have to throw away soon, i'd look at it as safeguard and analysis code to give us peace of mind while ironing out a very tricky format mapping and a difficult data store migration.  i don't think that code has a super clear ideal home, but pragmatically the choices for the fedora roundtrip validation code seem good to me so far.
* i can't currently think of an easier or less throwaway way to get the same peace of mind that thorough roundtrip testing has given us, and i'd be hesitant to have less thorough testing than that on any datastream we know we want to keep, given what that sort of testing has turned up for us so far.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+          },
+          registrationWorkflow: [
+            'goobiWF',
+            'registrationWF'
+          ],
+          collectionsForRegistration: [
+            'druid:ny719df8518'
+          ],
+          hasAdminPolicy: ur_apo_druid,
+          # referencesAgreement: 'druid:yr951qr4199',
```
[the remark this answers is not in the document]
ndushay: yes, precisely - RELS_EXT wil be needed for referencesAgreement and I was ... prepping.   Not perfectly agile, nor obvious from the PR.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
```diff
@@ -0,0 +1,529 @@
...
+        access: {
+          digitalRepository: [
+            {
+              value: 'Stanford Digital Repository'
+            }
+          ]
+        }
+      }
+    }
+  end
```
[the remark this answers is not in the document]
ndushay: agree with you, and yes, this will only be used in this file as administrativeMetadata.xml is only used for APOs.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
ndushay: actually, I don't know if it is needed or not at this point.  I copied this comment from https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_creator.rb#L64. (hahahaha!).   I'll remove it.

spec/services/cocina/mapping/administrative/apo_administrative_spec.rb
ndushay: Turns out it isn't needed.  Once again, I copied it from https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_creator.rb#L70 / https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_updater.rb#L84 (hahahaha)

justinlittman (approved):

ndushay:
@jmartin-sul FYI, I added your comment to the testing doc from the meeting:  https://docs.google.com/document/d/19NSubyerKKh5pjwkFrzBsUtHFZAlIla91C7h8jwiTnI

ndushay:
normalizations needed ticketed as https://github.com/sul-dlss/dor-services-app/issues/2707

PR #2681, opened 2021-04-13
Add a broken test for #2659

jcoyne:
## Why was this change made?
Illustrates the problem in #2659.  The existing test does not attempt to set `referencesAgreement`


ndushay:
closing this in favor of #2713

PR #2715, opened 2021-04-16
Fix typo when sending an updated message

### LINKED ISSUE #2714: Exception when updating APO
Reported by Andrew when trying to update an APO on qa.

see https://app.honeybadger.io/projects/49894/faults/78737084 which triggered https://app.honeybadger.io/projects/50568/faults/78909026

jcoyne:
## Why was this change made?

Fixes #2714

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -42,7 +42,7 @@ class ObjectsController < ApplicationController
     persisted_cocina_object = Cocina::ObjectUpdater.run(fedora_object, update_request)
 
     # Broadcast this update action to a topic
-    Notifications::ObjectUpdated.publish(model: cocina_object) if Settings.rabbitmq.enabled
+    Notifications::ObjectUpdated.publish(model: persisted_cocina_object) if Settings.rabbitmq.enabled
 
     render json: persisted_cocina_object
   rescue Cocina::RoundtripValidationError => e


app/controllers/objects_controller.rb
```diff
@@ -42,7 +42,7 @@ def update
     persisted_cocina_object = Cocina::ObjectUpdater.run(fedora_object, update_request)
 
     # Broadcast this update action to a topic
-    Notifications::ObjectUpdated.publish(model: cocina_object) if Settings.rabbitmq.enabled
+    Notifications::ObjectUpdated.publish(model: persisted_cocina_object) if Settings.rabbitmq.enabled
```
mjgiarlo, line 45: Should we add a test that fails when `cocina_object` (presumably a nil?) is passed here?

jcoyne: Are you suggesting we should add rabbitmq to the test stack?

peetucket: i think the problem is that `Settings.rabbitmq.enabled` is probably disabled in the tests, and so that line never gets called...the exception was an undefined local variable, not a nil exception

jcoyne: It also used to be `cocina_object`, so I think this broke when I rebased on https://github.com/sul-dlss/dor-services-app/commit/4ade72d765df36ee291f180fa47065e1f7337816

peetucket (approved):
I think we should get this fixed ASAP seeing as the commit is already on main

PR #2716, opened 2021-04-16
Fix bug with rights determinations when metadata has multiple access nodes

mjgiarlo:
Connects to sul-dlss/argo#2382

## Why was this change made?

Rights metadata containing e.g. file-specific nodes reveal a bug in the prior `FromFedora/Access` implementation, namely the assumption that the presence of a single access=read node is meaningful. Instead of hand-crafted, buggy XPaths, re-use the logic that is already in place in the dor-rights-auth gem.

This PR adds full object-level rights mappings on both the to- and from-fedora sides (32 cases a piece).

CI, locally, and tested in QA by Andrew, who comments below that this is good to go.

None

### DIFF
--- a/app/services/cocina/from_fedora/access.rb
+++ b/app/services/cocina/from_fedora/access.rb
@@ -18,13 +18,13 @@ module Cocina
       def props
         {
           access: access_rights,
-          download: download? ? access_rights : 'none',
+          download: download,
           readLocation: location,
           license: License.find(rights_metadata_ds),
           copyright: copyright,
           useAndReproductionStatement: use_statement
         }.compact.tap do |h|
-          h[:controlledDigitalLending] = true if cdl?
+          h[:controlledDigitalLending] = true if controlled_digital_lending?
         end
       end
 
@@ -32,8 +32,10 @@ module Cocina
 
       attr_reader :rights_metadata_ds
 
+      delegate :controlled_digital_lending?, :dark?, :citation_only?, :obj_lvl, to: :rights_object
+
       def rights_object
-        rights_metadata_ds.dra_object.obj_lvl
+        rights_metadata_ds.dra_object
       end
 
       def copyright
@@ -44,15 +46,73 @@ module Cocina
         rights_metadata_ds.use_statement.first.presence
       end
 
-      # @return [Bool] true unless the rule="no-download" has been set or if the access is citation-only or dark
-      def download?
-        return false if %w[citation-only dark].include? access_rights
+      # @note This method implements an algorithm for determining download
+      #       access based on both the object-level rights and the object-level
+      #       download, using helper methods that lean on the `dor-rights-auth`
+      #       gem.
+      def download
+        # Some access types dictate no downloading. Handle those cases first.
+        return 'none' if no_download? || stanford_no_download? || world_no_download? || location_no_download?
 
-        !rights_object.world.rule && !rights_object.group.fetch(:stanford).rule
+        # Then check to see if download is based on location
+        return 'location-based' if location_based_download?
+
+        # If no specific download rules have been found yet and there are no explicit download rules set, set download to access rights
+        return access_rights if no_world_or_group_download_rules?
+
+        # Finally: the only remaining case is when rights are stanford + world (no-download), so grant download to stanford
+        return 'stanford' if stanford_world_no_download?
+
+        raise "Unexpected download rights: #{obj_lvl}"
+      end
+
+      def no_download?
+        citation_only? || dark? || controlled_digital_lending?
+      end
+
+      def no_world_or_group_download_rules?
+        obj_lvl.world.rule.nil? && obj_lvl.group.fetch(:stanford).rule.nil?
+      end
+
+      def stanford_no_download?
+        stanford? &&
+          obj_lvl.group.fetch(:stanford).rule == 'no-download' &&
+          !(location && obj_lvl.location.fetch(location).value)
+      end
+
+      def world_no_download?
+        world? &&
+          obj_lvl.world.rule == 'no-download' &&
+          !stanford? &&
+          !(location && obj_lvl.location.fetch(location).value)
+      end
+
+      def location_no_download?
+        location &&
+          obj_lvl.location.fetch(location).value &&
+          obj_lvl.location.fetch(location).rule == 'no-download'
+      end
+
+      def location_based_download?
+        location &&
+          obj_lvl.location.fetch(location).value &&
+          obj_lvl.location.fetch(location).rule.nil?
+      end
+
+      def stanford_world_no_download?
+        stanford? && obj_lvl.world.rule == 'no-download'
+      end
+
+      def stanford?
+        obj_lvl.group.fetch(:stanford).value
+      end
+
+      def world?
+        obj_lvl.world.value
       end
 
       def location
-        @location ||= rights_object.location.keys.first
+        @location ||= obj_lvl.location.keys.first
       end
 
       # Map values from dor-services
@@ -62,35 +122,17 @@ module Cocina
         @access_rights ||=
           if world?
             'world'
-          elsif stanford?
+          elsif stanford? || controlled_digital_lending?
             'stanford'
+          elsif citation_only?
+            'citation-only'
           elsif dark?
             'dark'
           elsif location
             'location-based'
           else
-            'citation-only'
+            raise "Cannot interpret access rights from #{rights_metadata_ds.to_xml}"
           end
-      end
-
-      def rights_xml
-        @rights_xml ||= rights_metadata_ds.ng_xml
-      end
-
-      def stanford?
-        rights_xml.search('//rightsMetadata/access[@type=\'read\']/machine/group').length == 1
-      end
-
-      def world?
-        rights_xml.search('//rightsMetadata/access[@type=\'read\']/machine/world').length == 1
-      end
-
-      def dark?
-        rights_xml.search('//rightsMetadata/access[@type=\'discover\']/machine/none').length == 1
-      end
-
-      def cdl?
-        rights_object.controlled_digital_lending
       end
     end
   end
--- a/app/services/cocina/to_fedora/access.rb
+++ b/app/services/cocina/to_fedora/access.rb
@@ -19,12 +19,8 @@ module Cocina
       end
 
       def apply
-        # See https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb
-        Dor::RightsMetadataDS.upd_rights_xml_for_rights_type(rightsMetadata.ng_xml, Rights.rights_type(access))
-        # This invalidates the dra_object, which is necessary if re-mapping.
-        rightsMetadata.content = rightsMetadata.ng_xml.to_s
-        rightsMetadata.copyright = access.copyright if access.copyright
-        rightsMetadata.use_statement = access.useAndReproductionStatement if access.useAndReproductionStatement
+        RightsMetadataGenerator.generate(rights: rightsMetadata, access: access)
+        update_rights_statements!
         License.update(rightsMetadata, access.license) if access.license
       end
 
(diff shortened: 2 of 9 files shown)


app/services/cocina/from_fedora/access.rb
```diff
@@ -18,22 +18,24 @@ def initialize(rights_metadata_ds)
       def props
         {
           access: access_rights,
-          download: download? ? access_rights : 'none',
+          download: download,
```
mjgiarlo, line 21: This was too simplistic and did not cover all 32 object-level rights mapping cases we have.

app/services/cocina/from_fedora/access.rb
```diff
@@ -18,22 +18,24 @@ def initialize(rights_metadata_ds)
...
           readLocation: location,
           license: License.find(rights_metadata_ds),
           copyright: copyright,
           useAndReproductionStatement: use_statement
         }.compact.tap do |h|
-          h[:controlledDigitalLending] = true if cdl?
+          h[:controlledDigitalLending] = true if controlled_digital_lending?
```
mjgiarlo, line 27: Instead of implementing a bunch of bespoke and buggy XPaths, use the dor-rights-auth gem for heavy lifting.

app/services/cocina/from_fedora/access.rb
```diff
@@ -44,15 +46,57 @@ def use_statement
...
+        return 'location-based' if location_based_download?
+
+        # If no no-download rules found, set download to access rights
+        return access_rights if obj_lvl.world.rule.nil? && obj_lvl.group.fetch(:stanford).rule.nil?
+
+        return 'stanford' if obj_lvl.world.rule
+
+        return 'world' if obj_lvl.group.fetch(:stanford).rule
+
+        'foobar'
```
mjgiarlo, line 62: Cruft. Will remove.

app/services/cocina/from_fedora/access.rb
```diff
@@ -44,15 +46,57 @@ def use_statement
...
+        return 'none' if no_download? || stanford_no_download? || world_no_download? || location_no_download?
+
+        return 'location-based' if location_based_download?
+
+        # If no no-download rules found, set download to access rights
+        return access_rights if obj_lvl.world.rule.nil? && obj_lvl.group.fetch(:stanford).rule.nil?
+
+        return 'stanford' if obj_lvl.world.rule
+
+        return 'world' if obj_lvl.group.fetch(:stanford).rule
```
aaron-collier, line 60: Can you clarify why the return of `stanford` above on `obj_lvl.world.rule` and `world` here?

app/services/cocina/from_fedora/access.rb
aaron-collier, line 62: I like it, but assume we don't want to return `foobar` :-)

mjgiarlo: Yep! Removing now. :beers:

app/services/cocina/from_fedora/access.rb
mjgiarlo: I just tossed another commit at this to simplify it.

app/services/cocina/from_fedora/access.rb
```diff
@@ -44,15 +46,73 @@ def use_statement
...
+        return 'none' if no_download? || stanford_no_download? || world_no_download? || location_no_download?
+
+        # Then check to see if download is based on location
+        return 'location-based' if location_based_download?
+
+        # If no specific download rules have been found yet and there are no explicit download rules set, set download to access rights
+        return access_rights if no_world_or_group_download_rules?
+
+        # Finally: the only remaining case is when rights are stanford + world (no-download), so grant download to stanford
+        return 'stanford' if stanford_world_no_download?
```
mjgiarlo, line 60: Here's an algorithm for determining download access based on both the object-level rights and the object-level download, using helper methods that lean on dor-rights-auth. It errs on the side of explicitness by commenting every line.

app/services/cocina/from_fedora/access.rb
```diff
@@ -44,15 +46,73 @@ def use_statement
...
+
+        raise "Unexpected download rights: #{obj_lvl}"
+      end
+
+      def no_download?
+        citation_only? || dark? || controlled_digital_lending?
```
mjgiarlo, line 66: We do not allow downloads for any of these items.

app/services/cocina/to_fedora/access.rb
```diff
@@ -19,12 +19,8 @@ def initialize(item, access)
       end
 
       def apply
-        # See https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb
-        Dor::RightsMetadataDS.upd_rights_xml_for_rights_type(rightsMetadata.ng_xml, Rights.rights_type(access))
-        # This invalidates the dra_object, which is necessary if re-mapping.
-        rightsMetadata.content = rightsMetadata.ng_xml.to_s
```
justinlittman, line 25: Is it problematic that this was dropped?

justinlittman (approved):
Approve pending consideration of question.

app/services/cocina/to_fedora/access.rb
mjgiarlo: @justinlittman I don't think so. It was essentially replaced by [`Cocina::ToFedora::RightsMetadataGenerator#generate`](https://github.com/sul-dlss/dor-services-app/pull/2716/files#diff-fc3f0a6382213c47496252a88f5591b894a61d327791de94125c55d8f44d1f63R19-R35). Thanks for the review!

spec/services/cocina/from_fedora/dro_access_spec.rb
```diff
@@ -296,4 +199,322 @@
...
+            <access type="read">
+              <machine>
+                <none/>
+              </machine>
+            </access>
+          </rightsMetadata>
+        XML
+      end
+
+      it 'builds the hash' do
```
ndushay, line 223: same request regarding these descriptions

ndushay (changes requested):
in from_fedora/access line 7 - I am not fully clear on why it is explicitly called `collection_props` ... is it because the basic rights ALWAYS come from collections?   Maybe the comment could be improved?  (esp. since we're doing rights on ITEMs, no?)

andrewjbtw:
When setting rights from the blue button, I got the following errors on stage (the switch was from "World" to "World no-download"):

dor-services-app: https://app.honeybadger.io/projects/50568/faults/78908761
```
NameError: undefined local variable or method `cocina_object' for #<ObjectsController:0x00007f2230acd1a0>
```

Argo: https://app.honeybadger.io/projects/49894/faults/78940241
```
NoMethodError: undefined method `first' for nil:NilClass
```

spec/requests/update_metadata_spec.rb
```diff
@@ -20,18 +20,24 @@
     allow(AdministrativeTags).to receive(:for).and_return([])
 
     allow(EventFactory).to receive(:create)
+
+    allow(Cocina::ToFedora::RightsMetadataGenerator).to receive(:generate)
```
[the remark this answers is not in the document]
mjgiarlo: Agree. This was recommended by @jcoyne because, basically, it's either mock out the Fedora bits or add Fedora as a CI runtime dependency, which I was hoping to avoid for this PR.

app/services/cocina/from_fedora/access.rb
mjgiarlo: @ndushay When we dump Fedora, I don't think it'll be useful any longer, but apparently we're stuck with Fedora for another year at least.

mjgiarlo:
@ndushay :speech_balloon: 
> in from_fedora/access line 7 - I am not fully clear on why it is explicitly called `collection_props` ... is it because the basic rights ALWAYS come from collections? Maybe the comment could be improved? (esp. since we're doing rights on ITEMs, no?)

I don't know why. I ***suspect*** it was to add some clarity about the relationship between Access and DROAccess, which could use improvement. The relationship is that DROAccess inherits from Access, but Access is the class used for collections (even though it's not in the class name). Stuffing `collection` in the class method communicates that that method is collection-specific even if the class itself is not. It's admittedly messy. I'd like not to change that in this PR if I could avoid it.

spec/services/cocina/from_fedora/access_spec.rb
```diff
@@ -14,157 +14,406 @@
...
+                <none/>
+              </machine>
+            </access>
+          </rightsMetadata>
+        XML
+      end
 
-    it 'builds the hash' do
-      expect(access).to eq(access: 'world')
+      it 'builds the hash' do
```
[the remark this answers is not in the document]
mjgiarlo: done.

spec/services/cocina/from_fedora/dro_access_spec.rb
mjgiarlo: done

mjgiarlo:
@ndushay replied to all your comments. Thanks for the review! Mind having another look?

andrewjbtw:
This looks good for object-level rights. I checked all of the menu choices, the special case of "world access, stanford can download", plus at least one location for:

- location-based, no download
- world access, location-download
- stanford access, location-download

From the Fedora perspective, this includes both rights settings with one "read" block and rights settings with two "read" blocks.

app/services/cocina/from_fedora/access.rb
ndushay: i think a sideways thumb might have been the best emoticon.  dor-rights-auth is not my fav gem, but it does serve a purpose decently, i suppose.

app/services/cocina/from_fedora/access.rb
ndushay: "Here's an algorithm for determining download access based on both the object-level rights and the object-level download, using helper methods that lean on dor-rights-auth."

spec/requests/update_metadata_spec.rb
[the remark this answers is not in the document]
ndushay: understood.  Wishing for the moon.

ndushay (approved):

ndushay:
> @ndushay 💬
> 
> > in from_fedora/access line 7 - I am not fully clear on why it is explicitly called `collection_props` ... is it because the basic rights ALWAYS come from collections? Maybe the comment could be improved? (esp. since we're doing rights on ITEMs, no?)
> 
> I don't know why. I _**suspect**_ it was to add some clarity about the relationship between Access and DROAccess, which could use improvement. The relationship is that DROAccess inherits from Access, but Access is the class used for collections (even though it's not in the class name). Stuffing `collection` in the class method communicates that that method is collection-specific even if the class itself is not. It's admittedly messy. I'd like not to change that in this PR if I could avoid it.

agree.  Just trying to puzzle it out as a side effect of this review.   Maybe we should consider renaming it in cocina-models ... maybe ...

app/services/cocina/from_fedora/access.rb
mjgiarlo: Added! Thanks, @ndushay

ndushay:
I think my LEAST favorite lines are not yours, but I find them very confusing - maybe you've figured it out and can improve on them:

https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/from_fedora/access_spec.rb#L8-L10

```
let(:item) do
    Dor::Collection.new
end
```

mjgiarlo:
@ndushay :speech_balloon: 
> I think my LEAST favorite lines are not yours, but I find them very confusing - maybe you've figured it out and can improve on them:
> 
> [`main`/spec/services/cocina/from_fedora/access_spec.rb#L8-L10](https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/from_fedora/access_spec.rb?rgh-link-date=2021-04-19T20%3A39%3A56Z#L8-L10)
> 
> ```
> let(:item) do
>     Dor::Collection.new
> end
> ```

Yeah, we should prolly change those!

PR #2744, opened 2021-04-20
[HOLD] Updates cocina for access rights.

justinlittman:
## Why was this change made?
* Support the updated cocina model for embargoes by providing a full mapping.
* Refactor access/rights mapping so that mapping code could be shared between embargo and access when appropriate.

Unit

openapi


app/services/cocina/from_fedora/access.rb
```diff
@@ -2,38 +2,38 @@
...
         props.delete(:download)
+        props.delete(:controlledDigitalLending)
+        props.delete(:readLocation)
+        props[:access] = 'world' unless props[:access] == 'dark'
         props
       end
 
+      def self.props(rights_metadata_ds)
+        new(rights_metadata_ds).props
+      end
```
ndushay, line 19: currently addressed in Cocina::FromFedora::DROAccess  ??   Not that I'm in love with that approach ... but heretofore the cocina-models had both flavors.   Delighted if they go away ... but wonder if this should be dro_props just to be clearer in cocina language?

app/services/cocina/from_fedora/embargo.rb
```diff
@@ -17,24 +17,21 @@ def props
...
       attr_reader :embargo_metadata_ds
 
-      def build_embargo_access
-        access_node = embargo_metadata_ds.release_access_node.xpath('//access[@type="read"]/machine/*[1]').first
-        return 'dark' if access_node.nil?
-        return 'world' if access_node.name == 'world'
-        return access_node.content if access_node.name == 'group'
-
-        'dark'
+      def dra_object
```
ndushay, line 30: typo - dro not dra. ?   or don't use DRA abbreviation if it means something other than DRO

app/services/cocina/from_fedora/access_helper.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
+
+module Cocina
+  module FromFedora
+    # Helper for building part of the Access subschema
+    class AccessHelper
```
ndushay, line 6: I like this refactoring!

app/services/cocina/to_fedora/dro_access.rb
```diff
@@ -6,19 +6,10 @@ module ToFedora
     # Fedora 3 data model rightsMetadata
     class DROAccess < Access
```
ndushay, line 7: yay! - you caught it for ToFedora, but what about FromFedora?

app/services/embargo_service.rb
```diff
@@ -1,65 +0,0 @@
-# frozen_string_literal: true
-
-# Sets an embargo for an item.
-class EmbargoService
```
ndushay, line 4: whoa!  I think we _might_ want to keep an EmbargoService?   I'm thinking of someone, in Argo, wanting to change an embargo date.   Are you envisioning that working a different way?   Because we don't always consider changing embargos requiring mapping to and from fedora, do we?

I see that THIS implementation is fedora-centric.   But I think we'll want a way to do this native cocina.

spec/services/cocina/to_fedora/access_spec.rb
```diff
@@ -3,20 +3,20 @@
 require 'rails_helper'
 
 RSpec.describe Cocina::ToFedora::Access do
-  subject(:apply) { described_class.apply(collection, access) }
+  subject(:apply) { described_class.apply(item, access) }
```
ndushay, line 6: shouldn't we check both item and collection?

ndushay (changes requested):
i'm officially confused:   if cocina-models is now DROAccess and CollectionAccess (which I like), 

are we keeping DROAccess for mapping code and renaming  what was xxxFedora::Access to xxxFedora::CollectionAccess to match?   Or are we being clever and having a single xxxFedora::Access mapping class which creates the correct model flavor?

it's confusing, but current DSA has TWO mapping classes for each flavor:

one for collections
one for dro ... which inherits from the collections one

your changes here aren't in keeping with the current code -- should the dro_access classes be removed from the from_fedora and to_fedora areas?

It's possible I'm completely confused.  It's also possible that some of my comments are useful anyway.

justinlittman:
Wow, I missed the forest for the weeds (and entirely skipped the trees). Thanks for catching @ndushay. I'm refactoring some of the classes and tests to consolidate.

app/services/cocina/from_fedora/embargo.rb
justinlittman: I don't know what it means: https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/rights_metadata_ds.rb#L89

ndushay: how about `dor_rights_auth_object` ?

app/services/embargo_service.rb
justinlittman: `EmbargoService` became `Cocina::ToFedora::EmbargoGenerator`. I didn't make sense to have it as a service, when it is really to fedora mapping code.

justinlittman: Oh, and to be clear there is no longer an embargo endpoint in dor-services-app. It was subsumed by the update endpoint.

spec/services/cocina/to_fedora/access_spec.rb
justinlittman: Now moved to separate tests.

app/services/embargo_service.rb
ndushay: I didn't look at it closely enough to understand which code addresses the embargo date change stuff ... but we have integration tests for that now (deposit_via_hydrus_spec) so if we break it, we should be able to find out before deploying to prod!

ndushay: oh yeah - the famous update endpoint will be used for embargo kerfuffle.  w00t.  good on you for remembering.

app/services/cocina/from_fedora/access.rb
```diff
@@ -2,38 +2,24 @@
...
-          access: access_rights,
-          download: download,
-          readLocation: location,
           license: License.find(rights_metadata_ds),
           copyright: copyright,
           useAndReproductionStatement: use_statement
-        }.compact.tap do |h|
-          h[:controlledDigitalLending] = true if controlled_digital_lending?
-        end
+        }.compact.merge(AccessHelper.props(dra_object: rights_metadata_ds.dra_object))
```
mjgiarlo, line 16: Similar to what I was working on in #2748, it looks like?

app/services/cocina/from_fedora/collection_access.rb
```diff
@@ -0,0 +1,22 @@
...
+        new(rights_metadata_ds).props
+      end
+
+      def props
+        super.tap do |props|
+          # Collection access not include all props
+          props.delete(:download)
+          props.delete(:controlledDigitalLending)
+          props.delete(:readLocation)
+          props[:access] = 'world' unless props[:access] == 'dark'
```
ndushay, line 17: I'm not 100% convinced it's better to have the DRO properties in the super class, vs having them only specified in the DROAccess class ...

app/services/cocina/from_fedora/embargo.rb
```diff
@@ -17,24 +17,21 @@ def props
         return {} if embargo_metadata_ds.status != 'embargoed' # We don't need to map any released embargos
 
         {
-          releaseDate: embargo_metadata_ds.release_date.first.utc.iso8601,
-          access: build_embargo_access
+          releaseDate: embargo_metadata_ds.release_date.first.utc.iso8601
```
ndushay, line 20: I'm starting to wonder if we come out ahead having this be a DateTime here rather than a string.  Is that required because we make it a date object in the API?

app/services/cocina/to_fedora/embargo_generator.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Cocina
+  module ToFedora
+    # Builds the embargoMetadata xml from cocina
+    class EmbargoGenerator
```
ndushay, line 6: so more accurately:  `EmbargoMetadataGenerator` ?

ndushay (commented):
dropped a couple more line comments.  Would love to be at meeting between you and Mike, but since I have a doc appt in the morning, I think you should proceed without me.

app/services/cocina/from_fedora/embargo.rb
justinlittman: Yes. Either we provide it with a DateTime or dry types will coerce it to a DateTime.

app/services/cocina/from_fedora/embargo.rb
justinlittman: Works for me.

app/services/cocina/to_fedora/embargo_generator.rb
justinlittman: Yes. Renamed.

PR #2748, opened 2021-04-21
Map file-level rights metadata between Fedora and Cocina

mjgiarlo:
**HOLD** until this can be thoroughly tested in QA/stage

Fixes sul-dlss/argo#2382

## Why was this change made?

These commits complete mapping file-level rights from Fedora to Cocina and back. This is the first SDR implementation of rights mapping at the file level.

CI, and will be tested in QA/stage

None

### DIFF
--- a/app/services/cocina/from_fedora/access.rb
+++ b/app/services/cocina/from_fedora/access.rb
@@ -17,26 +17,17 @@ module Cocina
 
       def props
         {
-          access: access_rights,
-          download: download,
-          readLocation: location,
           license: License.find(rights_metadata_ds),
           copyright: copyright,
           useAndReproductionStatement: use_statement
-        }.compact.tap do |h|
-          h[:controlledDigitalLending] = true if controlled_digital_lending?
-        end
+        }
+          .merge(AccessRights.props(rights_metadata_ds.dra_object, rights_xml: rights_metadata_ds.to_xml))
+          .compact
       end
 
       private
 
       attr_reader :rights_metadata_ds
-
-      delegate :controlled_digital_lending?, :dark?, :citation_only?, :obj_lvl, to: :rights_object
-
-      def rights_object
-        rights_metadata_ds.dra_object
-      end
 
       def copyright
         rights_metadata_ds.copyright.first.presence
@@ -44,95 +35,6 @@ module Cocina
 
       def use_statement
         rights_metadata_ds.use_statement.first.presence
-      end
-
-      # @note This method implements an algorithm for determining download
-      #       access based on both the object-level rights and the object-level
-      #       download, using helper methods that lean on the `dor-rights-auth`
-      #       gem.
-      def download
-        # Some access types dictate no downloading. Handle those cases first.
-        return 'none' if no_download? || stanford_no_download? || world_no_download? || location_no_download?
-
-        # Then check to see if download is based on location
-        return 'location-based' if location_based_download?
-
-        # If no specific download rules have been found yet and there are no explicit download rules set, set download to access rights
-        return access_rights if no_world_or_group_download_rules?
-
-        # Finally: the only remaining case is when rights are stanford + world (no-download), so grant download to stanford
-        return 'stanford' if stanford_world_no_download?
-
-        raise "Unexpected download rights: #{obj_lvl}"
-      end
-
-      def no_download?
-        citation_only? || dark? || controlled_digital_lending?
-      end
-
-      def no_world_or_group_download_rules?
-        obj_lvl.world.rule.nil? && obj_lvl.group.fetch(:stanford).rule.nil?
-      end
-
-      def stanford_no_download?
-        stanford? &&
-          obj_lvl.group.fetch(:stanford).rule == 'no-download' &&
-          !(location && obj_lvl.location.fetch(location).value)
-      end
-
-      def world_no_download?
-        world? &&
-          obj_lvl.world.rule == 'no-download' &&
-          !stanford? &&
-          !(location && obj_lvl.location.fetch(location).value)
-      end
-
-      def location_no_download?
-        location &&
-          obj_lvl.location.fetch(location).value &&
-          obj_lvl.location.fetch(location).rule == 'no-download'
-      end
-
-      def location_based_download?
-        location &&
-          obj_lvl.location.fetch(location).value &&
-          obj_lvl.location.fetch(location).rule.nil?
-      end
-
-      def stanford_world_no_download?
-        stanford? && obj_lvl.world.rule == 'no-download'
-      end
-
-      def stanford?
-        obj_lvl.group.fetch(:stanford).value
-      end
-
-      def world?
-        obj_lvl.world.value
-      end
-
-      def location
-        @location ||= obj_lvl.location.keys.first
-      end
-
-      # Map values from dor-services
-      # https://github.com/sul-dlss/dor-services/blob/b9b4768eac560ef99b4a8d03475ea31fe4ae2367/lib/dor/datastreams/rights_metadata_ds.rb#L221-L228
-      # to https://github.com/sul-dlss/cocina-models/blob/main/docs/maps/DRO.json#L102
-      def access_rights
-        @access_rights ||=
-          if world?
-            'world'
-          elsif stanford? || controlled_digital_lending?
-            'stanford'
-          elsif citation_only?
-            'citation-only'
-          elsif dark?
-            'dark'
-          elsif location
-            'location-based'
-          else
-            raise "Cannot interpret access rights from #{rights_metadata_ds.to_xml}"
-          end
       end
     end
   end
--- a/app/services/cocina/from_fedora/access/access_rights.rb
+++ b/app/services/cocina/from_fedora/access/access_rights.rb
@@ -1,0 +1,154 @@
+# frozen_string_literal: true
+
+module Cocina
+  module FromFedora
+    class Access
+      # builds the access rights portion of the Access subschema
+      class AccessRights
+        def self.props(rights_object, rights_xml:)
+          new(rights_object, rights_xml: rights_xml).props
+        end
+
+        def initialize(rights_object, rights_xml:)
+          @rights_object = rights_object
+          @rights_xml = rights_xml
+        end
+
+        def props
+          {
+            access: access_rights,
+            download: download,
+            readLocation: location
+          }.compact.tap do |h|
+            h[:controlledDigitalLending] = true if controlled_digital_lending?
+          end
+        end
+
+        private
+
+        attr_reader :rights_object, :rights_xml
+
+        # @note This method implements an algorithm for determining download
+        #       access for objects and files based on both the rights setting
+        #       and the download setting, using helper methods that lean on the
+        #       `dor-rights-auth` gem.
+        def download
+          # Some access types dictate no downloading. Handle those cases first.
+          return 'none' if no_download? || stanford_no_download? || world_no_download? || location_no_download?
+
+          # Then check to see if download is based on location
+          return 'location-based' if location_based_download?
+
+          # If no specific download rules have been found yet and there are no explicit download rules set, set download to access rights
+          return access_rights if no_world_or_group_download_rules?
+
+          # Finally: the only remaining case is when rights are stanford + world (no-download), so grant download to stanford
+          return 'stanford' if stanford_world_no_download?
+
+          raise "Unexpected download rights: #{contextual_rights}"
+        end
+
+        # These next few methods are admittedly a little gross. But! They allow
+        # objects to use the full `Dor::RightsAuth` struct, which is returned by
+        # a method on the rights metadata datastream, while allowing files to
+        # use the `Dor::EntityRights` substruct of `Dor::RightsAuth`, which is
+        # how file rights are parsed by `dor-rights-auth`. And the algorithms
+        # for determining access and download are complex, so I'm OK with the
+        # timidness of these checks given the value we get in return.
+        def contextual_rights
+          return rights_object.obj_lvl if rights_object.respond_to?(:obj_lvl)
+
+          rights_object
+        end
+
+        def citation_only?
+          return false unless rights_object.respond_to?(:citation_only?)
+
+          rights_object.citation_only?
+        end
+
+        def controlled_digital_lending?
+          return contextual_rights.controlled_digital_lending unless contextual_rights.controlled_digital_lending.respond_to?(:value)
+
+          contextual_rights.controlled_digital_lending.value
+        end
+
+        def dark?
+          !(world? ||
+            stanford? ||
+            controlled_digital_lending? ||
+            location)
+        end
+
+        def no_download?
+          citation_only? || dark? || controlled_digital_lending?
+        end
+
+        def no_world_or_group_download_rules?
+          contextual_rights.world.rule.nil? && contextual_rights.group.fetch(:stanford).rule.nil?
+        end
+
+        def stanford_no_download?
+          stanford? &&
+            contextual_rights.group.fetch(:stanford).rule == 'no-download' &&
+            !(location && contextual_rights.location.fetch(location).value)
+        end
+
+        def world_no_download?
+          world? &&
+            contextual_rights.world.rule == 'no-download' &&
+            !stanford? &&
+            !(location && contextual_rights.location.fetch(location).value)
+        end
+
+        def location_no_download?
+          location &&
+            contextual_rights.location.fetch(location).value &&
+            contextual_rights.location.fetch(location).rule == 'no-download'
+        end
+
+        def location_based_download?
+          location &&
+            contextual_rights.location.fetch(location).value &&
+            contextual_rights.location.fetch(location).rule.nil?
+        end
+
+        def stanford_world_no_download?
+          stanford? && contextual_rights.world.rule == 'no-download'
+        end
+
+        def stanford?
+          contextual_rights.group.fetch(:stanford).value
+        end
+
+        def world?
+          contextual_rights.world.value
+        end
+
+        def location
+          @location ||= contextual_rights.location.keys.first
+        end
+
+        # Map values from dor-services
+        # https://github.com/sul-dlss/dor-services/blob/b9b4768eac560ef99b4a8d03475ea31fe4ae2367/lib/dor/datastreams/rights_metadata_ds.rb#L221-L228
+        # to https://github.com/sul-dlss/cocina-models/blob/main/docs/maps/DRO.json#L102
+        def access_rights
+          @access_rights ||=
+            if world?
+              'world'
+            elsif stanford? || controlled_digital_lending?
+              'stanford'
+            elsif citation_only?
+              'citation-only'
+            elsif dark?
+              'dark'
+            elsif location
+              'location-based'
+            else
+              raise "Cannot interpret access rights from #{rights_xml}"
+            end
+        end
+      end
+    end
+  end
+end
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -27,7 +27,13 @@ module Cocina
           has_member_orders = build_has_member_orders
           structural[:hasMemberOrders] = has_member_orders if has_member_orders.present?
 
-          contains = FileSets.build(item.contentMetadata, version: item.current_version.to_i, ignore_resource_type_errors: project_phoenix?)
+          # To build file sets, we need to consider both content metadata and
+          # rights metadata, the latter of which is used to map file-specific
+          # access/rights.
+          contains = FileSets.build(item.contentMetadata,
+                                    rights_metadata: item.rightsMetadata,
+                                    version: item.current_version.to_i,
+                                    ignore_resource_type_errors: project_phoenix?)
           structural[:contains] = contains if contains.present?
 
           structural[:hasAgreement] = item.identityMetadata.agreementId.first unless item.identityMetadata.agreementId.empty?
--- a/app/services/cocina/from_fedora/file_sets.rb
+++ b/app/services/cocina/from_fedora/file_sets.rb
@@ -4,14 +4,30 @@ module Cocina
   module FromFedora
     # builds the FileSet instance from a Dor::Item
     class FileSets
-      def self.build(content_metadata_ds, version:, ignore_resource_type_errors: false)
+      def self.build(content_metadata_ds, rights_metadata:, version:, ignore_resource_type_errors: false)
+        new(
+          content_metadata_ds,
+          rights_metadata: rights_metadata,
+          version: version,
+          ignore_resource_type_errors: ignore_resource_type_errors
+        ).build
+      end
+
+      def initialize(content_metadata_ds, rights_metadata:, version:, ignore_resource_type_errors:)
+        @content_metadata_ds = content_metadata_ds
+        @rights_metadata = rights_metadata
+        @version = version
+        @ignore_resource_type_errors = ignore_resource_type_errors
+      end
+
+      def build
         content_metadata_ds.ng_xml.xpath('//resource[file]').map do |resource_node|
-          files = build_files(resource_node.xpath('file'), version: version)
+          files = build_files(resource_node.xpath('file'))
           structural = {}
           structural[:contains] = files if files.present?
           {
             externalIdentifier: IdGenerator.generate_or_existing_fileset_id(resource_node['id']),
-            type: resource_type(resource_node, ignore_resource_type_errors: ignore_resource_type_errors),
+            type: resource_type(resource_node),
             version: version,
             structural: structural
           }.tap do |attrs|
@@ -20,7 +36,15 @@ module Cocina
         end
       end
 
-      def self.resource_type(resource_node, ignore_resource_type_errors:)
+      private
+
+      attr_reader :content_metadata_ds, :rights_metadata, :version, :ignore_resource_type_errors
+
+      def rights_object
+        rights_metadata.dra_object
+      end
+
+      def resource_type(resource_node)
         val = resource_node['type']&.underscore
         val = 'three_dimensional' if val == '3d'
         if val && Cocina::Models::Vocab::Resources.respond_to?(val)
@@ -32,7 +56,7 @@ module Cocina
         end
       end
 
-      def self.digests(node)
+      def digests(node)
         [].tap do |digests|
           # The old google books use upcased versions. See https://argo.stanford.edu/view/druid:dd116zh0343
           # Web archive crawls use SHA1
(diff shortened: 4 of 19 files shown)


app/services/cocina/from_fedora/access.rb
```diff
@@ -17,123 +17,25 @@ def initialize(rights_metadata_ds)
...
       private
 
       attr_reader :rights_metadata_ds
 
-      delegate :controlled_digital_lending?, :dark?, :citation_only?, :obj_lvl, to: :rights_object
-
-      def rights_object
-        rights_metadata_ds.dra_object
-      end
-
```
mjgiarlo, line 40: All the stuff that shows as deleted was moved into a new `AccessRights` class so that objects and files can share logic.

app/services/cocina/from_fedora/access/access_rights.rb
```diff
@@ -0,0 +1,147 @@
...
+          return false unless rights_object.respond_to?(:citation_only?)
+
+          rights_object.citation_only?
+        end
+
+        def controlled_digital_lending?
+          return obj_lvl.controlled_digital_lending unless obj_lvl.controlled_digital_lending.respond_to?(:value)
+
+          obj_lvl.controlled_digital_lending.value
+        end
```
mjgiarlo, line 67: This stuff is a little gross. But! It allows objects to use the full `Dor::RightsAuth` struct, which is returned by a method on the rights metadata datastream, while allowing files to use the `Dor::EntityRights` struct (a substruct of `Dor::RightsAuth`), which is how file rights are parsed by dor-rights-auth. And the algorithms for determining access and download are complex, so I'm OK with it. It looks much better (less duplication, fewer checks) than it did before I made this refactoring.

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -27,7 +27,10 @@ def props
           has_member_orders = build_has_member_orders
           structural[:hasMemberOrders] = has_member_orders if has_member_orders.present?
 
-          contains = FileSets.build(item.contentMetadata, version: item.current_version.to_i, ignore_resource_type_errors: project_phoenix?)
+          contains = FileSets.build(item.contentMetadata,
+                                    rights_metadata: item.rightsMetadata,
```
mjgiarlo, line 31: To build file sets, we now must include the rights metadata so that file-specific access schemas can be generated if needed.

app/services/cocina/from_fedora/file_sets.rb
```diff
@@ -4,14 +4,30 @@ module Cocina
   module FromFedora
     # builds the FileSet instance from a Dor::Item
     class FileSets
-      def self.build(content_metadata_ds, version:, ignore_resource_type_errors: false)
+      def self.build(content_metadata_ds, rights_metadata:, version:, ignore_resource_type_errors: false)
```
mjgiarlo, line 7: Step one of using this class is to convert it to do all the heavy lifting in instances instead of class methods, which allows for state (attrs) instead of passing so much information around.

app/services/cocina/from_fedora/file_sets.rb
```diff
@@ -4,14 +4,30 @@ module Cocina
...
+        @content_metadata_ds = content_metadata_ds
+        @rights_metadata = rights_metadata
+        @version = version
+        @ignore_resource_type_errors = ignore_resource_type_errors
+      end
+
+      def build
         content_metadata_ds.ng_xml.xpath('//resource[file]').map do |resource_node|
-          files = build_files(resource_node.xpath('file'), version: version)
+          files = build_files(resource_node.xpath('file'))
```
mjgiarlo, line 25: Version is an attr now so no need to pass it around.

app/services/cocina/from_fedora/file_sets.rb
```diff
@@ -4,14 +4,30 @@ module Cocina
...
           structural = {}
           structural[:contains] = files if files.present?
           {
             externalIdentifier: IdGenerator.generate_or_existing_fileset_id(resource_node['id']),
-            type: resource_type(resource_node, ignore_resource_type_errors: ignore_resource_type_errors),
+            type: resource_type(resource_node),
```
mjgiarlo, line 30: now an attr so don't pass it

app/services/cocina/from_fedora/file_sets.rb
```diff
@@ -57,7 +81,7 @@ def self.build_files(file_nodes, version:)
             size: node['size'].to_i,
             version: version,
             hasMessageDigests: digests(node),
-            access: access(node),
+            access: access(node['id']),
```
mjgiarlo, line 84: We only need the filename from contentMD to look it up in rightsMD; we don't need the whole xml node.

app/services/cocina/from_fedora/file_sets.rb
```diff
@@ -72,12 +96,11 @@ def self.build_files(file_nodes, version:)
...
 
-      def self.access(node)
-        if node['publish'] == 'yes'
-          { access: 'world', download: 'world' }
-        else
-          { access: 'dark', download: 'none' }
-        end
+      def access(filename)
+        file_specific_rights = rights_object.file[filename]
+        return {} if file_specific_rights.nil?
```
mjgiarlo, line 101: If a file does not have its own access metadata, return an empty. This brings our cocina implementation into parity with our fedora implementation (where files get object rights if they don't have their own rights).

spec/services/cocina/from_fedora/dro_structural_spec.rb
```diff
@@ -191,12 +191,12 @@
       expect(file1[:hasMessageDigests].first[:type]).to eq 'sha1'
       expect(file1[:administrative][:shelve]).to eq true
       expect(file1[:administrative][:sdrPreserve]).to eq false
-      expect(file1[:access][:access]).to eq('world')
+      expect(file1[:access][:access]).to be_nil
```
mjgiarlo, line 194: These files don't have file-specific rights, so these tests changed.

app/services/cocina/from_fedora/access/access_rights.rb
```diff
@@ -0,0 +1,147 @@
...
+          # If no specific download rules have been found yet and there are no explicit download rules set, set download to access rights
+          return access_rights if no_world_or_group_download_rules?
+
+          # Finally: the only remaining case is when rights are stanford + world (no-download), so grant download to stanford
+          return 'stanford' if stanford_world_no_download?
+
+          raise "Unexpected download rights: #{obj_lvl}"
+        end
+
+        def obj_lvl
```
ndushay, line 51: i would prefer this as `object_level_rights` everywhere so we don't have to scratch our brains so much.  especially since "L" looks like "1" in lowercase.    `object_level` ... is ... what???   it's a rights_object ??  I don't want to have to think so hard.  

and obviously, this applies to `xxx_lvl` anywhere

app/services/cocina/from_fedora/access/access_rights.rb
```diff
@@ -0,0 +1,147 @@
...
+            h[:controlledDigitalLending] = true if controlled_digital_lending?
+          end
+        end
+
+        private
+
+        attr_reader :rights_object, :rights_xml
+
+        # @note This method implements an algorithm for determining download
+        #       access based on both the object-level rights and the object-level
```
ndushay, line 32: does this also get used for file-level rights?  because you mention object-level here but never file-level ...   can/should that be clarified or not?

app/services/cocina/from_fedora/access/access_rights.rb
ndushay: putting this here in the review comments makes me wonder if it should go in code comments somewhere.  Not sure.   Maybe a class comment or a method comment?  Or maybe just ... not.

app/services/cocina/from_fedora/access/access_rights.rb
```diff
@@ -0,0 +1,147 @@
...
+        end
+
+        def location
+          @location ||= obj_lvl.location.keys.first
+        end
+
+        # Map values from dor-services
+        # https://github.com/sul-dlss/dor-services/blob/b9b4768eac560ef99b4a8d03475ea31fe4ae2367/lib/dor/datastreams/rights_metadata_ds.rb#L221-L228
+        # to https://github.com/sul-dlss/cocina-models/blob/main/docs/maps/DRO.json#L102
+        def access_rights
```
ndushay, line 128: no idea if this relates to JLitt's work that says that collections should only be "world" or "dark"

app/services/cocina/from_fedora/dro_structural.rb
ndushay: Your comment would make a good code comment.

app/services/cocina/to_fedora/access.rb
```diff
@@ -9,24 +9,28 @@ class Access
       #       See https://argo.stanford.edu/view/druid:bb142ws0723 as an example
       # @param [Dor::Item, Dor::Collection] item
       # @param [Cocina::Models::DROAccess, Cocina::Models::Access] access
-      def self.apply(item, access)
-        new(item, access).apply
+      # @param [Cocina::Models::DROStructural] structural
+      def self.apply(item, access, structural = nil)
```
ndushay, line 13: given all the refactoring, would now be a good time to rename `apply` to something like `update_rights_metadata_from_cocina_access` (obviously too long, but maybe there's some nice middle ground?  `update_rights_metadata` ?    "apply" tells us ... NOTHING.)

Also, would it be too much to ask that the comments clarify which objects are providing data and what will be changed?  Especially since I have to look through a lot of indirection to figure that out?

spec/requests/create_object_spec.rb
```diff
@@ -449,7 +449,7 @@
                     size: 0, version: 1,
                     hasMimeType: 'image/jp2',
                     hasMessageDigests: [],
-                    access: { access: 'world', download: 'world' },
+                    access: { access: 'world', download: 'none' },
```
ndushay, line 452: it always worries me when I see isolated lines like this ... why would the permissions need to change here?

spec/requests/update_metadata_spec.rb
```diff
@@ -780,7 +780,7 @@
                       label: '00002.html', filename: '00002.html', size: 0,
                       version: 1, hasMimeType: 'text/html',
                       hasMessageDigests: [],
-                      access: { access: 'dark', download: 'none' },
+                      access: { access: 'world', download: 'world' },
```
ndushay, line 783: more places for me to worry about permissions changes

spec/services/cocina/from_fedora/dro_access_spec.rb
```diff
@@ -200,7 +197,7 @@
     end
   end
 
-  describe 'access and download rights' do
+  describe 'access and download rights for objects' do
```
ndushay, line 200: do you mean object level?   do you mean non-collections?

spec/services/cocina/from_fedora/dro_structural_spec.rb
```diff
@@ -191,7 +191,7 @@
       expect(file1[:hasMessageDigests].first[:type]).to eq 'sha1'
       expect(file1[:administrative][:shelve]).to eq true
       expect(file1[:administrative][:sdrPreserve]).to eq false
-      expect(file1[:access][:access]).to eq('world')
+      expect(file1[:access][:access]).to eq('dark')
```
ndushay, line 194: another "eek" - what is happening?

spec/services/cocina/from_fedora/file_sets_spec.rb
```diff
@@ -27,17 +79,300 @@
...
         it 'does not notify Honeybadger' do
-          described_class.resource_type(node, ignore_resource_type_errors: true)
+          instance.send(:resource_type, node)
           expect(Honeybadger).not_to have_received(:notify)
         end
       end
     end
   end
+
+  describe 'file-level access rights' do
```
ndushay, line 98: i surely would like to see many of these tests be in a more "roundtrippy" place such as #2753 or #2755.

spec/services/cocina/to_fedora/rights_metadata_generator_spec.rb
```diff
@@ -24,263 +24,587 @@
...
-  context 'when citation-only' do
-    let(:cocina) do
-      <<~JSON
-        {
-          "access": "citation-only",
-          "download": "none"
-        }
-      JSON
-    end
+  describe 'object-level generation' do
```
ndushay, line 29: it would be good to break out unit tests needed for the class vs. roundtrippy tests we want going through all the possibilities ...

ndushay (commented):
dropped a bunch of inline comments.

app/services/cocina/from_fedora/access/access_rights.rb
mjgiarlo: Yes, this is also used for files. I'll update the comment.

app/services/cocina/from_fedora/access/access_rights.rb
mjgiarlo: Like this. Will make the change.

app/services/cocina/from_fedora/access/access_rights.rb
mjgiarlo: I could be convinced. I don't have strong inclinations here.

app/services/cocina/from_fedora/access/access_rights.rb
mjgiarlo: The code here is doing mapping w/o validation, so not related at the moment.

app/services/cocina/from_fedora/dro_structural.rb
mjgiarlo: OK, happy to add it.

app/services/cocina/to_fedora/access.rb
mjgiarlo: I'm meh on the rename. I think `apply` in this case is defensible since we know `ToFedora::Access` it for mapping access from cocina to Fedora.

spec/requests/create_object_spec.rb
mjgiarlo: I might want to get some pairing on this. Request specs in the rights mapping work have felt more like a hindrance than a help. I've been content to stub and mock to get past them, but it may be time to consider adding Fedora (etc.) to the build. :unamused:

spec/services/cocina/from_fedora/dro_access_spec.rb
mjgiarlo: in this case, I mean DROs. perhaps this should change to `items`?

spec/services/cocina/from_fedora/dro_structural_spec.rb
mjgiarlo: we're actually mapping file level rights now :laughing: 

I'm happy to pair on this or do a deep dive on my own to add a more detailed explanation.

app/services/cocina/to_fedora/access.rb
justinlittman: Yes, and a generic name like `apply` or `build` is consistent with other code.

app/services/cocina/to_fedora/rights/file_level.rb
```diff
@@ -0,0 +1,143 @@
...
+
+module Cocina
+  module ToFedora
+    class Rights
+      # Builds the rightsMetadata xml from cocina for a file
+      class FileLevel
+        # @param [Nokogiri::XML::Document] rights_xml an XML document representing rights metadata
+        # @param [Cocina::Models::DROAccess, Cocina::Models::Access] access access/rights metadata in Cocina
+        # @param [Cocina::Models::DROStructural] structural structural metadata in Cocina
+        def self.generate(rights_xml:, access:, structural:)
```
justinlittman, line 11: What does this return?

app/services/cocina/to_fedora/rights/object_level.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
+
+module Cocina
+  module ToFedora
+    class Rights
+      # Builds the rightsMetadata xml from cocina for an object (item/collection)
+      class ObjectLevel
+        # @param [Nokogiri::XML::Document] rights_xml an XML document representing rights metadata
+        # @param [Cocina::Models::DROAccess, Cocina::Models::Access] access access/rights metadata in Cocina
+        def self.generate(rights_xml:, access:)
```
justinlittman, line 10: What does this return?

app/services/cocina/object_creator.rb
```diff
@@ -89,7 +89,7 @@ def create_dro(cocina_item)
         add_dro_tags(pid, cocina_item)
 
         apply_default_access(fedora_item)
-        Cocina::ToFedora::DROAccess.apply(fedora_item, cocina_item.access) if cocina_item.access
+        Cocina::ToFedora::DROAccess.apply(fedora_item, cocina_item.access, cocina_item.structural) if cocina_item.access || cocina_item.structural
```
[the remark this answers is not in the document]
justinlittman: Oh, because what's actually being built here is rights metadata DS. Would this be less confusing if named by datastream instead of cocina subschema (when maping to fedora)?

spec/requests/create_object_spec.rb
justinlittman: This is trying to test the dark validator. Given that, I don't know why these should need to change.

app/services/cocina/from_fedora/access/access_rights.rb
mjgiarlo: done

app/services/cocina/from_fedora/dro_structural.rb
mjgiarlo: done

app/services/cocina/from_fedora/access/access_rights.rb
mjgiarlo: Done. I went with `contextual_rights` since the method can return either object-level rights or file-level rights.

app/services/cocina/from_fedora/access/access_rights.rb
mjgiarlo: updated

app/services/cocina/object_creator.rb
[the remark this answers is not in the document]
mjgiarlo: @justinlittman Could be. I'm open to renaming changes in DSA. What would you think about making renaming changes in a separate PR once it's clear to us what all renames we should make?

app/services/cocina/to_fedora/rights/file_level.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
+
+module Cocina
+  module ToFedora
+    class Rights
+      # Builds the rightsMetadata xml from cocina for a file
+      class FileLevel
+        # @param [Nokogiri::XML::Document] rights_xml an XML document representing rights metadata
+        # @param [Cocina::Models::DROAccess, Cocina::Models::Access] access access/rights metadata in Cocina
```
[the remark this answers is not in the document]
mjgiarlo: I don't believe `FileLevel` is ever passed a FileAccess. It can be passed access nodes attached to DROs and Collections (hence `Access`).

app/services/cocina/to_fedora/rights/file_level.rb
```diff
@@ -0,0 +1,143 @@
...
+          @rights_xml = rights_xml
+          @access = access
+          @structural = structural
+        end
+
+        def generate
+          [].tap do |nodes|
+            file_sets.each do |file_set|
+              file_set.structural.contains.each do |file|
+                next unless file_access_different_than_item_access?(file.access)
```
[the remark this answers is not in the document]
mjgiarlo: Correct. When a file is missing from the rights metadata XML, it is assumed to have the same rights as the containing object.

app/services/cocina/to_fedora/rights/file_level.rb
mjgiarlo: Added more YARD doc to cover this.

app/services/cocina/to_fedora/rights/object_level.rb
mjgiarlo: Added YARD doc. Thanks!

app/services/cocina/object_creator.rb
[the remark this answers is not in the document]
justinlittman: Yes.

spec/services/cocina/from_fedora/dro_structural_spec.rb
mjgiarlo: @ndushay @justinlittman This changed because the test was previously testing that a file in contentMD with `@publish="yes"` gets `world` access. We no longer use contentMD to determine the rights for a file; this was always the wrong implemenation, put in place to have something working without going down the rights :rabbit: hole. So this test change is legit and not a bug. I changed the tests to test the publish attr directly instead of the access level, which has nothing to do with contentMD.

spec/requests/create_object_spec.rb
mjgiarlo: @ndushay @justinlittman This is an expected change, I believe, not a bug. If you look up at the data that is posted, particularly the access metadata attached to each file (lines 280, 308, 327, and 346), you'll see that the client asks for the four files to be:

1. dark (download unspecified, thus `none`)
2. stanford (download unspecified, thus `none`)
3. dark  (download unspecified, thus `none`)
4. world (download unspecified, thus `none`)

Thus, this is our code behaving as it should, as I understand it. Concur?

spec/requests/update_metadata_spec.rb
mjgiarlo: @ndushay @justinlittman AFAICT, here's what's going on here. Check out these lines in the object_updater:

https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_updater.rb#L52-L54

If you add a block argument to `Mapper.build(...).tap` (say, `mapped`) and then inspect `mapped.to_h`, you can see all the file-level access nodes are world/world. And yet, if you inspect `cocina_object.to_h`, you see:

```ruby
# ...
:structural=>                                                                                                                                        
  {:contains=>                                                                                                                                        
    [{:type=>"http://cocina.sul.stanford.edu/models/resources/file.jsonld",                                                                           
      :externalIdentifier=>                                                                                                                           
       "http://cocina.sul.stanford.edu/fileSet/234-567-890",                                                                                          
      :label=>"Page 1",                                                                                                                               
      :version=>1,                                                                                                                                    
      :structural=>                                                                                                                                   
       {:contains=>                                                                                                                                   
         [{:type=>"http://cocina.sul.stanford.edu/models/file.jsonld",                                                                                
           :externalIdentifier=>                                                                                                                      
            "http://cocina.sul.stanford.edu/file/123-456-789",                                                                                        
           :label=>"00001.html",                                                                                                                      
           :filename=>"00001.html",                                                                                                                   
           :version=>1,                                                                                                                               
           :hasMimeType=>"text/html",                                                                                                                 
           :use=>"transcription",                                                                                                                     
           :hasMessageDigests=>                                                                                                                       
            [{:type=>"sha1",                                                                                                                          
              :digest=>"cb19c405f8242d1f9a0a6180122dfb69e1d6e4c7"},                                                                                   
             {:type=>"md5", :digest=>"e6d52da47a5ade91ae31227b978fb023"}],                                                                            
           :access=>{:access=>"dark", :download=>"none"},                                                                                             
           :administrative=>                                                                                                                          
            {:publish=>false, :sdrPreserve=>true, :shelve=>false}},
          {:type=>"http://cocina.sul.stanford.edu/models/file.jsonld",                                                                                
           :externalIdentifier=>                                                                                                                      
            "http://cocina.sul.stanford.edu/file/223-456-789",                                                                                        
           :label=>"00001.jp2",                                                                                                                       
           :filename=>"00001.jp2",
           :version=>1,
           :hasMimeType=>"image/jp2",
           :hasMessageDigests=>[],
           :access=>{:access=>"stanford", :download=>"stanford"},
           :administrative=>
            {:publish=>true, :sdrPreserve=>true, :shelve=>true}}]}},
     {:type=>"http://cocina.sul.stanford.edu/models/resources/file.jsonld", 
      :externalIdentifier=>
       "http://cocina.sul.stanford.edu/fileSet/334-567-890",
      :label=>"Page 2",
      :version=>1,
      :structural=>
       {:contains=>
         [{:type=>"http://cocina.sul.stanford.edu/models/file.jsonld",
           :externalIdentifier=>
            "http://cocina.sul.stanford.edu/file/323-456-789",
           :label=>"00002.html",
           :filename=>"00002.html",
           :version=>1,
           :hasMimeType=>"text/html",
           :hasMessageDigests=>[],
           :access=>{:access=>"world", :download=>"world"},
           :administrative=>
            {:publish=>false, :sdrPreserve=>true, :shelve=>false}},
          {:type=>"http://cocina.sul.stanford.edu/models/file.jsonld",
           :externalIdentifier=>
            "http://cocina.sul.stanford.edu/file/423-456-789",
           :label=>"00002.jp2",
           :filename=>"00002.jp2",
           :version=>1,
           :hasMimeType=>"image/jp2", 
           :hasMessageDigests=>[],
           :access=>{:access=>"world", :download=>"world"},
           :administrative=>
            {:publish=>true, :sdrPreserve=>true, :shelve=>true}}]}}]}}
# ...
```

So the file-level accesses are what we expect here and they're being mapped correctly, but the **re-mapping from Fedora** there, noting that these tests do not mutate Fedora state because we don't have a Fedora container running, naturally returns what was in Fedora in the first place, which is to say... NO specific file-level mappings, just the item-level rights.

Thus, as far as I can tell, this is not a bug. It's a shortcoming of how we're doing CI, and using request specs as not-quite-unit-tests and also not-quite-feature-tests. Happy to pair if you prefer to dig into more.

spec/services/cocina/to_fedora/rights_metadata_generator_spec.rb
mjgiarlo: Yes, happy to move these into round-trippy tests after it's merged. Would prefer not to rejigger all the tests in this PR after having spent a bunch of time making sure I have all the cases written up and passing. OK to work with me on getting this done after this is merged and after #2744 is in? cc: @justinlittman

spec/requests/update_metadata_spec.rb
mjgiarlo: Nope, the problem was needing to update the stubbed in rightsMD content.

justinlittman (approved):

app/services/cocina/object_creator.rb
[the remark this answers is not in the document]
ndushay: yes, perhaps a renaming confab later would be a well spent 15 min.

spec/services/cocina/from_fedora/file_sets_spec.rb
ndushay: Maybe we can discuss this separately.

Issue #2701, opened 2021-04-14
Fedora rights / Cocina access mappings - symmetrical (unit-ish) specs set up


ndushay:
somewhat akin to what we do for MODS mapping roundtrip tests, write rspec shared_examples and a few starter tests to check symmetry of rightsMetadata.xml (fedora) / access (cocina model) mapping code (to and from fedora):

- fedora (A) to cocina (B)
- cocina (B) to roundtrip fedora (C) where C and A should match, or C is explicitly specified as different per approved normalizations (e.g. don't care about useless vestigial attributes in the XML).
- roundtrip fedora (C) to roundtrip cocina (D) where D and B should match.

rightsMetadata.xml is the Fedora datastream, but it may be that the cocina-model access object https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/access.rb gets info from additional places, such as embargoMetadata.xml.

embargoMetadata.xml is also represented in cocina-model access object!

Note that the mappings may be different for APO vs collection vs item, and in other dimensions as well.

This ticket is not to get ALL of the wrinkles in specs;  it's to set up a pattern that can be used for additional rights <--> access specs, as needed.

See https://github.com/sul-dlss/cocina-models/issues/236 for some good examples of problems

ndushay:
tagging @mjgiarlo and @andrewjbtw as they may already have some ideas of 
- what sorts of wrinkles we should eventually test for (which can/will/should be written up as separate tickets, once this one is done) AND 
- what sorts of unimportant differences in roundtripped rightsMetadata.xml and cocina access are okay and should be accommodated with normalization code and/or tolerant specs ...

ndushay:
See https://github.com/sul-dlss/cocina-models/issues/236 for some examples.

ndushay:
Shore up tests after major cocina access rewrite

ndushay:
### Problems/Bugs with IdentityMetadata.xml <--> Cocina::Identification Mapping code

Make issues for each of these?
- [ ] accessCondition element in rightsMetadata  should error/notify?   (https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/access/dro_access_spec.rb#L396)


### Normalizations needed for IdentityMetadata.xml <--> Cocina::Identification Mapping code

[...]

PR #2753, opened 2021-04-21
DROAccess symmetrical mapping unit tests - a pattern and some examples

ndushay:
## Why was this change made?

Part of #2701, this PR presents a new spec file that can eventually become the replacement for many of the tests in:

- services/cocina/from_fedora/dro_access_spec.rb
- services/cocina/from_fedora/license_spec.rb
- services/cocina/to_fedora/dro_access_spec.rb
- services/cocina/to_fedora/license_spec.rb
- services/cocina/to_fedora/rights_metadata_generator_spec.rb

Mike and I have solemnly sworn to remove test duplication before #2701 and file level access implementation is done.

This approach, as well as this specific PR, may also ease some testing for @justinlittman as he works on updating embargo and access models in cocina and all the other changes that fall out of it.

it adds a spec.

### DIFF
--- a/app/services/cocina/from_fedora/dro_access.rb
+++ b/app/services/cocina/from_fedora/dro_access.rb
@@ -15,7 +15,7 @@ module Cocina
 
       def props
         super.tap do |access|
-          access[:embargo] = embargo unless embargo.empty?
+          access[:embargo] = embargo if embargo&.any?
           access[:useAndReproductionStatement] = use_statement
           access[:copyright] = copyright
           access[:license] = License.find(rights_metadata_ds)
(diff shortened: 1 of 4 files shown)


app/services/cocina/from_fedora/dro_access.rb
```diff
@@ -15,7 +15,7 @@ def initialize(rights_metadata_ds, embargo:)
 
       def props
         super.tap do |access|
-          access[:embargo] = embargo unless embargo.empty?
+          access[:embargo] = embargo if embargo&.any?
```
ndushay, line 18: this change allows embargo to be nil, and seems to have no ill effect.

mjgiarlo (approved):

PR #2762, opened 2021-04-23
Add an endpoint for displaying the public XML

mjgiarlo:
## Why was this change made?

Connects to sul-dlss/infrastructure-integration-test#241

CI

none

### DIFF
--- a/spec/requests/metadata_spec.rb
+++ b/spec/requests/metadata_spec.rb
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe 'Display metadata' do
-  let(:object) { Dor::Item.new(pid: 'druid:1234') }
+  let(:object) { Dor::Item.new(pid: 'druid:mk420bs7601') }
 
   before do
     object.descMetadata.title_info.main_title = 'Hello'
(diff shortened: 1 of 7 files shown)


ndushay (approved):

spec/requests/metadata_spec.rb
```diff
@@ -48,4 +48,75 @@
...
+          <identityMetadata/>
+          <rightsMetadata>
+            <access type="discover">
+              <machine>
+                <none/>
+              </machine>
+            </access>
+            <access type="read">
+              <machine>
+                <none/>
```
jmartin-sul, line 91: it seems odd to me to have public XML for a dark object (read and discover both being none is dark, right?), but i don't know enough to know whether that's actually something we shouldn't produce, so probably fine, esp in a unit test? 🤷

jmartin-sul (approved):
noticed one thing that surprised me, but don't know enough to know whether it's actually wrong.  LGTM, approving, feel free to merge if my question is of no concern.

spec/requests/metadata_spec.rb
mjgiarlo: @jmartin-sul Yeah, I don't think the public xml service validates that itself. We could "fix" this with more mocking. Worth the effort or nah?

jmartin-sul: nah

PR #2767, opened 2021-04-26
[WIP] Default object rights normalization

### LINKED ISSUE #2709: APO defaultObjectRights.xml needs normalization code
NOTE:  Do #2797 first.

PR  #2677 has an xit-ed spec on line 421 that shows how license information changes for defaultObjectRights.xml for APO objects.

for an APO, `defaultObjectRights.xml`  might contain

```xml
  <rightsMetadata>
            <access type="discover">
              <machine>
                <world/>
              </machine>
            </access>
            <access type="read">
              <machine>
                <world rule="no-download"/>
              </machine>
            </access>
            <use>
              <human type="useAndReproduction">Use at will.</human>
              <human type="creativeCommons">CC BY-NC Attribution-NonCommercial</human>
              <machine type="creativeCommons">by-nc</machine>
            </use>
</rightsMetadata>
```

but after roundtripping through cocina, it becomes something like this:

```
 <rightsMetadata>
(snip)
            <use>
              <human type="useAndReproduction">Use at will.</human>
              <human type="creativeCommons">https://whatever_the-url-is.org</human>
              <machine type="creativeCommons">https://whatever_the-url-is.org</machine>
            </use>
</rightsMetadata>
```

We should have "normalization" code, such as we do for MODS/descriptive, to ignore this difference when comparing roundtripped Fedora xml (fedora -> cocina -> roundtripped fedora).  (right @andrewjbtw ???)

The normalization could/should be executed by bin/(validate-cocina-roundtrip) scripts as appropriate and it should also be executed by spec/services/cocina/mapping/administrative/apo_administrative_spec.rb so the specs compare the actual APO roundtripped Fedora defaultObjectRights.xml to the defaultObjectRights.xml that is normalized by executing said normalization code.

Note that this may also come into play for rightsMetadata.xml !!!

peetucket:
## Why was this change made?

Fixes #2709 - default object rights normalization

### DIFF
--- a/app/services/cocina/apo_default_object_rights_normalizer.rb
+++ b/app/services/cocina/apo_default_object_rights_normalizer.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+module Cocina
+  # Normalizes a Fedora APO default object rights, accounting for differences between Fedora APOs and APOs generated from Cocina.
+  class ApoDefaultObjectRightsNoramlizer
+    # @param [Nokogiri::Document] rights_ng_xml default_object_rights_xml to be normalized
+    # @param [String] druid
+    # @return [Nokogiri::Document] normalized default_object_rights
+    def self.normalize(rights_ng_xml:, druid:)
+      ApoDefaultObjectRightsNoramlizer.new(rights_ng_xml: rights_ng_xml, druid: druid).normalize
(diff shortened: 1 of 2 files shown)


peetucket:
closed in favor of #2790 and #2791

Issue #2787, opened 2021-04-28
move rights/access normalization code out of bin/validate-rights and into a normalization service class for rights


ndushay:
It will be best to have a consistent approach for "normalizing" our data transforms for any and all fedora datastreams -- that is, accommodating the pre-approved differences between the original xml in a fedora datastream and the roundtrip datastream.

descMetadata / mods is already doing something like this, but we will need to add this for

- rightsMetadata
- identityMetadata
- administrativeMetadata (APOs only)
... (probably all datastreams, but :shrug: 

rightsMetadata currently has such normalization code baked into the bin/xxx-rights-xxx  script -- this code should MOVE to app/services/cocina/normalization/xxxx and the script should be updated there.

I suggest talking to me or JLitt before starting on this ticket as it might make the work clearer.

Note that rightsMetadata ALSO needs to properly normalize the license/use changes (e.g from "by-nd" in original xml to URI) rather than strip elements and ignore them.  (What if we have additional use elements ?   stripping them would prevent us from knowing this)

So this ticket has 3 pieces (3 PRs?  or combine them as you see fit):
- [x] refactor descMetadata normalization code to slightly new approach (or not) so all is under app/services/cocina/normalizion (and make sure bin/xxx scripts work with this change, too)
- [x] create rightsMetadata normalization code in consistent way to mods normalizations, pulling it out of bin/**rights*** roundtrippings script and calling the newly located code properly, as bin/**descMetadata** does
- [ ] fix the rightsMetadata licensing normalization code to o properly normalize the license/use changes (e.g from "by-nd" in original xml to URI) rather than strip elements and ignore them.

ndushay:
after this is done, ping Naomi to add the rights normalization to the rights/access roundtrip unit tests

peetucket:
Part 1 is done (though I used the folder name "normalizers") in https://github.com/sul-dlss/dor-services-app/pull/2790

There is test coverage for that code already, but it would be nice to test the branch...is deploying to qa and running a few druids through a good way to do this?

ndushay:
yay!   i think to test the bin/validate type scripts, you only need to test on sdr-deploy.  that is where the cache lives, so its the only place we run that code.

peetucket:
Part 2 is now done: #2791

peetucket:
I think part 3 in your list is actually in this ticket #2709, so I think we can now close this one

ndushay:
[...]

PR #2790, opened 2021-04-28
move mods normalizers into their own folder

peetucket:
## Why was this change made?

Part of #2787 - move mods normalizer into its own folder for better organization...

No code actually changed -- it all just got moved around (so ignore all the deletion/addition).

Basically:

1. created a new "normalizers" folder under `app/services/cocina`
2. moved the `ModsNormalizer` class into this folder 
3. moved the `mods_normalizer` folder that all of the mods specific normalizers into this folder and renamed it `mods`
4. adjusted all the spots in the code for the new module names/paths
5. adjusted the spec locations and names to account for the moving
6. adjust rubocop paths 

Existing test suite passes (and it failed spectacularly when i had typos/issues so the coverage actually seems to be good)


.rubocop_todo.yml
```diff
@@ -35,27 +54,27 @@ Lint/SuppressedException:
...
 # Configuration parameters: IgnoredMethods, CountRepeatedAttributes.
 Metrics/AbcSize:
   Max: 38
 
-# Offense count: 5
+# Offense count: 8
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
-  Max: 119
+  Max: 261
```
jcoyne, line 65: I really would prefer if this change didn't go in.  This raises the global threshold to something far higher than I would like. Can you restore the exception for the classes that brought this up?

jcoyne (changes requested):

.rubocop_todo.yml
```diff
@@ -1,12 +1,31 @@
...
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
 # Offense count: 1
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, IndentationWidth.
+# SupportedStyles: aligned, indented, indented_relative_to_receiver
+Layout/MultilineMethodCallIndentation:
+  Exclude:
+    - 'app/services/cocina/normalizers/mods.rb'
```
jcoyne, line 15: Does this file still exist?

peetucket: oops relic of the changes that got out of sync...let me undo the rubocop config changes and the exceptions and see if I can target to just what i moved around

.rubocop_todo.yml
peetucket: Reverted all the rubocop changes, which are now targeted to just the new class locations

jcoyne (approved):

PR #2791, opened 2021-04-28
move rights normalization into a class

peetucket:
## Why was this change made?

Part of #2787 - refactor and adding normalization for round-trip validation

Note: this does not change any of the normalization by design... it is just a refactor of the existing code to put the existing normalization into a class and remove it from the bin scripts.  It also adds a simple unit test, which we can then modify when the actual normalization changes in follow on PRs.

Existing unit tests, new unit test for rights normalizer class and running roundtrip rights validation on sdr-deploy


bin/validate-rights-cocina-roundtrip
```diff
@@ -133,8 +117,16 @@ def validate_druid(druid, loader)
...
 
-  normalized_original_ng_xml = normalize_original_ng_xml(original_ng_xml)
   normalized_roundtrip_ng_xml = normalize_roundtrip_ng_xml(roundtrip_ng_xml, original_ng_xml)
+
+  begin
+    # Perform approved XML normalization changes to avoid noise in roundtrip failures
+    normalized_original_ng_xml = Cocina::Normalizers::RightsNormalizer.normalize(rights_ng_xml: original_ng_xml)
+  rescue StandardError => e
+    write_error(druid, original_ng_xml, cocina_props, e)
+    return :rights_normalizer_error
```
justinlittman, line 127: You're going to want to count this and report.

app/services/cocina/normalizers/rights_normalizer.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      private
+
+      attr_reader :ng_xml
+
+      def normalize_original_xml
+        new_ng_xml = ng_xml.dup
+        ['openDataCommons', 'creativeCommons'].each do |license_type|
+          new_ng_xml.root.xpath("//use/machine[@type='#{license_type}' and text()]").each(&:remove)
+          new_ng_xml.root.xpath("//use/human[@type='#{license_type}' and text()]").each(&:remove)
```
justinlittman, line 35: We're going to want to change this rather than remove. Ask @ndushay or me.

peetucket: yes, this PR was just to move existing stuff into a new class but all the same functionality -- further changes in a new PR

bin/validate-rights-cocina-roundtrip
peetucket: got it, added to the counts

ndushay (commented):
I see nothing alarming.  I'll defer to @justinlittman

justinlittman (approved):

Issue #2786, opened 2021-04-28
get timings for roundtripping actual data - ALL datastreams vs single datastream (e.g. rights or desc metadata)


ndushay:
Before we embark on separating how we get roundtrip data for separate datastreams, @justinlittman suggests we find out if using a single script for ALL the datastreams is performant enough.

justinlittman:
Roundtripping all datastreams is much slower. Why? `object_updater` is slow. Why? It maps to cocina 2X and mapping to cocina is slow. Why? Getting administrative tags is slow because it requires retrieving the tags from dor-services-app, which is slow. (Recall that since this isn't being run on dor-services server, administrative tags are monkey patched to get them using dor-services-client instead of from the db.)

So:
* Changing so that `object_updater` only maps to cocina once.
* Adding some caching of administrative tags.

If this isn't fast enough, we can cache administrative tags in a local db so that don't need to retrieve from dor-services-app at all.

justinlittman:
OK.

With performance improvements from #2793, `validate-cocina-roundtrip` goes from 7m 35sec for 10K objects to 3m 58sec. Compare this to 34sec for `validate-cocina-roundtrip`.

Why still so slow? Some items have a large number of files, and this slows roundtripping down a lot.

ndushay:
To me the implication is that large numbers of files is likely to affect contentMetadata only, yes?   So maybe we could NOT do contentMetadata by default?  And have a switch to turn it on?

Also, your previous comment makes no sense.  What takes 3 sec?   vs 3m 58 sec?

Can you get timings with contentMetadata turned off?   2 orders of magnitude to get all the data is sorta compelling for splitting things up ...

justinlittman:
I added a flag for skipping content metadata. The timing goes down to 1m 15sec. (It might actually be faster, but another process is running on sdr-deploy, which might be slowing things down a bit.)

Testing locally (much slower), it went from 2m 21sec for 2500 objects to 40sec.

ndushay:
so, 10K takes 1min 15 seconds;  3M will take 6-7 hours, by my math.  (75 secs x 300 / 3600 (secs -> hours))  That seems tractable for full data.  What do you think?   

To speed it up even more:  should we profile to see where it spends most of its time?   Or are we done?

justinlittman:
I'd say done for now.

ndushay:
[...]

PR #2793, opened 2021-04-29
Performance improvements for validate-cocina-roundtrip.

justinlittman:
refs #2786

## Why was this change made?
Make validation faster.

sdr-deploy

NA

### DIFF
--- a/lib/fedora_cache.rb
+++ b/lib/fedora_cache.rb
@@ -38,6 +38,8 @@ class FedoraCache
       datastreams[dsid] = datastream if datastream
     end
 
+    tags = fetch_tags(druid)
+
     zip_path = zip_path_for(druid)
     path = File.dirname(zip_path)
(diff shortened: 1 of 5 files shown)


lib/fedora_cache.rb
```diff
@@ -67,7 +70,16 @@ def label_and_datastreams(druid)
     return result if result.failure?
 
     contents = result.value!
-    Success([label_for(contents['object']), contents.except('object')])
+    Success([label_for(contents['object']), contents.except('object', 'tags')])
+  end
+
+  def label_and_datastreams_and_tags(druid)
```
ndushay, line 76: I assume you're keeping the other methods on purpose (label_and_datastreams) ??

ndushay (approved):
LGTM; one question

lib/fedora_cache.rb
justinlittman: Yes-ish

Issue #2703, opened 2021-04-14
Fedora identity / Cocina identification mappings - symmetrical (unit-ish) specs set up


ndushay:
somewhat akin to what we do for MODS mapping roundtrip tests, write rspec shared_examples and a few starter tests to check symmetry of identityMetadata.xml (fedora) / identification (cocina model) mapping code (to and from fedora):

- fedora (A) to cocina (B)
- cocina (B) to roundtrip fedora (C) where C and A should match, or C is explicitly specified as different per approved normalizations (e.g. don't care about useless vestigial attributes in the XML).
- roundtrip fedora (C) to roundtrip cocina (D) where D and B should match.

identityMetadata.xml is the Fedora datastream, but it may be that the cocina-model identification object https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/identification.rb gets info from additional places, such as RELS-EXT or somewhere.

Note that the mappings may be different for APO vs collection vs item, and in other dimensions as well.

This ticket is not to get ALL of the wrinkles in specs; it's to set up a pattern that can be used for additional identity info roundtrip unit specs, as needed.

(analogous to #2701)

ndushay:
another example is apo_administrative_metadata mapping spec.

ndushay:
This ticket will be complete when tickets are created for identified bugs and normalizations

ndushay:
### Problems/Bugs with IdentityMetadata.xml <--> Cocina::Identification Mapping code

Make issues for each of these?

- [ ] APO objects do need to support sourceId
    - [ ] raise an error if (which object types?) do not have a sourceId
    - [ ] ETDs (early ones) may be missing a sourceId, in which case one is created from the dissertationid.
- [ ] disssertationId should be retained even when there is a separate sourceID that is not a dissertationId (?)
- [ ] releaseTags need to roundtrip back into identityMetadata.xml  - they should not disappear.  (affects a bunch of tests in dro and collection identification mapping specs)
- [ ] if a Fedora item has 2 catkeys in identityMetadata.xml - what should the code do?  (zw844wz5427 - it's an old ETD)
- [ ] do not double escape ampersands in objectLabel (bb001zc5754 - DRO)
- [ ] ensure fedora items with no collections do not error when going through object_creator OR object_updater (may be true now, didn't check)
- [ ] agreementId ... Honeybadger notify if it's found on non APO object?

[...]

PR #2796, opened 2021-04-29
Roundtrip identification mapping specs

ndushay:
Part of #2703, this PR presents four new spec files that are focused on the mapping code (from_fedora, to_fedora) that affects identityMetadata.xml for Fedora objects, and Cocina::Identification models.

There is a teeny bit of refactoring included -- trying to get more of the code to follow the naming conventions we agreed on in an infradev meeting ("fedora_collection" "cocina_dro" etc.)

it adds specs.

### DIFF
--- a/app/services/cocina/from_fedora/administrative.rb
+++ b/app/services/cocina/from_fedora/administrative.rb
@@ -2,34 +2,34 @@
 
 module Cocina
   module FromFedora
-    # Creates Cocina Administrative objects from Fedora objects
+    # Creates Cocina::Administrative object properties from Fedora objects
     class Administrative
-      # @param [Dor::Item,Dor::Etd] item
-      # @return [Hash] a hash that can be mapped to a cocina administrative model
-      def self.props(item)
-        new(item).props
+      # @param [Dor::Item,Dor::Etd,Dor::Collection] fedora_object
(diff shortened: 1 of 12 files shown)


app/services/cocina/from_fedora/administrative.rb
```diff
@@ -4,32 +4,32 @@ module Cocina
   module FromFedora
     # Creates Cocina Administrative objects from Fedora objects
     class Administrative
-      # @param [Dor::Item,Dor::Etd] item
+      # @param [Dor::Item,Dor::Etd,Dor::Collection] fedora_object
```
ndushay, line 7: calling this "item" was confusing, since this class is also used for Dor::Collection

app/services/cocina/from_fedora/collection.rb
```diff
@@ -4,39 +4,39 @@ module Cocina
   module FromFedora
     # Creates Cocina Collection objects from Fedora objects
     class Collection
-      # @param [Dor::Item,Dor::Etd] item
+      # @param [Dor::Collection] fedora_collection
```
ndushay, line 7: calling this "item" was misleading, as was the existing comment which was incorrect about the object type of the param

aaron-collier (approved):

PR #2799, opened 2021-04-30
Normalizes reading order.

justinlittman:
## Why was this change made?
Better roundtrip validation.

Unit, local

NA


ndushay (approved):
LGTM!

PR #2807, opened 2021-04-30
Handle leading and trailing spaces in source id.

### LINKED ISSUE #2804: Handle leading space from source id
For example druid:sn266tw5387:

```
<identityMetadata>
  <sourceId source="sul"> M0443_S2_D-K_B9_F33_011</sourceId>
</identityMetadata>
```

justinlittman:
closes #2804

## Why was this change made?
Handle bad data when mapping,

Unit, local

NA

### DIFF
--- a/app/services/cocina/from_fedora/identification.rb
+++ b/app/services/cocina/from_fedora/identification.rb
@@ -28,7 +28,7 @@ module Cocina
 
       def source_id
         if fedora_object.source_id
-          fedora_object.source_id
+          fedora_object.source_id.strip.sub(/ *: */, ':')
         elsif fedora_object.is_a? Dor::Collection
           nil
         else
--- a/spec/services/cocina/mapping/identification/dro_spec.rb
+++ b/spec/services/cocina/mapping/identification/dro_spec.rb
@@ -963,7 +963,7 @@ RSpec.describe 'Fedora Item identityMetadata <--> Cocina DRO Identification mapp
           label: label,
           version: 1,
           identification: {
-            sourceId: "#{source_id_source}:#{source_id}"
+            sourceId: "#{source_id_source.strip}:#{source_id.strip}"
           },
           administrative: {
             hasAdminPolicy: admin_policy_id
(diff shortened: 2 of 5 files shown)


app/services/cocina/from_fedora/identification.rb
```diff
@@ -28,7 +28,7 @@ def props
 
       def source_id
         if fedora_object.source_id
-          fedora_object.source_id
+          fedora_object.source_id.strip.sub(/ *: */, ':')
```
mjgiarlo, line 31: Is the intent here to remove all space characters from the source id regardless of location? If so, I wonder how you might feel about `fedora_object.source_id.delete(' ')` instead of the regex (and the strip). This will only handle space characters, of course, not newlines or tabs, etc.

spec/services/cocina/mapping/identification/dro_spec.rb
```diff
@@ -963,7 +963,7 @@
           label: label,
           version: 1,
           identification: {
-            sourceId: "#{source_id_source}:#{source_id}"
+            sourceId: "#{source_id_source.strip}:#{source_id.strip}"
```
mjgiarlo, line 966: Hmm, can you say more about this change?

app/services/cocina/from_fedora/identification.rb
justinlittman: No, it is to handle spaces where indicated in the following by an underscore: `_sul_:_abc123_`.

I don't know if a space is OK in other places, no leaving intact.

spec/services/cocina/mapping/identification/dro_spec.rb
justinlittman: Yes, before this PR leading and trailing spaces weren't being stripped when mapping from cocina. Now they are.

app/services/cocina/from_fedora/identification.rb
mjgiarlo: @justinlittman OK, perfect.

spec/services/cocina/mapping/identification/dro_spec.rb
mjgiarlo: I see. So this is ensuring the mapping is doing the space normalization. Thanks!

mjgiarlo (approved):

PR #2811, opened 2021-04-30
adjust rights normalization for licenses

### LINKED ISSUE #2797: rightsMetadata.xml needs license normalization code
There are many examples in services/cocina/mapping/access/dro_access_spec.rb  and .../collection_access_spec.rb 

Here is one example:

```
        <rightsMetadata>
            <access type="discover">
stuff
            </access>
            <use>
              <human type="useAndReproduction">blah blah</human>
              <human type="creativeCommons">Attribution Non-Commercial, No Derivatives 3.0 Unported</human>
              <machine type="creativeCommons">by-nc-nd</machine>
            </use>
            <copyright>
              <human>&#xA9;2021 Wingnut and Vinsky publishing</human>
            </copyright>
          </rightsMetadata>
```

becomes, after roundtripping through cocina:

```
        <rightsMetadata>
            <access type="discover">
stuff  
          </access>
            <use>
              <human type="useAndReproduction">blah blah</human>
              <license>https://creativecommons.org/licenses/by-nc-nd/3.0/</license>
            </use>
            <copyright>
              <human>&#xA9;2021 Wingnut and Vinsky publishing</human>
            </copyright>
          </rightsMetadata>
```

We should have "normalization" code, to ignore this difference when comparing roundtripped Fedora xml .  This normalization code should also be used for APO `defaultObjectRights.xml` which is ticketed as #2709.

The normalization could/should be executed by bin/(validate-cocina-roundtrip) scripts as appropriate and it should also be executed by spec/services/cocina/mapping/access/xxx_access_spec.rb so the specs compare the actual roundtripped Fedora rightsMetadata.xml to the rightsMetadata.xml that is normalized by executing said normalization code.

peetucket:
## Why was this change made?

Fixes #2797 - adjust rights normalization to deal with license code style differences

1. remove the "roundtrip" normalizer (which used to normalize the XML that came back from the roundtrip), which is no longer necessary, and eliminate some methods called by it
2. change the normalizer to take in a datastream as a parameter instead of xml, since the method I am calling to convert to license URIs takes a datastream

Updated tests and tested on sdr-deploy for 100 objects

### DIFF
--- a/app/services/cocina/normalizers/rights_normalizer.rb
+++ b/app/services/cocina/normalizers/rights_normalizer.rb
@@ -9,6 +9,8 @@ module Cocina
 
       # @param [Nokogiri::Document] the rights datastream to be normalized
       # @return [Nokogiri::Document] normalized rights xml
+      # Note: this is different than other normalizers in that it takes in a datastream as a parameter instead of the XML
+      #  because we use an existing class below to fetch the license URI and this class requires a datastream
       def self.normalize(datastream:)
         new(datastream: datastream).normalize
       end
@@ -44,7 +46,7 @@ module Cocina
         end
         # now add new <license> node
         license_uri = Cocina::FromFedora::Access::License.find(datastream)
-        new_license_node = Nokogiri::XML::Node.new('license', Nokogiri::XML(nil))
+        new_license_node = Nokogiri::XML::Node.new('license', ng_xml)
         new_license_node.content = license_uri
         ng_xml.at('//use') << new_license_node if license_uri.present?
         ng_xml.root.xpath('//use[count(*) = 0]').each(&:remove)
--- a/app/services/cocina/normalizers/rights_normalizer.rb
+++ b/app/services/cocina/normalizers/rights_normalizer.rb
@@ -7,25 +7,20 @@ module Cocina
     class RightsNormalizer
       include Cocina::Normalizers::Base
 
-      # @param [Nokogiri::Document] rights_ng_xml rights XML to be normalized
+      # @param [Nokogiri::Document] the rights datastream to be normalized
       # @return [Nokogiri::Document] normalized rights xml
-      def self.normalize(rights_ng_xml:)
-        new(rights_ng_xml: rights_ng_xml).normalize
+      def self.normalize(datastream:)
+        new(datastream: datastream).normalize
       end
 
-      # @param [Nokogiri::Document] roundtripped rights_ng_xml rights XML to be normalized
-      # @return [Nokogiri::Document] normalized roundtripped rights xml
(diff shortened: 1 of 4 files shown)


peetucket:
FYI @ndushay and @jmartin-sul , this is the start the work we discussed...will pick up and actually do the bit to add the license code later, but I found a class that may do what we need: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/to_fedora/license.rb

peetucket:
Note the test failure is expected ... it will pass when the implementation of the license transformation is complete

app/services/cocina/normalizers/rights_normalizer.rb
```diff
@@ -7,61 +7,45 @@ module Normalizers
     class RightsNormalizer
       include Cocina::Normalizers::Base
 
-      # @param [Nokogiri::Document] rights_ng_xml rights XML to be normalized
+      # @param [Nokogiri::Document] the rights datastream to be normalized
       # @return [Nokogiri::Document] normalized rights xml
-      def self.normalize(rights_ng_xml:)
-        new(rights_ng_xml: rights_ng_xml).normalize
+      def self.normalize(datastream:)
```
peetucket, line 12: passing in the datatream here instead of the XML itself since the license URI adding method I want to use takes the datastream as the input

app/services/cocina/normalizers/rights_normalizer.rb
```diff
@@ -7,61 +7,46 @@ module Normalizers
...
-      def remove_license
+      def normalize_license_to_uri
+        # first remove old style license nodes
         ['openDataCommons', 'creativeCommons'].each do |license_type|
           ng_xml.root.xpath("//use/machine[@type='#{license_type}' and text()]").each(&:remove)
           ng_xml.root.xpath("//use/human[@type='#{license_type}' and text()]").each(&:remove)
         end
+        # now add new <license> node
+        license_uri = Cocina::FromFedora::Access::License.find(datastream)
+        new_license_node = Nokogiri::XML::Node.new('license', Nokogiri::XML(nil))
```
ndushay, line 47: I *think* the second argument here should be ng_xml -- https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node#new-class_method implies the second argument is about telling ng when it can garbage collect on the new node

ndushay (changes requested):
one small thing and a larger ... wait ... where are we? as I'm trying to do something that may depend on this work.

ndushay (changes requested):
@peetucket did you have to do anything to validate-cocina-roundtrip script for the rights normalizer to take the datastream?

I think that is needed or it will break that script.

ndushay (approved):

PR #2819, opened 2021-05-03
rightsMetadata.xml mapping "unit" specs: wire in rights_normalizer

ndushay:
## Why was this change made?

- to keep future changes to rights_normalizer from inadvertently breaking other expected rightsMetadata.xml mappings
- to ensure our mapping tests are aligned with current rights_normalizer code

part of #2701

it's all unit tests;  wiring in the unit tests found an error in the rights_normalizer code, which is fixed and included here.

### DIFF
--- a/app/services/cocina/normalizers/rights_normalizer.rb
+++ b/app/services/cocina/normalizers/rights_normalizer.rb
@@ -34,10 +34,6 @@ module Cocina
 
       attr_reader :ng_xml, :datastream
 
-      def license_nodes(xml)
-        xml.root.xpath('//license')
-      end
-
       def normalize_license_to_uri
         # first remove old style license nodes
         ['openDataCommons', 'creativeCommons'].each do |license_type|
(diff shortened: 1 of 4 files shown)


app/services/cocina/normalizers/rights_normalizer.rb
```diff
@@ -34,10 +34,6 @@ def normalize
 
       attr_reader :ng_xml, :datastream
 
-      def license_nodes(xml)
-        xml.root.xpath('//license')
-      end
-
```
ndushay, line 40: this has become dead code.

spec/services/cocina/mapping/access/collection_access_spec.rb
```diff
@@ -117,6 +133,16 @@
...
+              <access type="discover">
+                <machine>
+                  <none/>
+                </machine>
+              </access>
+              <access type="read">
+                <machine>
+                  <none/>
+                </machine>
+              </access>
```
ndushay, line 143: the original test was just taking the defaults from cocina, but I suspect we don't have any such data in Fedora, so I changed the data.

app/services/cocina/normalizers/rights_normalizer.rb
```diff
@@ -46,9 +42,15 @@ def normalize_license_to_uri
         end
         # now add new <license> node
         license_uri = Cocina::FromFedora::Access::License.find(datastream)
-        new_license_node = Nokogiri::XML::Node.new('license', ng_xml)
-        new_license_node.content = license_uri
-        ng_xml.at('//use') << new_license_node if license_uri.present?
+        if license_uri.present?
+          existing_use_element = ng_xml.at('//use')
+          if existing_use_element.content.strip != license_uri
```
ndushay, line 47: we only want one license_uri node

justinlittman (approved):

PR #2820, opened 2021-05-03
update identity metadata specs to reflect what is happening with escaping

### LINKED ISSUE #2816: identityMetadata roundtripping:  do not double escape ampersands in objectLabel
do not double escape ampersands in objectLabel (bb001zc5754 - DRO).

(or -- check for and remove double escaping before writing out XML)

test written in dro identification mapping specs

peetucket:
## Why was this change made?

Fixes #2816 - avoid double escaping in identity metadata on roundtrip

Unxited an existing test

### DIFF
--- a/spec/services/cocina/mapping/identification/dro_spec.rb
+++ b/spec/services/cocina/mapping/identification/dro_spec.rb
@@ -1175,10 +1175,10 @@ RSpec.describe 'Fedora Item identityMetadata <--> Cocina DRO Identification mapp
   end
 
   context 'with 2 collections and objectLabel with xml encoding' do
-    # it_behaves_like 'DRO Identification Fedora Cocina mapping' do
-    xit 'to be implemented: do not double XML escape ampersands in objectLabel' do
+    it_behaves_like 'DRO Identification Fedora Cocina mapping' do
       let(:item_id) { 'druid:bb001zc5754' }
-      let(:label) { 'French Grand Prix &amp; 12 Hour Rheims 7/4/1954' }
+      let(:label) { 'French Grand Prix & 12 Hour Rheims 7/4/1954' }
+      let(:xml_escaped_label) { 'French Grand Prix &amp; 12 Hour Rheims 7/4/1954' }
       let(:admin_policy_id) { 'druid:qv648vd4392' } # from RELS-EXT
       let(:collection_ids) { ['druid:nt028fd5773', 'druid:wy149zp6932'] } # from RELS-EXT
       let(:source_id_source) { 'Revs' }
@@ -1189,7 +1189,7 @@ RSpec.describe 'Fedora Item identityMetadata <--> Cocina DRO Identification mapp
             <sourceId source="#{source_id_source}">#{source_id}</sourceId>
             <objectId>#{item_id}</objectId>
             <objectCreator>DOR</objectCreator>
-            <objectLabel>#{label}</objectLabel>
+            <objectLabel>#{xml_escaped_label}</objectLabel>
             <objectType>item</objectType>
             <adminPolicy>#{admin_policy_id}</adminPolicy>
             <otherId name="uuid">b9af2444-2525-11e2-a4c7-0050569b52d5</otherId>
@@ -1204,7 +1204,7 @@ RSpec.describe 'Fedora Item identityMetadata <--> Cocina DRO Identification mapp
             <sourceId source="#{source_id_source}">#{source_id}</sourceId>
             <objectId>#{item_id}</objectId>
             <objectCreator>DOR</objectCreator>
-            <objectLabel>#{label}</objectLabel>
+            <objectLabel>#{xml_escaped_label}</objectLabel>
             <objectType>item</objectType>
           </identityMetadata>
         XML
@@ -1334,7 +1334,8 @@ RSpec.describe 'Fedora Item identityMetadata <--> Cocina DRO Identification mapp
   context 'with &#x escaped chars in objectLabel (FRDA)' do
     it_behaves_like 'DRO Identification Fedora Cocina mapping' do
       let(:item_id) { 'druid:bb016nw8128' }
-      let(:label) { 'E&#x301;ve&#x301;nement du 19 fevrier 1790' }
+      let(:label) { 'Événement du 19 fevrier 1790' }
+      let(:xml_escaped_label) { 'E&#x301;ve&#x301;nement du 19 fevrier 1790' }
       let(:admin_policy_id) { 'druid:ht275vw4351' } # from RELS-EXT
       let(:collection_ids) { ['druid:jh957jy1101'] } # from RELS-EXT
       let(:source_id_source) { 'French Revolution Digital Archive' }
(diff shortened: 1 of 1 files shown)


app/services/cocina/to_fedora/identity.rb
```diff
@@ -11,7 +11,7 @@ def self.apply(fedora_object, label:, agreement_id: nil)
         fedora_object.objectId = fedora_object.pid
         fedora_object.objectCreator = 'DOR'
         # Label may have already been set when setting descriptive metadata.
-        fedora_object.objectLabel = label if fedora_object.objectLabel.empty?
+        fedora_object.objectLabel = CGI.unescapeHTML(label) if fedora_object.objectLabel.empty?
```
peetucket, line 14: I think the `objectLabel` setting on fedora object is escaping, so we want to be sure we start with an unescaped value before we set it in fedora.

spec/services/cocina/mapping/identification/dro_spec.rb
```diff
@@ -1354,14 +1353,13 @@
...
         XML
       end
-      # NOTE: label: xml escapes ampersands in unicode hex characters
       let(:roundtrip_identity_metadata_xml) do
         <<~XML
           <identityMetadata>
             <sourceId source="#{source_id_source}">#{source_id}</sourceId>
             <objectId>#{item_id}</objectId>
             <objectCreator>DOR</objectCreator>
-            <objectLabel>E&amp;#x301;ve&amp;#x301;nement du 19 fevrier 1790</objectLabel>
```
peetucket, line 1364: This test actually had an object label explicitly coming back as different after the roundtripping before this change, and now it comes back exactly the same.  So I'm not sure... shouldn't it come back exactly the same after roundtripping?

spec/services/cocina/mapping/identification/dro_spec.rb
```diff
@@ -1175,8 +1175,7 @@
   end
 
   context 'with 2 collections and objectLabel with xml encoding' do
-    # it_behaves_like 'DRO Identification Fedora Cocina mapping' do
-    xit 'to be implemented: do not double XML escape ampersands in objectLabel' do
+    it_behaves_like 'DRO Identification Fedora Cocina mapping' do
```
peetucket, line 1178: the test that was unskipped that now passes

app/services/cocina/to_fedora/identity.rb
peetucket: Here is the definition in the datastream in `dor-services`: https://github.com/sul-dlss/dor-services/blob/main/lib/dor/datastreams/identity_metadata_ds.rb#L17

and where it gets delegated to from the object level: https://github.com/sul-dlss/dor-services/blob/main/lib/dor/models/abstract.rb#L81

Perhaps ActiveFedora is doing the encoding as it generates the HTML

jcoyne: I'm curious why you are bringing HTML into this.  The original ticket was about ampersand escape in XML.

jcoyne (changes requested):
The problem is not in this code.  The problem is that someone injected a literal `&amp;` in the JSON source (on line 69 and originally https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/identification/dro_spec.rb#L1180)  JSON doesn't think this is a special escape sequence, so this code was behaving correctly.

peetucket:
> The problem is not in this code. The problem is that someone injected a literal `&amp;` in the JSON source (on line 69, 56, 41 and originally https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/identification/dro_spec.rb#L1180) JSON doesn't think this is a special escape sequence, so this code was behaving correctly.

I think the problem is that we have some data with an `&amp;` in the source which is getting double encoded?  That example you reference is from an actual object we have: https://argo.stanford.edu/view/druid:bb001zc5754

I unskipped the spec that had it and made it pass by unescaping. But the HTML unescape is probably overkill, I believe there are only a handful of characters than need to be XML escaped.

jcoyne:
@peetucket but that `:label` value is being used both as an xml snippet, where it should be encoded and as the stubbed return value from the label method on an object, where it wouldn't be encoded.

peetucket:
> @peetucket but that `:label` value is being used both as an xml snippet, where it should be encoded and as the stubbed return value from the label method on an object, where it wouldn't be encoded.

Yeah, I see what you mean...though i guess I'd expect the starting XML snippet (the stubbed `identity_metadata_xml`) which has `<objectLabel>#{label}</objectLabel>` to equal the round-tripped XML snippet (the stubbed `roundtrip_identity_metadata_xml`), which is what the unskipped spec is asserting: https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/identification/dro_spec.rb#L1177

 Isn't it just trying to assert that whatever XML you start with (with encoding, etc.) , is what you end up with after going through cocina and back again?  Sorry if I'm confused, maybe easier to do in person .

jcoyne:
If you were to use a real ActiveFedora object (rather than a stub), you get the un-escaped value. Which would be the same as doing `Nokogiri::XML(xml).xpath('//objectLabel').first.content`

spec/services/cocina/mapping/identification/dro_spec.rb
```diff
@@ -1334,7 +1334,8 @@
   context 'with &#x escaped chars in objectLabel (FRDA)' do
     it_behaves_like 'DRO Identification Fedora Cocina mapping' do
       let(:item_id) { 'druid:bb016nw8128' }
-      let(:label) { 'E&#x301;ve&#x301;nement du 19 fevrier 1790' }
+      let(:label) { 'Événement du 19 fevrier 1790' }
+      let(:escaped_label) { 'E&#x301;ve&#x301;nement du 19 fevrier 1790' }
```
ndushay, line 1338: what do you think about renaming this to `utf8_escaped_label` ?   "escaped" could be HTML, XML, UTF-8 ...

spec/services/cocina/mapping/identification/dro_spec.rb
ndushay: I think we see the ? UTF-8 escaped ? characters in the lightbox in Argo;  Actual Fedora UI show them as NOT UTF-8 escaped:  https://sul-dor-prod.stanford.edu/fedora/objects/druid:bb016nw8128/datastreams/identityMetadata/content   but that is also a web browser view.   So no idea how they are in native Fedora.

equivalent-xml  and/or our code is set up to successfully match the string whether utf-8 escaped or not.

So ultimately, we will either find problems for this roundtripping actual data with the scripts or, this test, as it stands ... is fine???

As to your question:  Because there is a diff between Argo lightbox and Fedora UI display ... "exactly the same" may be tricky, and if we use the definition from equivalent-xml and our encoding settings, that may be sufficient.

ndushay (changes requested):
request for a var name change, o.w. looks good.

spec/services/cocina/mapping/identification/dro_spec.rb
jcoyne: I would say this is an `xml_escaped_label`  if this were just utf-8 bytes they would be `"\u0301"`

ndushay: ok, I'm good with xml_escaped_label :-)

ndushay (approved):

PR #2821, opened 2021-05-03
Publish Cocina models to purl

### LINKED ISSUE #734: Publish should send cocina model to PURL
We can leave all XML in place for now, but also send the cocina json similarly to how we send public xml right now. https://github.com/sul-dlss/dor-services-app/blob/master/app/services/publish_metadata_service.rb#L36

we will want to filter out any non-public files similar to this: https://github.com/sul-dlss/dor-services-app/blob/master/app/services/public_xml_service.rb#L82-L83

Then we can add a .json endpoint in purl
https://github.com/sul-dlss/purl/blob/master/app/controllers/purl_controller.rb#L19-L21

jcoyne:
## Why was this change made?

So that we can begin sharing the new Cocina models with the Access team.   This change sanitizes the structural by removing any non-published fileset/files just as is done in the public xml.

Fixes #734

To be tested on stage.

### DIFF
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -30,6 +30,7 @@ module Publish
 
     # @raise [Dor::DataError]
     def transfer_metadata(release_tags)
+      transfer_to_document_store(PublicCocinaService.create(item), 'cocina.json')
       transfer_to_document_store(PublicXmlService.new(item, released_for: release_tags).to_xml, 'public')
       transfer_to_document_store(PublicDescMetadataService.new(item).to_xml, 'mods')
     end
--- a/app/services/publish/public_cocina_service.rb
+++ b/app/services/publish/public_cocina_service.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module Publish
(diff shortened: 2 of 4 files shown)


andrewjbtw:
In keeping with the change we made to remove tags from the identityMetadata, we should also remove the Project tag from the published Cocina.

app/services/publish/metadata_transfer_service.rb
```diff
@@ -30,6 +30,7 @@ def publish
 
     # @raise [Dor::DataError]
     def transfer_metadata(release_tags)
+      transfer_to_document_store(PublicCocinaService.create(item), 'cocina')
```
cbeer, line 33: I'd add a file extension for developer happiness; I think the current lack of extensions pre-dated purl-the-application (or otherwise is an artifact of previous system design goals).

```suggestion
      transfer_to_document_store(PublicCocinaService.create(item), 'cocina.json')
```

jcoyne: I've added it.  Thanks for the suggestion.

aaron-collier (approved):

PR #2822, opened 2021-05-04
Avoid using fixture data

jcoyne:
This only affects the test suite.

## Why was this change made?

In one place here it was using item data as fixture for a collection.


spec/services/publish/metadata_transfer_service_spec.rb
```diff
@@ -120,7 +126,7 @@
       end
 
       context 'with a collection object' do
-        let(:item) { instantiate_fixture('druid:bc123df4567', Dor::Collection) }
+        let(:item) { Dor::Collection.new }
```
jmartin-sul, line 129: should this `let` var be renamed to `collection`?

jcoyne: No, it needs to be item for https://github.com/sul-dlss/dor-services-app/blob/df8a348ba66e57de9b84fc2274b1b4aa825ac2c3/spec/services/publish/metadata_transfer_service_spec.rb#L18

jmartin-sul (approved):

PR #2824, opened 2021-05-04
Use filter_map rather than map.compact

jcoyne:
## Why was this change made?

filter_map has better performance and is recommended by rubocop


ndushay (approved):
filter_map ... to the rescue!!! 🦸   ⛑️  

(without seeing this PR, I wouldn't have known about filter_map ... except from next time rubocop tells me)

PR #2825, opened 2021-05-04
identityMetadata.xml mapping "unit" specs: wire in identity_normalizer

### LINKED ISSUE #2814: identityMetadata normalizations needed
Examples for these can be found in
- https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/identification/xxx_spec.rb

When identityMetadata.xml has:
- [x] if no sourceId and yes dissertationId, convert it to a sourceId

Note that identityMetadata keeps cruft like this, because it is updated on a roundtrip, not created:
- [x] otherId of type uuid -> normalize out (keep catkey, barcode, dissertationId ...)
- [x] (regular) tags -> normalize out
- [x] agreementId, adminPolicy, collectionId -> normalize out (we use RELS-EXT)
- [x] for collections with multiple objectType:  keep collection type and drop set type (set is vestigial early SDR junk)
- [x] displayType -> normalize out (vestigial early SDR junk)
- [x] objectAdminClass -> normalize out (vestigial early SDR junk)
- [x] citationTitle, citationCreator -> normalize out (vestigial early SDR/ETD junk)

It's okay to ticket these separately or to do some of them but not all -- just be clear as to which are done and don't close this ticket until all are accounted for.

ndushay:
## Why was this change made?

- to keep future changes to identity_normalizer from inadvertently breaking other expected identityMetadata.xml mappings
- to ensure our mapping tests are aligned with current identity_normalizer code

Closes #2814
Part of #2703

it's all specs; wiring in the unit tests found some errors in the identity_normalizer and mapping code, which are fixed and included here.

### DIFF
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -27,8 +27,6 @@ module Cocina
                                     ignore_resource_type_errors: project_phoenix?)
           structural[:contains] = contains if contains.present?
 
-          structural[:hasAgreement] = fedora_item.identityMetadata.agreementId.first unless fedora_item.identityMetadata.agreementId.empty?
-
           begin
             # Note that there is a bug with fedora_item.collection_ids that returns [] until fedora_item.collections is called.
             # Below side-steps this.
--- a/app/services/cocina/normalizers/identity_normalizer.rb
+++ b/app/services/cocina/normalizers/identity_normalizer.rb
@@ -19,20 +19,97 @@ module Cocina
       end
 
       def normalize
-        normalize_source_id
+        normalize_dissertation_id_to_source_id
+        normalize_out_uuid
+        normalize_out_admin_tags
+        normalize_out_admin_policy
+        normalize_out_agreement_id
+        normalize_out_set_object_type
+        normalize_out_display_type
+        normalize_out_object_admin_class
+        normalize_out_citation_elements
+        normalize_out_call_sequence_ids
+        normalize_out_empty_other_ids
+        normalize_source_id_whitespace
 
-        regenerate_ng_xml(ng_xml.to_s)
+        regenerate_ng_xml(ng_xml.to_xml)
       end
 
       private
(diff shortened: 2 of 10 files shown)


app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -27,8 +27,6 @@ def props
                                     ignore_resource_type_errors: project_phoenix?)
           structural[:contains] = contains if contains.present?
 
-          structural[:hasAgreement] = fedora_item.identityMetadata.agreementId.first unless fedora_item.identityMetadata.agreementId.empty?
```
ndushay, line 30: Per Andrew:

> items don’t need to have agreements anywhere
> the reference back to agreements is just supposed to be in the APO RELS-EXT

jcoyne: Would you mind making this change in a separate PR. I think this is a really big change that deserves a good PR description just about the motivation for this.  I have many questions about this and I wouldn't want this discussion to hold up the rest of this PR.

justinlittman (commented):
Agree with @jcoyne. Otherwise, I approve.

app/services/cocina/from_fedora/dro_structural.rb
ndushay: the property is needed for APOs, just not for items.  APOs has the property already in administrative as referencesAgreement in administrative.  

I'm inclined to leave it in here, as the tests all pass, blah blah, and to make a ticket for cocina-models and come at it that way.  Will that work for you?

https://github.com/sul-dlss/cocina-models/issues/259

ndushay:
@justinlittman, @jcoyne  I added a commit that removes the agreementId from to_fedora/identity.rb mapping.  There is a ticket to remove agreementID from DROStructural: sul-dlss/cocina-models/issues/259

I think we agreed that makes this mergable.

jcoyne (approved):

PR #2830, opened 2021-05-06
New license public descriptive

jcoyne:
## Why was this change made?
So that searchworks can show the new licenses.


justinlittman (approved):

andrewjbtw:
Looking at this fixture object: https://sul-purl-stage.stanford.edu/dc224fz4940

The public-facing display looks good with the Apache license. But in the non-displayed `<rightsMetadata>` there are references to creative commons:

```xml
<use>
    <human type="creativecommons">Attribution Non-Commercial Share Alike license</human>
    <machine type="creativecommons">by-nc-sa</machine>
    <license>http://www.apache.org/licenses/LICENSE-2.0</license>
</use>
```
My understanding from the meeting with the Access team is that we need to supply additional `human` and `machine` information for the new licenses.

jcoyne:
@andrewjbtw what I found out by looking at the code is that they are looking in the accessConditions for that data. And that's what this PR manipulated.


That extra use license is odd, but that's not a part of this PR.  That does look like a bug though.  That is because when we update from cocina, it's not completely overwriting the default license (which was an APO default)

jcoyne:
Also needs to resolve https://app.honeybadger.io/projects/50568/faults/79245511

PR #2838, opened 2021-05-07
Add support for roundtrip validation using object create.

### LINKED ISSUE #2837: Roundtrip validation of object create
To complement roundtrip validation of object update and help identify mapping errors and unmapped data.

justinlittman:
closes #2837

## Why was this change made?
Object create will help find unmapped, existing data that object update does not.

Unit, local

NA

### DIFF
--- a/bin/validate-cocina-roundtrip
+++ b/bin/validate-cocina-roundtrip
@@ -11,11 +11,12 @@ require 'diffy'
 require 'dor/services/client'
 require 'equivalent-xml'
 
-options = { random: false, druids: [], fast: false }
+options = { random: false, druids: [], fast: false, create: false }
 parser = OptionParser.new do |option_parser|
   option_parser.banner = 'Usage: bin/validate-cocina-roundtrip [options]'
 
   option_parser.on('-sSAMPLE', '--sample SAMPLE', Integer, 'Sample size, otherwise all druids.')
+  option_parser.on('-c', '--create', 'Run object create instead of object update.')
   option_parser.on('-r', '--random', 'Select random druids.')
   option_parser.on('-f', '--fast', 'Without content metadata.')
   option_parser.on('-dDRUIDS', '--druids DRUIDS', Array, 'List of druids (instead of druids.txt).')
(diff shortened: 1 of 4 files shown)


bin/validate-cocina-roundtrip
```diff
@@ -144,7 +151,13 @@ def validate_druid(druid, loader, fast: false)
     return :unmapped
   end
 
-  fedora_obj.contentMetadata.content = "<contentMetadata type='#{fedora_obj.contentMetadata.contentType.first}' />" if fast && fedora_obj.datastreams.include?('contentMetadata')
+  if fast && fedora_obj.datastreams.include?('contentMetadata')
```
jcoyne, line 154: Is this just a fancy way of saying `fedora_obj.is_a?(Dor::Item)`?

bin/validate-cocina-roundtrip
```diff
@@ -144,7 +151,13 @@ def validate_druid(druid, loader, fast: false)
...
+    fedora_obj.contentMetadata.content = if create
+                                           "<contentMetadata type='#{fedora_obj.contentMetadata.contentType.first}' objectId='#{druid}'/>"
```
jcoyne, line 156: I'm not sure I understand this.  You are reading out of the xml and then overwriting the very same? Why?

bin/validate-cocina-roundtrip
justinlittman: Hmmm ... I have no idea. This is really just about manipulating the content metadata datastream.

bin/validate-cocina-roundtrip
justinlittman: Yes, because processing the content metadata datastream is REALLY SLOW because it is often quite large. To speed up testing, the `--fast` flag replaces the content metadata with a dummy that processes much faster. Of course, this means the content metadata datastream isn't being tested, but sometimes that's OK.

bin/validate-cocina-roundtrip
ndushay: we have items without files ...

jcoyne: All Items have content metadata even if they don't have files.  In that case the xml looks like this:
```xml
<contentMetadata objectId="druid:zb652jb8517" type="file"/>
```

justinlittman:
@ndushay  What do you mean by "disruptive in production"?

ndushay:
@justinlittman if this will cause more validations to fail, thus making existing objects less usable in whatever way they are manipulated.

ndushay (approved):

PR #2842, opened 2021-05-07
Maps release tags from cocina to fedora.

### LINKED ISSUE #2834: map release tags from cocina to fedora (identityMetadata)
Currently we map releaseTags into cocina, but we don't round trip them.  We currently use the releaseTags API for doing writing.  Once we have the releaseTags updated from cocina, then we could deprecate the special purpose releaseTags API.

See #2815

justinlittman:
closes #2834

## Why was this change made?
Complete roundtrip mapping.

Unit

NA

### DIFF
--- a/app/services/cocina/to_fedora/identity.rb
+++ b/app/services/cocina/to_fedora/identity.rb
@@ -5,15 +5,52 @@ module Cocina
     # This transforms the cocina identification information to the Fedora 3 data model identityMetadata
     class Identity
       # @param [Dor::Item,Dor::Collection,Dor::Etd,Dor::AdminPolicyObject] fedora_object
+      def self.initialize_identity(fedora_object)
+        new(fedora_object).initialize_identity
+      end
+
+      # @param [Dor::Item,Dor::Collection,Dor::Etd,Dor::AdminPolicyObject] fedora_object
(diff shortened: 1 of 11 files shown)


app/services/cocina/to_fedora/identity.rb
```diff
@@ -4,16 +4,53 @@ module Cocina
   module ToFedora
     # This transforms the cocina identification information to the Fedora 3 data model identityMetadata
     class Identity
+      # @param [Dor::Item,Dor::Collection,Dor::Etd,Dor::AdminPolicyObject] fedora_object
```
justinlittman, line 7: Splitting the single apply method that got data from various parts of cocina object into separate methods made this less of a muddled mess and made it clearer in object_updater and object_creator what was happening.

spec/requests/update_metadata_spec.rb
```diff
@@ -562,10 +562,7 @@
         expect(response.status).to eq 200
 
         # Identity metadata set correctly.
-        expect(item.objectId).to eq(druid)
```
justinlittman, line 565: This shouldn't be set on an update, so gone.

jcoyne (approved):

PR #2836, opened 2021-05-06
updated abstract mapping

peetucket:
## Why was this change made?

Fixes #2728  - updated abstract mapping

Note that a number of other tests had to be updated to reflect the updated mappings, see https://github.com/sul-dlss/dor-services-app/issues/2728#issuecomment-833898287

## How was this change tested?

Unskipped tests in #2727 and updated other existing tests as needed with Arcadia's approval.

PR #2846, opened 2021-05-12
Add note tests

arcadiafalcone:
## Why was this change made?
Add mappings for notes that become abstracts

n/a

n/a

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/notes.rb
+++ b/app/services/cocina/from_fedora/descriptive/notes.rb
@@ -28,14 +28,14 @@ module Cocina
         def abstracts
           all_abstract_nodes = resource_element.xpath('mods:abstract', mods: DESC_METADATA_NS).select { |node| note_present?(node) }
           altrepgroup_abstract_nodes, other_abstract_nodes = AltRepGroup.split(nodes: all_abstract_nodes)
-          other_abstract_nodes.map { |node| common_note_for(node).merge({ type: abstract_type(node) }) } + \
+          other_abstract_nodes.map { |node| common_note_for(node).merge(abstract_type(node)) } + \
             altrepgroup_abstract_nodes.map { |parallel_nodes| parallel_abstract_for(parallel_nodes) }
         end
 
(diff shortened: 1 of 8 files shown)


app/services/cocina/from_fedora/descriptive/notes.rb
```diff
@@ -70,7 +70,7 @@ def abstract_type(node, parallel: false)
 
         def note_type(node)
           if node['type'].present? && node['type'] != 'summary'
-            { type: node['type'] }
+            { type: node['type'].downcase }
```
peetucket, line 73: I actually had this downcased in my original PR, but removed it after running validations against 10k random druids and finding new roundtripping errors for note nodes.  It is not currently downcased in the main branch.

 I then added tests to ensure the case is preserved: https://github.com/sul-dlss/dor-services-app/pull/2836/files#diff-ea4ff0bb6abd64bf28b19eaa84a65730ea4ee136b92a758f7f42f3e3c1e852b6R47-R94

peetucket:
I think the general problem I was having was that even though the unskipped tests were passing, when I ran the actual validation script, the error rate increased.  Are these new tests coming from the failed druids.

arcadiafalcone:
Yes, they're based on the errors that I saw in looking through the validation test results -- previously the tests didn't represent all cases where notes correctly turn into abstracts when round-tripping back to MODS,

peetucket:
> Yes, they're based on the errors that I saw in looking through the validation test results -- previously the tests didn't represent all cases where notes correctly turn into abstracts when round-tripping back to MODS,

Ok, that's cool.  Based on what I saw before, I'm thinking the downcasing of the note type in this PR may cause the roundtripping errors to come back again though.

arcadiafalcone:
I'll remove it - I didn't think it should be there based on my understanding of the code, but only made the edit because I was trying to resolve conflicts.

peetucket:
I'm going to try and unskip the new tests on the this branch and push commits here before merging to the other branch

peetucket (approved):

PR #2847, opened 2021-05-12
Update H2 form spec to match new mapping

arcadiafalcone:
## Why was this change made?
Part of https://github.com/sul-dlss/happy-heron/issues/1392

n/a

n/a

Implementation ticket #2869


ndushay (approved):
LGTM;  is there going to be a ticket for a developer to implement this?  Is it a priority for release 1.0 of h2?

PR #2857, opened 2021-05-24
Upgrade puma

ndushay:
## Why was this change made?

there was a CVE

unit tests so far (integration tests after this is merged)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,20 +3,6 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-gem 'rails', '~> 5.2.0'
-
-# Use Puma as the app server
-gem 'puma', '~> 3.12'
-
-group :development do
-  gem 'dor-services-client' # used by bin/validate-cocina-roundtrip
-  gem 'listen', '~> 3.0.5'
-  gem 'rubyzip', '>= 1.0.0'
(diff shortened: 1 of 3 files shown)


jcoyne (changes requested):
can we upgrade to puma 5.3?

Gemfile
```diff
@@ -3,20 +3,6 @@
...
-
-group :development do
-  gem 'dor-services-client' # used by bin/validate-cocina-roundtrip
-  gem 'listen', '~> 3.0.5'
-  gem 'rubyzip', '>= 1.0.0'
-  # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
-  gem 'spring'
-  gem 'spring-watcher-listen', '~> 2.0.0'
-end
-
```
ndushay, line 19: having this stuff separate made it harder to find - i put it where it is for most of our apps ...

jcoyne: I'm seeing this stuff at the top in argo: https://github.com/sul-dlss/argo/blob/main/Gemfile#L6-L9

ndushay:
I'll see if tests pass with puma 5.3, sure.

Gemfile
jcoyne: It is this way in h2 and dor_indexing_app though. 🤷

jcoyne (approved):

PR #2858, opened 2021-05-25
EventsController: return event list sorted by creation date (newest first)

jmartin-sul:
## Why was this change made?

I got confused when i went looking for a replication event on a test object that I'd versioned this morning, and the newest event seemed to be 26 days old.  Turns out the latest events were at the bottom, so i figured there was a SQL query that might need an `order by`.

I guess the consumer of the list could sort it (Argo for example just displays the list in the order it comes back, AFAICT).  But this seems like a fine deterministic default for most consumers, and it was slightly easier to make and test this change here, so I went ahead and sorted in DSA.

Compare this random QA object to this random prod object:

Screen Shot 2021-05-24 at 7 28 16 PM
Screen Shot 2021-05-24 at 7 28 30 PM
Screen Shot 2021-05-24 at 7 55 15 PM
Screen Shot 2021-05-24 at 8 03 17 PM

I suspect we mostly get a consistent sort by chance, because SQL results often return in creation order if no other order is specified (but of course, that's not guaranteed).

### DIFF
--- a/spec/requests/events_spec.rb
+++ b/spec/requests/events_spec.rb
@@ -32,5 +32,21 @@ RSpec.describe 'Add and retrieve events' do
       expect(json[0]['data']).to eq('description' => 'stuff', 'size' => 1900)
       expect(json[0]).to have_key 'created_at'
     end
+
+    context 'when there are multiple events' do
+      before do
+        create(:event, druid: druid, event_type: 'publish')
+        create(:event, druid: druid, event_type: 'unpublish')
+      end
+
+      it 'returns events in chronological order, newest first' do
+        get "/v1/objects/#{druid}/events",
+            headers: { 'Authorization' => "Bearer #{jwt}" }
+        expect(response).to have_http_status(:ok)
+        json = JSON.parse(response.body)
+        expect(json[0]['event_type']).to eq 'unpublish'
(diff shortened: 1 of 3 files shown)


spec/requests/events_spec.rb
```diff
@@ -32,5 +44,21 @@
       expect(json[0]['data']).to eq('description' => 'stuff', 'size' => 1900)
       expect(json[0]).to have_key 'created_at'
     end
+
+    it 'reutrns events in chronological order, newest first' do
```
jcoyne, line 48: typo.

spec/requests/events_spec.rb
```diff
@@ -32,5 +44,21 @@
...
+      post "/v1/objects/#{druid}/events",
```
jcoyne, line 49: Can these be a before block that just makes Event instances?  That way there is only a single request under test.

jcoyne (changes requested):

spec/requests/events_spec.rb
jmartin-sul: fixed

spec/requests/events_spec.rb
jmartin-sul: good call, done

jcoyne (approved):

PR #2873, opened 2021-06-03
allow publish to continue with HB data error for invalid license URL

### LINKED ISSUE #2872: Error in publish step when license URL is not valid
When an invalid license URL is present in an object, the publish step will fail.

see https://argo.stanford.edu/view/druid:dt303hw3940 as an example object

see https://app.honeybadger.io/projects/50568/faults/79616003 and https://app.honeybadger.io/projects/50568/faults/79779343

we have a check to try and avoid this: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/publish/access_conditions.rb#L56

but since we have a URL, and it's just not valid, that check is insufficient

peetucket:
## Why was this change made?

Fixes #2872 -- do not error out in the publish step when an invalid license URL is found and also provide the druid in the HB data error alert

Existing tests

### DIFF
--- a/app/services/publish/access_conditions.rb
+++ b/app/services/publish/access_conditions.rb
@@ -68,9 +68,13 @@ module Publish
     end
 
     def license
+      # The "none" licnese URI is a valid value in SDR (for legacy ETDs), but we don't want to publish that to PURL.
+      return if license_url == Cocina::FromFedora::Access::License::NONE_LICENSE_URI
+
       @license ||= License.new(url: license_url)
     rescue License::LegacyLicenseError
-      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license in object #{pid}")
+      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license",
+                         context: { druid: pid })
       if license_url.include?('creativecommons.org')
         # Try to find a workable license:
         @license = License.new(url: "#{license_url}legalcode")
--- a/app/services/publish/access_conditions.rb
+++ b/app/services/publish/access_conditions.rb
@@ -5,14 +5,13 @@ module Publish
   # These are derived from the rightsMetadata and are consumed by the
   # Searchworks item view via the mods_display gem.
   class AccessConditions
-    def self.add(public_mods:, rights_md:, pid:)
-      new(public_mods: public_mods, rights_md: rights_md, pid: pid).add
+    def self.add(public_mods:, rights_md:)
+      new(public_mods: public_mods, rights_md: rights_md).add
     end
 
-    def initialize(public_mods:, rights_md:, pid:)
+    def initialize(public_mods:, rights_md:)
       @public_mods = public_mods
       @rights_md = rights_md
-      @pid = pid
     end
 
     def add
@@ -24,7 +23,7 @@ module Publish
 
     private
 
-    attr_reader :rights_md, :public_mods, :pid
+    attr_reader :rights_md, :public_mods
 
     def rights
       @rights ||= rights_md.ng_xml
@@ -73,8 +72,7 @@ module Publish
 
       @license ||= License.new(url: license_url)
     rescue License::LegacyLicenseError
-      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license",
-                         context: { druid: pid })
+      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license")
       if license_url.include?('creativecommons.org')
         # Try to find a workable license:
         @license = License.new(url: "#{license_url}legalcode")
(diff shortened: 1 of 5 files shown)


app/services/publish/access_conditions.rb
```diff
@@ -69,15 +70,16 @@ def add_license
     def license
       @license ||= License.new(url: license_url)
     rescue License::LegacyLicenseError
-      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license")
+      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license in object #{pid}")
```
jcoyne, line 73: Can we put this in the Honeybadger context rather than the message? Then it won't be a new error each time.

app/services/publish/access_conditions.rb
```diff
@@ -69,15 +70,16 @@ def add_license
...
       if license_url.include?('creativecommons.org')
         # Try to find a workable license:
         @license = License.new(url: "#{license_url}legalcode")
       end
     end
 
+    # verify we have both a license_url and we were able to map that to a license with a description
     def license?
-      license_url.present?
+      license_url.present? && license.present?
```
jcoyne, line 82: Please add a test for this change.

jcoyne (changes requested):

app/services/publish/access_conditions.rb
peetucket: Yup, will do.

app/services/publish/access_conditions.rb
peetucket: Suggestions on where it should go?  It's not obvious to me.

jcoyne: spec/services/publish/public_desc_metadata_service_spec.rb is one place you could put it. Otherwise you could create a unit test for this class

app/services/publish/access_conditions.rb
jcoyne: I think we need a special trap here for the "none" license, because that is a valid value in SDR (for legacy ETDs), but we don't want to publish that to PURL.

spec/services/publish/access_conditions_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+RSpec.describe Publish::AccessConditions do
+  subject(:access) { described_class.new(public_mods: public_mods, rights_md: rights_md, pid: pid) }
+
+  let(:pid) { 'druid:bc123df4567' }
+
+  describe '#license' do
+    let(:public_mods) {  }
+    let(:rights_md) {  }
+
+    xit 'should alert HB on an unmapped license URL' do
```
jcoyne, line 14: Can you either complete these or remove them?  If you don't do it now, it's unlikely anyone will ever decide to do them later, and in the mean time they clutter up the test output.

jcoyne (changes requested):

spec/services/publish/access_conditions_spec.rb
peetucket: these were just placeholders, I was intending to fill in these in now, just wanted to commit to github -- just have switched the PR to draft

app/services/publish/access_conditions.rb
```diff
@@ -67,17 +68,22 @@ def add_license
...
 
     def license
+      # The "none" licnese URI is a valid value in SDR (for legacy ETDs), but we don't want to publish that to PURL.
+      return if license_url == Cocina::FromFedora::Access::License::NONE_LICENSE_URI
+
       @license ||= License.new(url: license_url)
     rescue License::LegacyLicenseError
-      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license")
+      Honeybadger.notify("[DATA ERROR] #{license_url} is not a supported license",
+                         context: { druid: pid })
```
jcoyne, line 77: What I was suggesting here was that we add `Honeybadger.context(druid: pid)` in `PublicDescMetadataService` and then we don't need to pass pid to this class.

jcoyne (changes requested):

app/services/publish/access_conditions.rb
jcoyne: Putting it in PublishJob might even be better.  Then we have the druid for any error that can occur in that context.

peetucket: Ahh, was not aware you could set the context earlier in the thread and have it added by HB later.  Changed.

jcoyne (changes requested):
Is this still necessary? The DRUID on the ticket seems to already be accessioned: https://argo.stanford.edu/view/druid:dt303hw3940

peetucket:
This particular druid was manually remediated to allow it to proceed and not wait for this code, but I think its possible this particular license may still exist in other objects and would become a problem, hence the reason for making this code hcange.  Do you know @andrewjbtw ?

peetucket:
We do not think this is needed anymore based on the fact that we are now validating all new licenses being added and have remediated all existing licenses.

PR #2882, opened 2021-06-18
Return a 400 bad request when catkey isn't found

jcoyne:
## Why was this change made?

When you send a catkey that symphony doesn't know about, it should be a you problem, not a me problem.


ndushay (approved):

ndushay:
Thanks - looks awesome.   Just wondering if we'll have consumers of this request using dor-services-client and hoping the error message will be clear enough coming through.

PR #2885, opened 2021-06-18
Cleanup as job

peetucket:
## Why was this change made?

Part of sul-dlss/argo#2752

Experiment in switching the end-accession/cleanup to a job.

We also need to changes to ~~dor-services-client: sul-dlss/dor-services-client#221 and~~ common-accessioning: sul-dlss/common-accessioning#770. Hold for these changes since they all need to be released together.

Todo: fix up `app/jobs/cleanup_job.rb`


peetucket:
Cool, didn't realize this was now covered (no longer a problem? or done in other work?)

ndushay:
per andrew, no longer a problem AFAICT.

PR #2905, opened 2021-07-08
Store DOI in identityMetadata and map it to Cocina

jcoyne:
and updated cocina-models to 0.61.2

### DIFF
--- a/app/services/cocina/from_fedora/identification.rb
+++ b/app/services/cocina/from_fedora/identification.rb
@@ -17,8 +17,9 @@ module Cocina
 
       def props
         {
-          sourceId: source_id,
-          barcode: fedora_object.identityMetadata.barcode
+          barcode: fedora_object.identityMetadata.barcode,
+          doi: doi,
+          sourceId: source_id
         }.compact
       end
 
@@ -38,6 +39,18 @@ module Cocina
 
           dissertation
         end
+      end
+
+      def doi
+        # We began to record DOI names in identityMetadata in the Summer of 2021
+        value = fedora_object.identityMetadata.ng_xml.xpath('//doi').first
+        return value.text if value
+
+        # Prior to that we only had DOI links in the descMetadata
+        value = fedora_object.descMetadata.ng_xml.xpath('//identifier[@type="doi"]').first
+        return unless value
+
+        value.text.delete_prefix('https://doi.org/')
       end
     end
   end
--- a/app/services/cocina/object_updater.rb
+++ b/app/services/cocina/object_updater.rb
@@ -116,13 +116,15 @@ module Cocina
         Cocina::ToFedora::Identity.apply_label(fedora_object, label: cocina_object.label)
       end
 
+      identity_updater = Cocina::ToFedora::Identity.new(fedora_object)
       if has_changed?(:identification)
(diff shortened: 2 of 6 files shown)


app/services/cocina/from_fedora/identification.rb
```diff
@@ -39,6 +40,18 @@ def source_id
...
+      def doi
+        # We began to record DOI names in identityMetadata in the Summer of 2021
+        value = fedora_object.identityMetadata.ng_xml.xpath('//doi').first
+        return value.text if value
+
+        # Prior to that we only had DOI links in the descMetadata
+        value = fedora_object.identityMetadata.ng_xml.xpath('//identifier[type="doi"]').first
+        return unless value
+
+        value.text.delete_prefix('https://doi.org/')
```
justinlittman, line 53: How confident are we that these are the only DOI prefixes in existing data?

app/services/cocina/from_fedora/identification.rb
```diff
@@ -17,8 +17,9 @@ def initialize(fedora_object)
 
       def props
         {
-          sourceId: source_id,
-          barcode: fedora_object.identityMetadata.barcode
+          barcode: fedora_object.identityMetadata.barcode,
+          doi: doi,
```
justinlittman, line 21: Are there cases in which may have multiple DOIs?

justinlittman (changes requested):
Does object_creator need to be changed as well?

app/services/cocina/from_fedora/identification.rb
jcoyne: not that we support.  This will currently be coming from H2, so we have full control over all of this.

app/services/cocina/from_fedora/identification.rb
jcoyne: I don't know. I don't know how we could check.

justinlittman: I'll write a report.

app/services/cocina/from_fedora/identification.rb
```diff
@@ -39,6 +40,18 @@ def source_id
...
         end
       end
+
+      def doi
+        # We began to record DOI names in identityMetadata in the Summer of 2021
+        value = fedora_object.identityMetadata.ng_xml.xpath('//doi').first
+        return value.text if value
+
+        # Prior to that we only had DOI links in the descMetadata
+        value = fedora_object.identityMetadata.ng_xml.xpath('//identifier[type="doi"]').first
```
justinlittman, line 50: I suspect this should be `//identifier[@type="doi"]`.

Also, looking in the identity metadata specs I don't see `<identifier>`. Are you sure this is correct?

justinlittman:
Also, the mapping needs a spec in https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/identification/dro_identity_spec.rb

app/services/cocina/from_fedora/identification.rb
jcoyne: ah, this is meant to be in descMetadata, not identityMetadata

justinlittman (approved):

PR #2915, opened 2021-07-14
Supports assigning DOI.

### LINKED ISSUE #2914: Fill in the doi field in cocina when doing register
refs https://github.com/sul-dlss/sdr-api/issues/334

justinlittman:
closes #2914

## Why was this change made?
Support assigning DOIs in H2.

Unit

openapi.yml


app/services/cocina/object_creator.rb
```diff
@@ -86,7 +88,7 @@ def create_apo(cocina_admin_policy, trial:)
     # @return [Dor::Item] a persisted Item model
```
jcoyne, line 88: please add parameter documentation for assign_doi

jcoyne (approved):
Approved, but please consider adding  that one bit of documentation.

Issue #2912, opened 2021-07-13
Create an HTTP endpoint that updates Datacite DOI metadata


jcoyne:
This should kick off a background job that handles the communication with Datacite.  This should be a sidekiq job because we want to make this resilient from Datacite being temporarily offline.  This endpoint will be called by the robots.  This will not have a callback to update the workflow.  If there is a failure, it should log in HB, but then rely on the sidekiq mechanism to retry as necessary.

ndushay:
@jcoyne is there code that does this already somewhere?

ndushay:
@arcadiafalcone can you let me know where you are in the mapping process, and where I should look for what you have so far?  (I can set up the plumbing and a partial mapping now.)

arcadiafalcone:
Mappings in progress are here: https://github.com/sul-dlss/dor-services-app/tree/desc-spec-datacite/spec/services/cocina/mapping/descriptive/h2_datacite

I'm currently revamping contributor and event based on a conversation with Amy, but the others should be mostly stable.

jcoyne:
@arcadiafalcone the datacite API that I'm looking at is in JSON, but it looks like you're making XML. https://support.datacite.org/reference/dois-2#post_dois

arcadiafalcone:
XML is what we use in the current workflow and the format of DataCite's example records, but I can switch it to JSON.

arcadiafalcone:
Though it looks like the metadata post API does take XML: https://support.datacite.org/reference/metadata#create-metadata-record-2.

jcoyne:
I think we'd prefer to work with JSON if it's the same to you.

jcoyne:
> Though it looks like the metadata post API does take XML: https://support.datacite.org/reference/metadata#create-metadata-record-2.

that's the MDS API, but we are not using that API.

arcadiafalcone:
OK. I'll finish up the event spec first and then start translating the specs to JSON.

ndushay:
noting that I will now work on a placeholder endpoint in DSA as the first part of this ticket.

ndushay:
https://github.com/sul-dlss/dor-services-app/pull/2920 provides the mappings;  I will now implement them.

ndushay:
Arcadia says:
> the required attributes are title, creator, identifier with identifierType, publisher, publication year, and resourceType with resourceTypeGeneral

- [ ] So if all of those aren’t populated after mappings, then there should be an HB error

[...]

PR #2917, opened 2021-07-14
update_doi_metadata endpoint added

ndushay:
## Why was this change made?

We will need to be able to update DOI metadata at DataCite;  the mechanism chosen is for DSA to perform these requests.  

This PR is for a dummy endpoint so other tickets can be unblocked as we wait for definitive DataCite mappings from Arcadia.

part of #2912

request spec added.

openapi.yml

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -285,6 +285,23 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+  '/v1/objects/{id}/update_doi_metadata':
+    post:
+      tags:
+        - integrations
+      summary: Update Datacite DOI metadata
+      description: 'Starts a background job to do this, from DOI and metadata in the repository'
+      operationId: 'objects#update_doi_metadata'
+      responses:
+        '202':
+          description: Accepted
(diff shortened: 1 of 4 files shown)


openapi.yml
```diff
@@ -285,6 +285,23 @@ paths:
...
             $ref: '#/components/schemas/Druid'
+  '/v1/objects/{id}/update_doi_metadata':
+    post:
+      tags:
+        - integrations
+      summary: Update Datacite DOI metadata
+      description: 'Starts a background job to do this, from DOI and metadata in the repository'
+      operationId: 'objects#update_doi_metadata'
+      responses:
+        '200':
```
jcoyne, line 296: this should be 202 ( to match `:accepted`).

jcoyne (changes requested):

openapi.yml
ndushay: oh!  yup.   Wonder if any of the other similar background job API calls should actually be 202.

jcoyne (approved):

PR #2929, opened 2021-07-20
Datacite mappings - part 1

ndushay:
## Why was this change made?

Implementation of some of the mappings of Arcadia's PRs #2920 and #2927;  these mappings need to be implemented so we can perform HTTP put requests to DataCite (per #2912)

Rather that have a PR for everything at once, this PR should be enough to get it started.

unit tests.


spec/services/cocina/mapping/descriptive/h2_datacite/note_h2_datacite_spec.rb
```diff
@@ -3,41 +3,76 @@
 require 'rails_helper'
 
 RSpec.describe 'Cocina --> DataCite mappings for note' do
+  # Note that this instantiation of Description does NOT validate against OpenAPI due to title validation issues.
```
jcoyne, line 6: I'm not sure I understand which issue you are talking about here.

jcoyne (changes requested):

app/services/cocina/to_datacite/attributes.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require_relative 'form'
+require_relative 'note'
+require_relative 'related_resource'
```
jcoyne, line 5: None of these should be necessary since this is a Rails app.

jcoyne (changes requested):

spec/services/cocina/mapping/descriptive/h2_datacite/note_h2_datacite_spec.rb
ndushay: because it's a cocina fragment, it complains on validation that there is no title.  This language was copied from MODS mapping spec code somehwere.

app/services/cocina/to_datacite/attributes.rb
ndushay: I'll try without.

ndushay: removed.

spec/services/cocina/mapping/descriptive/h2_datacite/note_h2_datacite_spec.rb
```diff
@@ -3,41 +3,76 @@
 require 'rails_helper'
 
 RSpec.describe 'Cocina --> DataCite mappings for note' do
+  # Note that this instantiation of Description does NOT validate against OpenAPI due to missing title.
```
jcoyne, line 6:
```suggestion
  # NOTE: Because we haven't set a title in this Description, it is not valid per the openapi.
```

jcoyne (changes requested):

spec/services/cocina/mapping/descriptive/h2_datacite/note_h2_datacite_spec.rb
ndushay: better!  will use.

ndushay: reworded.

jcoyne (approved):

PR #2933, opened 2021-07-21
datacite mappings: cocina identifier and purl -> datacite identifiers

ndushay:
## Why was this change made?

Arcadia's specs and my additions to them.


jcoyne (approved):

jcoyne:
@ndushay there is a merge conflict that needs to be resolved.

PR #2938, opened 2021-07-22
datacite mappings: cocina DROAccess.license -> DataCite rightsList

ndushay:
## Why was this change made?

Arcadia's specs and my additions to them.

@arcadiafalcone just want you to be aware that I moved `spec/services/cocina/mapping/descriptive/h2_datacite/rights_list_h2_datacite_spec.rb` to `spec/services/cocina/mapping/access/h2_datacite/dro_access_h2_datacite_spec.rb` so it would be more consistent with the directory structure.


spec/services/cocina/mapping/descriptive/h2_datacite/access_h2_datacite_spec.rb
```diff
@@ -2,27 +2,42 @@
...
+              type: 'email',
+              displayLabel: 'Contact'
+            }
+          ]
         }
-      end
+      }
+    end
 
+    xit 'not implemented' do
```
ndushay, line 25: still xit -- currently punting on implementing "we ignore this" tests

jcoyne (approved):

PR #2941, opened 2021-07-22
Cocina serialize

jcoyne:


peetucket (approved):

app/services/cocina/serializer.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Cocina
+  # Serializes Cocina::Model objects for active_job
+  # See https://blog.saeloun.com/2019/09/11/rails-6-custom-serializers-for-activejob-arguments.html
+  # class Serializer < ActiveJob::Serializers::ObjectSerializer
+  class Serializer
+    def serialize?(argument)
```
jmartin-sul, line 8: presumably this'll get invoked once things get re-wired a bit after the rails 6 upgrade?

jmartin-sul: ah, yeah, appears so from that blog post

jmartin-sul (approved):

app/services/cocina/serializer.rb
jcoyne: Yes

PR #2946, opened 2021-07-23
Add H2-MODS mappings for ORCID

arcadiafalcone:
## Why was this change made?
To provide necessary mappings

CircleCI

n/a


ndushay:
This appears to duplicate what is here:  https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/h2/contributor_h2_spec.rb#L1365-L1525   ?     If so, please close and delete the branch for this??

ndushay:
Arcadia agrees these are duplicate specs.

PR #2949, opened 2021-07-26
datacite mappings: cocina event -> DataCite publicationYear with unit tests

ndushay:
## Why was this change made?

This PR implements extracting correct publication year for DataCite mapping from cocina events

unit tests adapted from Arcadia's tests

### DIFF
--- a/app/services/cocina/to_datacite/event.rb
+++ b/app/services/cocina/to_datacite/event.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+module Cocina
+  module ToDatacite
+    # Transform the Cocina::Models::DRO.description.event attributes to appropriate DataCite attributes
+    #  see https://support.datacite.org/reference/dois-2#put_dois-id
+    class Event
+      def initialize(cocina_item)
+        @cocina_item = cocina_item
+      end
+
+      # For DataCite publicationYear, use year embargo lifted if present,
+      #  otherwise use deposit year, regardless of publication date entered
+      #
+      # @return [String] publication year, conforming to the expectations of HTTP PUT request to DataCite
+      def pub_year
+        if embargo?
+          embargo_release_date = cocina_dro_access&.embargo&.releaseDate
+          embargo_release_date&.year&.to_s
+        elsif deposit_event_publication_date
+          DateTime.parse(deposit_event_publication_date).year&.to_s
+        end
+      end
+
+      # Add Stanford Digital Repository as publisher to cocina release event if present, otherwise deposit event
+      def publisher
+        # TODO: implement this
+      end
+
+      # For DataCite dates
+      # H2 publication date > cocina event/date type publication > DataCite date type Issued
+      # H2 deposit date > cocina event type deposit > DataCite date type Submitted
+      ## If no embargo, > cocina date type publication
+      ## If embargo, > cocina date type deposit
+      # H2 embargo end date > cocina event type release and date type publication > DataCite date type Available
+      def dates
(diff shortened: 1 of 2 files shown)


app/services/cocina/to_datacite/event.rb
```diff
@@ -0,0 +1,66 @@
...
+      def deposit_event
+        cocina_events&.find { |event| event&.type == 'deposit' }
+      end
+
+      def cocina_events
+        @cocina_events ||= cocina_item.description.event
+      end
+
+      # embargo is in Cocina::Models::DROAccess -- the top level access, not description.access
+      def cocina_dro_access
```
justinlittman, line 61: This indirection isn't buying anything.

app/services/cocina/to_datacite/event.rb
```diff
@@ -0,0 +1,66 @@
...
+      ## If no embargo, > cocina date type publication
+      ## If embargo, > cocina date type deposit
+      # H2 embargo end date > cocina event type release and date type publication > DataCite date type Available
+      def dates
+        # TODO: implement this
+      end
+
+      private
+
+      attr :cocina_item, :cocina_events
```
justinlittman, line 42: I don't think `cocina_events` is an attr.

app/services/cocina/to_datacite/event.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      def deposit_event_publication_date
+        deposit_event&.date&.find { |date| date&.type == 'publication' }&.value
+      end
+
+      def deposit_event
+        cocina_events&.find { |event| event&.type == 'deposit' }
+      end
+
+      def cocina_events
```
justinlittman, line 56: This is only used one placed. Is the indirection buying anything?

justinlittman (changes requested):
Note that some of the comments are intended as comments, not requested changes.

app/services/cocina/to_datacite/event.rb
```diff
@@ -0,0 +1,66 @@
...
+      def dates
+        # TODO: implement this
+      end
+
+      private
+
+      attr :cocina_item, :cocina_events
+
+      def embargo?
+        cocina_dro_access&.embargo&.releaseDate.presence
```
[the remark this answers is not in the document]
ndushay: I think I ended up with `.presence` due to our rubocop masters.  I don't have strong feelings about it but every time I do a rubocop exception in this app, a certain individuals says "why did you do this"

app/services/cocina/to_datacite/event.rb
ndushay: more code is in progress that uses it.  Much more to be implemented in this class.

app/services/cocina/to_datacite/event.rb
ndushay: you are correct;  fixing.

app/services/cocina/to_datacite/event.rb
ndushay: I disagree - it is used in 2 places and makes it clear we are not in `cocina_item.description.access` which also exists and is confusing.

app/services/cocina/to_datacite/event.rb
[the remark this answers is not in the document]
ndushay: presence is like `.present?` except it return the value if it is there, and returns nil if it isn't there.

ndushay:
@justinlittman I responded to all of your suggestions and took one of them.

justinlittman (approved):

PR #2950, opened 2021-07-27
datacite mappings: cocina event -> DataCite dates with unit tests

ndushay:
~~[Hold until Arcadia clarifies Submitted date algorithm and Created date algorithm and until PR #2949 is merged so this can be rebased.]~~

## Why was this change made?

This PR implements extracting correct date info for DataCite from cocina events

unit tests adapted from Arcadia's tests

### DIFF
--- a/app/services/cocina/to_datacite/event.rb
+++ b/app/services/cocina/to_datacite/event.rb
@@ -23,7 +23,12 @@ module Cocina
         end
       end
 
+      # H2 publisher role > same cocina event as publication date > see DataCite contributor mappings
       # Add Stanford Digital Repository as publisher to cocina release event if present, otherwise deposit event
+      #
+      # sdr is the publisher for the event where the content becomes public via purl -- deposit if no embargo, release if embargo present.
+      # if it's not public via purl, sdr should not be the publisher;
+      #   the user may enter someone with the publisher role in h2, referring to publication in another venue, regardless of the purl status.
       def publisher
         # TODO: implement this
       end
--- a/app/services/cocina/to_datacite/event.rb
+++ b/app/services/cocina/to_datacite/event.rb
@@ -18,8 +18,8 @@ module Cocina
         if embargo?
           embargo_release_date = cocina_dro_access&.embargo&.releaseDate
           embargo_release_date&.year&.to_s
-        elsif deposit_event_publication_date
-          DateTime.parse(deposit_event_publication_date).year&.to_s
+        elsif deposit_event_publication_date_value
+          DateTime.parse(deposit_event_publication_date_value).year&.to_s
         end
       end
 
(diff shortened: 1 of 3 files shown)


app/services/cocina/to_datacite/event.rb
```diff
@@ -17,24 +17,30 @@ def pub_year
...
-      # H2 publication date > cocina event/date type publication > DataCite date type Issued
-      # H2 deposit date > cocina event type deposit > DataCite date type Submitted
-      ## If no embargo, > cocina date type publication
-      ## If embargo, > cocina date type deposit
-      # H2 embargo end date > cocina event type release and date type publication > DataCite date type Available
+      # DataCite date (YYYY-MM-DD) is repeatable and each DataCite data has an associated type attribute
+      # @return [Array<Hash>] DataCite date hashs, conforming to the expectations of HTTP PUT request to DataCite
       def dates
-        # TODO: implement this
+        dates = []
```
justinlittman, line 38: Could be a `[].tap`.

app/services/cocina/to_datacite/event.rb
```diff
@@ -45,12 +51,124 @@ def embargo?
...
-        deposit_event&.date&.find { |date| date&.type == 'publication' }&.value
+      # If embargo,
+      #   Cocina event type deposit, date type deposit maps to DataCite date type Submitted
+      # If no embargo
+      #   Cocina event type deposit, date type publication maps to DataCite date type Submitted
+      # If no embargo and no deposit event with date type publication,
+      #   Cocina event type publication, date type publication maps to DataCite date type Submitted
+      def submitted_date
+        @submitted_date ||= begin
+          submitted_date = {}
```
justinlittman, line 62: Could be `{}.tap` which would avoid `begin` and `end`.

app/services/cocina/to_datacite/event.rb
```diff
@@ -45,12 +51,124 @@ def embargo?
...
+
+          if embargo? && deposit_event_deposit_date_value.present?
```
justinlittman, line 64: Could be `submitted_date[:date] = if embargo? && deposit_event_deposit_date_value.present? ...`

app/services/cocina/to_datacite/event.rb
```diff
@@ -45,12 +51,124 @@ def embargo?
...
+      # from Arcadia:
+      #   Cocina event type release, date type publication maps to DataCite date type Available
+      # In actuality:
+      #   embargo release date is in DROAccess, not in an event
+      def available_date
+        return unless embargo?
+
+        @available_date ||=
+          {
+            date: cocina_dro_access&.embargo&.releaseDate,
```
justinlittman, line 85: Are the `&` checks necessary?

app/services/cocina/to_datacite/event.rb
```diff
@@ -45,12 +51,124 @@ def embargo?
...
+          submitted_date
+        end
+      end
+
+      # from Arcadia:
+      #   Cocina event type release, date type publication maps to DataCite date type Available
+      # In actuality:
+      #   embargo release date is in DROAccess, not in an event
+      def available_date
+        return unless embargo?
```
justinlittman, line 81: Does having the return outside the `||=` defeat the purposes of the `||=`?

app/services/cocina/to_datacite/event.rb
```diff
@@ -45,12 +51,124 @@ def embargo?
...
+                        start_date
+                      elsif end_date.present?
+                        end_date
+                      end
+
+        {
+          date: result_date
+        }.tap do |attributes|
+          attributes[:dateInformation] = info if info.present?
+        end.compact
```
justinlittman, line 171: Does the `.compact` do anything here?

justinlittman (approved):
Style comments only (for your consideration).

app/services/cocina/to_datacite/event.rb
ndushay: definitely - embargo might not exist

app/services/cocina/to_datacite/event.rb
ndushay: tried it;  got an error;  plenty of other fish to fry.

app/services/cocina/to_datacite/event.rb
ndushay: no because if it's nil, it would recompute anyway.

app/services/cocina/to_datacite/event.rb
ndushay: result_date is not guaranteed to have a value.

PR #2955, opened 2021-07-27
[Hold] h2 event mapping tweaks to and from MODS for h2

ndushay:
[HOLD - need clarification from Arcadia on what to do for MODS -> Cocina for uncited publisher]

## Why was this change made?

a) We need to map cocina generated by h2 to datacite format so we can update DOI data;  
b) we need cocina generated by h2 to be stored as MODS with the correct info so we can reconstitute cocina from MODS and send the right stuff to DataCite

This PR implements specs Arcadia asked for in PRs #2925 and #2954, and is also related to her PRs #2920 and #2927 ... all this for H2 which requires DSA to be able to update DOI information  ...

Arcadia's specs.

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/event.rb
+++ b/app/services/cocina/from_fedora/descriptive/event.rb
@@ -26,6 +26,7 @@ module Cocina
           'copyright',
           'creation',
           'degree conferral',
+          'deposit',
           'development',
           'distribution',
           'generation',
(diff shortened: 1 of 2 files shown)


app/services/cocina/from_fedora/descriptive/event.rb
```diff
@@ -366,9 +367,20 @@ def add_publisher_info(event, publisher_nodes, origin_info_node)
...
 
           event[:contributor] ||= []
+          # rubocop:disable Metrics/BlockLength
           publisher_nodes.each do |publisher_node|
             next if publisher_node.text.blank?
 
+            contrib_type = nil
+            roles = [role_for(event)]
+            if origin_info_node['eventType'] == 'deposit' || origin_info_node['eventType'] == 'release'
+              roles << {
```
justinlittman, line 377: Always adding this role is going to break roundtripping if non-H2. Since `release` is already an event value, I suspect this is problematic.

ndushay:
turns out the mappings being implemented here were flawed.

PR #2956, opened 2021-07-28
[Hold] H2 contrib mapping tweaks to and from MODS for h2

ndushay:
[HOLD - need clarification from Arcadia on what to do for MODS -> Cocina for DataCite and H2 roles ... and a bunch of other work]

## Why was this change made?

a) We need to map cocina generated by h2 to datacite format so we can update DOI data;  
b) we need cocina generated by h2 to be stored as MODS with the correct info so we can reconstitute cocina from MODS and send the right stuff to DataCite

This PR implements specs Arcadia asked for in PRs #2925 and #2954, and is also related to her PRs #2920 and #2927 ... all this for H2 which requires DSA to be able to update DOI information  ...

Arcadia's specs - the newly changed and the existing.

### DIFF
--- a/app/services/cocina/from_fedora/descriptive/identifier_builder.rb
+++ b/app/services/cocina/from_fedora/descriptive/identifier_builder.rb
@@ -40,6 +40,7 @@ module Cocina
             else
               attrs[:type] = cocina_type
               attrs[:value] = identifier_element.text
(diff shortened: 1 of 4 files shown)


ndushay:
Turns out the mappings for which these changes were written ... weren't okay.

PR #2965, opened 2021-07-29
Update H2-Cocina-MODS event mapping to match H2 output

arcadiafalcone:
## Why was this change made?
H2 can't currently provide some of the data the mapping expected

CircleCI

n/a

### DIFF
--- a/spec/services/cocina/mapping/descriptive/h2/event_h2_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/h2/event_h2_spec.rb
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe 'Cocina --> MODS mappings for event (h2 specific)' do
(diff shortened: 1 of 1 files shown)


ndushay (approved):

jcoyne:
Is there a reason we chose this route over updating h2 to provide the necessary format?

arcadiafalcone:
When I last raised the issue, there was a desire to keep H2 dates encoded as edtf, though in practice they are identical to w3cdtf.

jcoyne:
I think we want to keep h2 dates encoded as edtf in the database, but we could make the cocina data that h2 exports to SDR map to usew3cdtf.

arcadiafalcone:
Yes, we're not actually using any of the edtf-specific encodings, and our systems aren't set up to parse them anyway. I think it's fine to keep them as edtf - they appear to be displaying correctly, and LC has moved towards preferring it.

PR #2978, opened 2021-08-03
datacite form mappings hardcoded as a stopgap

ndushay:
NOTE:  **It may be advantageous NOT to merge this and wait for correct, roundtrippable mappings from @arcadiafalcone.**

## Why was this change made?

To be able to debug `update_doi_metadata` endpoint and `UpdateDoiMetadataJob`   

If this is merged, issue #2979 is about removing the hardcodings.

created some H2 objects in stage.

### DIFF
--- a/app/services/cocina/to_datacite/form.rb
+++ b/app/services/cocina/to_datacite/form.rb
@@ -21,6 +21,9 @@ module Cocina
 
         {}.tap do |attribs|
           # FIXME: the following were coded without Arcadia's mappings as a stopgap to be able to debug update_doi_metadata
+          warning_msg = 'DataCite mappings result in empty resourceTypeGeneral and/or empty resourceType; using hardcoded default values'
+          Honeybadger.notify(warning_msg) if resource_type_general.blank? || resource_type.blank?
+
           attribs[:resourceTypeGeneral] = resource_type_general || 'Text'
           attribs[:resourceType] = resource_type || 'Testing'
         end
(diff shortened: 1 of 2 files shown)


app/services/cocina/to_datacite/form.rb
```diff
@@ -20,8 +20,9 @@ def type_attributes
         return {} if cocina_desc&.form.blank?
 
         {}.tap do |attribs|
-          attribs[:resourceTypeGeneral] = resource_type_general if resource_type_general
-          attribs[:resourceType] = resource_type if resource_type
+          # FIXME: the following were coded without Arcadia's mappings as a stopgap to be able to debug update_doi_metadata
+          attribs[:resourceTypeGeneral] = resource_type_general || 'Text'
```
jcoyne, line 24: How about a check here that sends `Honeybadger.notify` if the default has to be used?

jcoyne (changes requested):

app/services/cocina/to_datacite/form.rb
ndushay: It will always have to be used until Arcadia does some updates -- the current code looks for resource_type  with DataCite as a source, and we have no way to represent that in MODS.

jcoyne: I think I'm okay with it having to be ripped out later, but I don't want to merge code that we know is producing bad data without some kind of pain for us that will get us to change it later.

ndushay:
@jcoyne added Honeybadger notification.

jcoyne (approved):

PR #2984, opened 2021-08-06
Add warnings when in an infinite loop

jcoyne:
## Why was this change made?

To point people who are debugging to the right place in the code.

### DIFF
--- a/app/services/administrative_tags.rb
+++ b/app/services/administrative_tags.rb
@@ -104,7 +104,9 @@ class AdministrativeTags
     # If two threads are creating the same tag, one will get an exception.
     # We must catch this outside the transaction block, because once a constraint
     # is violated, PG will permit no more statements in that transaction.
+    # When we go to rails 6 we can replace find_or_create_by with create_or_find_by: https://sikac.hu/use-create-or-find-by-to-avoid-race-condition-in-rails-6-0-f44fca97d16b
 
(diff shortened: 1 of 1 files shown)


app/services/administrative_tags.rb
```diff
@@ -127,7 +129,9 @@ def update(current:, new:)
     # If two threads are creating the same tag, one will get an exception.
     # We must catch this outside the transaction block, because once a constraint
     # is violated, PG will permit no more statements in that transaction.
+    # When we go to rails 6 we can replace find_or_create_by with create_or_find_by: https://sikac.hu/use-create-or-find-by-to-avoid-race-condition-in-rails-6-0-f44fca97d16b
 
+    Rails.logger.warn("Possible race condition updating tags: #{tags}.  This should only happen one time, otherwise this might be an error")
```
jmartin-sul, line 134: apologies if i'm missing something, but is there actually a var called `tags` in scope here?

jmartin-sul: i see it as a param to the method in the other added logging statement

jmartin-sul (changes requested):
i think one of the logging statements might reference an unavailable variable

app/services/administrative_tags.rb
jcoyne: Good catch, should be `new`

jcoyne:
Superseded by #3048

PR #2986, opened 2021-08-09
Map MODS DataCite extension for contributor

arcadiafalcone:
## Why was this change made?

CircleCI

n/a

Implementation ticket https://github.com/sul-dlss/happy-heron/issues/1911


arcadiafalcone:
Different approach taken.

PR #2987, opened 2021-08-09
Map MODS DataCite extension for event

arcadiafalcone:
## Why was this change made?

CircleCI

n/a

Implementation ticket https://github.com/sul-dlss/happy-heron/issues/1914


arcadiafalcone:
Different approach taken.

PR #2989, opened 2021-08-10
Handles empty copyright data.

### LINKED ISSUE #2988: NoMethodError: undefined method `compact' for nil:NilClass
## Backtrace

    line 87 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/event.rb: block in build_ungrouped_origin_infos
    line 81 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/event.rb: each
    line 81 of [PROJECT_ROOT]/app/services/cocina/from_fedora/descriptive/event.rb: filter_map

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/80851900)

justinlittman:
closes #2988

## Why was this change made?
The code was only partially handling empty copyright data.

Unit


jcoyne:
@andrewjbtw Should we handle this, or should this be a data error that is remediated?

andrewjbtw:
Is there an example of what triggered this error? Generally, a copyright statement is not required.

andrewjbtw:
Oh, I see this is in the MODS, not the the rightsMetadata. In that case, I don't know. It's a question for Arcadia.

justinlittman:
druid:wh374xb2997. In general, it would have considered this a data remediation issue. However, since our code half-handled the case (it handles the empty copyright and returns a nil, but then the calling method chokes on the nil), I opted for a code fix as well.

jcoyne:
@justinlittman I'm wondering if its acceptable that this gets normalized out in a round trip.

justinlittman:
Removing empty elements is one of the most common forms of normalization. Apparently, the existing normalization already handles this case.

spec/services/cocina/mapping/descriptive/mods/origin_info_event_type_spec.rb
```diff
@@ -60,6 +60,22 @@
       end
     end
 
+    describe 'empty copyright data' do
```
jcoyne, line 63: Please add a comment that this is not valid, but we are handling it nevertheless.

app/services/cocina/from_fedora/descriptive/event.rb
```diff
@@ -84,7 +84,7 @@ def build_ungrouped_origin_infos(origin_infos)
                     origin_info.xpath('//*[@xlink:href]', xlink: XLINK_NS).empty?
 
             event = build_event_for_origin_info(origin_info)
-            event.compact
+            event&.compact
```
jcoyne, line 87: Do we even need to call compact here?

jcoyne (changes requested):

app/services/cocina/from_fedora/descriptive/event.rb
jcoyne: I removed this call and all the tests still pass.

jcoyne (approved):

PR #2993, opened 2021-08-10
Add MODS>DataCite mappings for form

arcadiafalcone:
## Why was this change made?

CircleCI

n/a


jcoyne (changes requested):
I think there's a misunderstanding here @arcadiafalcone. This is testing a mapping between cocina and datacite.  There isn't any MODS involved in this transformation.

arcadiafalcone:
@jcoyne I was going off of this convo yesterday: https://stanfordlib.slack.com/archives/C69UZCJ8M/p1628535051337200

jcoyne:
@arcadiafalcone I'm sorry, I didn't have a full understanding of what you were asking there. We need h2>cocina>mods and mods > cocina > datacite-json.  That latter can be decomposed into mods > cocina (which is already done) and cocina > datacite-json (which I belive is partially done)

arcadiafalcone:
@jcoyne Thanks, that clarifies matters. I hadn't realized that the DataCite was coming out of cocina roundtripped from MODS.

PR #3000, opened 2021-08-11
[REFACTOR] Remove unnecessary conditionals from code

jcoyne:
## Why was this change made?

Fewer conditionals makes the code easier to reason about.

### DIFF
--- a/app/services/cocina/to_datacite/dro_access.rb
+++ b/app/services/cocina/to_datacite/dro_access.rb
@@ -6,7 +6,7 @@ module Cocina
     #  see https://support.datacite.org/reference/dois-2#put_dois-id
     class DROAccess
       # @param [Cocina::Models::DROAccess] cocina_item_access
-      # @return [Hash] Hash of DataCite rightsList attributes, conforming to the expectations of HTTP PUT request to DataCite
+      # @return [NilClass,Array<Hash>] list of DataCite rightsList attributes, conforming to the expectations of HTTP PUT request to DataCite
       def self.rights_list_attributes(cocina_item_access)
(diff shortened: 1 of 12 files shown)


spec/services/cocina/mapping/descriptive/h2_datacite/title_h2_datacite_spec.rb
```diff
@@ -20,56 +20,11 @@
...
-      expect(title_attributes).to eq(
+      expect(title_attributes).to eq [
         {
           title: 'Tales of a brooding sea star'
         }
-      )
-    end
-  end
-
-  ### --------------- specs below added by developers ---------------
```
jcoyne, line 36: These are not possible https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L536

app/services/cocina/to_datacite/dro_access.rb
```diff
@@ -15,13 +15,13 @@ def initialize(cocina_item_access)
         @cocina_item_access = cocina_item_access
       end
 
-      # @return [Hash] Hash of DataCite rightsList attributes, conforming to the expectations of HTTP PUT request to DataCite
+      # @return [Array<Hash>] list of DataCite rightsList attributes, conforming to the expectations of HTTP PUT request to DataCite
```
justinlittman, line 18: or NilClass

justinlittman (approved):
On minor suggested change.

app/services/cocina/to_datacite/dro_access.rb
jcoyne: Yes. Good catch. I have fixed it.

PR #3016, opened 2021-08-18
Map the creation date to datacite publicationYear

jcoyne:

Tested on stage


spec/services/cocina/to_datacite/attributes_spec.rb
```diff
@@ -92,6 +100,14 @@
                                     ]
                                   }
                                 ],
+                                event: [
```
arcadiafalcone, line 103: The DataCite publication year is not the creation or publication date entered by the user - it's (our best guess at) the year the digital object is published to the SDR, similar to the autogenerated citation.

app/services/cocina/to_datacite/attributes.rb
```diff
@@ -42,6 +42,12 @@ def mapped_from_cocina
 
       attr :access, :description
 
+      def publication_year
+        creation_event = description.event.find { |event| event.type == 'creation' }
```
justinlittman, line 46: Would like to see a guard against a missing creation event.

jcoyne: And then what?  This is a required field.  if it's missing here, the API call will fail.

spec/services/cocina/to_datacite/attributes_spec.rb
jcoyne: So, current year if not embargoed? I believe that's what autogenerated citation uses.

arcadiafalcone: Correct - embargo year if embargoed, otherwise current year.

justinlittman (approved):

PR #3051, opened 2021-09-16
Add a full description of the error response from the tags service

jcoyne:
## Why was this change made?
Related to https://github.com/sul-dlss/argo/issues/2813

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -44,7 +44,12 @@ module DorServices
     )
 
     # TODO: we can uncomment this at a later date to ensure we are passing back valid responses
-    # config.middleware.use Committee::Middleware::ResponseValidation, schema: schema
+    # config.middleware.use(
+    #   Committee::Middleware::ResponseValidation,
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -44,7 +44,12 @@ class Application < Rails::Application
     )
 
     # TODO: we can uncomment this at a later date to ensure we are passing back valid responses
-    # config.middleware.use Committee::Middleware::ResponseValidation, schema: schema
+    # config.middleware.use(
```
jcoyne, line 47: I was able to uncomment this and run the request specs to validate these changes.

justinlittman: For Sinopia API we run response validation only for specs. Maybe a similar approach here?

justinlittman (approved):

config/application.rb
jcoyne: Currently not all of our tests pass with this enabled, just the ones relating to the tags.

PR #3064, opened 2021-10-19
Upgrade to moab-versioning 5 and Faraday 2

jcoyne:
## Why was this change made?
To keep dependencies up to date.


Gemfile
```diff
@@ -35,7 +35,6 @@ gem 'iso-639', '~> 0.2.8'
 gem 'jbuilder'
 gem 'jwt' # json web token
 gem 'okcomputer'
-gem 'openapi_parser'
```
jcoyne, line 38: This is not a direct dependency, so it isn't required.

jmartin-sul (approved):

PR #3086, opened 2021-11-30
Handle a nil process content type tag

jcoyne:
## Why was this change made?

This was causing a problem in ObjectUpdater when it tries to add a tag with a nil value

### DIFF
--- a/app/services/cocina/object_updater.rb
+++ b/app/services/cocina/object_updater.rb
@@ -189,7 +189,12 @@ module Cocina
     end
 
     def add_tags(pid, cocina_object)
-      add_tag(pid, ToFedora::ProcessTag.map(cocina_object.type, cocina_object.structural&.hasMemberOrders&.first&.viewingDirection), 'Process : Content Type') if has_changed?(:structural)
+      if has_changed?(:structural)
+        # This is necessary so that the content type tag for a book can get updated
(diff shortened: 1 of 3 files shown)


app/services/cocina/to_fedora/process_tag.rb
```diff
@@ -5,30 +5,29 @@ module ToFedora
     # This transforms the DRO.type attribute to the process tag value
     class ProcessTag
       # TODO: add Software
+      MAPPING = {
```
jcoyne, line 8: This set of changes is just a refactor so we don't have a huge case statement for what is essentially a hash.

app/services/cocina/object_updater.rb
```diff
@@ -189,7 +189,12 @@ def catkey_for(cocina_object)
     end
 
     def add_tags(pid, cocina_object)
-      add_tag(pid, ToFedora::ProcessTag.map(cocina_object.type, cocina_object.structural&.hasMemberOrders&.first&.viewingDirection), 'Process : Content Type') if has_changed?(:structural)
+      if has_changed?(:structural)
+        # This is necessary so that the content type tag for a book can get updated
+        # to reflect the new direction if the direction hash changed in the structural metadata.
```
mjgiarlo, line 194: 👏🏻

mjgiarlo (approved):

Issue #3105, opened 2021-12-10
Determine what unmapped objects are and if they should be mapped


justinlittman:
![image](https://user-images.githubusercontent.com/588335/145637285-5b80ca12-68e6-4127-b434-c38439a6c359.png)

where unmapped is determined here: https://github.com/sul-dlss/dor-services-app/blob/main/lib/fedora_loader.rb#L100-L110

justinlittman:
Here's a list of unmapped:
```
druid:vc840ct8051=unmapped
druid:kk870jg0101=unmapped
druid:qy077yy0209=unmapped
druid:rb635cy4341=unmapped
druid:nf196xm8079=unmapped
druid:bb298tt5624=unmapped
druid:zb435nj9560=unmapped
druid:gk226jc8488=unmapped
druid:dp408hj5698=unmapped
druid:rh330wv2236=unmapped
druid:pf875qf2918=unmapped
druid:hd466vz4698=unmapped
druid:xw555bc6541=unmapped
druid:km957st7637=unmapped
druid:fn347js8478=unmapped
druid:ct978kq2178=unmapped
druid:ng490dx4393=unmapped
druid:ns523jd4552=unmapped
druid:gh691hq4438=unmapped
druid:rh699zm6525=unmapped
druid:pz744fc3517=unmapped
druid:zg374rc4591=unmapped
druid:rm249nq5059=unmapped
druid:jd716fw5543=unmapped
druid:cn256rg8583=unmapped
druid:zg837jb8453=unmapped
druid:rg264mg9570=unmapped
druid:xg439xy3864=unmapped
druid:kr102pj2548=unmapped
druid:ty923dj7361=unmapped
druid:gm482nj6139=unmapped
druid:qp161bd9459=unmapped
druid:fx315nj1109=unmapped
druid:bw925nc9437=unmapped
druid:rm419pw1964=unmapped
druid:dk634hw8530=unmapped
druid:tw617yc6610=unmapped
druid:fx216pc8450=unmapped
druid:tr997gp6153=unmapped
druid:qb155wd5139=unmapped
druid:bv407fr7895=unmapped
druid:fj829nc6769=unmapped
druid:wv445bd2710=unmapped
druid:zb265zb6765=unmapped
druid:rm870tr0400=unmapped
druid:pq376sd3515=unmapped
druid:wp608tf5441=unmapped
druid:bf580ss8395=unmapped
druid:sk476ph9817=unmapped
[...]

PR #3110, opened 2021-12-14
Improves mapping to Fedora items when loading from cache.

justinlittman:
refs #3105

## Why was this change made?
Make sure that performing correct mapping when validating.

Local

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -3,6 +3,12 @@
 require 'simplecov'
 SimpleCov.start :rails do
   add_filter '/spec/'
+  add_filter '/lib/data_error_notifier.rb'
+  add_filter '/lib/fedora_cache.rb'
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -3,6 +3,8 @@
 require 'simplecov'
 SimpleCov.start :rails do
   add_filter '/spec/'
+  add_filter '/lib/'
```
justinlittman, line 6: We don't write tests for this code, so removing from code coverage.

jcoyne (approved):

spec/spec_helper.rb
jcoyne: Can we just exclude fedora_loader.rb and not any file in lib?

justinlittman: Yup. Done.

PR #3134, opened 2022-01-05
Moved to using inline orb for circleci.

justinlittman:
## Why was this change made?
Clearer configuration, less boilerplate.


.circleci/config.yml
```diff
@@ -1,193 +1,364 @@
...
+                } || {
+                echo $result
+                exit 1
+                }
     jobs:
-    - test
-
+      build:
+        parameters:
+          tag:
```
jcoyne, line 27: Can this be `ruby_tag`?

edsu:
Is this no longer needed since #3136 has been  merged?

PR #3136, opened 2022-01-10
Switched circle to use ruby-rails orb.

justinlittman:
## Why was this change made?
Less config

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,193 +1,33 @@
 version: 2.1
-executors:
-  docker-publisher:
-    environment:
-      IMAGE_NAME: suldlss/dor-services-app
-    docker:
-    - image: circleci/buildpack-deps:stretch
-jobs:
-  test:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,193 +1,25 @@
...
-          exit 0
-          } || {
-          echo $result
-          exit 1
-          }
-    # Save test results for timing analysis
-    - store_test_results:
-        path: test_results
-
-  build-image:
```
ndushay, line 111: FYI:  see https://github.com/sul-dlss/dlme/pull/1382/files and/or talk to @aaron-collier re: what we need for docker image stuff to work.

justinlittman: Underlying our custom orb for docker operations is the docker orb. This makes docker stuff just work with much less config / complexity in our circle configs.

mjgiarlo (approved):
🚀

PR #3162, opened 2022-01-16
Omit empty mimetypes from cocina.

### LINKED ISSUE #3155: Map file nodes missing mimeType and size (contentMetadata)
For example:
```
<contentMetadata objectId="druid:hv992ry2431" type="book">
          <resource id="ck831vq4558_1" sequence="1" type="file">
            <file id="sul-logo.png" publish="yes" preserve="yes" shelve="yes"/>
          </resource>
        </contentMetadata>
```

Currently, when mapping to Cocina, `size` is set to 0 and `mimeType` is omitted.

And when roundtripping:
```
<contentMetadata objectId="druid:hv992ry2431" type="book">
        <resource id="ck831vq4558_1" sequence="1" type="file">
          <file id="sul-logo.png" mimetype="" size="0" publish="yes" preserve="yes" shelve="yes"/>
        </resource>
      </contentMetadata>
```

Is this correct? Should size be set to 0? On roundtrip, should mimeType be set to "" or omitted?

justinlittman:
closes #3155

## Why was this change made?
Mapping.

Before:
```
Status (n=10000; not using Missing for success/different/error stats):
  Success:   9409 (94.184%)
  Different: 545 (5.455%)
  Mapping error:     0 (0.0%)
  Create error:     36 (0.36%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     10 (0.1%)
  Bad cache:     0 (0.0%)
```

After:
```
Status (n=10000; not using Missing for success/different/error stats):
  Success:   9409 (94.184%)
  Different: 545 (5.455%)
  Mapping error:     0 (0.0%)
  Create error:     36 (0.36%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     10 (0.1%)
  Bad cache:     0 (0.0%)
```

### DIFF
--- a/app/services/cocina/from_fedora/file_sets.rb
+++ b/app/services/cocina/from_fedora/file_sets.rb
@@ -88,8 +88,8 @@ module Cocina
               shelve: node['shelve'] == 'yes'
             }
           }.tap do |attrs|
-            # Files from Goobi and Hydrus don't have mimetype until they hit exif-collect in the assemblyWF
-            attrs[:hasMimeType] = node['mimetype'] if node['mimetype']
+            # Files from Goobi don't have mimetype until they hit exif-collect in the assemblyWF
+            attrs[:hasMimeType] = node['mimetype'] if node['mimetype'].present?
             attrs[:presentation] = { height: height, width: width } if height && width
             attrs[:use] = use if use
           end
--- a/spec/services/cocina/mapping/structural/dro_structural_spec.rb
+++ b/spec/services/cocina/mapping/structural/dro_structural_spec.rb
@@ -308,4 +308,40 @@ RSpec.describe 'Fedora item content metadata <--> Cocina DRO structural mappings
       end
     end
   end
+
+  context 'when blank mimetype' do
+    it_behaves_like 'DRO Structural Fedora Cocina mapping' do
+      let(:content_xml) do
+        <<~XML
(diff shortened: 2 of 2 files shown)


jcoyne:
No change in the sample?

app/services/cocina/from_fedora/file_sets.rb
```diff
@@ -89,7 +89,7 @@ def build_files(file_nodes)
             }
           }.tap do |attrs|
             # Files from Goobi and Hydrus don't have mimetype until they hit exif-collect in the assemblyWF
-            attrs[:hasMimeType] = node['mimetype'] if node['mimetype']
+            attrs[:hasMimeType] = node['mimetype'] if node['mimetype'].present?
```
jcoyne, line 92: I thought we agreed on `application/octet-stream` no?

justinlittman: We did, but changed when I saw comment in the code that indicates that files from Goobi and Hydrus don't have mimetype until they hit exif-collect in the assemblyWF. That led me to conclude that a blank is meant to mean not yet present rather than missing.

jcoyne: That sounds good, can you remove "Hydrus" though. It's been decommissioned.

PR #3181, opened 2022-01-19
update_metadata_spec fixes.

justinlittman:
## Why was this change made?
Correct tests.

### DIFF
--- a/spec/requests/update_metadata_spec.rb
+++ b/spec/requests/update_metadata_spec.rb
@@ -106,8 +106,7 @@ RSpec.describe 'Update object' do
         "identification":#{identification.to_json},
         "structural":{
           "hasMemberOrders":[{"viewingDirection":"right-to-left"}],
-          "isMemberOf":["druid:xx888xx7777"],
-          "hasAgreement":"druid:cd777df7777"
+          "isMemberOf":["druid:xx888xx7777"]
         }
       }
     JSON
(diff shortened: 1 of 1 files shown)


spec/requests/update_metadata_spec.rb
```diff
@@ -106,8 +106,7 @@
         "identification":#{identification.to_json},
         "structural":{
           "hasMemberOrders":[{"viewingDirection":"right-to-left"}],
-          "isMemberOf":["druid:xx888xx7777"],
-          "hasAgreement":"druid:cd777df7777"
```
justinlittman, line 110: hasAgreement is NOT in the data that is submitted.

spec/requests/update_metadata_spec.rb
```diff
@@ -402,8 +400,11 @@
   end
 
   context 'when tags change' do
+    # In other tests, viewing direction starts at and remains right-to-left.
+    # For this test, starting at left-to-right and changing to right-to-left.
     before do
       allow(AdministrativeTags).to receive(:for).and_return(['Project : Tom Swift', 'Process : Content Type : Book (ltr)'])
+      allow(AdministrativeTags).to receive(:content_type).and_return(['Book (ltr)'], ['Book (rtl)'])
```
justinlittman, line 407: For this test, content type is switching from ltr to rtl.

spec/requests/update_metadata_spec.rb
```diff
@@ -131,7 +130,6 @@
     expect(item).to have_received(:admin_policy_object_id=).with(apo_druid)
 
     # Tags are created.
-    expect(AdministrativeTags).to have_received(:create).with(pid: druid, tags: ['Process : Content Type : Book (rtl)'])
```
justinlittman, line 134: structural isn't changing, so this tag doesn't get created.

spec/requests/update_metadata_spec.rb
justinlittman: That was making it seem like structural was changing, which affects other tests.

ndushay (approved):

PR #3184, opened 2022-01-19
Standardize on Ruby 2.7.5

mjgiarlo:
This is the version we have in CI, so I naïvely reckon it's the same we should have in the image?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.7.3-alpine
+FROM ruby:2.7.5-alpine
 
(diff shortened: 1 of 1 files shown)


ndushay (approved):
Is it what is on the VMs too?  All should be consistent, in a perfect world.

mjgiarlo:
@ndushay Dunno. But at least the codebase itself is consistent now?

ndushay:
I think it's a simple matter of rvm install , bundle install, puppet change default to 2.75 ... except the techstack is perhaps in use elsewhere?

PR #3185, opened 2022-01-19
Adds CocinaObjectPersister.

### LINKED ISSUE #3143: Create cocina object persister service
The role of the service is to support finding and saving Cocina objects. The service (called `CocinaObjectPersister`) will have the following methods and initial implementations:
* `save`: Save will retrieve the Fedora item, perform an update with Cocina::ObjectUpdater, and save the item. For reference, see ObjectsController.update.
* `find`: Find will retrieve the Fedora item and map to a Cocina object with Cocina::Mapper. For reference, see ObjectsController.show.

As DSA services are updated to operate on Cocina objects instead of Fedora objects, CocinaObjectPersister will be used to abstract finding and saving.

justinlittman:
closes #3143

## Why was this change made?
Provide an abstraction around finding / saving Cocina objects.

Unit

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -61,4 +61,8 @@ class ApplicationController < ActionController::API
   def load_item
     @item = Dor.find(params[:object_id] || params[:id])
   end
+
+  def load_cocina_object
+    @cocina_object = CocinaObjectStore.find(params[:object_id] || params[:id])
+  end
 end
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -2,19 +2,41 @@
 
 # rubocop:disable Metrics/ClassLength
 class ObjectsController < ApplicationController
-  before_action :load_item, except: [:create]
+  before_action :load_cocina_object, only: %i[show update_doi_metadata]
+  before_action :load_item, only: %i[show accession update_marc_record destroy notify_goobi]
 
+  # No longer be necessary when remove Fedora.
   rescue_from(Cocina::ObjectUpdater::NotImplemented) do |e|
     json_api_error(status: :unprocessable_entity, message: e.message)
   end
 
-  rescue_from(ActiveFedora::ObjectNotFoundError) do |e|
+  rescue_from(CocinaObjectStore::CocinaObjectNotFoundError) do |e|
     json_api_error(status: :not_found, message: e.message)
   end
 
   rescue_from(Dry::Struct::Error) do |e|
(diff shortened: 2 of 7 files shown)


app/controllers/objects_controller.rb
```diff
@@ -40,14 +57,11 @@ def create
   end
 
   def update
```
justinlittman, line 59: update switched to use CocinaObjectPersister.

app/controllers/objects_controller.rb
```diff
@@ -56,21 +70,8 @@ def update
   end
 
   def show
```
justinlittman, line 72: show (via load_cocina before action) switched to use CocinaObjectPersister.

app/controllers/objects_controller.rb
```diff
@@ -2,13 +2,15 @@
...
-  before_action :load_item, except: [:create]
+  before_action :load_cocina_object, only: %i[show update_doi_metadata]
+  before_action :load_item, only: %i[show accession update_marc_record destroy notify_goobi]
 
+  # No longer be necessary when remove Fedora.
   rescue_from(Cocina::ObjectUpdater::NotImplemented) do |e|
     json_api_error(status: :unprocessable_entity, message: e.message)
   end
 
-  rescue_from(ActiveFedora::ObjectNotFoundError) do |e|
```
jcoyne, line 11: I don't think you want to remove this clause until there are no more usages of `Dor.find` (e.g. `load_item`)

jcoyne (changes requested):

app/controllers/objects_controller.rb
justinlittman: Good call. Fixed.

app/controllers/application_controller.rb
```diff
@@ -61,4 +61,8 @@ def http_auth_header
   def load_item
     @item = Dor.find(params[:object_id] || params[:id])
   end
+
+  def load_cocina_object
+    @cocina_object = CocinaObjectPersister.find(params[:object_id] || params[:id])
```
jcoyne, line 66: I'm wondering about the name "persister".  this is also doing loading. Can we change the name of the class to CocinaObjectRepository?

jcoyne (approved):
Approved pending consideration of class name change.

app/controllers/application_controller.rb
justinlittman: Renaming to CocinaObjectStore (prefer over Repository).

jcoyne: That works for me.

PR #3188, opened 2022-01-19
[PROOF OF CONCEPT] Support round-tripping italics for some fields

mjgiarlo:
Connects to sul-dlss/exhibits#1981

This is a proof of concept to support shipping italicized data to public XML, so that fields render as expected. This is being done to support the Martin Wong collection work. It has several limitations at this point:

* Only applies italics to abstracts, and only "basic values" (not structured or parallel values, etc.)
* Has not been tested in different systems to see what affect it has (e.g., Argo, PURL)

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/note.rb
+++ b/app/services/cocina/to_fedora/descriptive/note.rb
@@ -5,6 +5,8 @@ module Cocina
     class Descriptive
       # Maps notes from cocina to MODS XML
       class Note
+        ITALIC_TAG_NAMES = %i[abstract].freeze
+
         # @params [Nokogiri::XML::Builder] xml
         # @params [Array<Cocina::Models::DescriptiveValue>] notes
         # @params [IdGenerator] id_generator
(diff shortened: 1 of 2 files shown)


app/services/cocina/to_fedora/descriptive/note.rb
```diff
@@ -65,7 +67,18 @@ def tag(note, tag_name, attributes)
...
                   end
-          xml.public_send tag_name, value, attributes
+
+          xml.public_send tag_name, attributes do
+            if ITALIC_TAG_NAMES.include?(tag_name) && value&.match?('<i>')
+              segments = value.split(%r{</?i>})
+              segments.each.with_index(1) do |segment, i|
+                xml.text segment
+                xml.cdata i.odd? ? '<i>' : '</i>' unless i == segments.size
+              end
```
mjgiarlo, line 77: Maybe there's a better way to create a set of text/cdata nodes in an XML builder, but if so, I didn't find it.

app/services/cocina/to_fedora/descriptive/note.rb
```diff
@@ -65,7 +67,22 @@ def tag(note, tag_name, attributes)
...
+            end
+          end
+        end
+
+        def create_text_cdata_structure(value, xml)
+          segments = value.split(%r{</?i>})
+          segments.each.with_index(1) do |segment, i|
+            xml.text segment
+            xml.cdata i.odd? ? '<i>' : '</i>' unless i == segments.size # avoid creating <i> for the trailing segment
+          end
```
mjgiarlo, line 85: Maybe there's a better way to create a set of text/cdata nodes in an XML builder, but if so, I didn't find it. 🤷🏻‍♂️

PR #3189, opened 2022-01-20
Report to find APOs without agreementids

edsu:
## Why was this change made?

To find APOs without agreement ids

On sdr-deploy:

    bin/reports/report-rels-apo_agreement_ids

using: 

    /opt/app/deploy/dor-services-app/druids.txt

I've attached the report here:

[rels_apo_without_agreement_ids.csv](https://github.com/sul-dlss/dor-services-app/files/7901797/rels_apo_without_agreement_ids.csv)

Here is one of the items it found:

https://argo.stanford.edu/view/druid:cw247ty3859

Being new to the various data models I'm not confident this is finding what was needed, so please check [the logic](https://github.com/sul-dlss/dor-services-app/blob/apo-without-agreementids/bin/reports/report-rels-apo_agreement_ids#L12).

### DIFF
--- a/bin/reports/report-rels-apo_agreement_ids
+++ b/bin/reports/report-rels-apo_agreement_ids
@@ -1,0 +1,15 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative '../../config/environment'
+require_relative '../../lib/report'
+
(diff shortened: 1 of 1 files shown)


justinlittman (approved):
Can you combine into a single commit before merging?

PR #3192, opened 2022-01-20
[HOLD] Upgrades cocina.

justinlittman:
## Why was this change made?
Better cocina normalization.

Unit


justinlittman:
Before:
```
  Success:   94587 (94.665%)
  Different: 5331 (5.335%)
  Mapping error:     0 (0.0%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     82 (0.082%)
  Bad cache:     0 (0.0%)
```

After:
```
  Success:   94587 (94.665%)
  Different: 5331 (5.335%)
  Mapping error:     0 (0.0%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     82 (0.082%)
  Bad cache:     0 (0.0%)
```

jcoyne:
Why does a Cocina upgrade require all these changes to descriptive metadata? Should this be more than one PR?

mjgiarlo (approved):

justinlittman:
@jcoyne Precisely. And we have a lot of code that does `if foo` instead of `if foo.present?`.

PR #3212, opened 2022-01-24
Normalizing missing objectId in contentMetadata

### LINKED ISSUE #3127: Normalize content metadata nodes without objectIds
If a content metadata node is missing the objectId attribute, it should be added.

For example, druid:hk850sn4437 is currently:
```
Normalized original XML for contentMetadata:
<?xml version="1.0" encoding="UTF-8"?>
<contentMetadata type="image"/>
```

but should be:
```
Normalized roundtripped XML for contentMetadata:
<?xml version="1.0" encoding="UTF-8"?>
<contentMetadata objectId="druid:hk850sn4437" type="image"/>
```

lwrubel:
## Why was this change made?

Resolves #3127. Normalization to put content Id at root level.

Before
```
  Success:   94844 (94.922%)
  Different: 5074 (5.078%)
  Mapping error:     0 (0.0%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     82 (0.082%)
  Bad cache:     0 (0.0%)
```
After
```
  Success:   95156 (95.234%)
  Different: 4762 (4.766%)
  Mapping error:     0 (0.0%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     82 (0.082%)
  Bad cache:     0 (0.0%)


```

### DIFF
--- a/app/services/cocina/normalizers/content_metadata_normalizer.rb
+++ b/app/services/cocina/normalizers/content_metadata_normalizer.rb
@@ -31,7 +31,7 @@ module Cocina
         remove_sequence
         remove_location
         remove_format
-        normalize_object_id
+        normalize_object_id(druid)
         normalize_reading_order(druid)
         normalize_label_attr
         normalize_attr
@@ -69,11 +69,16 @@ module Cocina
         ng_xml.root.xpath('//location[@type="url"]').each(&:remove)
       end
 
-      def normalize_object_id
+      def normalize_object_id(druid)
         object_id = ng_xml.root['objectId']
-        return if object_id.nil? || object_id.start_with?('druid:')
 
-        ng_xml.root['objectId'] = "druid:#{object_id}"
+        if object_id
+          return if object_id.start_with?('druid:')
+
+          ng_xml.root['objectId'] = "druid:#{object_id}"
+        else
+          ng_xml.root['objectId'] = druid
+        end
       end
 
       def remove_format
--- a/spec/services/cocina/normalizers/content_metadata_normalizer_spec.rb
+++ b/spec/services/cocina/normalizers/content_metadata_normalizer_spec.rb
@@ -80,6 +80,22 @@ RSpec.describe Cocina::Normalizers::ContentMetadataNormalizer do
     end
   end
 
+  context 'when normalizing missing objectId' do
+    let(:original_xml) do
+      <<~XML
+        <contentMetadata type="image" />
+      XML
+    end
+
+    it 'adds druid as objectId' do
+      expect(normalized_ng_xml).to be_equivalent_to(
+        <<~XML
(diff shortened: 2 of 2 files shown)


app/services/cocina/normalizers/content_metadata_normalizer.rb
```diff
@@ -68,11 +68,16 @@ def remove_location
...
         object_id = ng_xml.root['objectId']
-        return if object_id.nil? || object_id.start_with?('druid:')
 
-        ng_xml.root['objectId'] = "druid:#{object_id}"
+        if object_id
+          return if object_id.start_with?('druid:')
+
+          ng_xml.root['objectId'] = "druid:#{object_id}"
+        else
+          ng_xml.root['objectId'] = druid
```
ndushay, line 79: are we sure the druid variable value always has prefix "druid:" ?

spec/services/cocina/normalizers/content_metadata_normalizer_spec.rb
```diff
@@ -80,6 +80,22 @@
...
+      <<~XML
+        <contentMetadata type="image" />
+      XML
+    end
+
+    it 'adds druid as objectId' do
+      expect(normalized_ng_xml).to be_equivalent_to(
+        <<~XML
+          <contentMetadata objectId="druid:bb035tg0974" type="image" />
+        XML
```
ndushay, line 94: Should there be a test case added for bare druid getting the "druid:" prefix?

app/services/cocina/normalizers/content_metadata_normalizer.rb
lwrubel: Thanks. Sounds like sometimes there might be a druid variable that lacks the "druid" prefix and that needs to be checked before creating that objectId attribute. Will work on that.

spec/services/cocina/normalizers/content_metadata_normalizer_spec.rb
lwrubel: Got it, will likewise add a test for that prefixless scenario.

app/services/cocina/normalizers/content_metadata_normalizer.rb
justinlittman: Yes, the druid variable always has the druid prefix.

lwrubel: Got it, thanks.

spec/services/cocina/normalizers/content_metadata_normalizer_spec.rb
lwrubel: Test added.

ndushay (approved):

justinlittman (approved):

PR #3234, opened 2022-01-26
Cocina-izes Goobi service.

### LINKED ISSUE #3173: Rewrite Dor::Goobi service to use Cocina objects

justinlittman:
closes #3173

## Why was this change made?
Switch Goobi service from using Fedora objects to Cocina objects.

Unit

### DIFF
--- a/app/models/dor/goobi.rb
+++ b/app/models/dor/goobi.rb
@@ -12,8 +12,8 @@ module Dor
                             Faraday::TimeoutError,
                             Faraday::RetriableResponse].freeze
 
-    def initialize(druid_obj)
-      @druid_obj = druid_obj
+    def initialize(cocina_obj)
+      @cocina_obj = cocina_obj
     end
 
     def register
@@ -31,6 +31,8 @@ module Dor
 
     private
 
+    attr_reader :cocina_obj
+
     # We send all tags to Goobi, but "DPG : Workflow : xxx" is the one tag that Goobi uses
     def goobi_xml_tags
       goobi_tag_list.map(&:to_xml).join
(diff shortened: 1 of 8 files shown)


app/models/dor/goobi.rb
```diff
@@ -39,14 +41,14 @@ def goobi_xml_tags
     def xml_request
       <<-END
         <stanfordCreationRequest>
-            <objectId>#{@druid_obj.id}</objectId>
-            <objectType>#{@druid_obj.object_type}</objectType>
-            <sourceID>#{@druid_obj.source_id.encode(xml: :text)}</sourceID>
+            <objectId>#{cocina_obj.externalIdentifier}</objectId>
+            <objectType>item</objectType>
```
justinlittman, line 45: @andrewjbtw Can this be hardcoded to item?

app/services/cocina/to_fedora/descriptive.rb
```diff
@@ -23,7 +23,7 @@ def transform
           xml.mods(mods_attributes) do
             Descriptive::DescriptiveWriter.write(xml: xml, descriptive: descriptive, druid: druid)
           end
-        end
+        end.doc
```
justinlittman, line 26: This was returning a Builder, not a Document (as the method doc specified). Correcting this required removing `.doc` in a handful of locations below.

app/models/dor/goobi.rb
andrewjbtw: What does this do? Is this part of the communication with goobi?

justinlittman: We've determined that this can be hardcoded.

peetucket (approved):

peetucket:
LGTM - not sure if we need to hold for another review or testing on stage before merging

PR #3237, opened 2022-01-26
MODS recordInfoNote mapping

### LINKED ISSUE #2689: Implement recordInfoNote spec
https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/record_info_spec.rb#L491

See #2688

lwrubel:
## Why was this change made?

Resolves #2689 to map recordInfoNote from MODS. 

Unit tests. Validating on sdr-deploy with `bin/validate-desc-cocina-roundtrip`:

Before
```Status (n=100000; not using Missing for success/different/error stats):
  Success:   99984 (99.984%)
  Different: 16 (0.016%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
```

After
```Status (n=100000; not using Missing for success/different/error stats):
  Success:   99984 (99.984%)
  Different: 16 (0.016%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
```

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/admin_metadata.rb
+++ b/app/services/cocina/to_fedora/descriptive/admin_metadata.rb
@@ -24,7 +24,7 @@ module Cocina
             build_content_source
             build_description_standard
             build_record_origin
-            build_record_info_note
+            build_record_info_notes
             build_event
             build_identifier
           end
@@ -40,11 +40,18 @@ module Cocina
           end
         end
 
-        def build_record_info_note
-          record_info_note = Array(admin_metadata.note).find(&:valueAt)
-          return unless record_info_note
+        def build_record_info_notes
+          record_info_notes = Array(admin_metadata.note)
+          # return unless record_info_notes
+          return if record_info_notes.empty?
 
-          xml.recordInfoNote nil, { 'xlink:href' => record_info_note.valueAt }
+          record_info_notes.each do |info_note|
+            if info_note.valueAt.present?
+              xml.recordInfoNote nil, { 'xlink:href' => info_note.valueAt }
+            elsif info_note.type == 'record information'
+              xml.recordInfoNote info_note.value
+            end
+          end
         end
 
         def build_content_source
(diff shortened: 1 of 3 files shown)


app/services/cocina/to_fedora/descriptive/admin_metadata.rb
```diff
@@ -40,11 +40,17 @@ def build_record_origin
           end
         end
 
-        def build_record_info_note
-          record_info_note = Array(admin_metadata.note).find(&:valueAt)
-          return unless record_info_note
+        def build_record_info_notes
+          record_info_notes = Array(admin_metadata.note)
+          return unless record_info_notes
```
mjgiarlo, line 45: I *think* `record_info_notes` will always be truthy here, since `Array(nil)` is truthy. I assume the intent here is to return when `record_info_notes` is an empty array? Would `return if record_info_notes.empty?` work here?

mjgiarlo (commented):
Have a question about a conditional

app/services/cocina/to_fedora/descriptive/admin_metadata.rb
lwrubel: Thanks for catching that. Looks like `.empty?` would work and also would be better because would account for possible empty arrays. Will push a commit to address that.

mjgiarlo: @lwrubel Sounds good, thanks!

mjgiarlo: @lwrubel Do you think it's worth re-running the roundtripper on sdr-deploy or nah? I'll approve now and leave to your judgment.

mjgiarlo (approved):

lwrubel:
It doesn't take long to run for descriptive metadata and I'd prefer to be thorough, so am running it to confirm if anything changes.
________________________________

lwrubel:
Looks like nothing changed/got worse with using `.empty?` so that's good.

```Status (n=100000; not using Missing for success/different/error stats):
  Success:   99984 (99.984%)
  Different: 16 (0.016%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
```

PR #3239, opened 2022-01-26
Fix validate-to-desc-cocina to handle a missing Fedora item.

### LINKED ISSUE #3216: validate-to-desc-cocina report fails if druid not found in Fedora
I'm trying to run the validate-to-desc-cocina report on sdr-deploy, and am finding that the report breaks when it gets to a druid that isn't in Fedora. I tried refreshing the cache for those druids but nothing happens because the druid does not exist. Improved behavior would be to report the error and continue running the report, and/or to remove nonexistent druids from the cache. This is slowing down work on pre-migration remediation of descriptive metadata.

justinlittman:
closes #3216

## Why was this change made?
Handle the case where an item is deleted from Fedora, but not deleted from cache.

sdr-deploy

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,9 +210,12 @@ Using the druids from `druids.txt` and the cache, this will create a Fedora item
 ### Validate mapping to Cocina from MODS (descriptive metadata only)
 ```
 $ bin/validate-to-desc-cocina -h
-Usage: bin/validate-to-cocina [options]
+Usage: bin/validate-to-desc-cocina [options]
     -s, --sample SAMPLE              Sample size, otherwise all druids.
     -u, --unique-filename            Result file named for branch and runtime
+    -a, --apo                        Include APO in report (slower).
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):
Worth also tossing a tweak to https://github.com/sul-dlss/dor-services-app#validate-mapping-to-cocina-from-mods-descriptive-metadata-only in the README as part of this PR? Or a follow-on? approving either way

justinlittman:
Good call @mjgiarlo. Updated README.

README.md
```diff
@@ -210,9 +210,12 @@ Using the druids from `druids.txt` and the cache, this will create a Fedora item
 ### Validate mapping to Cocina from MODS (descriptive metadata only)
 ```
 $ bin/validate-to-desc-cocina -h
-Usage: bin/validate-to-cocina [options]
+Usage: bin/validate-to-desc-cocina [options]
```
mjgiarlo, line 213: Good catch!

PR #3250, opened 2022-01-27
Cocina-ize the AdminPolicyDefaultsController

### LINKED ISSUE #3168: Rewrite AdminPolicyDefaultsController for Cocina objects

mjgiarlo:
And apply APO defaults only to items that are in the registered or opened workflow states.

## Why was this change made?

Fixes #3168
Fixes sul-dlss/argo#3019

This commit changes the AdminPolicyDefaultsController so that it saves and loads Cocina objects instead of Fedora ones. It tests both cases of default defaultAccess and custom defaultAccess. It also copies over an RSpec matcher from Argo allowing comparison of Cocina substructures, in this case the access substructure.

CI. Will also test in stage.

None

### DIFF
--- a/app/controllers/admin_policy_defaults_controller.rb
+++ b/app/controllers/admin_policy_defaults_controller.rb
@@ -32,7 +32,28 @@ class AdminPolicyDefaultsController < ApplicationController
   end
 
   def updated_cocina_object
-    @cocina_object.new(access: @cocina_object.access.new(**default_access_from_apo))
+    @cocina_object.new(
+      access: @cocina_object.access.new(**default_access_from_apo),
+      structural: @cocina_object.structural.new(
+        contains: @cocina_object.structural.contains.map do |file_set|
+          file_set.new(
+            structural: file_set.structural.new(
+              contains: file_set.structural.contains.map do |file|
+                file.new(
+                  access: file.access.new(
+                    default_access_from_apo.to_h.slice(*file_access_props)
+                  )
+                )
+              end
+            )
+          )
+        end
+      )
+    )
+  end
+
+  def file_access_props
+    %i[access controlledDigitalLending download readLocation]
   end
 
   def default_access_from_apo
--- a/app/controllers/admin_policy_defaults_controller.rb
+++ b/app/controllers/admin_policy_defaults_controller.rb
@@ -7,7 +7,8 @@ class AdminPolicyDefaultsController < ApplicationController
   before_action :load_cocina_object, only: :apply
 
   def apply
-    return error_response unless current_workflow_state.in?(ALLOWED_WORKFLOW_STATES)
+    return type_error_response unless @cocina_object.dro? || @cocina_object.collection?
+    return workflow_error_response unless current_workflow_state.in?(ALLOWED_WORKFLOW_STATES)
 
     CocinaObjectStore.save(updated_cocina_object)
     head :no_content
@@ -22,7 +23,15 @@ class AdminPolicyDefaultsController < ApplicationController
       .display_simplified
   end
 
-  def error_response
+  def type_error_response
+    json_api_error(
+      status: :bad_request,
+      message: "#{@cocina_object.externalIdentifier} is a #{@cocina_object.class} and this type cannot currently have APO access defaults applied",
+      title: 'Object cannot inherit APO access defaults'
+    )
+  end
+
+  def workflow_error_response
     json_api_error(
       status: :unprocessable_entity,
       message: "#{@cocina_object.externalIdentifier} is in a state in which it cannot be modified (#{current_workflow_state}): " \
@@ -32,17 +41,19 @@ class AdminPolicyDefaultsController < ApplicationController
   end
 
   def updated_cocina_object
-    @cocina_object.new(
-      access: @cocina_object.access.new(**default_access_from_apo),
+    access_updated = @cocina_object.new(
+      access: @cocina_object.access.new(default_access_from_apo)
+    )
+    return access_updated unless access_updated.dro? && access_updated.structural&.contains&.any?
+
+    access_updated.new(
       structural: @cocina_object.structural.new(
         contains: @cocina_object.structural.contains.map do |file_set|
           file_set.new(
             structural: file_set.structural.new(
               contains: file_set.structural.contains.map do |file|
                 file.new(
-                  access: file.access.new(
-                    default_access_from_apo.to_h.slice(*file_access_props)
-                  )
+                  access: file.access.new(default_access_from_apo(file_level: true))
                 )
               end
             )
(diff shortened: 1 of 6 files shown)


spec/requests/apply_admin_policy_defaults_spec.rb
```diff
@@ -4,22 +4,91 @@
 
 RSpec.describe 'Get the object' do
   before do
-    allow(Dor).to receive(:find).and_return(object)
-    allow(object).to receive(:save!)
+    allow(Dor).to receive(:find).with(object_pid).and_return(object)
```
justinlittman, line 7: I would expect the spec to not have Fedora in it. Should you be mocking `CocinaObjectStore.find` instead?

justinlittman (changes requested):

spec/requests/apply_admin_policy_defaults_spec.rb
mjgiarlo: @justinlittman I changed the stubs to use `CocinaObjectStore.fedora_find`.

spec/requests/apply_admin_policy_defaults_spec.rb
```diff
@@ -2,24 +2,92 @@
...
+    allow(object).to receive(:admin_policy_object_id).and_return(apo_pid)
+    allow(admin_policy).to receive(:admin_policy_object_id).and_return(ur_apo_pid)
   end
 
-  let(:object) { Dor::Item.new(admin_policy_object: admin_policy) }
-  let(:admin_policy) do
-    Dor::AdminPolicyObject.new.tap do |coll|
-      coll.defaultObjectRights.content = '<foo/>'
+  let(:apo_pid) { 'druid:df123cd4567' }
+  let(:object) do
```
justinlittman, line 15: I still think this test should be entirely abstracted from Fedora. Ping me and let's discuss.

justinlittman (changes requested):

spec/requests/apply_admin_policy_defaults_spec.rb
mjgiarlo: @justinlittman Oh, no worries, I can do that. No discussion needed on my part.

I was hesitant to start off with a pure unit test, mostly because I hadn't been in DSA for a long time and had low confidence that I could put a bunch of mocks and stubs in and feel good that the changes I made would work as intended in a deployed environment. So even if we back out those bits, they were a valuable demonstration to me that things are wired up right.

justinlittman (dismissed):

mjgiarlo:
TO-DO:

* UI reports “defaults applied” even when not applied;
* Files should have rights of “Stanford,Stanford” but ended up with “dark,none” (for test item: https://argo-stage.stanford.edu/view/druid:bf345my0487)

app/controllers/admin_policy_defaults_controller.rb
```diff
@@ -2,11 +2,64 @@
...
 class AdminPolicyDefaultsController < ApplicationController
-  before_action :load_item, only: :apply
+  ALLOWED_WORKFLOW_STATES = %w[Registered Opened].freeze
+
+  before_action :load_cocina_object, only: :apply
 
   def apply
-    @item.rightsMetadata.content = @item.admin_policy_object.defaultObjectRights.content
-    @item.save!
+    return error_response unless current_workflow_state.in?(ALLOWED_WORKFLOW_STATES)
```
jcoyne, line 10: Does Argo already do this check?

mjgiarlo: @jcoyne Not that I saw, anyway.

app/controllers/admin_policy_defaults_controller.rb
```diff
@@ -1,12 +1,116 @@
 # frozen_string_literal: true
 
+# TODO: This controller is doing way too much now. Consider extracting most of what is below into a new service object.
```
mjgiarlo, line 3: Dear reviewer(s): take note! I am inclined to take up this refactoring after this PR is merged, ideally.

app/controllers/admin_policy_defaults_controller.rb
```diff
@@ -1,12 +1,116 @@
...
+    'stanford' => 'world',
+    'world' => 'world'
+  }.freeze
+  FILE_ACCESS = {
+    'citation-only' => 'dark',
+    'dark' => 'dark',
+    'location-based' => 'location-based',
+    'stanford' => 'stanford',
+    'world' => 'world'
+  }.freeze
```
mjgiarlo, line 20: It would be lovely to only have one place in DSA to validate/map rights, and I can investigate that possibility in the refactoring PR.

app/controllers/admin_policy_defaults_controller.rb
```diff
@@ -1,12 +1,116 @@
...
+
+  before_action :load_cocina_object, only: :apply
 
   def apply
-    @item.rightsMetadata.content = @item.admin_policy_object.defaultObjectRights.content
-    @item.save!
+    return type_error_response unless @cocina_object.dro? || @cocina_object.collection?
+    return workflow_error_response unless current_workflow_state.in?(ALLOWED_WORKFLOW_STATES)
```
mjgiarlo, line 26: @andrewjbtw confirmed this is the right behavior.

app/services/workflow_client_factory.rb
```diff
@@ -6,7 +6,10 @@ def self.build
     logger = if Settings.workflow.logfile == 'rails'
                Rails.logger
              else
-               Logger.new(Settings.workflow.logfile, Settings.workflow.shift_age)
+               Logger.new(
+                 Rails.root.join(Settings.workflow.logfile),
+                 Settings.workflow.shift_age
+               )
```
mjgiarlo, line 12: Unrelated to the issue I'm working on, but it's tiny and could nuke an HB alert we've been seeing forever. 🤷🏻‍♂️

config/application.rb
```diff
@@ -42,13 +42,13 @@ class Application < Rails::Application
...
       query_hash_key: 'action_dispatch.request.query_parameters'
     )
-
     # Ensure we are passing back valid responses when running tests
     if Rails.env.test?
       config.middleware.use(
         Committee::Middleware::ResponseValidation,
         schema_path: 'openapi.yml',
         parse_response_by_content_type: true,
+        query_hash_key: 'rack.request.query_hash',
```
mjgiarlo, line 51: Nukes a deprecation warning in the test suite. Has no effect on production apps.

spec/support/cocina_matchers.rb
```diff
@@ -32,6 +32,25 @@ module CocinaMatchers
...
+            if expected_value.is_a?(Hash) && actual_value.is_a?(Hash)
+              expected_value.all? { |pair| actual_value.to_a.include?(pair) }
+            else
+              actual_key == expected_key && actual_value == expected_value
+            end
+          end
+        end
+      end
+    end
+  end
```
mjgiarlo, line 53: I wonder if the cocina-models should ship with this matcher? But maybe a thing for later.

justinlittman (approved):

PR #3254, opened 2022-01-28
Remove the license if the user has said so

mjgiarlo:
Fixes sul-dlss/argo#3018

## Why was this change made?

Without this change, the APO edit page in Argo (e.g., https://argo-stage.stanford.edu/view/druid:sc012gz0974) fails to do what the user expects when they set the default use license for an APO to none.

CI + @andrewjbtw tested successfully IN QA

None

### DIFF
--- a/app/services/cocina/to_fedora/license.rb
+++ b/app/services/cocina/to_fedora/license.rb
@@ -16,9 +16,10 @@ module Cocina
       def update
         initialize_use_node!
         clear_licenses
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

mjgiarlo:
Cannot test right now because QA is still borked for some reason. Looking into this tomorrow morning.

aaron-collier (approved):

PR #3255, opened 2022-01-28
FileSets: if HB alerting on an invalid type, tag alert as data error

### LINKED ISSUE #3232: clarify HB alert for "Invalid resource type"
From druid:zn580kw6428

## Backtrace

    line 54 of [PROJECT_ROOT]/app/services/cocina/from_fedora/file_sets.rb: resource_type
    line 30 of [PROJECT_ROOT]/app/services/cocina/from_fedora/file_sets.rb: block in build
    line 24 of [PROJECT_ROOT]/app/services/cocina/from_fedora/file_sets.rb: map

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/79080969)

jmartin-sul:
## Why was this change made?

closes #3232

after grepping around to double-check that i was using the right alert format, i realized i'd missed some applicable centralized code for this sort of alert, but also that the existing code wanted a druid, which wasn't immediately available from where the alert was being sent.  i threw on a second commit passing in the druid from the caller, and switching `FileSets` to use the centralized notifier code, but if it turns out that's bad for some reason, i can knock that commit off (or otherwise tweak as needed, of course).

unit tests

n/a

### DIFF
--- a/app/services/cocina/from_fedora/file_sets.rb
+++ b/app/services/cocina/from_fedora/file_sets.rb
@@ -4,20 +4,22 @@ module Cocina
   module FromFedora
     # builds the FileSet instance from a Dor::Item
     class FileSets
-      def self.build(content_metadata_ds, rights_metadata:, version:, ignore_resource_type_errors: false)
+      def self.build(content_metadata_ds, rights_metadata:, version:, druid:, ignore_resource_type_errors: false)
         new(
           content_metadata_ds,
           rights_metadata: rights_metadata,
           version: version,
+          druid: druid,
           ignore_resource_type_errors: ignore_resource_type_errors
         ).build
       end
 
-      def initialize(content_metadata_ds, rights_metadata:, version:, ignore_resource_type_errors:)
+      def initialize(content_metadata_ds, rights_metadata:, version:, druid:, ignore_resource_type_errors:)
         @content_metadata_ds = content_metadata_ds
         @rights_metadata = rights_metadata
         @version = version
         @ignore_resource_type_errors = ignore_resource_type_errors
+        @data_err_notifier = DataErrorNotifier.new(druid: druid)
       end
 
       def build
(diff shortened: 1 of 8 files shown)


app/services/cocina/from_fedora/file_sets.rb
```diff
@@ -4,20 +4,22 @@ module Cocina
...
         ).build
       end
 
-      def initialize(content_metadata_ds, rights_metadata:, version:, ignore_resource_type_errors:)
+      def initialize(content_metadata_ds, rights_metadata:, version:, druid:, ignore_resource_type_errors:)
         @content_metadata_ds = content_metadata_ds
         @rights_metadata = rights_metadata
         @version = version
         @ignore_resource_type_errors = ignore_resource_type_errors
+        @data_err_notifier = DataErrorNotifier.new(druid: druid)
```
justinlittman, line 22: This should be passed as a parameter rather than instantiated.

justinlittman (approved):

app/services/cocina/from_fedora/file_sets.rb
justinlittman: You can pass from Cocina::FromFedora::DRO: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/from_fedora/dro.rb#L64

justinlittman (changes requested):

jmartin-sul:
ok, i switched to passing along the existing notifier instance, instead of instantiating anew based on a druid given to `FileSets.build`.  that resulted in a bit more churn than i'd have expected, so left that as a separate commit just in case we want to walk that back.  but since it sounds like it's fine to use the centralized notifier one way or the other, i squashed that change into the first commit, where the `data_error` tag was given to an inline `Honeybadger.notify` call.

i tested the validation script change by running `bin/validate-rights-cocina-roundtrip -s 100000` on sdr-deploy using this branch.  that completed with no problem, but it also didn't detect any errors mapping to either cocina or fedora, so i'm not sure that change got exercised.  but i was just following the pattern i saw in `bin/validate-cocina-roundtrip`, so hopefully that's ok.

i think this is ready for another look.

justinlittman (approved):

PR #3261, opened 2022-01-28
We currently use the objectId in ThumbnailService over resourceId

aaron-collier:
## Why was this change made?

This should be reviewed by @andrewjbtw, @arcadiafalcone, @jcoyne, and @justinlittman to verify that this small change is ok.

When working in ThumbnailService, I noticed we expect the objectId (that includes a druid) to be used instead of resourceId which does not.

https://github.com/sul-dlss/dor-services-app/blob/main/app/services/thumbnail_service.rb#L50

For example, with content metadata of:

```
<contentMetadata objectId="tf960jc7725" type="image">
  <resource id="tf960jc7725_1" sequence="1" type="image">
    <externalFile fileId="2576001.jp2" mimetype="image/jp2" objectId="druid:pj563wy4429" resourceId="pj563wy4429_1"/>
    <relationship objectId="druid:pj563wy4429" type="alsoAvailableAs"/>
  </resource>
  <resource id="tf960jc7725_2" sequence="2" type="image">
    <externalFile fileId="2576002.jp2" mimetype="image/jp2" objectId="druid:jv233bw7647" resourceId="jv233bw7647_1"/>
    <relationship objectId="druid:jv233bw7647" type="alsoAvailableAs"/>
  </resource>
</contentMetadata>
```

we currently map as:
```
pj563wy4429_1/2576001.jp2
jv233bw7647_1/2576002.jp2
```

However, ThumbnailService expects:
```
druid:pj563wy4429/2576001.jp2
druid:jv233bw7647/2576002.jp2
```

If this was a design decision that I can ignore, please close. 

Unit.

### DIFF
--- a/app/services/cocina/from_fedora/dro_structural.rb
+++ b/app/services/cocina/from_fedora/dro_structural.rb
@@ -61,7 +61,7 @@ module Cocina
       # @return [Array<String>] the identifiers of files in a sequence for a virtual object
       def build_sequence(content_metadata_ds)
         content_metadata_ds.ng_xml.xpath('//resource/externalFile').map do |resource_node|
-          "#{resource_node['resourceId']}/#{resource_node['fileId']}"
+          "#{resource_node['objectId']}/#{resource_node['fileId']}"
         end
       end
     end
--- a/spec/requests/show_object_spec.rb
+++ b/spec/requests/show_object_spec.rb
@@ -257,17 +257,17 @@ RSpec.describe 'Get the object' do
                                     hasMemberOrders: [
                                       {
                                         members: [
-                                          'kq126jw7402_1/1592A.jp2',
-                                          'cv761kr7119_1/1592B.jp2',
-                                          'kn300wd1779_1/1592c.jp2',
-                                          'rz617vr4473_1/1592001.jp2',
-                                          'sd322dt2118_1/1592002.jp2',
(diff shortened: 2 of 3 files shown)


justinlittman:
I don't know anything about the intent / design.

app/services/cocina/from_fedora/dro_structural.rb
```diff
@@ -61,7 +61,7 @@ def create_member_order
       # @return [Array<String>] the identifiers of files in a sequence for a virtual object
       def build_sequence(content_metadata_ds)
         content_metadata_ds.ng_xml.xpath('//resource/externalFile').map do |resource_node|
-          "#{resource_node['resourceId']}/#{resource_node['fileId']}"
+          "#{resource_node['objectId']}/#{resource_node['fileId']}"
```
jcoyne, line 64: This was expected to be the resource id, not the object id filename.

jcoyne (changes requested):

app/services/cocina/from_fedora/dro_structural.rb
aaron-collier: Are we sure the `pj563wy4429_1/filename.jp2` works? From what I can tell in existing tests the `_1` is not included, but happy to ditch this if it's not needed.

PR #3264, opened 2022-01-29
Cocina-ize ThumbnailService

### LINKED ISSUE #3253: Rewrite ThumbnailService to use Cocina objects
https://github.com/sul-dlss/dor-services-app/blob/main/app/services/thumbnail_service.rb

aaron-collier:
## Why was this change made?

Closes #3253 

- Rewrites ThumbnailService to take a Cocina object
- Updates PublicXmlService to use a Cocina object for thumbnails
- Updates UpdateMarcRecordService to use a Cocina object for thumbnails

UpdateMarcRecordService (in progress) and PublicXmlService will be completely rewritten for cocina in follow up issues/PRs.

Updated unit tests

### DIFF
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -9,10 +9,11 @@ module Dor
     # objects goverened by these APOs (ETD and EEMs) will get indicator 2 = 0, else 1
     BORN_DIGITAL_APOS = %w(druid:bx911tp9024 druid:jj305hm5259).freeze
 
-    def initialize(druid_obj)
+    def initialize(druid_obj, thumbnail_service:)
       @druid_obj = druid_obj
       @druid_id = Dor::PidUtils.remove_druid_prefix(druid_obj.id)
       @dra_object = druid_obj.rightsMetadata.dra_object
+      @thumbnail_service = thumbnail_service
     end
 
     def update
@@ -241,8 +242,7 @@ module Dor
     # the @id attribute of resource/file elements including extension
     # @return [String] thumbnail filename (nil if none found)
     def thumb
-      cocina_object = CocinaObjectStore.find(@druid_obj.pid)
-      @thumb ||= ERB::Util.url_encode(ThumbnailService.new(cocina_object).thumb).presence
+      @thumb ||= ERB::Util.url_encode(@thumbnail_service.thumb).presence
     end
   end
   # rubocop:enable Metrics/ClassLength
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -241,7 +241,8 @@ module Dor
     # the @id attribute of resource/file elements including extension
     # @return [String] thumbnail filename (nil if none found)
     def thumb
-      @thumb ||= ERB::Util.url_encode(ThumbnailService.new(@druid_obj).thumb).presence
+      cocina_object = CocinaObjectStore.find(@druid_obj.pid)
+      @thumb ||= ERB::Util.url_encode(ThumbnailService.new(cocina_object).thumb).presence
     end
   end
   # rubocop:enable Metrics/ClassLength
--- a/app/services/publish/public_xml_service.rb
+++ b/app/services/publish/public_xml_service.rb
@@ -31,7 +31,9 @@ module Publish
       # Note we cannot base this on if an individual object has release tags or not, because the collection may cause one to be generated for an item,
       # so we need to calculate it and then look at the final result.
 
-      thumb = ThumbnailService.new(object).thumb
+      cocina_object = CocinaObjectStore.find(object.pid)
+      thumb = ThumbnailService.new(cocina_object).thumb
+
       pub.add_child(Nokogiri("<thumb>#{thumb}</thumb>").root) unless thumb.nil?
 
       new_pub = Nokogiri::XML(pub.to_xml, &:noblanks)
--- a/app/services/publish/public_xml_service.rb
+++ b/app/services/publish/public_xml_service.rb
@@ -7,9 +7,10 @@ module Publish
 
     # @param [Dor::Item] object
     # @param [Hash{String => Boolean}] released_for keys are Project name strings, values are boolean
-    def initialize(object, released_for:)
+    def initialize(object, released_for:, thumbnail_service:)
       @object = object
       @released_for = released_for
+      @thumbnail_service = thumbnail_service
     end
 
     # @raise [Dor::DataError]
@@ -31,9 +32,7 @@ module Publish
       # Note we cannot base this on if an individual object has release tags or not, because the collection may cause one to be generated for an item,
       # so we need to calculate it and then look at the final result.
 
-      cocina_object = CocinaObjectStore.find(object.pid)
-      thumb = ThumbnailService.new(cocina_object).thumb
-
+      thumb = @thumbnail_service.thumb
       pub.add_child(Nokogiri("<thumb>#{thumb}</thumb>").root) unless thumb.nil?
 
       new_pub = Nokogiri::XML(pub.to_xml, &:noblanks)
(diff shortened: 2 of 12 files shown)


app/services/thumbnail_service.rb
```diff
@@ -34,24 +14,21 @@ def initialize(object)
...
-      break # break out of the loop so we stop searching
+    return unless object.respond_to?(:structural) && object.structural.present?
+
+    object.structural.contains.each do |attachment|
+      next unless attachment.respond_to?(:structural) && attachment.structural.present?
+
+      attachment.structural.contains.each do |file|
+        next unless file.hasMimeType.include?(MIME_TYPE)
+
+        # if we find one, return the filename based on whether it is a local file or external file
```
jcoyne, line 25: How is this supporting local files?

aaron-collier: Yeah probably, it was an older comment.

app/models/dor/update_marc_record_service.rb
```diff
@@ -241,7 +241,10 @@ def previous_ckeys
     # the @id attribute of resource/file elements including extension
     # @return [String] thumbnail filename (nil if none found)
     def thumb
-      @thumb ||= ERB::Util.url_encode(ThumbnailService.new(@druid_obj).thumb).presence
+      return unless @druid_obj
```
jcoyne, line 244: Why is this check necessary?

app/models/dor/update_marc_record_service.rb
```diff
@@ -241,7 +241,10 @@ def previous_ckeys
...
+
+      cocina_object = CocinaObjectStore.find(@druid_obj.pid)
```
jcoyne, line 246: Is it necessary to do this find if `@thumb` is already defined?

app/services/publish/public_xml_service.rb
```diff
@@ -31,7 +31,9 @@ def to_xml(**)
       # Note we cannot base this on if an individual object has release tags or not, because the collection may cause one to be generated for an item,
       # so we need to calculate it and then look at the final result.
 
-      thumb = ThumbnailService.new(object).thumb
+      cocina_object = CocinaObjectStore.find(object.pid)
```
jcoyne, line 34: would this cause an extra load?  Is it possible to do dependency injection with the ThumbnailService instance when instantiating PublicXMLService to avoid that?

app/services/thumbnail_service.rb
```diff
@@ -34,24 +14,20 @@ def initialize(object)
...
-      next if thumb_files.empty?
-
-      # if we find one, return the filename based on whether it is a local file or external file
-      thumb_image = if search_path[:image_type] == 'local'
-                      "#{Dor::PidUtils.remove_druid_prefix(object.pid)}/#{thumb_files[0]['id']}"
-                    else
-                      "#{Dor::PidUtils.remove_druid_prefix(thumb_files[0]['objectId'])}/#{thumb_files[0]['fileId']}"
-                    end
-      break # break out of the loop so we stop searching
+    return unless object.respond_to?(:structural) && object.structural.present?
```
jcoyne, line 17: Why is `present?` necessary here?

app/services/thumbnail_service.rb
```diff
@@ -34,24 +14,20 @@ def initialize(object)
...
+
+    object.structural.contains.each do |attachment|
```
jcoyne, line 19: Please use `file_set` rather than `attachment` here, as it is an instance of `Cocina::FileSet`

app/services/thumbnail_service.rb
```diff
@@ -34,24 +14,20 @@ def initialize(object)
...
+      next unless attachment.respond_to?(:structural) && attachment.structural.present?
```
jcoyne, line 20: Is there a case where FileSet would not respond to structural? is `present?` necessary?

jcoyne (changes requested):

app/services/thumbnail_service.rb
aaron-collier: perhaps mostly being extra cautious... happy to remove any unnecessary checks.

app/models/dor/update_marc_record_service.rb
aaron-collier: It may not be. I recall in one test that it was `nil` but that could have been related to something else. I'll try without.

app/services/thumbnail_service.rb
aaron-collier: Probably not, following the existing pattern, which is likely over cautious now.

app/models/dor/update_marc_record_service.rb
aaron-collier: @jcoyne I'm not sure I'm following. In this case, we're starting with a `Dor::Item` and need to pass a `Cocina::Model` to the ThumbnailService.

app/services/publish/public_xml_service.rb
aaron-collier: This is somewhat temporary as once PublicXmlService is rewritten for cocina we won't need this. I'd be happy to pair through dependency injection if you think that's the write path forward, but I'm not grokking how it may help in this case.

app/services/thumbnail_service.rb
aaron-collier: I get a load of failures without it. I'll think through a bit of a refactor to improve this though.

aaron-collier: I suspect both of these checks are required. If a `Cocina::Models::Collection` is passed in, the model does not include `structural`, so a check for `present?` will fail. But also, if a DRO is passed in that has does not include `structural` in its data (as it's not required) then the loop over `contains` fails. I'm happy to suggestions and pair though if I'm missing anything.

app/models/dor/update_marc_record_service.rb
jcoyne: How about refactoring this class so that it's initializer takes both a Dor::Item and an instace of ThumbnailService (depenency injection).  Then the class doesn't need to know how to find a cocina object.

app/services/publish/public_xml_service.rb
jcoyne: Sounds good, lets pair together this morning.

app/services/thumbnail_service.rb
jcoyne: The existing pattern is looking for nil or an empty string. The new model makes an empty string impossible. 
I'm pretty sure `return unless object.respond_to?(:structural) && object.structural` would be sufficient, right?

aaron-collier: Ah, yeah - that is correct. :+1:

jcoyne (approved):
Looks great! Thanks for the changes.

app/services/thumbnail_service.rb
```diff
@@ -34,24 +14,18 @@ def initialize(object)
...
+        next unless file.hasMimeType.include?(MIME_TYPE)
+
+        return "#{Dor::PidUtils.remove_druid_prefix(object.externalIdentifier)}/#{file.filename}"
+      end
     end
 
-    thumb_image
+    return if object.structural.hasMemberOrders.empty?
+
+    object.structural.hasMemberOrders.first.members.first
```
mjgiarlo, line 29: Now that I'm touching member orders as part of cocinization of virtual objects, I'm wondering how we know the first member of the first member order is the thumb. @aaron-collier @jcoyne @justinlittman

mjgiarlo: @andrewjbtw this looks like the source of the bug in #4456 which suggests it's been out there for a yearish 😬

andrewjbtw: Hmm. I think this only affects places where a thumbnail appears in SW (not many), and not the virtual object functionality itself, which might be how we haven't heard any reports. I feel like it's taken me the year since the migration to fully understand all the possibilities in an 856.

PR #3265, opened 2022-01-29
Apply APO defaults only to items that are in the registered or opened states

mjgiarlo:
## Why was this change made?

Fixes sul-dlss/argo#3019

CI, will also test in stage

None


mjgiarlo:
Superseded by #3250.

PR #3272, opened 2022-01-31
Implement previous catkey mapping

### LINKED ISSUE #3235: Map otherId[type="previous_catkey"]
Depends on https://github.com/sul-dlss/cocina-models/issues/296

edsu:
Closes #3235

## Why was this change made?

To map previous catkeys to Cocina.


app/services/cocina/from_fedora/identification.rb
```diff
@@ -55,11 +55,13 @@ def doi
...
-          .map { |id| { catalog: 'symphony', catalogRecordId: id.text } }
-          .uniq { |clink| clink[:catalogRecordId] }
-          .presence
+        results = []
+        fedora_object.identityMetadata.ng_xml.xpath('//otherId[@name="catkey" or @name="previous_catkey"]').each do |clink|
+          catalog = clink['name'] == 'catkey' ? 'symphony' : 'previous symphony'
+          results << { catalog: catalog, catalogRecordId: clink.text }
+        end
+
+        results.uniq { |clink| clink[:catalogRecordId] }.presence
```
justinlittman, line 64: If there is a previous catkey and a catkey with the same value, might this accidentally retain the previous catkey and remove the catkey?

edsu: I'll add a test for that -- it does seem possible.

edsu: @justinlittman I've added [a test](https://github.com/sul-dlss/dor-services-app/blob/891956879cb193d19aa26836eb2c35adf83c4195/spec/services/cocina/mapping/identification/dro_identity_spec.rb#L357-L403) and you're right it does look like the current catkey can get stomped on when the order of the catkeys is just right:

```
  1) Fedora Item identityMetadata <--> Cocina DRO Identification mappings with duplicate current/previous catkeys behaves like DRO Identification Fedora Cocina mapping when mapping from Cocina to (roundtrip) Fedora identityMetadata roundtrips thru cocina model to expected roundtrip identityMetadata.xml
     Failure/Error: expect(roundtrip_fedora_item.identityMetadata.ng_xml).to be_equivalent_to(roundtrip_identity_md_xml)

       expected:
       <identityMetadata>
         <sourceId source="sul">naxos_nac_8.557501</sourceId>
         <otherId name="previous_catkey">8501137</otherId>
         <otherId name="catkey">8501137</otherId>
         <objectLabel>The rite of spring</objectLabel>
         <objectId>druid:bb010dx6027</objectId>
         <objectType>item</objectType>
       </identityMetadata>

       got:
       <?xml version="1.0"?>
       <identityMetadata>
         <sourceId source="sul">naxos_nac_8.557501</sourceId>
         <objectId>druid:bb010dx6027</objectId>
         <objectType>item</objectType>
         <objectLabel>The rite of spring</objectLabel>
         <otherId name="previous_catkey">8501137</otherId>
       </identityMetadata>
     Shared Example Group: "DRO Identification Fedora Cocina mapping" called from ./spec/services/cocina/mapping/identification/dro_identity_spec.rb:358
```

edsu: @justinlittman Is the desired behavior to map both to Cocina and round trip them back?

```javascript
  catalogLinks: [
    {
      catalog: 'symphony',
      catalogRecordId: '12345'
    },
    {
      catalog: 'previous symphony',
      catalogRecordId: '12345'
    }
  ]
```

justinlittman: Correct.

edsu:
I'm rerunning the roundtrip tests to make sure that I didn't disturb those results with my latest commits.

edsu:
Here are the new mapping stats:

Before:

```
  Success:   96092 (96.174%)
  Different: 3822 (3.825%)
  Mapping error:     1 (0.001%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     85 (0.085%)
  Bad cache:     0 (0.0%)
```

After:

```
  Success:   96226 (96.308%)
  Different: 3688 (3.691%)
  Mapping error:     1 (0.001%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     85 (0.085%)
  Bad cache:     0 (0.0%)
```

The bit I'm most confused about is why [this](https://github.com/sul-dlss/dor-services-app/blob/fc898b76e78d30121e4fc95810d7849dbb2197b4/app/services/cocina/to_fedora/identity.rb#L75) is needed.

app/services/cocina/to_fedora/identity.rb
```diff
@@ -68,14 +68,17 @@ def apply_catalog_links(catalog_links)
...
         identity_md.ng_xml_will_change!
-        catalog_links
-          # only symphony links are catkeys
-          .filter { |clink| clink.catalog == 'symphony' }
-          # filter out object.catkey to prevent duplicates
-          .filter { |clink| clink.catalogRecordId != identity_md.catkey }
-          .each do |clink|
+
+        catalog_links.each do |clink|
+          # guard against adding an otherid catkey if it has already been set
```
justinlittman, line 73: According to https://github.com/sul-dlss/dor-services/blob/c2c7eeb84cfae6e84a7b636f2e08159838cfe699/lib/dor/datastreams/identity_metadata_ds.rb, `identity_md.catkey` will only return the first catkey and `identity_md.add_value` will add even if it already exists. Thus, the below logic will allow dupes.

What I would suggest instead is something like:
```
identity_md.remove_other_id('catkey')
identity_md.remove_other_id('previous catkey')
catalog_links.each do |clink|
   if clink.catalog == 'symphony'
      identity_md.add_value(:otherId, clink.catalogRecordId, { name: 'catkey' })
    elsif clink.catalog == 'previous symphony'
        identity_md.add_value(:otherId, clink.catalogRecordId, { name: 'previous_catkey' })
  end
end
```

Alternatively, you can use the `.add_other_id()` method instead of `.add_value()`.

justinlittman (changes requested):

app/services/cocina/to_fedora/identity.rb
edsu: @justinlittman I like the idea of wiping them clean with `.remove_other_Id()` and then adding them again with `.add_other_Id()`.

Would it be ok to [relax](https://github.com/sul-dlss/dor-services/blob/c2c7eeb84cfae6e84a7b636f2e08159838cfe699/lib/dor/datastreams/identity_metadata_ds.rb#L97) `.add_other_Id()` to allow multiple other ids of the same type to be added? From our previous discussions it seems like that should be allowed?

edsu: I thought about using `.add_other_Id()` but it prevents adding otherIds with the same type, and we want to allow for multiple otherId catkeys. I've pushed a change to use `.remove_other_Id()` to wipe the slate clean before adding them again following your first suggestion.

justinlittman (approved):

PR #3273, opened 2022-02-01
Add Cocina ActiveRecord models.

### LINKED ISSUE #3138: Implement new object models and object store
* Implement new AdminPolicy, Collection, and DRO models as shown in https://github.com/sul-dlss/dor-services-app/pull/2413
* Implement object store (called `CocinaObjectStore`) that provides an abstraction around these models as shown in https://github.com/sul-dlss/dor-services-app/blob/a181ada6db35151a26b3acc4ace207198ef988cb/app/services/cocina/object_store.rb. (That is, it allows finding, deleting, and saving cocina models without knowing about the underlying ActiveRecord models.)

Note that this does not include adding versioning.

justinlittman:
closes #3138

## Why was this change made?
~~Fedora~~

Unit

### DIFF
--- a/app/models/admin_policy.rb
+++ b/app/models/admin_policy.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+# Model for an AdminPolicy.
+class AdminPolicy < ApplicationRecord
+  # @return [Cocina::Models::AdminPolicy] Cocina Administrative Policy
+  def to_cocina
+    Cocina::Models::AdminPolicy.new({
+      cocinaVersion: cocina_version,
+      type: Cocina::Models::Vocab.admin_policy,
+      externalIdentifier: external_identifier,
+      label: label,
+      version: version,
+      administrative: administrative,
+      description: description
+    }.compact)
+  end
+
+  # @param [Cocina::Models::AdminPolicy] Cocina Administrative Policy
+  # @return [AdminPolicy] ActiveRecord Administative Policy
+  def self.from_cocina(admin_policy)
+    new(to_model_hash(admin_policy))
+  end
+
+  # @param [Cocina::Models::AdminPolicy] Cocina Administrative Policy
+  # @return [Hash] Hash representation of ActiveRecord Administrative Policy
+  def self.to_model_hash(cocina_admin_policy)
+    admin_policy_hash = cocina_admin_policy.to_h
+    admin_policy_hash[:external_identifier] = admin_policy_hash.delete(:externalIdentifier)
+    admin_policy_hash[:cocina_version] = admin_policy_hash.delete(:cocinaVersion)
+    admin_policy_hash.delete(:type)
+    admin_policy_hash[:description] ||= nil
+    admin_policy_hash
+  end
+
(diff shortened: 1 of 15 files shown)


db/migrate/20220131194912_create_collection.rb
```diff
@@ -0,0 +1,18 @@
+class CreateCollection < ActiveRecord::Migration[5.2]
+  def change
+    create_table :collections do |t|
+      t.string :external_identifier, null: false
+      t.string :cocina_version, null: false
+      # Type is reserved by Rails.
+      t.string :content_type, null: false
```
jcoyne, line 7: Why is content_type necessary for collections? Isn't this constant?  If not, can this be changed to collection_type as content_type is an attribute that has specific meaning in sdr.

spec/factories/admin_policies.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :admin_policy do
+    cocina_version { '0.0.1' }
```
jcoyne, line 5: Should this be Cocina::Models::VERSION?

spec/factories/collections.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :collection do
+    cocina_version { '0.0.1' }
```
jcoyne, line 5: Cocina::Models::VERSION?

spec/factories/dros.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :dro do
+    cocina_version { '0.0.1' }
```
jcoyne, line 5: Cocina::Models::VERSION?

jcoyne (changes requested):

db/migrate/20220131194912_create_collection.rb
justinlittman: Not a constant: https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L394-L399

I will rename to `collection_type`.

spec/factories/admin_policies.rb
justinlittman: Could be, but doesn't have to be.

My general strategy for test fixtures is to use `Cocina::Models::VERSION` when the version matters and `0.0.1` when it doesn't. However, if you feel strongly happy to change in these tests.

jcoyne (approved):

app/models/admin_policy.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+# Model for an AdminPolicy.
+class AdminPolicy < ApplicationRecord
+  # @return [Cocina::Models::AdminPolicy] Cocina Administrative Policy
+  def to_cocina
+    Cocina::Models::AdminPolicy.new({
+      cocinaVersion: cocina_version,
+      type: Cocina::Models::Vocab.adminPolicy,
```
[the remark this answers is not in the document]
justinlittman: Fixing.

PR #3287, opened 2022-02-02
Name subjects with subelements, role, displayForm, and subject subdivision

### LINKED ISSUE #2733: Implement subject/name with additional subelements and subject/name without name type (Cocina>MODS)
* https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/subject_name_spec.rb#L736

Note that the below is the first of 3 specs that are related to each other, and not entirely related to the above.

* https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/subject_name_spec.rb#L1142

See #2732

lwrubel:
## Why was this change made?
Resolves #2733. (Second half was already addressed by #3276.) Maps back to MODS when Cocina has a name subject with structuredValue > parallelValue > structuredValue.

Added spec and ran unit tests. I had to disable Metrics/CyclometricComplexity on one existing method. (It's disabled on other methods in this mapping already.) Wondering if I should be trying to refactor that, or create a ticket for it?

Ran `bin/validate-desc-cocina-roundtrip -s 100000 -i druids.testbed.txt` and no change. 

Before:
```Status (n=100000; not using Missing for success/different/error stats):
  Success:   99985 (99.985%)
  Different: 15 (0.015%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
```

After:
```Status (n=100000; not using Missing for success/different/error stats):
  Success:   99985 (99.985%)
  Different: 15 (0.015%)
  To Cocina error:     0 (0.0%)
  To Fedora error:     0 (0.0%)
```

### DIFF
--- a/app/services/cocina/to_fedora/descriptive/subject.rb
+++ b/app/services/cocina/to_fedora/descriptive/subject.rb
@@ -435,22 +435,12 @@ module Cocina
           parallel_subject_values = Array(value.parallelValue)
           display_values, parallel_subject_values = parallel_subject_values.partition { |par_value| par_value.type == 'display' }
 
-          if parallel_subject_values.size == 1
-            parallel_subject_value = parallel_subject_values.first
-            if parallel_subject_value.structuredValue.present?
-              write_structured_person(value, parallel_subject_value, type: value.type, display_values: display_values)
-            else
-              write_person(value, parallel_subject_value, display_values: display_values)
-            end
-          end
-          return unless parallel_subject_values.size > 1
-
-          parallel_subject_values.each do |parallel_value|
-            if parallel_value.structuredValue.present?
-              write_structured_person(value, parallel_value, type: value.type, display_values: display_values)
-            else
-              write_person(value, parallel_value, display_values: display_values)
-            end
+          # there will not be more than one parallelValue within a structuredValue
+          parallel_subject_value = parallel_subject_values.first
+          if parallel_subject_value.structuredValue.present?
+            write_structured_person(value, parallel_subject_value, type: value.type, display_values: display_values)
+          else
+            write_person(value, parallel_subject_value, display_values: display_values)
           end
         end
       end
--- a/app/services/cocina/to_fedora/descriptive/subject.rb
+++ b/app/services/cocina/to_fedora/descriptive/subject.rb
@@ -89,6 +89,7 @@ module Cocina
         end
 
         # rubocop:disable Metrics/PerceivedComplexity
+        # rubocop:disable Metrics/CyclomaticComplexity
         def write_structured_or_grouped(subject, subject_value, alt_rep_group: nil, type: nil, display_values: nil)
           type ||= subject_value.type || subject.type
           xml.subject(structured_attributes_for(subject_value, type, alt_rep_group: alt_rep_group)) do
@@ -106,6 +107,8 @@ module Cocina
                 if FromFedora::Descriptive::Contributor::ROLES.value?(value.type)
                   if value.structuredValue.present?
                     write_structured_person(subject, value, display_values: display_values)
+                  elsif value.parallelValue.present?
+                    write_parallel_structured_person(value)
                   else
                     write_person(subject, value, display_values: display_values)
                   end
@@ -118,6 +121,7 @@ module Cocina
           end
         end
         # rubocop:enable Metrics/PerceivedComplexity
+        # rubocop:enable Metrics/CyclomaticComplexity
 
         def write_title(subject_value)
           title = subject_value.to_h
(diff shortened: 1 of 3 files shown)


spec/services/cocina/mapping/descriptive/mods/subject_name_spec.rb
```diff
@@ -733,14 +733,14 @@
...
-    xit 'not implemented in cocina>MODS direction' do
+    it_behaves_like 'MODS cocina mapping' do
       let(:mods) do
         <<~XML
           <subject>
             <name type="personal">
               <role>
-                <roleTerm type="text" authority="marcrelator" authorityURI="http://id.loc.gov/vocabulary/relators" valueURI="http://id.loc.gov/vocabulary/relators/dpc">Depicted</roleTerm>
-                <roleTerm type="code" authority="marcrelator" authorityURI="http://id.loc.gov/vocabulary/relators" valueURI="http://id.loc.gov/vocabulary/relators/dpc">dpc</roleTerm>
+                <roleTerm type="text" authority="marcrelator" authorityURI="http://id.loc.gov/vocabulary/relators/" valueURI="http://id.loc.gov/vocabulary/relators/dpc">Depicted</roleTerm>
```
lwrubel, line 742: Adding trailing slash to allow this to pass, in conversation with @justinlittman. Normalizing a trailing slash on authorityURI is already handled in name_builder.rb

spec/services/cocina/to_fedora/descriptive/subject_name_spec.rb
```diff
@@ -275,4 +275,108 @@
       XML
     end
   end
+
```
lwrubel, line 278: @arcadiafalcone I've added some specs below for subject name metadata, specifically to cover parallelValue within a structuredValue. The intent is to test 1) a parallelValue with more than one non-display type element and 2) when there is no structuredValue within the parallelValue. 

Do these make sense? Are these scenarios we should be addressing in mapping? If you have better specs I could use, happy to swap these out.

arcadiafalcone: For both, the parallelValue would only be inside the structuredValue if it were a multipart subject - i.e. if the person object was followed by a topic. In the first one, because the name parallelValues are all for the same person object and language/script isn't involved, they should all be within the same MODS name element, which isn't possible when "Andrada, L. F." has type "surname." In a real record it would be type "alternative" and map to alternativeName in the MODS. But I don't think we have any cases of alternativeName being used with a name subject in our data.

lwrubel: Thanks! Can you think of a situation where (1) or (2) would exist and help me fix the spec to better reflect that? Or if parallelValues inside a structuredValue are never going have more than one non-display type element, I could remove that logic from the code.

arcadiafalcone: I think (1) can be dropped, but (2) is worth keeping. You can add a sibling element to the parallelValue with value "Homes and haunts" (my fave LCSH subject heading) and type "topic" to make the structuredValue work.

lwrubel: I like "Homes and haunts" :)

I'm trying to create a spec in line with what you described for scenario 2, but I don't think I have it right. Can you review? 

```
    let(:subjects) do
      [
        Cocina::Models::DescriptiveValue.new(
          {
            type: 'person',
            structuredValue: [
              {
                parallelValue: [
                  {
                    value: 'Holbein, Han, 1497-1543',
                    type: 'surname'
                  },
                  {
                    value: 'Holbein, Han, 1497-1543',
                    type: 'display'
                  }
                ]
              },
              {
                value: 'Homes and haunts',
                type: 'topic'
              }
            ]
          }
        )
      ]
    end
```

lwrubel: I think I corrected the spec in the latest commit, @arcadiafalcone, but would still appreciate your eyes on it.

lwrubel: I think I fixed this, @arcadiafalcone but would still appreciate your eyes on it.

arcadiafalcone: That looks right to me.

lwrubel: Great, thanks!

app/services/cocina/to_fedora/descriptive/subject.rb
```diff
@@ -426,6 +430,18 @@ def name_type_for(type)
         def edition(version)
           version.split.first.gsub(DEORDINAL_REGEX, '')
         end
+
+        def write_parallel_structured_person(value)
+          parallel_subject_values = Array(value.parallelValue)
+          display_values, parallel_subject_values = parallel_subject_values.partition { |par_value| par_value.type == 'display' }
+
+          parallel_subject_value = parallel_subject_values.first
```
justinlittman, line 438: Why only the first? Shouldn't all be mapped?

lwrubel: I had started trying to account for multiple parallelValues inside a structuredValue and when struggling to create a spec, reached out to Arcadia. See conversation thread on the spec in this PR, where she said that scenario could be dropped. Let me know if I should add a comment to that effect or refactor in some way.

justinlittman (approved):
Yeah, just add a comment then good to go.

lwrubel:
Thanks @justinlittman. Added a comment and will merge.

PR #3289, opened 2022-02-02
Rewrite UpdateMarcRecordService to use Cocina.

### LINKED ISSUE #3171: Rewrite UpdateMarcRecordService to use Cocina objects

aaron-collier:
## Why was this change made?

Closes #3171  be rewriting UpdateMarcRecordService to work on cocina objects.

Unit


spec/support/datastreams.rb
```diff
@@ -22,6 +22,33 @@ def build_identity_metadata_1
 </identityMetadata>'
 end
 
+def build_cocina_identity_metadata_basic
```
jcoyne, line 25: Please don't put cocina in the support file for "datastreams". If these are only used in one spec file, is there a reason not to keep them in the spec file?  That makes the code a lot easier to read.

jcoyne (changes requested):

jcoyne (approved):

PR #3304, opened 2022-02-04
Fix the openapi response code and improve testing

jcoyne:


app/controllers/metadata_controller.rb
```diff
@@ -16,10 +16,6 @@ def descriptive
     render xml: service
   end
 
-  def mods
```
jcoyne, line 19: This is just dead code now. This moved to the ModsController

justinlittman (approved):

PR #3311, opened 2022-02-04
Cocina-izes VersionService. Stores version data in db.

### LINKED ISSUE #3170: Rewrite VersionService and ObjectsController.accession to use Cocina objects

justinlittman:
closes #3170

## Why was this change made?
Version metadata is being moved from the version datastream to the DB.

Unit

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -1,17 +1,22 @@
 # frozen_string_literal: true
 
 class VersionsController < ApplicationController
-  before_action :load_item
+  before_action :load_cocina_object
 
   def index
-    return render json: {} unless @item.respond_to?(:versionMetadata)
+    # This can be removed after migration.
+    VersionMigrationService.find_and_migrate(@cocina_object.externalIdentifier)
+
+    object_versions = ObjectVersion.where(druid: @cocina_object.externalIdentifier)
+
+    return render json: {} if object_versions.empty?
 
     # add an entry with version id, tag and description for each version
-    versions = (1..@item.current_version.to_i).map do |version_number|
+    versions = object_versions.map do |object_version|
       {
-        versionId: version_number,
-        tag: @item.versionMetadata.tag_for_version(version_number.to_s),
-        message: @item.versionMetadata.description_for_version(version_number.to_s)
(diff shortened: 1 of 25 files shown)


app/controllers/versions_controller.rb
```diff
@@ -1,45 +1,50 @@
 # frozen_string_literal: true
 
 class VersionsController < ApplicationController
-  before_action :load_item
+  before_action :load_cocina_object
 
   def index
-    return render json: {} unless @item.respond_to?(:versionMetadata)
+    # This can be removed after migration.
```
jcoyne, line 7: Do we need to keep a list of what has and what hasn't migrated?

justinlittman: My thought was to create an Uber Migration job that would migrate versions, events, and persist to cocina and then invoke that for every object. So I don't think there is a need for a separate tracking mechanism for versions.

jcoyne:
Holding until when?

justinlittman:
Testing completed, so ready for review / merge.

spec/requests/authorization_spec.rb
```diff
@@ -3,10 +3,10 @@
 require 'rails_helper'
 
 RSpec.describe 'Authorization' do
-  let(:object) { instance_double(Dor::Item, current_version: '5') }
+  let(:cocina_object) { instance_double(Cocina::Models::DRO, version: '5') }
```
jcoyne, line 6: Version should be an integer. https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L197-L198

jcoyne (changes requested):

jcoyne (changes requested):
Shouldn't this line have been updated too? https://github.com/sul-dlss/dor-services-app/blob/main/app/services/embargo_release_service.rb#L43

justinlittman:
@jcoyne Yes, changes are needed to EmbargoReleaseService. In fact, I made them. And then I forgot to hit save in the editor. That showed up here as well: https://app.honeybadger.io/projects/50568/faults/83969890

jcoyne (approved):

PR #3315, opened 2022-02-05
QueriesController: rewrite to use Cocina objects

### LINKED ISSUE #3174: Rewrite QueriesController to use Cocina objects

jmartin-sul:
## Why was this change made?

close #3174 

- [x] unit tests
- [x] test on stage (using collection edit UI in argo, which hits this endpoint)

n/a

### DIFF
--- a/app/controllers/queries_controller.rb
+++ b/app/controllers/queries_controller.rb
@@ -2,11 +2,13 @@
 
 # Responds to queries about objects
 class QueriesController < ApplicationController
-  before_action :load_item, only: [:collections]
+  before_action :load_cocina_object, only: [:collections]
 
   # Returns a list of collections this object is in.
   def collections
-    # If we move to Valkyrie this can be find_inverse_references_by
-    @collections = @item.collections.map { |collection| Cocina::Mapper.build(collection) }
+    # isMemberOf may be nil, in which case we want to return an empty array
+    @collections = Array(@cocina_object.structural.isMemberOf).map do |collection_id|
+      CocinaObjectStore.find(collection_id)
+    end
   end
 end
--- a/spec/factories/collections.rb
+++ b/spec/factories/collections.rb
@@ -12,6 +12,10 @@ FactoryBot.define do
     end
   end
 
+  factory :collection_unique_druid, parent: :collection do
+    external_identifier { generate(:unique_druid) }
+  end
+
   trait :with_collection_identification do
     identification do
       { sourceId: 'googlebooks:999999' }
--- a/spec/factories/collections.rb
+++ b/spec/factories/collections.rb
@@ -3,17 +3,13 @@
 FactoryBot.define do
   factory :collection do
     cocina_version { '0.0.1' }
-    external_identifier { 'druid:hp308wm0436' }
+    external_identifier { generate(:unique_druid) }
     collection_type { Cocina::Models::Vocab.collection }
     label { 'Test Collection' }
     version { 1 }
     access do
       { access: 'world' }
     end
-  end
-
-  factory :collection_unique_druid, parent: :collection do
-    external_identifier { generate(:unique_druid) }
   end
 
   trait :with_collection_identification do
--- a/spec/factories/dros.rb
+++ b/spec/factories/dros.rb
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :dro do
     cocina_version { '0.0.1' }
-    external_identifier { 'druid:xz456jk0987' }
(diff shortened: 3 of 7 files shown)


app/controllers/queries_controller.rb
```diff
@@ -2,11 +2,13 @@
...
-  before_action :load_item, only: [:collections]
+  before_action :load_cocina_object, only: [:collections]
 
   # Returns a list of collections this object is in.
   def collections
-    # If we move to Valkyrie this can be find_inverse_references_by
-    @collections = @item.collections.map { |collection| Cocina::Mapper.build(collection) }
+    # isMemberOf may be nil, in which case we want to return an empty array
+    @collections = Array(@cocina_object.structural.isMemberOf).map do |collection_id|
+      CocinaObjectStore.find(collection_id)
```
jmartin-sul, line 11: this lookup mirrors the automatic lookup that `@item.collections` did to fetch the collection info from fedora (lazily, when `#collections` was invoked).

spec/factories/collections.rb
```diff
@@ -12,6 +12,10 @@
     end
   end
 
+  factory :collection_unique_druid, parent: :collection do
+    external_identifier { generate(:unique_druid) }
+  end
```
jmartin-sul, line 17: playing with a follow on branch/commit that makes this the default behavior, so we can get rid of this extra factory, since that change only breaks four tests.

spec/factories/dros.rb
```diff
@@ -15,6 +15,10 @@
     end
   end
 
+  factory :dro_unique_druid, parent: :dro do
+    external_identifier { generate(:unique_druid) }
+  end
```
jmartin-sul, line 20: same comment here as the other unique ID comment (changing both breaks four tests total)

spec/factories/dros.rb
```diff
@@ -60,7 +68,9 @@
             ]
           }
         }
-      ] }
+      ] }.tap do |h|
+        h.merge!({ isMemberOf: isMemberOf }) if isMemberOf.present?
+      end
```
jmartin-sul, line 73: i'm not sure if there's a nicer looking or more idiomatic way to do this sort of optional fixture field.  couldn't quite figure out how to do it w/ nested traits (if that's possible), and this seemed better than a copypasta trait for a one field difference.

spec/factories/druids.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  sequence :unique_druid do |n|
+    # this should give us 10000000 (1000 * 10000) unique druids:
+    # * start the counter from 0.
+    # * div the counter so that the 3 digit cluster increments every time the 4 digit cluster rolls over.
+    idx = n - 1
+    format('druid:zx%03dwv%04d', idx / 10000, idx) # rubocop:disable Style/FormatStringToken annotations make this string less readable
+  end
```
jmartin-sul, line 10: seemed like it might be useful to easily get unique/differentiable druids across the test suite

spec/requests/collections_for_object_spec.rb
```diff
@@ -4,48 +4,53 @@
...
+    allow(CocinaObjectStore).to receive(:find) do |druid|
+      found_obj = ([dro] + collections).find { |cocina_obj| cocina_obj.externalIdentifier == druid }
+      raise CocinaObjectStore::CocinaObjectNotFoundError if found_obj.blank?
 
-  let(:object) { instance_double(Dor::Item, collections: [collection]) }
-  let(:collection_id) { 'druid:bc123df4567' }
-  let(:collection) do
-    Dor::Collection.new(pid: collection_id, label: 'collection #1').tap do |coll|
-      coll.descMetadata.title_info.main_title = 'Hello'
+      found_obj
```
jmartin-sul, line 11: i'm not sure if this is more clever than test code should be.  happy to do something more verbose if that seems less brittle.

jmartin-sul:
for your consideration: i'm not sure if this is an improvement, but this follow-on makes the `dro` and `collection` fixtures use unique druids by default:  https://github.com/sul-dlss/dor-services-app/pull/3316

spec/factories/collections.rb
jmartin-sul: https://github.com/sul-dlss/dor-services-app/pull/3316

spec/factories/druids.rb
justinlittman: Yeah!

justinlittman (approved):

PR #3355, opened 2022-02-08
Translate dro and collection uris to terms for the 856 output

### LINKED ISSUE #3353: 856 fields from SW release should use the short form of the object type, not the Cocina URL
When an item with a catkey is released from SDR, an 856 field is added to the MARC record for that catkey. The 856 includes the item's object type. This should be expressed in the short form, like "item" as seen here (the second to last field):

`13972869     pk343gn7701     .856. 40|zAvailable to Stanford-affiliated users.|uhttps://purl.stanford.edu/pk343gn7701|xSDR-PURL|xitem|xrights:group=stanfor`

After a recent change, the object type is now being expressed as a Cocina URL:

`13972869      pk343gn7701     .856. 40|zAvailable to Stanford-affiliated users.|uhttps://purl.stanford.edu/pk343gn7701|xSDR-PURL|xhttp://cocina.sul.stanford.edu/models/object.jsonld|xrights:group=stanford`

This is probably not what SW is expecting. SW uses this field to determine whether to display a collection or an item.

aaron-collier:
## Why was this change made?

Fixes #3353 

Updated unit test.

### DIFF
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -12,7 +12,7 @@ module Dor
     def initialize(cocina_object, thumbnail_service:)
       @cocina_object = cocina_object
       @druid_id = Dor::PidUtils.remove_druid_prefix(cocina_object.externalIdentifier)
-      @access = cocina_object.access
+      @access = cocina_object.access if cocina_object.respond_to?(:access)
       @thumbnail_service = thumbnail_service
     end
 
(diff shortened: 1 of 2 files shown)


app/models/dor/update_marc_record_service.rb
```diff
@@ -97,6 +97,13 @@ def get_identifier(ckey)
       "#{ckey}\t#{@druid_id}\t"
     end
 
+    def get_object_type_from_uri
+      return 'item' if @cocina_object.dro?
+      return 'collection' if @cocina_object.collection?
+
+      raise "Error translating #{@cocina_object.type} to short form"
```
aaron-collier, line 104: Adding test coverage for this...

aaron-collier: Actually - I don't think we should ever get here. Going to switch to returning `nil`

justinlittman (approved):

PR #3365, opened 2022-02-08
add objectLabel node with value from fedora object if not in datastream

### LINKED ISSUE #3320: Add objectLabel when missing (identity metadata)
```
Original XML for identityMetadata:
<?xml version="1.0"?>
<identityMetadata>
  <objectId>druid:ck375kx3685</objectId>
  <objectType>item</objectType>
  <objectAdminClass>EEMs</objectAdminClass>
  <agreementId>druid:fn200hb6598</agreementId>
  <tag>EEM : 1.0</tag>
  <sourceId source="eems">eems_src_03118</sourceId>
</identityMetadata>


Normalized original XML for identityMetadata:
<?xml version="1.0" encoding="UTF-8"?>
<identityMetadata>
  <objectId>druid:ck375kx3685</objectId>
  <objectType>item</objectType>
  <sourceId source="eems">eems_src_03118</sourceId>
</identityMetadata>


Roundtripped XML for identityMetadata:
<?xml version="1.0"?>
<identityMetadata>
  <sourceId source="eems">eems_src_03118</sourceId>
  <objectId>druid:ck375kx3685</objectId>
  <objectType>item</objectType>
  <objectLabel>EEMs: A 2010 leadership imperative: the future built on broadband</objectLabel>
</identityMetadata>
```

If objectLabel is missing, add it as part of normalization.

peetucket:
## Why was this change made?

Fixes #3320 - if missing when normalizing, add objectLabel to identityMetadata using value pulled from fedora object

- Added new unit tests
- roundtrip validation on server

### DIFF
--- a/app/services/cocina/normalizers/identity_normalizer.rb
+++ b/app/services/cocina/normalizers/identity_normalizer.rb
@@ -8,12 +8,14 @@ module Cocina
       include Cocina::Normalizers::Base
 
       # @param [Nokogiri::Document] identity_ng_xml identity metadata XML to be normalized
+      # @param [String] label the object label to add when normalizing
       # @return [Nokogiri::Document] normalized identity metadata xml
-      def self.normalize(identity_ng_xml:)
-        new(identity_ng_xml: identity_ng_xml).normalize
+      def self.normalize(identity_ng_xml:, label:)
+        new(identity_ng_xml: identity_ng_xml, label: label).normalize
       end
 
-      def initialize(identity_ng_xml:)
+      def initialize(identity_ng_xml:, label:)
+        @label = label
         @ng_xml = identity_ng_xml.dup
         @ng_xml.encoding = 'UTF-8' if @ng_xml.respond_to?(:encoding=) # sometimes it's a String (?)
       end
@@ -36,6 +38,7 @@ module Cocina
         normalize_object_creator
         normalize_out_labels
         normalize_apo_source_id
+        normalize_label
 
         regenerate_ng_xml(ng_xml.to_xml)
       end
@@ -156,6 +159,15 @@ module Cocina
         object_creator_node = Nokogiri::XML::Node.new('objectCreator', ng_xml)
         object_creator_node.content = 'DOR'
         ng_xml.root << object_creator_node
+      end
+
+      # add fedora object label if no object label is present
+      def normalize_label
+        return if ng_xml.root.xpath('//objectLabel').present?
+
+        object_label_node = Nokogiri::XML::Node.new('objectLabel', ng_xml)
+        object_label_node.content = @label
+        ng_xml.root << object_label_node
       end
     end
   end
--- a/bin/validate-cocina-roundtrip
+++ b/bin/validate-cocina-roundtrip
@@ -97,7 +97,7 @@ def norm_orig_datastream_ng_xml_for(dsid, orig_datastream_ng_xml, druid, label, 
   when 'contentMetadata'
     Cocina::Normalizers::ContentMetadataNormalizer.normalize(druid: druid, content_ng_xml: orig_datastream_ng_xml)
   when 'identityMetadata'
-    Cocina::Normalizers::IdentityNormalizer.normalize(identity_ng_xml: orig_datastream_ng_xml, label: MetadataService.label_from_mods(fedora_obj.descMetadata.ng_xml))
+    Cocina::Normalizers::IdentityNormalizer.normalize(identity_ng_xml: orig_datastream_ng_xml, label: label)
   when 'embargoMetadata'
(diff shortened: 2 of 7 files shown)


peetucket:
Running on this PR:

```
bin/validate-cocina-roundtrip -s 100000 -i druids.testbed.txt -f

```

Running on main:

```
bin/validate-cocina-roundtrip -s 100000 -i druids.testbed.txt -f

```

bin/validate-cocina-roundtrip
```diff
@@ -97,7 +97,7 @@ def norm_orig_datastream_ng_xml_for(dsid, orig_datastream_ng_xml, druid, label,
   when 'contentMetadata'
     Cocina::Normalizers::ContentMetadataNormalizer.normalize(druid: druid, content_ng_xml: orig_datastream_ng_xml)
   when 'identityMetadata'
-    Cocina::Normalizers::IdentityNormalizer.normalize(identity_ng_xml: orig_datastream_ng_xml)
+    Cocina::Normalizers::IdentityNormalizer.normalize(identity_ng_xml: orig_datastream_ng_xml, label: MetadataService.label_from_mods(fedora_obj.descMetadata.ng_xml))
   when 'embargoMetadata'
```
peetucket, line 101: fetch the label from the MetadataService and pass to normalizer so we can use if needed

app/services/cocina/normalizers/identity_normalizer.rb
```diff
@@ -8,12 +8,14 @@ class IdentityNormalizer
       include Cocina::Normalizers::Base
 
       # @param [Nokogiri::Document] identity_ng_xml identity metadata XML to be normalized
+      # @param [String] label the object label to add when normalizing
       # @return [Nokogiri::Document] normalized identity metadata xml
-      def self.normalize(identity_ng_xml:)
-        new(identity_ng_xml: identity_ng_xml).normalize
+      def self.normalize(identity_ng_xml:, label:)
+        new(identity_ng_xml: identity_ng_xml, label: label).normalize
       end
```
peetucket, line 15: we need this label passed in since it comes from the fedora object, which we do not have access to inside the normalizer

ndushay:
I think the problem has to do with https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/to_fedora/identity.rb#L40-L44 -- this method is called for `object_updater` and in `object_creator` and would overwrite the existing label, right?

image

bin/validate-cocina-roundtrip
```diff
@@ -97,7 +97,7 @@ def norm_orig_datastream_ng_xml_for(dsid, orig_datastream_ng_xml, druid, label,
   when 'contentMetadata'
     Cocina::Normalizers::ContentMetadataNormalizer.normalize(druid: druid, content_ng_xml: orig_datastream_ng_xml)
   when 'identityMetadata'
-    Cocina::Normalizers::IdentityNormalizer.normalize(identity_ng_xml: orig_datastream_ng_xml)
+    Cocina::Normalizers::IdentityNormalizer.normalize(identity_ng_xml: orig_datastream_ng_xml, label: MetadataService.label_from_mods(fedora_obj.descMetadata.ng_xml))
```
justinlittman, line 100: Why not just use the label that is passed into norm_orig_datastream_ng_xml_for method?

justinlittman:
@ndushay This ticket is about normalizing when the original identity metadata is missing an objectLabel.

However, unless I'm misunderstanding the existing code is correct in overwriting the existing label.

bin/validate-cocina-roundtrip
peetucket: ahh, missed that it was available here, will use that instead

peetucket:
roundtrip tests updated - now basically identical (1 better in the PR)

peetucket:
Example from ticket (druid:ck375kx3685) succeeds on PR but fails on main.

ndushay:
I think this spec should be changed:

https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/identification/dro_identity_spec.rb#L745

and maybe some others???

ndushay (changes requested):
I found 3 specs where we're mocking the normalized identityMetadata and it would be awesome if all such specs were corrected to match this change.  :-P

peetucket:
> I found 3 specs where we're mocking the normalized identityMetadata and it would be awesome if all such specs were corrected to match this change. :-P

What changes would you suggest?  The new normalization code does not do anything about blank labels (though perhaps it should ... but was not in the original specification).  These particular specs you mention appear to simply have incoming blank labels in the mocked data, which are then preserved through the roundtrip, as expected.  The new specs I added in the PR confirm the new changes in the code.

ndushay (approved):
oh duh.  I don't know why I keep confusing myself over all this stuff.  (Maybe it's ... confusing?)

PR #3368, opened 2022-02-09
Unpack names with displayForm into parallelValue

### LINKED ISSUE #3215: Implement updated name_spec for displayForm
See #3214

edsu:
## Why was this change made?

To ensure that names with displayForm elements are unpacked into parallelValues.

closes #3215


justinlittman:
@edsu What is the status of this?

PR #3374, opened 2022-02-09
Enhances delete to take into account ActiveRecord objects.

### LINKED ISSUE #3370: ActiveRecord delete
When deleting an item, ActiveRecord objects should also be deleted. This includes DROs, AdminPolicies, Collections, Versions, Events, and AdministrativeTags.

justinlittman:
closes #3370

## Why was this change made?
* Prep for switching to persist to DB.
* Actually clear out all items stored in DB when deleting.

Unit


app/services/cocina_object_store.rb
```diff
@@ -65,20 +77,33 @@ def fedora_find(druid)
     raise CocinaObjectNotFoundError
   end
 
+  def fedora_destroy(druid)
+    ActiveFedora::Base.connection_for_pid(0).api.purge_object(pid: druid)
```
jcoyne, line 81: You probably want to do `Dor.find(druid).destroy` as that will execute any hooks and will remove it from the Solr index too.

jcoyne (changes requested):

app/services/cocina_object_store.rb
justinlittman: This was from the existing DeleteService: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/delete_service.rb#L53

And the DeleteService continues to delete from Solr: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/delete_service.rb#L54-L55

But I defer to your expertise. Do you want me to change?

jcoyne: Yeah, Willy originally wrote that code 8 years ago, and I wouldn't suggest copying it: https://github.com/sul-dlss/dor-services/commit/057a65dd8827e0e88ccb8ded2288f294ad10bb6e

app/services/administrative_tags.rb
```diff
@@ -157,6 +165,10 @@ def destroy(tag:)
     end
   end
 
+  def destroy_all
+    send(:for).each { |tag| destroy(tag: tag) }
```
jcoyne, line 169: why `send(:for)` and not just call the method?

app/services/cocina_object_store.rb
```diff
@@ -122,4 +145,8 @@ def cocina_to_ar_save(cocina_object)
     model_clazz.upsert_cocina(cocina_object)
     cocina_object
   end
+
+  def ar_destroy(druid)
+    ar_find(druid).destroy
```
jcoyne, line 150: We probably need a RabbitMQ message here so that the Solr document is removed too.

jcoyne (changes requested):

jcoyne (approved):

PR #3381, opened 2022-02-09
apo identityMetadatata.xml objectLabel used for cocina label (when objectLabel exists)

ndushay:
## Why was this change made?

part of #3336 (the same changes for Collection records will be in a separate PR)

APOs should use the existing `objectLabel` in `identityMetadata` for the value of cocina `label` if the `objectLabel` exists.

## How was this change tested?

unit tests and this data:

### This branch (NEW)

Note:  most of the failing APOs still have other roundtrip problems.

```sh
$ bin/validate-cocina-roundtrip -i tmp/druid-lists/druids.apos.txt -f -n

Status (n=1270; not using Missing for success/different/error stats):
  Success:   1055 (83.071%)
  Different: 205 (16.142%)
  Mapping error:     10 (0.787%)
  Create error:     0 (0.0%)
  Normalization error:     0 (0.0%)
  Missing from cache:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     0 (0.0%)
  Bad cache:     0 (0.0%)
````

### main branch (BEFORE)

```sh
Status (n=1270; not using Missing for success/different/error stats):
  Success:   1043 (82.126%)
  Different: 217 (17.087%)
  Mapping error:     10 (0.787%)
  Create error:     0 (0.0%)
  Normalization error:     0 (0.0%)
  Missing from cache:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     0 (0.0%)
  Bad cache:     0 (0.0%)
```

PR #3384, opened 2022-02-10
dro_identity_spec: add spec for missing objectLabel from identityMetadata.xml

ndushay:
## Why was this change made?

Inspired by PRs #3381 and #3383, this adds the analogous spec for DRO objects.

Tagging @peetucket not because he has to review, but because THIS is what I couldn't articulate nor figure out how to write for a while.

it's a test, and it passes.


jcoyne:
Should this be rebased? Isn't the first commit already on main?

justinlittman (approved):

PR #3350, opened 2022-02-08
Allow ReleaseTags to accept cocina objects.

justinlittman:
closes #3348

## Why was this change made?
UpdateMarcRecordService uses ReleaseTags, but has already been cocina-ized.

## How was this change tested?
Unit

PR #3386, opened 2022-02-10
Generate release tags from cocina instead of fedora

### LINKED ISSUE #3335: Rewrite ReleaseTags to use cocina object
https://github.com/sul-dlss/dor-services-app/blob/main/app/services/release_tags.rb

peetucket:
## Why was this change made?

Fixes #3335 

- rewrite `ReleaseTags` to use cocina instead of fedora
- adjust all tests that used to mock fedora to instead mock cocina
- get rid of unused `#create` method (that was only used in tests)
- some refactoring of the tests to avoid duplication of pids
- revert `legacy_for` methods added here since they should no longer be needed: https://github.com/sul-dlss/dor-services-app/pull/3350

Note: I believe release tags are created in argo here: https://github.com/sul-dlss/argo/blob/main/app/jobs/release_object_job.rb#L55-L83   

So we should not need to the `#create` method I removed.

Todo:
~~- fix up `./spec/requests/show_object_spec.rb` to remove the dependency on the `ReleaseTags#create` method and instead mock the data in some way~~
~~- fix up `./spec/services/release_tags/identity_metadata_spec.rb` to remove all of the mocked Fedora calls/data and replace with mocked cocina objects~~

Updated the relevant existing tests

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -17,7 +17,7 @@ class MetadataController < ApplicationController
   end
 
   def public_xml
-    release_tags = ReleaseTags.legacy_for(item: @item)
+    release_tags = ReleaseTags.for(dro_object: @cocina_object)
     service = Publish::PublicXmlService.new(@item, released_for: release_tags, thumbnail_service: ThumbnailService.new(@cocina_object))
     render xml: service
   end
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -17,7 +17,7 @@ class MetadataController < ApplicationController
   end
 
   def public_xml
-    release_tags = ReleaseTags.for(dro_object: @cocina_object)
+    release_tags = ReleaseTags.for(cocina_object: @cocina_object)
     service = Publish::PublicXmlService.new(@item, released_for: release_tags, thumbnail_service: ThumbnailService.new(@cocina_object))
     render xml: service
   end
--- a/app/services/release_tags.rb
+++ b/app/services/release_tags.rb
@@ -4,27 +4,9 @@
 class ReleaseTags
   # Retrieve the release tags for an item and all the collections that it is a part of
   #
-  # @param dro_object [Cocina::DRO] the DRO to list release tags for
+  # @param cocina_object [Cocina::Models::DRO, Cocina::Models::Collection] the object to list release tags for
   # @return [Hash] (see Dor::ReleaseTags::IdentityMetadata.released_for)
-  def self.for(dro_object:)
-    IdentityMetadata.for(dro_object).released_for({})
+  def self.for(cocina_object:)
+    IdentityMetadata.for(cocina_object).released_for({})
   end
-
-  # Determine if the supplied tag is a valid release node that meets all requirements
-  #
-  # @param tag [Boolean] True or false for the release node
-  # @param attrs [hash] A hash of attributes for the tag, must contain :when, a ISO 8601 timestamp and :who to identify who or what added the tag, :to,
-  # @raise [ArgumentError]  Raises an error of the first fault in the release tag
-  # @return [Boolean] Returns true if no errors found
-  def self.validate_release_attributes(tag, attrs = {})
-    raise ArgumentError, ':when is not iso8601' if attrs[:when].match('\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z').nil?
-
-    [:who, :to, :what].each do |check_attr|
-      next unless attrs[check_attr]
-      raise ArgumentError, "#{check_attr} not supplied as a String" unless attrs[check_attr].is_a? String
-    end
-    raise ArgumentError, ':what must be self or collection' unless %w[self collection].include? attrs[:what]
-    raise ArgumentError, 'the value set for this tag is not a boolean' unless [true, false].include? tag
-  end
-  private_class_method :validate_release_attributes
 end
--- a/app/services/release_tags.rb
+++ b/app/services/release_tags.rb
@@ -4,38 +4,10 @@
 class ReleaseTags
   # Retrieve the release tags for an item and all the collections that it is a part of
   #
-  # @param item [Cocina::DRO] the DRO to list release tags for
+  # @param dro_object [Cocina::DRO] the DRO to list release tags for
   # @return [Hash] (see Dor::ReleaseTags::IdentityMetadata.released_for)
   def self.for(dro_object:)
-    item = Dor.find(dro_object.externalIdentifier)
-    IdentityMetadata.for(item).released_for({})
-  end
-
-  # Retrieve the release tags for an item and all the collections that it is a part of
-  #
-  # @param item [Dor::Item] the item to list release tags for
-  # @return [Hash] (see Dor::ReleaseTags::IdentityMetadata.released_for)
-  def self.legacy_for(item:)
-    IdentityMetadata.for(item).released_for({})
-  end
-
-  # Add a release node for the item
-  # Will use the current time if timestamp not supplied. You can supply a timestap for correcting history, etc if desired
-  # Timestamp will be calculated by the function
-  #
-  # @param work [Dor::Item]  the work to create the release tag for
-  # @param [Hash] params all the other stuff
-  # @option params [Boolean] release: True or false for the release node
-  # @return [Nokogiri::XML::Element] the tag added if successful
-  # @raise [ArgumentError] Raised if attributes are improperly supplied
-  #
-  # @example
-  #  ReleaseTags.create(item, release: true, what: 'self', to: 'Searchworks', who: 'petucket')
-  def self.create(work, attrs)
-    release = attrs.delete(:release)
-    attrs[:when] ||= Time.now.utc.iso8601 # add the timestamp
-    validate_release_attributes(release, attrs)
-    work.identityMetadata.add_value(:release, release.to_s, attrs)
+    IdentityMetadata.for(dro_object).released_for({})
   end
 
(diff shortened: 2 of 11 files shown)


app/services/release_tags.rb
```diff
@@ -4,38 +4,10 @@
...
-  #
-  # @param work [Dor::Item]  the work to create the release tag for
-  # @param [Hash] params all the other stuff
-  # @option params [Boolean] release: True or false for the release node
-  # @return [Nokogiri::XML::Element] the tag added if successful
-  # @raise [ArgumentError] Raised if attributes are improperly supplied
-  #
-  # @example
-  #  ReleaseTags.create(item, release: true, what: 'self', to: 'Searchworks', who: 'petucket')
-  def self.create(work, attrs)
```
peetucket, line 34: this method was only used in tests, getting rid of it

app/services/release_tags/identity_metadata.rb
```diff
@@ -45,36 +45,47 @@ def released_for(released_hash)
...
+      tags = {}
+      dro_object.administrative.releaseTags.each do |tag|
+        tags[tag.to] ||= []
+        tags[tag.to] << { 'what' => tag.what, 'who' => tag.who, 'when' => tag.date.utc, 'release' => tag.release }
+      end
+      tags
+    end
 
-    delegate :release_tags, to: :item
+    private
```
peetucket, line 78: instead of sending `release_tags` to the fedora object, we have a new method which pulls them out of cocina and puts them into the identical tag structure so everything else can work the same

spec/services/release_tags_spec.rb
```diff
@@ -3,47 +3,33 @@
...
-  end
-
-  describe '.create' do
-    it 'creates a plain directory in the workspace when passed no source directory' do
-      expect(work.identityMetadata.to_xml).to be_equivalent_to <<-XML
-        <identityMetadata>
-          <release what="self" when="2018-11-30T22:41:35Z" to="SearchWorks">true</release>
-        </identityMetadata>
-      XML
     end
```
peetucket, line 33: this method is now gone

app/services/release_tags/identity_metadata.rb
```diff
@@ -45,36 +45,47 @@ def released_for(released_hash)
...
 
-        release_service = self.class.for(collection)
-        return_tags = combine_two_release_tag_hashes(return_tags, release_service.release_tags_for_item_and_all_governing_sets)
+      return return_tags unless dro_object.dro? # no need to continue if this is a collection, since they don't nest anymore
+
+      # now go through any collections it is a member of and add them
+      dro_object.structural.isMemberOf.each do |collection_druid|
+        release_service = self.class.for(CocinaObjectStore.find(collection_druid))
+        return_tags = combine_two_release_tag_hashes(return_tags, release_service.release_tags)
       end
```
peetucket, line 56: this used to do recursion to walk a nested tree of items in collections inside of other collections ... but I don't believe we support nested collections anymore, correct?  

so there is no need to do recursion and we can instead simply loop once over all of the collections that an item belongs to (which i assume most of the time is only one)

justinlittman: @andrewjbtw ^^

app/services/release_tags.rb
```diff
@@ -4,38 +4,10 @@
...
-    release = attrs.delete(:release)
-    attrs[:when] ||= Time.now.utc.iso8601 # add the timestamp
-    validate_release_attributes(release, attrs)
-    work.identityMetadata.add_value(:release, release.to_s, attrs)
+    IdentityMetadata.for(dro_object).released_for({})
   end
```
peetucket, line 12: the diff is confusing...this new line is not part of the deleted method...it is part of `self.for`

app/controllers/metadata_controller.rb
```diff
@@ -17,7 +17,7 @@ def descriptive
   end
 
   def public_xml
-    release_tags = ReleaseTags.legacy_for(item: @item)
+    release_tags = ReleaseTags.for(dro_object: @cocina_object)
```
mjgiarlo, line 20: Is `@cocina_object` here a DRO or a collection or possibly both? I have a general question that cuts across the changes made in this PR (and, actually, related code that wasn't touched in this PR but I don't want to bite off more than either of us wants to chew): can/should we look for opportunities to conform more closely to our Cocina naming conventions: https://github.com/sul-dlss/cocina-models/#usage-conventions (which we have admittedly been collectively pretty lax about).

I'm also wondering the same about the `dro_object` kwarg.

app/services/release_tags/identity_metadata.rb
```diff
@@ -3,14 +3,14 @@
 class ReleaseTags
   class IdentityMetadata
     # Determine projects in which an item is released
-    # @param [Dor::Item] item to get the release tags for
+    # @param dro_object [Cocina::DRO] the DRO to list release tags for
```
mjgiarlo, line 6: Rename to `cocina_item` if always a DRO?

spec/services/publish/metadata_transfer_service_spec.rb
```diff
@@ -24,18 +25,45 @@
...
+                                                {
+                                                  who: 'dhartwig',
+                                                  what: 'collection',
+                                                  date: '2020-01-18T17:03:35.000+00:00',
+                                                  to: 'Some_special_place',
+                                                  release: true
+                                                }
+                                              ] })
+  end
+  let(:cocina_object_collection) do
```
mjgiarlo, line 58: I think this should be `cocina_collection`

spec/services/release_tags/identity_metadata_spec.rb
```diff
@@ -101,41 +136,59 @@
     subject(:release_tags) { releases.release_tags }
 
     context 'for an item that does not have any release tags' do
-      let(:item) { instantiate_fixture('druid:qv648vd4392', Dor::Item) }
+      let(:dro_object) do
```
mjgiarlo, line 139: `cocina_item` ?

justinlittman (approved):
Approve pending consideration of Mike's comments.

app/services/release_tags.rb
```diff
@@ -4,55 +4,9 @@
...
-    raise ArgumentError, ':when is not iso8601' if attrs[:when].match('\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z').nil?
-
-    [:who, :to, :what].each do |check_attr|
-      next unless attrs[check_attr]
-      raise ArgumentError, "#{check_attr} not supplied as a String" unless attrs[check_attr].is_a? String
-    end
-    raise ArgumentError, ':what must be self or collection' unless %w[self collection].include? attrs[:what]
-    raise ArgumentError, 'the value set for this tag is not a boolean' unless [true, false].include? tag
-  end
-  private_class_method :validate_release_attributes
```
peetucket, line 57: Discovered this method is not used anymore either, getting rid of it.

spec/requests/show_object_spec.rb
```diff
@@ -69,11 +69,13 @@
...
-                                   to: 'Searchworks',
-                                   who: 'petucket',
-                                   when: '2014-08-30T01:06:28Z')
+        # add a release tag to our fedora object for this test
+        object.identityMetadata.add_value(:release, true, {
+                                            what: 'self',
+                                            to: 'Searchworks',
+                                            who: 'petucket',
+                                            when: '2014-08-30T01:06:28Z'
+                                          })
```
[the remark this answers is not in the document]
peetucket: This is because this particular test appears to be exercising some code that is not yet fully ported over from Fedora.  So this test still appears to fetch into the fedora object and produce cocina from it.  The fedora fixture doesn't have release tags in it, but our expected cocina output does.  Previously there was a method that did exactly what was shown above (the `#create` method I ditched in this PR since it was only used in tests like this).  So opted to do it inline for now, eventually this will go away in the test here too.

app/services/release_tags.rb
```diff
@@ -4,55 +4,9 @@
 class ReleaseTags
   # Retrieve the release tags for an item and all the collections that it is a part of
   #
-  # @param item [Cocina::DRO] the DRO to list release tags for
+  # @param cocina_object [Cocina::Models::DRO OR Cocina::Models::Collection] the object to list release tags for
```
mjgiarlo, line 7: I'm not sure it makes a ton of difference, but I think YARD wants to see a comma-separated list here rather than an ` OR `-separated one 🤷🏻‍♂️

peetucket: switched to commas

app/services/release_tags/identity_metadata.rb
andrewjbtw: SDR supported nested collections? Unless there are any already in existence, I think we should act like we don't support them. The only path I could see to creating one is editing the RELS-EXT.

peetucket: Excellent.  Well I just removed support for nested collections in those code here!

The Revs Archive Collection is nested.  There is a master Revs Collection: https://argo.stanford.edu/view/druid:nt028fd5773 and then each of the sub collections are part of it, e.g. https://argo.stanford.edu/view/druid:tf080wn0696  You can see that in the RELS-EXT

But then each item of the that sub collection is a member of both collections, e.g. https://argo.stanford.edu/view/druid:bc313gk7619

So the nesting was ultimately not important.  I was told to set it up this way back in the early days (likely by Lynn) but it didn't have any importance for our discovery environment (we had a custom indexer).  And now it's all dark, so it's not important to preserve that nesting anyway.

andrewjbtw: Ah, Revs. I saw those items were in two collections but didn't know they were nested.

Issue #3329, opened 2022-02-07
Rewrite content metadata in PublicXmlService to map from a cocina item


justinlittman:
Currently, content metadata that is including in the public XML is derived from a Fedora object. It should be re-written to derive it from a Cocina object. The public XML should remain unchanged.

You should be able to re-use the [ContentMetadataGenerator](https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/to_fedora/content_metadata_generator.rb) for this.

[See https://github.com/sul-dlss/dor-services-app/blob/main/app/services/publish/public_xml_service.rb#L24](https://github.com/sul-dlss/dor-services-app/blob/main/app/services/publish/public_xml_service.rb#L25)

refs https://github.com/sul-dlss/dor-services-app/issues/3098

andrewjbtw:
I'm seeing a change in the `<resource "id">` attribute in the public XML after a republish. I'm not sure if that's intended.

Example from https://argo-stage.stanford.edu/view/druid:fm539yc9854

Before:
```
    <resource id="cocina-fileSet-5b074b49-c431-4704-84b4-c7d61505bdc9" sequence="1" type="image">
```
After:
```
    <resource id="cocina-fileSet-fm539yc9854/5b074b49-c431-4704-84b4-c7d61505bdc9" sequence="1" type="image">
```

In the before, the `id` is `cocina-fileSet-` followed by a uuid, in the after it's `cocina-fileSet-` followed by the druid, followed by `/`, followed by the uuid. I know we're doing some filtering of the full Cocina URI so that it's compatible with Access applications, but I don't know what exactly is supposed to be filtered out besides the `http` prefix.

justinlittman:
This is intentional. It came from the need to keep our file / fileset identifiers stable (instead of changing them up every mapping).

andrewjbtw:
Testing on items:

Except for the identifiers, this looks good. We are publishing the "publish/shelve/preserve" settings where we didn't before, but there doesn't seem to be any harm in doing so. We're also publishing the checksums, which should satisfy the request in https://github.com/sul-dlss/dor-services-app/issues/3077

Testing on collections:

Collections don't have contentMetadata/structural metadata but I checked publishing them anyway. They're throwing an error, apparently because they don't have structural metadata: https://github.com/sul-dlss/dor-services-app/issues/3077

PR #3404, opened 2022-02-10
Remove the ActiveFedora dependency from public_cocina_service

jcoyne:
## Why was this change made?
This gets us closer to being able to remove the activefedora dependency from PublicXmlService
Ref #3329

### DIFF
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -28,11 +28,12 @@ module Publish
 
     private
 
-    attr_reader :item
+    attr_reader :item, :cocina_object
 
     # @raise [Dor::DataError]
     def transfer_metadata(release_tags)
-      transfer_to_document_store(PublicCocinaService.create(item), 'cocina.json')
+      public_cocina = PublicCocinaService.create(cocina_object)
+      transfer_to_document_store(public_cocina.to_json, 'cocina.json')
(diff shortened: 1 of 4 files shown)


app/services/publish/metadata_transfer_service.rb
```diff
@@ -28,11 +28,12 @@ def publish
 
     private
 
-    attr_reader :item
+    attr_reader :item, :cocina_object
 
     # @raise [Dor::DataError]
     def transfer_metadata(release_tags)
-      transfer_to_document_store(PublicCocinaService.create(item), 'cocina.json')
+      public_cocina = PublicCocinaService.create(cocina_object)
```
mjgiarlo, line 35: Is `cocina_object` always a DRO or can it also be a collection? If the former, would you consider renaming `cocina_item`? If the latter, :shipit:

app/services/publish/public_cocina_service.rb
```diff
@@ -3,8 +3,7 @@
 module Publish
   # Filters out the non-public parts of cocina for publishing to purl
   class PublicCocinaService
-    def self.create(item)
-      cocina = Cocina::Mapper.build(item)
+    def self.create(cocina)
```
mjgiarlo, line 6: Same question as above. Thinking about https://github.com/sul-dlss/cocina-models/#usage-conventions

justinlittman (approved):

mjgiarlo (approved):

PR #3411, opened 2022-02-11
Prevent invalid access from being copied from an APO to a collection

jcoyne:
## Why was this change made?

Previously creating a collection with an APO that had default access would cause the server to raise an error because collections have no embargo property. This arose in https://github.com/sul-dlss/dor-services-app/commit/6c93258bced43d07647fd60bfd01305b09be2dd9

Fixes https://github.com/sul-dlss/argo/issues/3097

unit tests

### DIFF
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -203,11 +203,14 @@ module Cocina
       apo = CocinaObjectStore.find(cocina_object.administrative.hasAdminPolicy)
       return cocina_object unless apo.administrative.respond_to?(:defaultAccess) && apo.administrative.defaultAccess
 
-      embargo = cocina_object.access&.embargo
       default_access = apo.administrative.defaultAccess
-      # Add embargo into the DROAccess node if one exists
-      default_access = Cocina::Models::DROAccess.new(default_access.to_h.merge({ embargo: embargo })) if embargo
-      cocina_object.new(access: default_access)
+      updated_access = if cocina_object.collection?
+                         # Collection access only supports dark or world, but default access is more complicated
+                         cocina_object.access.new(access: default_access.access == 'dark' ? 'dark' : 'world')
+                       else
+                         cocina_object.access.new(default_access.to_h)
+                       end
+      cocina_object.new(access: updated_access)
     end
 
     def truncate_label(label)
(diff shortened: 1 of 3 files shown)


app/services/cocina/object_creator.rb
```diff
@@ -203,11 +203,10 @@ def default_access_for(cocina_object)
       apo = CocinaObjectStore.find(cocina_object.administrative.hasAdminPolicy)
       return cocina_object unless apo.administrative.respond_to?(:defaultAccess) && apo.administrative.defaultAccess
 
-      embargo = cocina_object.access&.embargo
-      default_access = apo.administrative.defaultAccess
-      # Add embargo into the DROAccess node if one exists
-      default_access = Cocina::Models::DROAccess.new(default_access.to_h.merge({ embargo: embargo })) if embargo
-      cocina_object.new(access: default_access)
+      default_access = apo.administrative.defaultAccess.to_h
+      default_access.delete(:download) if cocina_object.collection?
```
mjgiarlo, line 207: Do we also need to delete `:controlledDigitalLending` and `:readLocation`? Those are the other two settings that can be in APO default access but cannot in collection access.

app/services/cocina/object_creator.rb
```diff
@@ -203,11 +203,10 @@ def default_access_for(cocina_object)
...
+      updated_access = cocina_object.access.new(default_access.to_h)
```
mjgiarlo, line 208: Do we need to map APO `defaultAccess.access` values (`stanford`, `location-based`, and `citation-only`) to values supported by collections (`world`, `dark`)?

I'm sure we do similar in 1-2 other spots in DSA already. 😤

jcoyne: I'm actually not even sure if Collections should be picking up access from the APO. This is a question for @andrewjbtw

mjgiarlo: Good question.

mjgiarlo (approved):

app/services/cocina/object_creator.rb
andrewjbtw: I see this is already merged, and I replied on Slack, but FWIW I'm pretty sure the Collection-APO relationship has been inconsistently applied in the system up to now. IIRC, creating a collection hasn't historically picked up default rights from the APO but "Apply APO defaults" has.

PR #3412, opened 2022-02-11
Identity normalizer refactor

ndushay:
## Why was this change made?

To reduce mental load when looking at this class - I missed something I should have seen.


ndushay (commented):
there are missing tests for some of the normalization methods.  This PR does not address that.

Issue #3417, opened 2022-02-11
Do a bulk reindex of virtual objects


justinlittman:
Upon completion of sul-dlss/dor-services-app#3419 and https://github.com/sul-dlss/dor_indexing_app/issues/743, perform a bulk reindex of virtual objects using Argo.

justinlittman:
Please make sure that https://github.com/sul-dlss/dor_indexing_app/issues/743 is deployed to production and verify that `has_constituents_ssim` is present after reindex.

jcoyne:
@justinlittman it was deployed on Friday last week.

mjgiarlo:
All 4,693 virtual objects in prod have been reindexed successfully and `has_constituents_ssim` is now present in the index.

PR #3428, opened 2022-02-11
Lookup parent from solr

### LINKED ISSUE #3418: Change PublicDescMetadataService to get constituency from Solr
Currently PublicDescMetadataService determines if an item is part of a virtual object from a relationship stored in RELS-EXT: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/publish/public_desc_metadata_service.rb#L63

This should be changed to query for virtual objects of which it is a constituent by searching on its druid in the `has_constituents_ssim` field. See https://github.com/sul-dlss/dor_indexing_app/issues/743

jcoyne:
Blocked by #3417

## Why was this change made?
Fixes #3418

### DIFF
--- a/app/services/publish/public_desc_metadata_service.rb
+++ b/app/services/publish/public_desc_metadata_service.rb
@@ -60,33 +60,46 @@ module Publish
     # expand constituent relations into relatedItem references -- see JUMBO-18
     # @return [Void]
     def add_constituent_relations!
-      object.relationships(:is_constituent_of).each do |parent|
-        # fetch the parent object to get title
-        druid = parent.gsub(%r{^info:fedora/}, '')
-        parent_item = Dor.find(druid)
-
+      find_virtual_object.each do |solr_doc|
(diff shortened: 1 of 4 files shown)


app/services/publish/public_desc_metadata_service.rb
```diff
@@ -60,11 +60,7 @@ def add_doi
     # expand constituent relations into relatedItem references -- see JUMBO-18
     # @return [Void]
     def add_constituent_relations!
-      object.relationships(:is_constituent_of).each do |parent|
-        # fetch the parent object to get title
-        druid = parent.gsub(%r{^info:fedora/}, '')
-        parent_item = Dor.find(druid)
-
+      find_parent_item.each do |parent_item|
```
mjgiarlo, line 63: Can we `s/parent_item/virtual_object/g` for consistency with the recent renames?

mjgiarlo (approved):

Issue #3408, opened 2022-02-10
H2 object labels


justinlittman:
Some H2 objects, e.g., druid:bb328jb5128 and druid:wg608dc4783 have "Hydrus" as identityMetadata objectLabel even though the Fedora item has a descriptive label.

Currently, when mapping to Cocina, "Hydrus" is used as the label, since objectLabel trumps Fedora item label. This should be changed so that Fedora item label trumps objectLabel when objectLabel="Hydrus".

andrewjbtw:
Strange. I would have expected the opposite to happen. These are both quite recent objects too. Both of these were Hydrus-registered and then H2-deposited and I suspect that has something to do with it.

jcoyne:
This sounds more like a remediation task than a mapping task

justinlittman:
Mapping is faster, easier, and more accurate than remediation. And by mapping it correctly, it will be remediated when we migrate to Cocina. (This approach is used widely for other problems as well.)

jcoyne:
Can we verify that this only applies to items registered in Hydrus and deposited by H2?  It's possible that H2 is doing something wrong, so mapping won't be sufficient if we keep making bad data.

justinlittman:
#3431 fixes the underlying problem that was causing this -- updates were not updating objectLabel.

justinlittman:
With @jcoyne label refactoring, these now map correctly. Win!

PR #3431, opened 2022-02-12
Overwrite identity objectLabel on update.

justinlittman:
refs #3408

## Why was this change made?
So that objectLabels can be updated.

Unit

### DIFF
--- a/app/services/cocina/to_fedora/identity.rb
+++ b/app/services/cocina/to_fedora/identity.rb
@@ -11,8 +11,9 @@ module Cocina
 
       # @param [Dor::Item,Dor::Collection,Dor::Etd,Dor::AdminPolicyObject] fedora_object
       # @param [String] label the label for the cocina object.
-      def self.apply_label(fedora_object, label:)
-        new(fedora_object).apply_label(label)
+      # @param [boolean] overwrite the objectLabel if it already exists
+      def self.apply_label(fedora_object, label:, overwrite: false)
+        new(fedora_object).apply_label(label, overwrite: overwrite)
       end
 
       # @param [Dor::Item,Dor::Collection,Dor::Etd,Dor::AdminPolicyObject] fedora_object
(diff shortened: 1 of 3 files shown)


app/services/cocina/to_fedora/identity.rb
```diff
@@ -11,8 +11,9 @@ def self.initialize_identity(fedora_object)
 
       # @param [Dor::Item,Dor::Collection,Dor::Etd,Dor::AdminPolicyObject] fedora_object
       # @param [String] label the label for the cocina object.
-      def self.apply_label(fedora_object, label:)
-        new(fedora_object).apply_label(label)
+      # @param [boolean] overwrite the objectLabel if it already exists
+      def self.apply_label(fedora_object, label:, overwrite: false)
```
jcoyne, line 15: I'm not seeing any place where this overwrite is false.  Can we just make overwriting be the default?

justinlittman: Overwrite is false when invoked in ObjectCreator, e.g., https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_creator.rb#L92

justinlittman: Sure, but what is the "it"? 

I tried to make sense of the title/label logic in object create -- based how extensive the tests were, it seemed quite deliberate. But I couldn't figure out the intent, so opted to leave it alone.

jcoyne: I think that alone means we should take a look at it. We shouldn't have code or tests that are so confusing that we can't understand the intent.

ndushay: ... so we merged this BEFORE we understood the intent?  Or was there a conversation with an outcome I missed (because it wasn't included here?)

justinlittman: The intent of ObjectCreator code was not understood, so it was left alone by this PR. This only affected the ObjectUpdater code (which set `overwrite` to true), which was the intent.

PR #3432, opened 2022-02-13
Adds feature for updating, creating, and deleting from postgres.

### LINKED ISSUE #3144: Save to / delete from Postgres
Add option (controlled by feature flag) to allow Cocina objects to be saved to Postgres (via `CocinaObjectStore`). Saving is in addition to the existing save to Fedora (via `item.save`).

This should be added to [Cocina::ObjectUpdater](https://github.com/sul-dlss/dor-services-app/blob/a181ada6db35151a26b3acc4ace207198ef988cb/app/services/cocina/object_updater.rb#L46) and [Cocina::ObjectCreator](https://github.com/sul-dlss/dor-services-app/blob/a181ada6db35151a26b3acc4ace207198ef988cb/app/services/cocina/object_creator.rb#L46). (Alternatively, it might be possible to just add to ActiveFedoraPersister.)

This same feature flag should also control deleting Cocina objects from Postgres (via `CocinaObjectStore`). Deleting is in addition to the existing delete from Fedora. This should be added to DeleteService.

The default flag setting should be to not save to Postgres.

justinlittman:
closes #3144

## Why was this change made?
Begin exercising persistence to postgres.

Unit

### DIFF
--- a/app/services/cocina/object_creator.rb
+++ b/app/services/cocina/object_creator.rb
@@ -4,7 +4,7 @@ module Cocina
   # Given a Cocina model, create an ActiveFedora model.
   class ObjectCreator
     # @raises SymphonyReader::ResponseError if symphony connection failed
-    def self.create(cocina_object, event_factory: EventFactory, persister: ActiveFedoraPersister, assign_doi: false, cocina_object_store: CocinaObjectStore)
+    def self.create(cocina_object, event_factory: EventFactory, persister: ActiveFedoraPersister, assign_doi: false, cocina_object_store: CocinaObjectStore.new)
       _fedora_object, cocina_object = new(cocina_object_store: cocina_object_store).create(cocina_object, event_factory: event_factory, persister: persister, assign_doi: assign_doi)
       cocina_object
     end
@@ -13,7 +13,7 @@ module Cocina
       new(cocina_object_store: cocina_object_store).create(cocina_object, event_factory: nil, persister: nil, trial: true, notifier: notifier)
     end
 
-    def initialize(cocina_object_store: CocinaObjectStore)
+    def initialize(cocina_object_store: CocinaObjectStore.new)
       @cocina_object_store = cocina_object_store
     end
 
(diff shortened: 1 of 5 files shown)


justinlittman:
Successfully tested in QA.

app/services/cocina/object_creator.rb
```diff
@@ -38,10 +41,9 @@ def create(cocina_object, event_factory:, persister:, trial: false, notifier: ni
 
         # This creates version 1.0.0 (Initial Version)
         ObjectVersion.increment_version(fedora_object.pid)
-      end
 
-      # This will rebuild the cocina model from fedora, which shows we are only returning persisted data
-      roundtrip_cocina_object = Mapper.build(fedora_object, notifier: notifier)
+        CocinaObjectStore.new.cocina_to_ar_save(roundtrip_cocina_object) if Settings.enabled_features.postgres.create
```
edsu, line 45: I was interested in why a new `CocinaObjectStore` is created here instead of using `@cocina_object_store`.

edsu: I guess it's because because CocinaObjectStore#save also saves to Fedora, and you want to avoid that here? It might be worth a comment?

justinlittman: Good catch. Fixing.

ndushay (approved):
I'm approving but would like another reviewer as I'm not taking the time to fully grok.

app/services/cocina/object_creator.rb
```diff
@@ -38,10 +41,9 @@ def create(cocina_object, event_factory:, persister:, trial: false, notifier: ni
 
         # This creates version 1.0.0 (Initial Version)
         ObjectVersion.increment_version(fedora_object.pid)
-      end
 
-      # This will rebuild the cocina model from fedora, which shows we are only returning persisted data
-      roundtrip_cocina_object = Mapper.build(fedora_object, notifier: notifier)
+        cocina_object_store.cocina_to_ar_save(roundtrip_cocina_object) if Settings.enabled_features.postgres.create
```
jcoyne, line 45: Could this go in the persister?  Should there be a `CombinedFedoraAndPostgresPersister`?

justinlittman: I considered that. However, this is the only place that the persister is used so it just hides the code and doesn't really buy anything. I chose to leave it here where it would be more visible.

Issue #3395, opened 2022-02-10
Remediate geo metadata objects


justinlittman:
Some objects (e.g., druid:vp514gs0717, druid:zn779hf4325) have geo metadata, but are not content geo content type. However, our current code only maps geo metadata when geo content type: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/from_fedora/dro.rb#L72

Action:

- write a report to identify other objects that look like this (appear to be geo but are not marked as geo).
- this is any object that has geoMetadata datastream but is not the geo content type

Remdiation: switch them to geo content type

jcoyne:
I disagree. I think these objects just have the wrong content type. They are in the geo dataset collection and went through the geo workflows. These objects just need remediation.

justinlittman:
This was @andrewjbtw's call.

peetucket:
Action for this ticket: 
- write a report to identify other objects that look like this (appear to be geo but are not marked as geo).
- this is any object that has geoMetadata datastream but is not the geo content type

Remdiation: switch them to geo content type

andrewjbtw:
I looked at some geo objects and I think what happens in the geo workflows is that at some point the content type is actually changed from file to geo. There are no registered objects with content type geo, which leads me to believe they're registered as "file" first and not as geo. So updating these to geo is probably what was supposed to happen but they stalled somewhere in the geo workflows and didn't hit that point.

justinlittman:
So the correct action would be to map geo metadata regardless of type (not remediate), right @andrewjbtw?

jcoyne:
I think allowing geo metadata on non geo items makes the data less useful and makes our systems more complicated. Why have object types if they do not convey anything useful? The generate content metadata in gisAssemblyWF already generates as type 'geo', so these  looks like bad data to me, not how it's typically done.

justinlittman:
https://github.com/sul-dlss/dor-services-app/pull/3433 follows the approach that we've been using for other non-fatal data errors: perform the mapping, but HB notify so that it can be remediated.

justinlittman:
@andrewjbtw Let's go with remediation and see if problem re-appears in testing.

andrewjbtw:
So the remediation is to make the content type geo? Is it just these two or have I lost track of a related report?

justinlittman:
[...]

PR #3433, opened 2022-02-13
Warn on non-geo items with geoMetadata ds.

justinlittman:
refs #3395

## Why was this change made?
Map existing items that are non-geo content type but have geo metadata and provide warning.

Unit

### DIFF
--- a/app/services/cocina/from_fedora/dro.rb
+++ b/app/services/cocina/from_fedora/dro.rb
@@ -53,6 +53,7 @@ module Cocina
       # rubocop:disable Metrics/AbcSize
       def props
         type = DRO.dro_type(fedora_item)
+        check_geo(type)
         {
           externalIdentifier: fedora_item.pid,
           type: type,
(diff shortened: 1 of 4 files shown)


justinlittman:
Before:
```
Status (n=10000; not using Missing for success/different/error stats):
  Success:   9978 (99.86%)
  Different: 14 (0.14%)
  Mapping error:     0 (0.0%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     8 (0.08%)
  Bad cache:     0 (0.0%)
```

After:
```
Status (n=10000; not using Missing for success/different/error stats):
  Success:   9978 (99.86%)
  Different: 14 (0.14%)
  Mapping error:     0 (0.0%)
  Create error:     0 (0.0%)
  Unmapped:     0 (0.0%)
  Expected unmapped:     8 (0.08%)
  Bad cache:     0 (0.0%)
```

jcoyne:
I don't think it's a good idea to map these.  I think we should consider these to be non-conforming objects and remediate them.  I would be surprised if there are more than these 30 objects (https://argo.stanford.edu/catalog?f%5Bprocessing_status_text_ssi%5D%5B%5D=Registered&f%5Bwf_wps_ssim%5D%5B%5D=gisAssemblyWF)  which we can remediate using the "Set content type" bulk update in Argo.

justinlittman:
Remediating and mapping/notifying are not mutually exclusive. However, as we've found in the past remediation is often incomplete, so best to have a backup.

jcoyne:
True, but this adds an extra API call against Fedora, and since the server is struggling, we should ask whether that API call is worth it if we can get by without it.

PR #3437, opened 2022-02-14
Require legacy updates to produce valid cocina before save.

### LINKED ISSUE #3426: When updating updating legacy datastreams, map to cocina before saving
Currently, datastreams that are updated via the [update_legacy_metadata](https://github.com/sul-dlss/dor-services-app/blob/10a11083f9075dea17a30ac14bc8e688da3de872/app/controllers/metadata_controller.rb#L26) and [mods update](https://github.com/sul-dlss/dor-services-app/blob/main/app/controllers/mods_controller.rb#L11) endpoints save the Fedora item before mapping the item to Cocina. This allows users to create Fedora items that cannot map to Cocina.

Instead, map the updated Fedora item to cocina before saving it.

justinlittman:
closes #3426

## Why was this change made?
To prevent users from producing bad data when editing datastreams.

Unit, qa

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,7 +16,7 @@ class ApplicationController < ActionController::API
     json_api_error(status: :unprocessable_entity,
                    title: 'Unexpected Cocina::Mapper.build error',
                    message: e.cause,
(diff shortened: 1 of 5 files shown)


justinlittman:
image

PR #3439, opened 2022-02-14
problematic virtual objects#3425

### LINKED ISSUE #3425: Write report for problematic virtual objects
Write a report that looks for items which contain content metadata that have `<resource[@type != 'image']>` and contain an `<externalFile>`.

For example, https://argo-prod-01.stanford.edu/view/druid:fw628pk2797:
```
<contentMetadata objectId="fw628pk2797" type="file">
  <resource id="fw628pk2797_1" sequence="1" type="file">
    <externalFile fileId="SC0193_2017-095_30_SupMat_4726_1997-09-25.pdf" mimetype="application/pdf" objectId="druid:kd296dc7239" resourceId="kd296dc7239_1"/>
    <relationship objectId="druid:kd296dc7239" type="alsoAvailableAs"/>
  </resource>
```

mjgiarlo:
Fixes #3425

## Why was this change made?

So we know how many virtual objects are invalid in this way.

Ran on sdr-deploy.

None


mjgiarlo:
Results of the report in prod: https://github.com/sul-dlss/dor-services-app/issues/3425#issuecomment-1039400749

PR #3450, opened 2022-02-14
Generate versionMetadata.xml from object versions.

### LINKED ISSUE #3448: Write versionMetadata.xml from ObjectVersion (AR) in SDR Ingest Service

justinlittman:
closes #3448

## Why was this change made?
Versions need to be included in the moab.

Unit

### DIFF
--- a/app/services/datastream_extractor.rb
+++ b/app/services/datastream_extractor.rb
@@ -61,6 +61,7 @@ class DatastreamExtractor
   def datastream_content(ds_name, required)
     ds = (ds_name == :relationshipMetadata ? 'RELS-EXT' : ds_name.to_s)
     return workflow_xml if ds_name == :workflows
+    return version_xml if ds_name == :versionMetadata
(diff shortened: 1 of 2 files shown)


jcoyne (approved):

app/services/datastream_extractor.rb
```diff
@@ -70,4 +71,21 @@ def datastream_content(ds_name, required)
...
     WorkflowClientFactory.build.all_workflows_xml(item.pid)
   end
+
+  def version_xml
+    # This can be removed after migration.
+    VersionMigrationService.migrate(item)
+
+    object_versions = ObjectVersion.where(druid: item.pid).order(:version)
+
+    Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
```
jcoyne, line 81: Consider moving to a class method on ObjectVersion

justinlittman: Sorry, missed that before merging. Will fix in follow-up PR.

PR #3482, opened 2022-02-16
[HOLD] normalize blank file directives to no

### LINKED ISSUE #3473: Normalize blank preserve, publish, and shelve attributes to "no" (contentMetadata)
For example, druid:ft113wv9260, druid:ns538tk9778:
```
<file mimetype="image/x-3ds" preserve="" format="" size="72078920" shelve="" publish="" id="ns538tk9778_00_0002.tif">
      <imageData width="4865" height="4933"/>
      <attr name="representation">uncropped</attr>
      <checksum type="sha1">29a93d8e1374511c552ba660a1bf99c2514085ac</checksum>
      <checksum type="md5">a2b5b59896277e7dcc0cd3fb72c444ed</checksum>
    </file>
```

All blank preserve, publish, or shelve attributes should be normalized to "no".

peetucket:

Fixes #3473 - normalize blank file directives (publish/preserve/shelve) to no

HOLD pending roundtrip validation testing

1. Added a new test
2. Running roundtrip validations, results in comment below

### DIFF
--- a/app/services/cocina/normalizers/content_metadata_normalizer.rb
+++ b/app/services/cocina/normalizers/content_metadata_normalizer.rb
@@ -6,6 +6,7 @@ module Cocina
     # when round-tripping.
     class ContentMetadataNormalizer
       include Cocina::Normalizers::Base
+      FILE_DIRECTIVES = %i[publish preserve shelve].freeze
 
       # @param [String] druid
       # @param [Nokogiri::Document] content_ng_xml content metadata XML to be normalized
@@ -47,6 +48,7 @@ module Cocina
         normalize_empty_xml
         normalize_content_file_type
         normalize_image_data
+        normalize_blank_file_directives
 
         regenerate_ng_xml(ng_xml.to_s)
       end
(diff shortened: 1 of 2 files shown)


app/services/cocina/normalizers/content_metadata_normalizer.rb
```diff
@@ -163,6 +165,13 @@ def normalize_label_attr
         end
       end
 
+      def normalize_blank_file_directives
+        query = FILE_DIRECTIVES.map { |attr| "@#{attr} = ''" }.join(' or ')
+        ng_xml.root.xpath("//file[#{query}]").each do |file|
+          FILE_DIRECTIVES.each { |attr| file[attr] = 'no' if file[attr] == '' }
+        end
```
peetucket, line 172: if we also want to deal with the case of normalizing the directive to "no" when missing and not just blank, we can easily do that with a `file[attr].blank?` instead of `file[attr] == ''`

justinlittman: The existing code is logically OK, but this might be a little cleaner:
```
FILE_DIRECTIVES.each |directive|
  ng_xml.root.xpath("//file[@#{directive} = '']").each do |file|
    file[directive] = "no"
  end
end
```

justinlittman (approved):
Approve pending consideration of suggestion and successful roundtrip testing.

peetucket:
Yes, your version is cleaner to read...adjusted my code.

peetucket:
Results are back -- improvement of 1 druid.

PR #3483, opened 2022-02-16
When updating, only save when fedora obj can be mapped to cocina.

justinlittman:
Prevent bad data.

Unit


jcoyne (changes requested):
Two small suggestions

.rubocop.yml
```diff
@@ -96,6 +96,10 @@ RSpec/ExampleLength:
 RSpec/MultipleExpectations:
   Max: 19
 
+RSpec/MultipleMemoizedHelpers:
```
jcoyne, line 99: Currently we have a threshold configured for this in .rubocop_todo.yml.  Should we just adjust that up or just disable this cop?

jcoyne (approved):

PR #3491, opened 2022-02-16
Push the ActiveFedora dependency out of the interface for Preservatio…

### LINKED ISSUE #3178: Rewrite PreserveJob / SdrIngestService to use Cocina object

jcoyne:
…nMetadataExtractor

Fixes #3178
Decoupling from ActiveFedora.

### DIFF
--- a/spec/services/sdr_ingest_service_spec.rb
+++ b/spec/services/sdr_ingest_service_spec.rb
@@ -37,18 +37,17 @@ RSpec.describe SdrIngestService do
 
   describe '.transfer' do
     let(:druid) { 'druid:dd116zh0343' }
-    let(:dor_item) { instance_double(Dor::Item, pid: druid) }
-    let(:cocina_object) { instance_double(Cocina::Models::DRO) }
+    let(:cocina_object) { instance_double(Cocina::Models::DRO, externalIdentifier: druid) }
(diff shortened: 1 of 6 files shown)


spec/services/sdr_ingest_service_spec.rb
```diff
@@ -113,17 +112,17 @@
 
   describe '.signature_catalog_from_preservation' do
     let(:druid) { 'druid:dd116zh0343' }
-    let(:dor_item) { instance_double(Dor::Item, pid: druid) }
+    let(:cocina) { instance_double(Cocina::Models::DRO, externalIdentifier: druid) }
```
jmartin-sul, line 115: does this let var actually get referenced?  cruft, or me just not seeing where it's used?

jmartin-sul (approved):
one question about a possibly extraneous let var in a test, but otherwise LGTM.

spec/services/sdr_ingest_service_spec.rb
jmartin-sul: (feel free to merge if i'm off-base, or after removing the var if it actually is superfluous)

justinlittman (approved):

spec/services/sdr_ingest_service_spec.rb
jcoyne: It doesn't seem to be used. I'll remove it.

PR #3492, opened 2022-02-16
content_metadata_normalizer refactors

ndushay:

I was confused by some things in the content_metadata_normalizer when I started to work on a ticket, so I got some clarifications, fixed a bug and did some other refactors.

### DIFF
--- a/app/services/cocina/normalizers/content_metadata_normalizer.rb
+++ b/app/services/cocina/normalizers/content_metadata_normalizer.rb
@@ -2,8 +2,9 @@
 
 module Cocina
   module Normalizers
-    # Normalizes a Fedora object content metadata datastream, accounting for differences between Fedora rights and cocina rights that are valid but different
-    # when round-tripping.
+    # Normalizes a Fedora object content metadata datastream, accounting for differences
+    #   between Fedora contentMetadata and cocina structural that are valid but different
+    #   when round-tripping.
     class ContentMetadataNormalizer
       include Cocina::Normalizers::Base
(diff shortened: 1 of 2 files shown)


app/services/cocina/normalizers/content_metadata_normalizer.rb
```diff
@@ -27,37 +29,37 @@ def initialize(content_ng_xml:)
...
         regenerate_ng_xml(ng_xml.to_s)
       end
 
+      # resource ids and sequence numbers are regenerated so they must be normalized out of the roundtrip comparison
       def normalize_roundtrip
-        remove_resource_id
+        remove_resource_id_attribute
         remove_external_resource_id
-        remove_sequence
-        normalize_checksum
```
ndushay, line 60: `normalize_checksum` should NOT be in this, and was put in here because we have a spec calling this method in a way that wasn't actually doing the roundtripping.

spec/services/cocina/normalizers/content_metadata_normalizer_spec.rb
```diff
@@ -593,7 +625,7 @@
     # Adapted from bb157dt2402
     let(:original_xml) do
       <<~XML
-        <contentMetadata type="file" stacks="/web-archiving-stacks/data/collections/mm553tf6423" id="druid:bb035tg0974">
```
ndushay, line 596: remove attribute irrelevant to this test.

justinlittman (approved):

PR #3500, opened 2022-02-17
Require cocina.json in BagVerifier

### LINKED ISSUE #3496: Add cocina.json to BagVerifier
`data/metadata/cocina.json` is one of the files that should be verified by Preserve::BagVerifier.

edsu:

The BagVerifier should ensure that `cocina.json` is present.

Unit tests.

closes #3496

### DIFF
--- a/spec/services/preserve/bag_verifier_spec.rb
+++ b/spec/services/preserve/bag_verifier_spec.rb
@@ -5,6 +5,8 @@ require 'rails_helper'
 RSpec.describe Preserve::BagVerifier do
   let(:fixtures) { Pathname(File.dirname(__FILE__)).join('../../fixtures') }
 
+  # NOTE: BagVerifier#verify is exercised by spec/services/sdr_ingest_service_spec.rb
+
   describe '#verify_pathname' do
     subject(:verify_pathname) { instance.verify_pathname(path) }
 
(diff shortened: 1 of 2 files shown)


spec/services/preserve/bag_verifier_spec.rb
```diff
@@ -30,4 +30,22 @@
       end
     end
   end
+
+  describe '#verify' do
+    subject(:verify) { instance.verify }
+
+    context 'with a valid bag' do
+      let(:instance) { described_class.new(directory: fixtures.join('bags/good')) }
```
jcoyne, line 38: Rather than adding 68 spec files, can you just check that it's in the list of required files?  This seems like a huge decrease in the signal-to-noise, for a small benefit.

jcoyne (changes requested):

spec/services/preserve/bag_verifier_spec.rb
edsu: Yes, I just wasn't sure if down the road we might want to verify more things about a bag.

edsu:
As noted in #3496 it turns out that `BagVerifier#verify` is exercised by another spec. So I've removed the test and the fixture data and left a note about that in `spec/services/preserve/bag_verifier.rb`.

PR #3503, opened 2022-02-17
Public desc no identity

jcoyne:


app/services/publish/public_desc_metadata_service.rb
```diff
@@ -44,15 +44,12 @@ def strip_comments!
...
-    # PURL to display the DOI
+    # Export DOI into the public descMetadata to allow PURL to display it
     def add_doi
-      value = object.identityMetadata.ng_xml.xpath('//doi').first
-      return unless value
+      return unless cocina_object.dro? && cocina_object.identification.doi
 
       identifier = doc.create_element('identifier', xmlns: MODS_NS)
-      identifier.content = "https://doi.org/#{value.text}"
+      identifier.content = "https://doi.org/#{cocina_object.identification.doi}"
```
justinlittman, line 52: Can this use https://github.com/sul-dlss/dor-services-app/blob/main/app/services/doi.rb#L5?

justinlittman (approved):

app/services/publish/public_desc_metadata_service.rb
justinlittman: No it can't.

jcoyne: No, that is for generating new DOIs, and this is just showing the DOI already in the record.

PR #3516, opened 2022-02-17
[HOLD] Cocinizes EmbargoReleaseService.

### LINKED ISSUE #3308: Rewrite EmbargoReleaseService to use cocina items
https://github.com/sul-dlss/dor-services-app/blob/main/app/services/embargo_release_service.rb#L37

This will depend on the VersionService (#3170) being rewritten first.

justinlittman:
closes #3308

Less Fedora

Unit

### DIFF
--- a/app/services/embargo_release_service.rb
+++ b/app/services/embargo_release_service.rb
@@ -10,7 +10,7 @@ class EmbargoReleaseService
   # Finds druids from solr based on the passed in query
   # It will then load each item from Dor, and call the block with the item
   # @param [String] query used to locate druids of items to release from solr
-  # @yield [Dor::Item] gets executed after loading the object from DOR and opening new version
+  # @yield [Cocina::Models::DRO] gets executed after loading the object from DOR and opening new version
   #  Steps needed to release the particular embargo from the item
   def self.release_items(query, &release_block)
     # Find objects to process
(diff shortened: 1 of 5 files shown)


app/services/embargo_release_service.rb
```diff
@@ -48,64 +75,35 @@ def self.release_item(druid)
...
-    end
+  def release_cocina_object(cocina_object)
+    # Copy access > embargo > useAndReproductionStatement, access, download, readLocation, controlledDigitalLending to access >
+    # Remove access > embargo
+    access_props = cocina_object.access.to_h.except(:access, :download, :readLocation, :controlledDigitalLending, :useAndReproductionStatement)
+    access_props.merge!(access_props[:embargo].except(:releaseDate))
+    # Remove embargo
+    access_props.delete(:embargo)
 
-    events.add_event('embargo', release_agent, 'Embargo released')
```
jcoyne, line 103: Are you intentionally dropping this event?

jcoyne (approved):
Looks good, so long as we don't care about that event record.  I suspect the version writing into the event is sufficient.

app/services/embargo_release_service.rb
justinlittman: This is for the events datastream, so intentionally dropping.

justinlittman:
Testing completed by @andrewjbtw

PR #3517, opened 2022-02-17
Include the objectLabel in the public xml

jcoyne:
This value is used by the searchworks indexer for collection titles.
See https://github.com/sul-dlss/searchworks_traject_indexer/blob/master/lib/sdr_stuff.rb#L54

### DIFF
--- a/app/services/publish/public_xml_service.rb
+++ b/app/services/publish/public_xml_service.rb
@@ -82,6 +82,7 @@ module Publish
 
     # catkeys are used by PURL
     # objectType is used by purl-fetcher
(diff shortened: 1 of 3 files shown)


app/services/publish/public_xml_service.rb
```diff
@@ -90,6 +91,7 @@ def public_identity_metadata
         <<~XML
           <identityMetadata>
             <objectType>#{public_cocina.collection? ? 'collection' : 'item'}</objectType>
+            <objectLabel>#{public_cocina.label}</objectLabel>
```
edsu, line 94: I think the answer is probably no, but does the newish [Label.for](https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/from_fedora/label.rb#L7-L10) need to be used here? Or will that already have been called I guess?

edsu (approved):

PR #3519, opened 2022-02-17
[refactor] each notification aborts itself if rabbitMQ is not enabled

### LINKED ISSUE #3507: [Refactor] Move enabled check to within Notifications
Whenever notifications are invoked, they are guarded by checking the setting flag. For example:
```
Notifications::ObjectCreated.publish(model: cocina_object, created_at: Time.zone.now, modified_at: Time.zone.now) if Settings.rabbitmq.enabled
```

This setting check should be moved inside the Notifications::* classes because:
* A developer shouldn't need to remember to add the guard whenever using a notification.
* It requires additional testing and test setup whenever a notification is used.

peetucket:

Fixes #3507 - make the notifications responsible for checking if rabbitMQ is enabled instead of the caller

Added new tests

NOTE: The diffs in the specs are less substantial than they appear.  The existing tests were moved into a new context of "RabbitMQ enabled", with mocking of that setting to true.  Then a new context block was added mocking RabbitMQ to disabled to verify the messages are not sent.

### DIFF
--- a/app/controllers/metadata_controller.rb
+++ b/app/controllers/metadata_controller.rb
@@ -55,11 +55,9 @@ class MetadataController < ApplicationController
     cocina_object = Cocina::Mapper.build(@item)
 
     @item.save!
-    if Settings.rabbitmq.enabled
-      Notifications::ObjectUpdated.publish(model: cocina_object,
-                                           created_at: @item.create_date,
-                                           modified_at: @item.modified_date)
-    end
+    Notifications::ObjectUpdated.publish(model: cocina_object,
+                                         created_at: @item.create_date,
+                                         modified_at: @item.modified_date)
   rescue LegacyMetadataService::DatastreamValidationError => e
     json_api_error(status: :unprocessable_entity, message: e.detail, title: e.message)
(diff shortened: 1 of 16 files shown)


peetucket:
> Can you remove from existing tests where no longer necessary? For example, `legacy_metadata_update_spec`?

That test is mocking the RabbitMQ setting to true and the call to publish so that it can verifying the call to publish... which still seems reasonable to me...if get rid of the mocking, we would need to get rid of the test expectaction too, e.g.

https://github.com/sul-dlss/dor-services-app/blob/main/spec/requests/legacy_metadata_update_spec.rb#L177

And in the other tests that do this as well.

justinlittman:
No, the test expectation would remain the same in that case -- you'd expect Notification::Updated to have been invoked.

peetucket:
> No, the test expectation would remain the same in that case -- you'd expect Notification::Updated to have been invoked.

ok, i just removed the rabbitmq setting mocking in those other tests (not needed anymore since it defaults to false) - everything else stays the same in the test

PR #3535, opened 2022-02-18
Docs: update scp user

edsu:

Updating the docs to make it easier to scp certs and keys. You need to scp as `root` in order to scp files from `/etc/pki/tls/private/` (which are read-only by root).

I tried the scp command on sul-dor-qa.stanford.edu and sul-dor-prod.stanford.edu


justinlittman (approved):
I'm surprised that root works, but I've you've tested then works for me.

edsu:
Yeah, I was surprised it worked too TBH.

PR #3539, opened 2022-02-18
Run rubocop auto-correct

mjgiarlo:

Try to make Rubocop configuration less bad?


ndushay (changes requested):
I don't want to review ALL of this in one shot;  at best I got halfway.  I added some review comments, but 

1.  I don't think this should get merged as is
2.  Can we separate out the whitespace changes into a separate PR, if we're doing to do it?
3.  I think we should turn some rules off or use exceptions rather than take autocorrect ... but there's too much here for me to figure out which ones.

PR #3550, opened 2022-02-19
add report-content-imagedata-not-num

### LINKED ISSUE #3540: Report on non-SVG with non-integer height or width
Create a report for any objects which contain content metadata that has `<file[@mimetype!="image/svg+xml">` and child `<imageData>` has a height or width attribute that is not blank and not an integer.

See #3323

ndushay:

To look beyond the tea leaves into the actual data, I presume.

Closes #3540 

I ran the code in this report against some contentMetadata in the rails console.

### DIFF
--- a/bin/reports/report-content-imagedata-not-num
+++ b/bin/reports/report-content-imagedata-not-num
@@ -1,0 +1,20 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative '../../config/environment'
+require_relative '../../lib/report'
+
+Report.new(name: 'content-missing-imagedata-non-num', dsid: 'contentMetadata').run do |ng_xml|
+  not_svg = '@mimetype!="image/svg+xml"'
+
+  width = ng_xml.xpath("/contentMetadata/resource/file[#{not_svg}]/imageData/@width").presence&.first&.content
+  height = ng_xml.xpath("/contentMetadata/resource/file[#{not_svg}]/imageData/@height").presence&.first&.content
+
+  # we are interested in the ones that cannot be parsed as a Integer value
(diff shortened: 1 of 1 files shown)


bin/reports/report-content-imagedata-not-num
```diff
@@ -0,0 +1,20 @@
...
+require_relative '../../lib/report'
+
+Report.new(name: 'content-missing-imagedata-non-num', dsid: 'contentMetadata').run do |ng_xml|
+  not_svg = '@mimetype!="image/svg+xml"'
+
+  width = ng_xml.xpath("/contentMetadata/resource/file[#{not_svg}]/imageData/@width").presence&.first&.content
+  height = ng_xml.xpath("/contentMetadata/resource/file[#{not_svg}]/imageData/@height").presence&.first&.content
+
+  # we are interested in the ones that cannot be parsed as a Float value
+  Float(width) if width.present?
```
justinlittman, line 14: Integer, not float.

justinlittman (changes requested):

bin/reports/report-content-imagedata-not-num
ndushay: er ... we have ones that are Float like 29.82334124mm -- I thought that was ok

ndushay: Is the goal to identify all the non-integer ones, or the ones with  "mm" or whatever?

justinlittman: Non-integer and non-blank.

ndushay: ok!  can do.

justinlittman (approved):

PR #3574, opened 2022-02-22
Fix check for world_discoverable

aaron-collier:

Fixes https://github.com/sul-dlss/common-accessioning/issues/840 (at least part of)

### DIFF
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -16,7 +16,7 @@ module Publish
     # Appends contentMetadata file resources from the source objects to this object
     # @raise [Dor::DataError]
     def publish
-      return unpublish unless world_discoverable?
+      return unpublish unless discoverable?
 
       # Retrieve release tags from identityMetadata and all collections this item is a member of
       release_tags = ReleaseTags.for(cocina_object: cocina_object)
@@ -46,8 +46,8 @@ module Publish
       publish_delete_on_success
     end
 
-    def world_discoverable?
-      ['world', 'citation-only'].include? cocina_object.access.access
+    def discoverable?
+      cocina_object.access.access != 'dark'
     end
 
     # Create a file inside the content directory under the stacks.local_document_cache_root


app/services/publish/metadata_transfer_service.rb
```diff
@@ -47,7 +47,9 @@ def unpublish
     end
 
     def world_discoverable?
-      ['world', 'citation-only'].include? cocina_object.access.access
+      return false if ['dark', 'none'].include? cocina_object.access.access
```
mjgiarlo, line 50: AFAIK, there are only five valid values for `access.access` across our models and those are world, stanford, location-based, citation-only, and dark. So I am wondering if `'none'` can be removed from this check. Cc: @andrewjbtw

aaron-collier: Oh, that's likely @mjgiarlo - I'd have to hunt it down, but I've seen `none`... somewhere... and thought to include it. But if it's unnecessary, happy to remove.

jcoyne: Since include? is a boolean, can you remove the `if` statement?

jcoyne (changes requested):

app/services/publish/metadata_transfer_service.rb
jcoyne: "none" is possible for download, but not for access. See https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml

aaron-collier: Ah, yeah, that's it.. Thanks @jcoyne

app/services/publish/metadata_transfer_service.rb
```diff
@@ -47,7 +47,7 @@ def unpublish
     end
 
     def world_discoverable?
```
jcoyne, line 49: Can we change the name please? "world_discoverable?" is misleading. Now "discoverable?" seems like a better name.

app/services/publish/metadata_transfer_service.rb
```diff
@@ -47,7 +47,7 @@ def unpublish
...
-      ['world', 'citation-only'].include? cocina_object.access.access
+      %w[world stanford location-based citation-only].include? cocina_object.access.access
```
jcoyne, line 50: Would it be easier to say `!= 'dark'`?

jcoyne (changes requested):

jcoyne (approved):

mjgiarlo (approved):

PR #3576, opened 2022-02-22
Update pull_request_template.md

ndushay:

In the vain hopes it will help us better recognize when to run integration tests.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,7 +4,7 @@
 
 ## How was this change tested? 🤨
 
-⚡ ⚠ If this change has cross service impact, including writes to file system, ***run [integration tests](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test in [stage|qa] environment, in addition to specs. ⚡
+⚡ ⚠ If this change has cross service impact, including data writes to shared file system, ***run [integration tests](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test in [stage|qa] environment, in addition to specs. ⚡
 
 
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -4,7 +4,7 @@
 
 ## How was this change tested? 🤨
 
-⚡ ⚠ If this change has cross service impact, ***run [integration tests](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test in [stage|qa] environment, in addition to specs. ⚡
+⚡ ⚠ If this change has cross service impact, including writes to file system, ***run [integration tests](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test in [stage|qa] environment, in addition to specs. ⚡
```
jcoyne, line 7: I'm not sure this is specific enough.  Can we say a "shared file system"?

ndushay: good idea, but I'm not sure that folks will know from the code when we're writing to a shared file system.

ndushay: I understand this;  I think everyone on the team understands this.  Not everyone on the team will know what is being written to a "shared file system" or which ones we mean, or want to go looking through shared_configs and puppet to find out

aaron-collier: I think for clarity, @jcoyne's suggestion makes sense to me. Or perhaps something that indicates non-logging based writes like `including writes to file system (exceptions for simple logging, etc.)`.

mjgiarlo: Agree that @jcoyne's tweak makes sense in this case.

ndushay: ok, updated!   Would be delighted if someone merged

aaron-collier (approved):

PR #3577, opened 2022-02-22
preservation_metadata_extractor: add back some yard doc info

ndushay:

The information was correct and I removed it under the impression it was incorrect.  My bad.

### DIFF
--- a/app/services/preservation_metadata_extractor.rb
+++ b/app/services/preservation_metadata_extractor.rb
@@ -3,7 +3,7 @@
 # This writes the object metadata files to the workspace metadata directory
 class PreservationMetadataExtractor
   # @param [Cocina::Models::DRO, Cocina::Models::AdminPolicy, Cocina::Models::Collection] cocina_object The representation of the digital object
-  # @param workspace The representation of the item's work area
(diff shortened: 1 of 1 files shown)


aaron-collier (approved):

app/services/preservation_metadata_extractor.rb
```diff
@@ -3,15 +3,15 @@
 # This writes the object metadata files to the workspace metadata directory
 class PreservationMetadataExtractor
   # @param [Cocina::Models::DRO, Cocina::Models::AdminPolicy, Cocina::Models::Collection] cocina_object The representation of the digital object
-  # @param workspace The representation of the item's work area
+  # @param DruidTools::Druid] workspace The representation of the item's work area
```
aaron-collier, line 6: Approved pending update to add the `[`

aaron-collier (changes requested):

app/services/preservation_metadata_extractor.rb
ndushay: thanks for the sharp eyes -- I missed that.  Fixed.

aaron-collier (approved):

Issue #3145, opened 2022-01-11
Find from Postgres


justinlittman:
Add option (controlled by feature flag) to allow CocinaObjectPersister to find the Cocina object from Postgres (via CocinaObjectStore) before Fedora (via Dor.find). If the Cocina object is not found in Postgres, it should be loaded from Fedora.

The default setting for the feature flag should be to disable finding from Postgres.

Depends on #3143 and #3138.

justinlittman:
Note that this work is proactive -- it cannot be enabled until cocinization and mapping work is completed.

justinlittman:
closed by #3598

PR #3578, opened 2022-02-23
public_desc_metadata_service: fix bug for APOs

ndushay:

To address https://app.honeybadger.io/projects/50568/faults/84175043, which happened in prod with an APO.  (not sure why we need public_desc_metadata for an APO.)

QUESTION:  do we need to put in some guards for Agreement objects also?   

QUESTION:  how do we find other places where a cocina object of unknown type is being queried for`structural` or `access` info?   as they don't all have it.

### DIFF
--- a/app/services/publish/public_desc_metadata_service.rb
+++ b/app/services/publish/public_desc_metadata_service.rb
@@ -102,7 +102,7 @@ module Publish
     # For use in published mods and mods-to-DC conversion.
     # @return [Void]
     def add_collection_reference!
-      return if cocina_object.collection? || cocina_object.structural&.isMemberOf.blank?
+      return if cocina_object.collection? || cocina_object.admin_policy? || cocina_object.structural&.isMemberOf.blank?
 
       collections = CocinaObjectStore.find_collections_for(cocina_object, swallow_exceptions: true)
 
--- a/spec/services/publish/public_desc_metadata_service_spec.rb
+++ b/spec/services/publish/public_desc_metadata_service_spec.rb
@@ -117,6 +117,28 @@ RSpec.describe Publish::PublicDescMetadataService do
         doc
       end
     end
+
+    context 'when the object is an apo' do
+      subject(:doc) { service.ng_xml(include_access_conditions: false) }
+
+      let(:cocina_object) do
+        Cocina::Models.build({
+                               'type' => Cocina::Models::Vocab.admin_policy,
+                               'label' => 'test',
+                               'externalIdentifier' => 'druid:bb666cc7777',
+                               'version' => 1,
+                               'administrative' => {
+                                 'hasAdminPolicy' => 'druid:ss666tt7777',
+                                 'hasAgreement' => 'druid:ff666kk6666'
+                               },
+                               'description' => description
+                             })
+      end
+
+      it 'has no errors' do
+        doc
+      end
+    end
   end
 
   describe '#to_xml' do


app/services/publish/public_desc_metadata_service.rb
```diff
@@ -102,7 +102,7 @@ def purl_url(druid)
     # For use in published mods and mods-to-DC conversion.
     # @return [Void]
     def add_collection_reference!
-      return if cocina_object.collection? || cocina_object.structural&.isMemberOf.blank?
+      return if cocina_object.collection? || cocina_object.admin_policy? || cocina_object.structural&.isMemberOf.blank?
```
aaron-collier, line 105: I wonder if `return unless cocina_object.dro?` would be sufficient here...

ndushay: I'm not sure about Agreements.  @andrewjbtw ?

jcoyne: I don't believe that AdminPolicies are ever published, so how did it get into this code in the first place?

jcoyne: This happened because someone pressed the "MODS" button in Argo on the APO show page.  @andrewjbtw is that something that we want to allow in Argo?

andrewjbtw: Why wouldn't we allow it? The MODS button and Dublin Core buttons display descriptive metadata and APOs have descriptive metadata. Is publish functionality somehow hooked up to this even though those buttons are purely Argo functions in terms of user interaction?

jcoyne: Those buttons are specifically hooked up to the endpoints that produce the public desc metadata.  Should they be or should it be the unprocessed version? Should the buttons indicate that it is the public version?

andrewjbtw: Ok, I didn't realize that. I can see the logic of it since it must be the closest we can get to a preview of what a page will look like. 

In that case, we can remove the buttons from APOs and agreements. I would still like APOs to handle descriptive metadata internally through the bulk metadata upload function. I'm adding notes to some of the old Hydrus APOs. But if those can only be seen via Cocina that's ok.

justinlittman: Let me just add that agreements are `Cocina::Models::DRO` so will respond to `.structural`.

ndushay: uh ... so is the consensus that this PR is:  

- no longer needed due to #3145?
- okay as is?
- needs to change to X?

jcoyne: I think we should put a check at the controller level such that it is not possible to call the `Publish::*` code on an APO.

ndushay: are there other types we shouldn't publish?   Or are DRO, Collection and Agreement all publish-able?

jcoyne: Agreements are not publishable either, but they are a subclass of Item.

ndushay:
@jcoyne was this effectively addressed with your PR of yesterday?  Or is this still a thing (no honeybadger errors for it since Tues 2/22 morning)

jcoyne:
Yeah, the PR prevents this code from being used for non-publishable objects.

PR #3582, opened 2022-02-23
[HOLD] Fixes logic for applying default object rights.

justinlittman:
closes https://github.com/sul-dlss/google-books/issues/753

The previous logic was causing APO default object rights to trump collection/DRO provided rights.

Unit

### DIFF
--- a/app/services/access_merge_service.rb
+++ b/app/services/access_merge_service.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+# Determines access by merging provided access with default access.
+class AccessMergeService
+  # @param [Cocina::Models::RequestDRO,Cocina::Models::RequestCollection] cocina_object for which to determine access.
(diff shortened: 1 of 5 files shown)


ndushay (commented):
I'm confused by the use of the word "Default".   Is "defaultAccess" actually more like "computedAccess"?   Is it a bad name for the class, with its description of "Determines access by merging provided access with default access."

And is "default_access_for" method also a bad name?  is it really "access" or "access_level"?

Also, I didn't look closely, but:

- the "create_collection_spec" set everything to "world" - would it not be better to have a variety?
- and in your new spec, most of them have the same starting default access in the APO - would it be better to vary that?  I see you do vary it for RequestCollection, but not so much for RequestDRO.

These are small quibbles, and shouldn't hold up fixing a production problem.  But It would be great if you would consider them for a follow-on PR if you can't get them into this one.

ndushay (approved):

PR #3584, opened 2022-02-24
Rewrite RefreshMetadataAction for Cocina objects

### LINKED ISSUE #3169: Rewrite RefreshMetadataAction and MetadataRefreshController to use Cocina objects

aaron-collier:

Fixes #3169 - rewrites RefreshMetadataAction to use cocina objects, Moves existing RefreshMetadataAction into LegacyRefreshMetadataAction for object creation where Draft cocina objects do not have the required externalIdentifier yet.

Updated unit tests.

### DIFF
--- a/app/controllers/metadata_refresh_controller.rb
+++ b/app/controllers/metadata_refresh_controller.rb
@@ -2,18 +2,18 @@
 
 # Gets the catkey or barcode from identityMetadata and returns the catalog info
 class MetadataRefreshController < ApplicationController
+  before_action :load_cocina_object
+
   rescue_from(SymphonyReader::ResponseError) do |e|
     render status: :internal_server_error, plain: e.message
   end
 
   def refresh
-    @item = Dor.find(params[:id])
-    status = RefreshMetadataAction.run(identifiers: identifiers,
-                                       fedora_object: @item)
-    return render status: :unprocessable_entity, plain: "#{@item.pid} had no resolvable identifiers: #{identifiers.inspect}" unless status
-    return render status: :internal_server_error, plain: "#{@item.pid} descMetadata missing required fields (<title>)" if missing_required_fields?
+    updated_cocina_object = RefreshMetadataAction.run(identifiers: identifiers,
+                                                      cocina_object: @cocina_object)
+    return render status: :unprocessable_entity, plain: "#{@cocina_object.externalIdentifier} had no resolvable identifiers: #{identifiers.inspect}" unless updated_cocina_object
 
-    @item.save!
+    CocinaObjectStore.save(updated_cocina_object)
   rescue SymphonyReader::NotFound => e
     json_api_error(status: :bad_request, title: 'Catkey not found in Symphony', message: e.message)
   end
@@ -21,10 +21,10 @@ class MetadataRefreshController < ApplicationController
   private
 
   def identifiers
-    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
-  end
+    return [] unless @cocina_object.identification&.catalogLinks
 
-  def missing_required_fields?
-    @item.descMetadata.mods_title.blank?
+    @identifiers ||= @cocina_object.identification.catalogLinks.filter_map { |clink| "catkey:#{clink.catalogRecordId}" if clink.catalog == 'symphony' }.tap do |id|
+      id << "barcode:#{@cocina_object.identification.barcode}" if @cocina_object.identification.barcode
+    end
   end
 end
--- a/app/services/refresh_metadata_action.rb
+++ b/app/services/refresh_metadata_action.rb
@@ -3,50 +3,40 @@
 # Look into identityMetadata for compliant ids and use them to fetch
 # descriptive metadata from Symphony.  Put the fetched value in the descMetadata
 class RefreshMetadataAction
-  # @return [NilClass,Object] returns nil if there was no resolvable metadata id.
+  include Dry::Monads[:result]
+
+  # @return [Dry::Monads::Result::Failure,Object] returns Failure if there was no resolvable metadata id, otherwise the Cocina object
   # @raises SymphonyReader::ResponseError
-  def self.run(identifiers:, fedora_object:)
-    new(identifiers: identifiers, fedora_object: fedora_object).run
+  def self.run(identifiers:, cocina_object:)
+    new(identifiers: identifiers, cocina_object: cocina_object).run
   end
 
   # @param [Array<String>] identifiers the set of identifiers that might be resolvable
-  # @param [Dor::Abstract] fedora_object to refresh
-  def initialize(identifiers:, fedora_object:)
+  # @param [Cocina::Models::DRO|Collection|APO|Agreement] cocina_object to refresh
+  def initialize(identifiers:, cocina_object:)
     @identifiers = identifiers
-    @fedora_object = fedora_object
-    @datastream = fedora_object.descMetadata
+    @cocina_object = cocina_object
   end
 
-  # Returns nil if it didn't retrieve anything
+  # @return [Dry::Monads::Results,Object] Returns Failure if it didn't retrieve anything, otherwise the Cocicna object
   # @raises SymphonyReader::ResponseError
   def run
-    content = fetch_datastream
-    return nil if content.nil?
+    metadata = fetch_metadata
+    return Failure() if metadata.nil?
 
-    datastream.dsLabel = 'Descriptive Metadata'
-    datastream.ng_xml = Nokogiri::XML(content)
-    datastream.ng_xml.normalize_text!
-    datastream.content = datastream.ng_xml.to_xml
+    description_props = Cocina::FromFedora::Descriptive.props(mods: Nokogiri::XML(fetch_metadata), druid: cocina_object.externalIdentifier)
+    return Failure() if description_props.nil?
 
-    validate
-
-    datastream
+    cocina_object.new(description: description_props)
   end
 
   private
 
-  attr_reader :identifiers, :datastream, :fedora_object
+  attr_reader :identifiers, :cocina_object
 
   # @raises SymphonyReader::ResponseError
-  def fetch_datastream
+  def fetch_metadata
     metadata_id = MetadataService.resolvable(identifiers).first
     metadata_id.nil? ? nil : MetadataService.fetch(metadata_id.to_s)
-  end
-
-  def validate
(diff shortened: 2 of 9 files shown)


app/services/refresh_metadata_action.rb
```diff
@@ -5,48 +5,34 @@
...
-    datastream.dsLabel = 'Descriptive Metadata'
-    datastream.ng_xml = Nokogiri::XML(content)
-    datastream.ng_xml.normalize_text!
-    datastream.content = datastream.ng_xml.to_xml
-
-    validate
-
-    datastream
+    updated_cocina_object = cocina_object.new(description: description_props)
+    CocinaObjectStore.save(updated_cocina_object)
```
justinlittman, line 26: I don't think we want to move save into the action.

justinlittman (changes requested):

app/services/refresh_metadata_action.rb
```diff
@@ -5,48 +5,33 @@
 class RefreshMetadataAction
   # @return [NilClass,Object] returns nil if there was no resolvable metadata id.
```
justinlittman, line 6: This is a slight refactoring, but would you consider returning a [dry-monads result](https://dry-rb.org/gems/dry-monads/1.3/result/) instead of using nil to indicate that no resolveable metadata id?

app/controllers/metadata_refresh_controller.rb
```diff
@@ -2,30 +2,25 @@
...
 
   private
 
   def identifiers
-    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
-  end
-
-  def missing_required_fields?
-    @item.descMetadata.mods_title.blank?
+    @identifiers ||= @cocina_object.identification&.catalogLinks&.collect { |clink| clink.catalogRecordId }
```
jcoyne, line 24: What about barcode?

spec/requests/metadata_refresh_spec.rb
```diff
@@ -3,69 +3,121 @@
...
 
 RSpec.describe 'Refresh metadata' do
-  let(:object) { Dor::Item.new(pid: 'druid:1234') }
+  let(:druid) { 'druid:bc753qt7345' }
+  let(:object) { Dor::Item.new(pid: druid) }
+  let(:apo_druid) { 'druid:pp000pp0000' }
+  let(:description) do
+    {
+      title: [{ value: 'However am I going to be' }],
+      purl: "https://purl.stanford.edu/#{Dor::PidUtils.remove_druid_prefix(druid)}"
```
jcoyne, line 12: Can we be certain that the MODS provided by Symphony will always be able to produce a purl? If it doesn't this descriptive won't validate.

app/services/refresh_metadata_action.rb
```diff
@@ -5,48 +5,33 @@
...
-  def self.run(identifiers:, fedora_object:)
-    new(identifiers: identifiers, fedora_object: fedora_object).run
+  def self.run(identifiers:, cocina_object:)
+    new(identifiers: identifiers, cocina_object: cocina_object).run
   end
 
   # @param [Array<String>] identifiers the set of identifiers that might be resolvable
-  # @param [Dor::Abstract] fedora_object to refresh
-  def initialize(identifiers:, fedora_object:)
+  # @param [Cocina::Models::DRO] cocina_object to refresh
```
jcoyne, line 13: It's not always a DRO.

jcoyne (changes requested):

app/services/refresh_metadata_action.rb
jcoyne: @justinlittman doesn't that make it hard to reuse this for the ObjectCreator (or whatever that evolves into?)

jcoyne: Ah, nevermind you said "don't"

spec/requests/metadata_refresh_spec.rb
jcoyne: I think this is a safe bet:
```
2.7.0 :010 > mods = MetadataService.fetch("catkey:1997249")
2.7.0 :011 > props = Cocina::FromFedora::Descriptive.props(mods: Nokogiri::XML(mods), druid: 'druid:bb000fg4892')
2.7.0 :012 > props[:purl]
 => "https://purl.stanford.edu/bb000fg4892"
 ```

File: app/controllers/metadata_refresh_controller.rb (hunk 3 above)
Comment (aaron-collier) [reply app/controllers/metadata_refresh_controller.rb]: Added barcode here. If adding the `barcode:` prefix is inaccurate, happy to address. Barcodes are unresolvable without it.

File: app/controllers/metadata_refresh_controller.rb [hunk 6]
```diff
@@ -2,30 +2,27 @@
...
 
   private
 
   def identifiers
-    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
-  end
-
-  def missing_required_fields?
-    @item.descMetadata.mods_title.blank?
+    @identifiers ||= @cocina_object.identification&.catalogLinks&.collect { |clink| clink.catalogRecordId }.tap do |id|
```
jcoyne, line 24: Should we filter out things where `clink.catalog  != 'symphony'` currently we also put previous catkeys in here too.

jcoyne (changes requested):

app/controllers/metadata_refresh_controller.rb
aaron-collier: Did we before? I can add that if necessary.

jcoyne: Oh, one other thing. Don't these need to be prefixed with `"catkey:#{clink.catalogRecordId}"`?

spec/requests/metadata_refresh_spec.rb
```diff
@@ -3,69 +3,140 @@
...
+    }
+  end
+  let(:identification) do
+    {
+      catalogLinks: [{
+        catalog: 'symphony',
+        catalogRecordId: 'catkey:10121797'
```
jcoyne, line 19: I don't think that most catkeys are formatted like this.  See https://argo.stanford.edu/view/druid:bb000fg4892
Screen Shot 2022-02-25 at 1 34 22 PM
.

app/controllers/metadata_refresh_controller.rb
```diff
@@ -2,30 +2,33 @@
...
-    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
+    @identifiers ||= @cocina_object.identification&.catalogLinks&.collect { |clink| prefixed_identifier(clink.catalogRecordId) }.tap do |id|
+      id << "barcode:#{@cocina_object.identification.barcode}" if @cocina_object.identification.barcode
+    end
   end
 
-  def missing_required_fields?
-    @item.descMetadata.mods_title.blank?
+  def prefixed_identifier(identifier)
+    return identifier if identifier.include?(':')
```
jcoyne, line 30: I don't think we have any restrictions against `:` in catkeys (e.g. 'foo:bar:123'), so I think you need to remove this.  It should always prefix with `catkey:`

app/controllers/metadata_refresh_controller.rb
```diff
@@ -2,30 +2,33 @@
...
+    CocinaObjectStore.save(updated_cocina_object)
   rescue SymphonyReader::NotFound => e
     json_api_error(status: :bad_request, title: 'Catkey not found in Symphony', message: e.message)
   end
 
   private
 
   def identifiers
-    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
+    @identifiers ||= @cocina_object.identification&.catalogLinks&.collect { |clink| prefixed_identifier(clink.catalogRecordId) }.tap do |id|
```
jcoyne, line 24: We definitely need to filter out previous catkeys from this or it can return bad data.

app/controllers/metadata_refresh_controller.rb
```diff
@@ -2,30 +2,27 @@
...
 
   private
 
   def identifiers
-    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
-  end
-
-  def missing_required_fields?
-    @item.descMetadata.mods_title.blank?
+    @identifiers ||= @cocina_object.identification&.catalogLinks&.collect { |clink| "catkey:#{clink.catalogRecordId}" }.tap do |id|
```
jcoyne, line 24: we still need to filter out previous catkeys or those will be considered the current identifiers and return bad data.

jcoyne (changes requested):

app/controllers/metadata_refresh_controller.rb
aaron-collier: shoot, i totally missed your comment above (shakes fist at github).

app/controllers/metadata_refresh_controller.rb
```diff
@@ -2,30 +2,30 @@
...
   rescue SymphonyReader::NotFound => e
     json_api_error(status: :bad_request, title: 'Catkey not found in Symphony', message: e.message)
   end
 
   private
 
   def identifiers
-    @identifiers ||= @item.identityMetadata.otherId.collect(&:to_s)
-  end
+    return nil unless @cocina_object.identification&.catalogLinks
```
jcoyne, line 24: This used to return an empty array, and now it returns nil.  Have you tested that `RefreshMetadataAction` can handle being called with nil?

app/controllers/metadata_refresh_controller.rb
```diff
@@ -2,30 +2,30 @@
...
 
-  def missing_required_fields?
-    @item.descMetadata.mods_title.blank?
+    @identifiers ||= @cocina_object.identification.catalogLinks.select { |clink| clink.catalog == 'symphony' }
```
jcoyne, line 26: This works, but `filter_map` might be an easier way to write this. nbd.

jcoyne (changes requested):

app/controllers/metadata_refresh_controller.rb
aaron-collier: Mistake, fixed.

app/controllers/metadata_refresh_controller.rb
aaron-collier: Oh yeah, I like filter_map better, thanks!

jcoyne (approved):

justinlittman (approved):

justinlittman:
Tested by Andrew.

PR #3598, opened 2022-02-24
Implement configurable find for postgres.

### LINKED ISSUE #3146: Save Cocina object to Postgres on show
Add option (controlled by feature flag) to save a Cocina object to Postgres (with CocinaObjectStore) if that object is not already saved in Postgres when ObjectsController.show is called.

Invoking show will provide a mechanism for migrating objects from Fedora to Postgres.

justinlittman:
closes #3146

Another step to ~~fedora~~.

Unit. This will require further testing when enabled.

### DIFF
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -84,12 +84,16 @@ class CocinaObjectStore
   end
 
   def find(druid)
-    fedora_to_cocina_find(druid).first
+    find_with_timestamps(druid).first
   end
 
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

jcoyne:
Why do you think this closes #3146?  Isn't that about "save to postgres" and this is about "find"?

justinlittman:
You're right, this should close #3145. (However, with the migration script #3146 isn't necessary, so going to leave closed.)

PR #3613, opened 2022-02-28
Cocinize PublishedRelationshipsFilter

### LINKED ISSUE #3608: Cocinize PublishedRelationshipsFilter

jcoyne:

Fixes #3608

### DIFF
--- a/spec/requests/metadata_spec.rb
+++ b/spec/requests/metadata_spec.rb
@@ -95,9 +95,8 @@ RSpec.describe 'Display metadata' do
     let(:now) { Time.now } # rubocop:disable Rails/TimeZone
     let(:relationships_xml) do
       <<~XML
-        <rdf:RDF xmlns:fedora-model="info:fedora/fedora-system:def/model#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:fedora="info:fedora/fedora-system:def/relations-external#" xmlns:hydra="http://projecthydra.org/ns/relations#">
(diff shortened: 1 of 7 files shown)


spec/requests/metadata_spec.rb
```diff
@@ -95,9 +95,8 @@
     let(:now) { Time.now } # rubocop:disable Rails/TimeZone
     let(:relationships_xml) do
       <<~XML
-        <rdf:RDF xmlns:fedora-model="info:fedora/fedora-system:def/model#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:fedora="info:fedora/fedora-system:def/relations-external#" xmlns:hydra="http://projecthydra.org/ns/relations#">
+        <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:fedora="info:fedora/fedora-system:def/relations-external#">
           <rdf:Description rdf:about="info:fedora/druid:ab123cd4567">
-            <fedora:isMemberOf rdf:resource="info:fedora/druid:xh235dd9059"/>
```
justinlittman, line 100: Why are we no longer mapping isMemberOf?

jcoyne: It's not used by the consumers, it's not mapped into Cocina and it's a duplicate of isMemberOfCollection.

jcoyne: I think this is a relic of when we had "set" objects

justinlittman (approved):

PR #3617, opened 2022-02-28
Return empty datastream when cocina exists in postgres.

justinlittman:
This will allow testing cocina-only objects (i.e., not saved to Fedora) without having to disable datastream editing in Argo. (It just won't be supported for cocina-only objects since Argo will receive an empty datastream list for those.)

Unit

### DIFF
--- a/app/controllers/datastreams_controller.rb
+++ b/app/controllers/datastreams_controller.rb
@@ -5,11 +5,19 @@
 # from the datastream abstraction.
 class DatastreamsController < ApplicationController
   def index
-    item = Dor.find(params[:object_id])
-    result = item.datastreams
-                 .reject { |name, instance| instance.new? || name == 'workflows' }
-                 .values
-                 .map { |ds| serialize_datastream(ds) }
+    druid = params[:object_id]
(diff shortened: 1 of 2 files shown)


app/controllers/datastreams_controller.rb
```diff
@@ -7,10 +7,14 @@ class DatastreamsController < ApplicationController
   before_action :load_item
 
   def index
-    result = @item.datastreams
-                  .reject { |name, instance| instance.new? || name == 'workflows' }
-                  .values
-                  .map { |ds| serialize_datastream(ds) }
+    result = if Settings.enabled_features.postgres.ar_find && CocinaObjectStore.new.ar_exists?(@item.pid)
```
jcoyne, line 10: Is there a reason we show the data streams if it isn't in PG?

justinlittman: It preserves the ability to edit the datastreams for objects that are stored in Fedora (e.g., for remediation).

jcoyne: Would you mind adding this as a comment here so that we don't forget the reason for this choice?

justinlittman: I don't mind. Done.

jcoyne (approved):

PR #3626, opened 2022-03-01
Move creating description from label to CocinaObjectStore

### LINKED ISSUE #3625: Move setting description when absent to CocinaObjectStore
https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/object_creator.rb#L157

justinlittman:
closes #3625

This action is required regardless of datastore.

Unit

### DIFF
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -124,6 +124,7 @@ class CocinaObjectStore
     updated_cocina_request_object = merge_access_for(cocina_request_object)
     druid = SuriService.mint_id
     updated_cocina_request_object = sync_from_symphony(updated_cocina_request_object, druid)
+    updated_cocina_request_object = add_description(updated_cocina_request_object)
     cocina_object = cocina_from_request(updated_cocina_request_object, druid)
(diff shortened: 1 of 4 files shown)


app/services/cocina_object_store.rb
```diff
@@ -349,7 +350,7 @@ def sync_from_symphony(cocina_request_object, druid)
     return cocina_request_object if catkeys.blank?
 
     result = RefreshMetadataAction.run(identifiers: catkeys, cocina_object: cocina_request_object, druid: druid)
-    return if result.failure?
+    return cocina_request_object if result.failure?
```
jmartin-sul, line 353: i don't think i understand the change here.  fix for an unrelated issue, or related to description in a way i'm not picking up on?

ndushay (approved):
again, not fully grokking, but approving.

jmartin-sul (approved):
this lgtm, a bit hesitant to merge since i'm not sure i fully grok the change, but seems fine afaict

app/services/cocina_object_store.rb
justinlittman: I'll be honest; I don't recall.

PR #3631, opened 2022-03-02
[HOLD] Do not generate slashes in the file/fileset resource ID segment

### LINKED ISSUE #3623: FileSet identifiers with slashes in them break IIIF manifests
This is a regression introduced in https://github.com/sul-dlss/dor-services-app/pull/3522

The extra slash cause the canvas in the IIIF manifest to be: https://sul-purl-stage.stanford.edu/fm539yc9854/iiif/canvas/cocina-fileSet-fm539yc9854/5b074b49-c431-4704-84b4-c7d61505bdc9
Which doesn’t resolve due to the addition of the slash.
Compare that to https://purl.stanford.edu/cb587tt9501/iiif/canvas/cb587tt9501_1 a canvas within https://purl.stanford.edu/cb587tt9501/iiif/manifest

mjgiarlo:

Fixes #3623

This broke IIIF links on the Access side.

CI + will be tested in stage where there are Access VMs integrated.

### DIFF
--- a/spec/fixtures/hj097bm8879_publicObject.xml
+++ b/spec/fixtures/hj097bm8879_publicObject.xml
@@ -12,14 +12,14 @@
 <release to="Searchworks">true</release>
 </identityMetadata>
 <contentMetadata objectId="druid:hj097bm8879" type="map">
-  <resource id="cocina-fileSet-hj097bm8879/9475bc2c-7552-43d8-b8ab-8cd2212d5873" sequence="1" type="image">
+  <resource id="cocina-fileSet-hj097bm8879-9475bc2c-7552-43d8-b8ab-8cd2212d5873" sequence="1" type="image">
     <label>(Covers to) Carey's American Atlas: Containing Twenty Maps And One Chart ... Philadelphia: Engraved For, And Published By, Mathew Carey, No. 118, Market Street. M.DCC.XCV. [Price, Plain, Five Dollars-Coloured, Six Dollars.]</label>
     <externalFile fileId="2542A.jp2" objectId="druid:cg767mn6478" resourceId="cocina-fileSet-9475bc2c-7552-43d8-b8ab-8cd2212d5873" mimetype="image/jp2">
       <imageData width="6475" height="4747"/>
     </externalFile>
     <relationship objectId="druid:cg767mn6478" type="alsoAvailableAs"/>
   </resource>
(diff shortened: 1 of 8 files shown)


spec/fixtures/hj097bm8879_publicObject.xml
```diff
@@ -12,14 +12,14 @@
...
-  <resource id="cocina-fileSet-hj097bm8879/9475bc2c-7552-43d8-b8ab-8cd2212d5873" sequence="1" type="image">
+  <resource id="cocina-fileSet-hj097bm8879-9475bc2c-7552-43d8-b8ab-8cd2212d5873" sequence="1" type="image">
     <label>(Covers to) Carey's American Atlas: Containing Twenty Maps And One Chart ... Philadelphia: Engraved For, And Published By, Mathew Carey, No. 118, Market Street. M.DCC.XCV. [Price, Plain, Five Dollars-Coloured, Six Dollars.]</label>
     <externalFile fileId="2542A.jp2" objectId="druid:cg767mn6478" resourceId="cocina-fileSet-9475bc2c-7552-43d8-b8ab-8cd2212d5873" mimetype="image/jp2">
       <imageData width="6475" height="4747"/>
     </externalFile>
     <relationship objectId="druid:cg767mn6478" type="alsoAvailableAs"/>
   </resource>
-  <resource id="cocina-fileSet-hj097bm8879/929604b0-00bc-40b1-af71-5c17f066e2fd" sequence="2" type="image">
+  <resource id="cocina-fileSet-hj097bm8879-929604b0-00bc-40b1-af71-5c17f066e2fd" sequence="2" type="image">
```
mjgiarlo, line 22: Calling out here that I needed to explicitly edit fixture data to make things work. I suspect this is because we generated fixture data before we put in this fix, but I'm bringing attention to it in case I'm mistaken.

andrewjbtw:
Since this affects the id generated behind the scenes, does that mean we need to re-generate resource ids - as in open a version, force id regeneration, close the version - to remediate all affected items accessioned since the change in ids was made nearly two weeks ago?

andrewjbtw:
Now that I'm back on my work machine and able to access stage, I see that I misunderstood this change. Only the published ids are changing and the stored ids remain the same. That looks good to me. We'll need to republish items accessioned in the past couple of weeks to fix remediate the existing data.

PR #3632, opened 2022-03-02
Add a SolrService rather than relying on the one from ActiveFedora

### LINKED ISSUE #3624: Add ActiveFedora::SolrService
ActiveFedora::SolrService is a wrapper around RSolr. It currently resides in ActiveFedora: https://github.com/samvera/active_fedora/blob/main/lib/active_fedora/solr_service.rb

The code for the service should be copied into DSA (dropping ActiveFedora) and all references to ActiveFedora::SolrService should be changed to use the new service.

jcoyne:
Fixes #3624


config/settings.yml
```diff
@@ -52,6 +52,7 @@ version_service:
 # URLs
 fedora_url: 'https://user:password@fedora.example.com:1000/fedora'
 solr:
+  select_path: 'select'
```
justinlittman, line 55: Will we ever configure a different value?

justinlittman: (In other words, does this really need to be configurable?)

app/services/solr_service.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+# Functions for querying solr
+class SolrService
+  include Singleton
+
+  def options
+    { timeout: 120, url: Settings.solr.url }
```
justinlittman, line 8: Should timeout be configurable.

justinlittman (approved):

config/settings.yml
jcoyne: I think yes, because it depends on how you have solr configured. See https://solr.apache.org/guide/6_6/requestdispatcher-in-solrconfig.html and https://github.com/sul-dlss/sul-solr-configs/blob/master/argo3_prod/solrconfig.xml#L136

PR #3652, opened 2022-03-02
Port SuriService from dor-services gem

### LINKED ISSUE #3636: Import SuriService from dor-services

mjgiarlo:

Fixes #3636

And, in so doing, tear out a bunch of functionality that is not even supported in the Suri service anymore.

CI, and successfully integration-tested in QA

### DIFF
--- a/app/services/suri_service.rb
+++ b/app/services/suri_service.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Use the SURI service to mint an identifier
+class SuriService
+  class MalformedDruidError < RuntimeError; end
+
+  def self.mint_id
+    # POST with no body
+    response = Faraday.post("#{Settings.suri.url}/identifiers")
+    druid = "druid:#{response.body}"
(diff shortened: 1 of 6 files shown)


jcoyne (approved):

app/services/suri_service.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SuriService
```
jcoyne, line 3: Rubocop and I would appreciate some class level docs here.

mjgiarlo: @jcoyne Oh, right, I forgot about our friend Rubocop.

mjgiarlo: Got it now. Thanks!

mjgiarlo:
@jcoyne Can you have another look? I made a couple small tweaks and thought you should 👁️ again before we merge this.

jcoyne (approved):

PR #3653, opened 2022-03-02
replace Dor::SearchService with SolrService

### LINKED ISSUE #3639: Stop using Dor::SearchService
Depends on #3632

peetucket:

Fixes #3639 - replace Dor::SearchService with new SolrService

Updated existing tests 

### DIFF
--- a/app/services/embargo_release_service.rb
+++ b/app/services/embargo_release_service.rb
@@ -7,32 +7,6 @@
 class EmbargoReleaseService
   RELEASEABLE_NOW_QUERY = 'embargo_status_ssim:"embargoed" AND embargo_release_dtsim:[* TO NOW]'
 
-  # Finds druids from solr based on the passed in query
-  # It will then load each item from Dor, and call the block with the item
-  # @param [String] query used to locate druids of items to release from solr
-  # @yield [Cocina::Models::DRO] gets executed after loading the object from DOR and opening new version
-  #  Steps needed to release the particular embargo from the item
-  def self.release_items(query, &release_block)
-    # Find objects to process
-    Rails.logger.info("***** Querying solr: #{query}")
-    solr = SolrService.get(query, 'rows' => '5000', 'fl' => 'id')
-
-    num_found = solr['response']['numFound'].to_i
-    if num_found.zero?
-      Rails.logger.info('No objects to process')
-      return
-    end
-    Rails.logger.info("Found #{num_found} objects")
-
-    count = 0
-    solr['response']['docs'].each do |doc|
-      release_item(doc['id'], &release_block)
-      count += 1
-    end
-
-    Rails.logger.info("Done! Processed #{count} objects out of #{num_found}")
-  end
-
   def self.release_all
     # Find objects to process
     Rails.logger.info("***** Querying solr: #{RELEASEABLE_NOW_QUERY}")
--- a/app/services/embargo_release_service.rb
+++ b/app/services/embargo_release_service.rb
@@ -15,7 +15,7 @@ class EmbargoReleaseService
   def self.release_items(query, &release_block)
     # Find objects to process
     Rails.logger.info("***** Querying solr: #{query}")
-    solr = Dor::SearchService.query(query, 'rows' => '5000', 'fl' => 'id')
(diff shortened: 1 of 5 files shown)


app/validators/cocina/unique_source_id_validator.rb
```diff
@@ -35,7 +35,11 @@ def duplicate_druid
     end
 
     def lookup_duplicate
-      Dor::SearchService.query_by_id(cocina_dro.identification.sourceId).first
+      solr_response = SolrService.get("identifier_tesim:\"#{@cocina_dro.identification.sourceId}\"", rows: 100, fl: 'id')
+
+      return if solr_response['response']['numFound'].zero?
+
+      solr_response['response']['docs'].map { |doc| doc['id'] }
     end
```
peetucket, line 43: the `query_by_id` method in dor-services (https://github.com/sul-dlss/dor-services/blob/main/lib/dor/services/search_service.rb#L29-L43) was more generalized and complicated ... but we only call it in this one place, so we can simplify a bit (basically, lookup by the single supplied source ID and return either an array of druids, or nil if none found)

spec/requests/create_object_spec.rb
```diff
@@ -27,14 +27,16 @@
...
+  let(:search_result) { no_result }
 
   before do
     allow(Dor::SuriService).to receive(:mint_id).and_return(druid)
     allow_any_instance_of(CocinaObjectStore).to receive(:find).with('druid:dd999df4567').and_return(apo)
     allow(Cocina::ActiveFedoraPersister).to receive(:store)
     stub_request(:post, 'https://dor-indexing-app.example.edu/dor/reindex/druid:gg777gg7777')
-    allow(Dor::SearchService).to receive(:query_by_id).and_return(search_result)
+    allow(SolrService).to receive(:get).and_return(search_result)
   end
```
peetucket, line 40: we now mock a solr response (see new lines 30-32 above) so that our updated `lookup_duplicate` method works as expected

mjgiarlo (approved):

app/services/embargo_release_service.rb
```diff
@@ -15,7 +15,7 @@ class EmbargoReleaseService
   def self.release_items(query, &release_block)
     # Find objects to process
     Rails.logger.info("***** Querying solr: #{query}")
-    solr = Dor::SearchService.query(query, 'rows' => '5000', 'fl' => 'id')
+    solr = SolrService.get(query, 'rows' => '5000', 'fl' => 'id')
```
mjgiarlo, line 18: This line is the source of the coverage ding. Not sure if it's bogus or legit.

peetucket: Yeah,  interesting -- I don't see this method being called from anywhere in the DSA codebase (it's also not tested).  I wonder if the whole method can just be removed?

mjgiarlo: @peetucket looks like dead code to me too. KILL IT

mjgiarlo: @peetucket This became dead code five days ago: https://github.com/sul-dlss/dor-services-app/commit/29f9d8907f729d43f36ac4858cbd8a6b092a45fc#diff-a5e0fdbbca23f671b0fba6425be8786707a402115eab792a3b8bcfa3ad449fbcL67

peetucket: Cool.  I removed it.

app/services/embargo_release_service.rb
```diff
@@ -7,36 +7,10 @@
...
-    count = 0
-    solr['response']['docs'].each do |doc|
-      release_item(doc['id'], &release_block)
-      count += 1
-    end
-
-    Rails.logger.info("Done! Processed #{count} objects out of #{num_found}")
-  end
-
   def self.release_all
```
peetucket, line 10: dead code - not called anywhere anymore

app/validators/cocina/unique_source_id_validator.rb
jcoyne: It was doing a term query there, but this logic would tokenize and match on the tokens which might match more than you expect. I think a term query here is the right choice.

jcoyne (changes requested):
Please use a term query when finding by identifier

PR #3658, opened 2022-03-03
Don't map project tag to partOfProject 

jcoyne:
It will be removed by the next cocina models release. See https://github.com/sul-dlss/cocina-models/pull/328


justinlittman:
No longer necessary.

PR #3672, opened 2022-03-07
Adds uniqueness constraint for sourceId for DROs and collections.

### LINKED ISSUE #3665: Add a unique constraint on sourceId
Now that we've switched to a postgres data store, add a constraint for sourceId.  This will avoid the problem where two registrations at the nearly the same time are able to use the same sourceId.

justinlittman:
closes #3665

This approach is more reliable then querying Solr (since updating Solr involves a latency).

Unit


jcoyne (changes requested):
Won't this have to be unique on `cocina_version, sourceId`?  Each version of an object should have the same sourceId, right?

justinlittman:
There is only ever a single version per object in the db, so I don't think so.

jcoyne (approved):

PR #3689, opened 2022-03-08
Move location of timestamp in migrate log.

justinlittman:
Per @ndushay suggestion.


ndushay (approved):
Thank you!

PR #3701, opened 2022-03-09
indicate which way version is mismatched

### LINKED ISSUE #3693: Update with version mismatch provides a confusing error
If you have an object with a version of 1
and then send a new version open request (version is now 2)
and then send an update with the original version (1),
Then this error is raised: https://app.honeybadger.io/projects/50568/faults/84382624

This occurs in this code: https://github.com/sul-dlss/dor-services-app/blob/09b8470101bec4b4e992cf3066b69f505a32a096/app/services/cocina/object_updater.rb#L178-L184

The code is assuming that any version difference is an increment on the provided item.  If the code checked to see if the version is less than the expected version (`fedora_object.current_version > cocina_object.version`), then it could raise a more helpful error (e.g. "It looks like you're trying to update an old version, this is not permitted").

This entire situation can be prevented by #3664 (and https://github.com/sul-dlss/dor-services-client/pull/253, and actually using this feature in all the clients)

peetucket:

Fixes #3693 - provide better error message indicating which way the version mismatches (is that what you were thinking Jcoyne?)

Added a new test

### DIFF
--- a/app/services/cocina/object_updater.rb
+++ b/app/services/cocina/object_updater.rb
@@ -180,7 +180,13 @@ module Cocina
 
       fedora_object.versionMetadata.increment_version
 
-      raise "Incremented version of #{fedora_object.current_version} is not expected version #{cocina_object.version}" unless version_match?
+      return if version_match?
+
+      if fedora_object.current_version.to_i > cocina_object.version
+        raise "Incremented version of #{fedora_object.current_version} is greater than expected version #{cocina_object.version} - it appears you are trying to update an old version."
+      end
+
+      raise "Incremented version of #{fedora_object.current_version} is less than the expected version #{cocina_object.version}"
     end
 
     def version_match?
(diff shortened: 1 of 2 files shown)


app/services/cocina/object_updater.rb
```diff
@@ -180,7 +180,13 @@ def update_version
 
       fedora_object.versionMetadata.increment_version
 
-      raise "Incremented version of #{fedora_object.current_version} is not expected version #{cocina_object.version}" unless version_match?
+      if fedora_object.current_version.to_i > cocina_object.version
```
jcoyne, line 183: I think it might be easier to follow the logic here if we have `return if version_match?` (happy path). Then do the comparison where we raise the appropriate message.

jcoyne (changes requested):

app/services/cocina/object_updater.rb
peetucket: ok - changed the logic a bit (i originally had it as a conditional only with no guard clause, but rubocop didn't like that)

PR #3704, opened 2022-03-10
Remove unnecessary exception

jcoyne:
We don't need this code.


justinlittman (approved):

justinlittman:
Thanks. That one has been bothering me.

PR #3713, opened 2022-03-11
Allow existing virtual objects to be changed.

### LINKED ISSUE #3675: allow the first druid in a virtual object CSV to be a virtual object
The "Create virtual objects" job checks whether any druids in the CSV are virtual objects. This check should be limited to constituent objects. It should be permissible for the virtual object itself - in the first column of the CSV - to already be a virtual object. 

Currently, the "Create virtual objects" job will fail if any of the druids are already virtual objects, not just the constituents.

**Impact**

It is not possible to edit the membership of an existing virtual object without having to remove all content in order to turn the virtual object back into a regular item.

**Expected behavior**

Submitting a CSV for an existing virtual object should:

- reorder the virtual object to match what's in the CSV
- remove any constituents not listed in the new CSV. They'll remain items but should be un-linked to the virtual object.

lwrubel:
Fixes #3675 to allow a virtual object to be recreated via CSV upload.

Ran the create virtual objects integration test on QA. Loaded a CSV with a current virtual object druid and only one of its constituents. Then the job completed, only the specified constituent was a member. The other was no longer attached. Also, specs passed and added to them.  

### DIFF
--- a/app/services/item_query_service.rb
+++ b/app/services/item_query_service.rb
@@ -14,8 +14,19 @@ class ItemQueryService
 
     errors[virtual_object] << "Item #{virtual_object} cannot be a constituent of itself" if constituents.include?(virtual_object)
 
-    ([virtual_object] + constituents).each do |druid|
-      find_combinable_item(druid)
+    # check virtual_object for combinability
+    begin
+      find_combinable_item(virtual_object)
+    rescue UncombinableItemError => e
+      errors[virtual_object] << e.message
+    end
+
+    # check constituents for combinability and whether they are already virtual objects
+    begin
+      constituents.each do |druid|
+        check_virtual(druid)
+        find_combinable_item(druid)
+      end
     rescue UncombinableItemError => e
       errors[virtual_object] << e.message
     end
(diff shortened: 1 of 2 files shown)


app/services/item_query_service.rb
```diff
@@ -14,8 +14,19 @@ def self.validate_combinable_items(virtual_object:, constituents:)
...
+
+    # check constituents for combinability and whether they are already virtual objects
+    begin
+      constituents.each do |druid|
+        check_virtual(druid)
+        find_combinable_item(druid)
+      end
     rescue UncombinableItemError => e
       errors[virtual_object] << e.message
     end
```
mjgiarlo, line 32: I could be mistaken, but I believe having the `rescue` outside of the `#each` loop means that the first exception will terminate the loop. Current behavior I believe is to loop over all constituents to see all the errors for a given batch of druids. So I might suggest something like:

```suggestion
# check constituents for combinability and whether they are already virtual objects
constituents.each do |druid|
  check_virtual(druid)
  find_combinable_item(druid)
rescue UncombinableItemError => e
  errors[virtual_object] << e.message
end
```

mjgiarlo: (This is a block-level `rescue` and does not require an explicit `begin` statement.)

mjgiarlo: And now that I'm typing this out, it's suboptimal that this was the prior behavior and we weren't testing it. If not too onerous, this might be a test worth adding (namely, that two or more non-combinables in a batch result in the same number of errors in the error hash). If you feel that's better for follow-up work, that's cool too.

lwrubel: I completely missed your comments for a few days. Thanks for catching these issues. Have added fixed the rescue statement placement and added a test. I think it makes sense to include it in this work.

mjgiarlo (approved):

PR #3719, opened 2022-03-14
Use cocina-models RSpec matchers

mjgiarlo:

This commit switches DSA so that it uses the RSpec matchers provided by the cocina-models gem.

CI, only affects testing


edsu (approved):
Yay, less code!

aaron-collier (approved):

PR #3728, opened 2022-03-18
Streamline identifier spec to be more consistent with other specs

arcadiafalcone:
Make descriptive metadata identifiers work like everything else.


ndushay:
um ... is this a priority descmd thingy?  is there a ticket to do the code changes?  @justinlittman

ndushay (approved):
I'm all for simpler;  wondering if @justinlittman should be in the loop.

PR #3732, opened 2022-03-21
Adds migration audit script.

### LINKED ISSUE #3712: Develop migration audit strategy
@andrewjbtw would like to be able to verify that after migration, the count of objects stored in postgres should equal the expected count.

This is tricky because:
* Project Phoenix objects are not being migrated.
* Some other objects are not being migrated. See https://github.com/sul-dlss/dor-services-app/blob/main/app/services/migration_filter.rb
* New objects are being added in the midst of migration.

One approach would be to develop a script that given a list of druids (that excludes Project Phoenix), for objects that are being migrated, verify that the object exists in postgres. For reference, see the existing migration script.

justinlittman:
closes #3712

Per @andrewjbtw allow verifying that all objects were migration from Fedora to PG.

On DSA migrate:
```
[dor_services@dor-services-migrate current]$ RAILS_ENV=production bin/migrate-audit -i ~/druids.20220219.txt -s 250000
Provided druids: 250000
Found: 246283
Missing: 3717
```

### DIFF
--- a/bin/migrate-audit
+++ b/bin/migrate-audit
@@ -1,0 +1,81 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Audits migration by determining if all druids are in the database.
+
+require_relative '../config/environment'
+require_relative '../lib/fedora_migrator'
+require 'optparse'
+require 'parallel'
+require 'tty-progressbar'
+
+options = { druids: [], input: 'druids.txt', processes: 4 }
+parser = OptionParser.new do |option_parser|
+  option_parser.banner = 'Usage: bin/migrate-audit [options]'
+
+  option_parser.on('-sSAMPLE', '--sample SAMPLE', Integer, 'Sample size, otherwise all druids.')
+  option_parser.on('-pPROCESSES', '--processes PROCESSES', Integer, "Number of processes. Default is #{options[:processes]}.")
+  option_parser.on('-dDRUIDS', '--druids DRUIDS', Array, 'List of druids (instead of druids.txt).')
+  option_parser.on('-iFILENAME', '--input FILENAME', String, 'File containing list of druids (instead of druids.txt).')
+  option_parser.on('-h', '--help', 'Displays help.') do
+    puts option_parser
+    exit
+  end
+end
+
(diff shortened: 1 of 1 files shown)


bin/migrate-audit
```diff
@@ -0,0 +1,81 @@
...
+
+results = []
+File.open("log/migration_results/#{results_file_name}", 'w') do |file|
+  progress_bar = tty_progress_bar(druids.size)
+  progress_bar.start
+  results = Parallel.map(druids,
+                         in_processes: options[:processes],
+                         finish: ->(druid, _, result) { on_finish(druid, result, progress_bar, file) }) do |druid|
+    cocina_object_store = CocinaObjectStore.new
+    cocina_object_store.ar_exists?(druid)
```
justinlittman, line 73: This is the crux: looking the object up in postgres.

jcoyne: This is fine, but I think you could do something 1000x times faster using raw SQL.  e.g. `SELECT druid FROM objects WHERE druid IN ('druid:123', 'druid:456', 'druid:783', 'druid:2932')` and comparing the number of rows returned with the number of ids passed in.

jcoyne (approved):

edsu (approved):

bin/migrate-audit
```diff
@@ -0,0 +1,81 @@
...
+# rubocop:enable Rails/TimeZone
+
+results = []
+File.open("log/migration_results/#{results_file_name}", 'w') do |file|
+  progress_bar = tty_progress_bar(druids.size)
+  progress_bar.start
+  results = Parallel.map(druids,
+                         in_processes: options[:processes],
+                         finish: ->(druid, _, result) { on_finish(druid, result, progress_bar, file) }) do |druid|
+    cocina_object_store = CocinaObjectStore.new
```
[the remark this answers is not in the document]
justinlittman: Minimal, and since this is run in a separate process, this avoids problems with shared variables.

bin/migrate-audit
justinlittman: Agree, but for the limited use of this not really worth additional optimization.

PR #3754, opened 2022-03-29
New required fields for collections and dros

justinlittman:
Since the DB records have been corrected in QA, it is no longer necessary to handle nulls in the code.

Unit


jcoyne (approved):
👏

PR #3774, opened 2022-03-31
Use standard ETag header

### LINKED ISSUE #3772: ETag headers are not standard.
It's `ETag` not `Etag`.  See https://datatracker.ietf.org/doc/html/rfc7232#section-2.3

https://github.com/sul-dlss/dor-services-app/blob/a2272bb71c3cdc8d20adadb13a479f8d24e65f5e/app/controllers/objects_controller.rb#L219

I'm not sure how this test is passing: https://github.com/sul-dlss/dor-services-app/blob/a2272bb71c3cdc8d20adadb13a479f8d24e65f5e/spec/requests/update_metadata_spec.rb#L153

jcoyne:

Fixes #3772

Test suite


edsu (approved):
Consistency is a good thing!

PR #3775, opened 2022-03-31
Make the ETag value opaque.

jcoyne:
The RFC (https://datatracker.ietf.org/doc/html/rfc7232#section-2.3) specifies opqaue ETags, so opaqify them.


edsu (approved):
An md5 hash is pretty opaque, and there's basically zero risk of collisions in this use case.

PR #3790, opened 2022-04-01
Fix bug with pruning Access druids

mjgiarlo:

Do not delete the parent when pruning Access druids, since these druids lack the extra leaf (ID) directory that non-Access druids have. This is causing unintentional pruning of the Stacks.

CI

### DIFF
--- a/spec/services/prune_service_spec.rb
+++ b/spec/services/prune_service_spec.rb
@@ -5,18 +5,48 @@ require 'rails_helper'
 RSpec.describe PruneService do
   let(:druid_str) { 'druid:cd456ef7890' }
   let(:strictly_valid_druid_str) { 'druid:cd456gh1234' }
+  let(:access_druid_str) { 'druid:cd456ef9999' }
 
   describe '#prune!' do
     let(:workspace) { Dir.mktmpdir }
     let(:dr1) { DruidTools::Druid.new(druid_str, workspace) }
     let(:dr2) { DruidTools::Druid.new(strictly_valid_druid_str, workspace) }
-    let(:pathname1) { dr1.pathname }
 
     after do
       FileUtils.remove_entry workspace
     end
 
+    context 'with an access druid sharing the first three path segments' do
+      let(:dr3) { DruidTools::StacksDruid.new(access_druid_str, workspace) }
+
+      before do
+        # Nil the create records for this context because we're in a known read only one
+        dr1.mkdir
+        dr2.mkdir
+        dr3.mkdir
+        described_class.new(druid: dr3).prune!
+      end
+
+      it 'deletes the outermost directory' do
+        expect(File).not_to exist(dr3.pruning_base)
+      end
+
+      it 'does not delete unrelated ancestor directories' do
+        expect(File).to exist(dr1.pruning_base)
+        expect(File).to exist(dr1.pruning_base.parent)
+      end
+
+      it 'stops at ancestor directories that have children' do
+        # 'cd/456/ef' should still exist because of dr1
+        shared_ancestor = dr1.pruning_base.parent
+        expect(shared_ancestor.to_s).to match(%r{cd/456/ef$})
+        expect(File).to exist(shared_ancestor)
+      end
+    end
+
     context 'when there is a shared ancestor' do
+      let(:pathname1) { dr1.pathname }
+
       before do
         # Nil the create records for this context because we're in a known read only one
         dr1.mkdir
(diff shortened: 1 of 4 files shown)


spec/services/prune_service_spec.rb
```diff
@@ -5,17 +5,47 @@
...
+      end
+
+      it 'deletes the outermost directory' do
+        expect(File).not_to exist(dr3.path)
+      end
+
+      it 'does not delete empty ancestor directories' do
+        expect(File).to exist(pathname1.parent)
+        expect(File).to exist(pathname1.parent.parent)
+      end
```
mjgiarlo, line 39: This behavior is different from non-Access druids. Can someone confirm this is right?

spec/services/prune_service_spec.rb
```diff
@@ -5,17 +5,47 @@
...
+
+      it 'stops at ancestor directories that have children' do
+        # 'cd/456/ef' should still exist because of dr1
+        shared_ancestor = pathname1.parent.parent
+        expect(shared_ancestor.to_s).to match(%r{cd/456/ef$})
+        expect(File).to exist(shared_ancestor)
+      end
```
mjgiarlo, line 46: This is the test that failed before I put in the patch.

jmartin-sul (approved):

jmartin-sul:
just going through this explicitly to confirm my own understanding of how things fit together:
* [`PruneService`](https://github.com/sul-dlss/dor-services-app/blob/2d741afde56ad244ad9b3dbd5a11312ead46be08/app/services/prune_service.rb) (fixed by this PR) is called directly by [`DeleteService`](https://github.com/sul-dlss/dor-services-app/blob/2d741afde56ad244ad9b3dbd5a11312ead46be08/app/services/delete_service.rb#L35-L43) and also indirectly via [`CleanupService`](https://github.com/sul-dlss/dor-services-app/blob/2d741afde56ad244ad9b3dbd5a11312ead46be08/app/services/cleanup_service.rb#L17-L19), which is called by [`DeleteService`](https://github.com/sul-dlss/dor-services-app/blob/2d741afde56ad244ad9b3dbd5a11312ead46be08/app/services/delete_service.rb#L23-L29).  And `DeleteService` is called by [`ObjectsController#destroy`](https://github.com/sul-dlss/dor-services-app/blob/2d741afde56ad244ad9b3dbd5a11312ead46be08/app/controllers/objects_controller.rb#L167) is called by Argo's "purge" commands (via bulk action or item show page).
* `PruneService.prune!` is also called by [`MetadataTransferService#unpublish`](https://github.com/sul-dlss/dor-services-app/blob/fc66e987742001ab4bccce52bf0c5aa1f8d0f6af/app/services/publish/metadata_transfer_service.rb#L42-L45) using a `PurlDruid` (which, like `StacksDruid`, is an alias of [`AccessDruid`](https://github.com/sul-dlss/druid-tools/blob/59571a3676e14efc42778e7fbd4d3101a760151f/lib/druid_tools/access_druid.rb#L20-L21)).

So, it seems like purge was how @andrewjbtw got a lead on the misbehavior @mjgiarlo fixed here, but maybe unpublish could be causing trouble too?

jmartin-sul:
i noticed that code climate seems to think that the meat of the `DeleteService` logic is uncovered.  if that's a valid assessment, it makes me think that might be a good place to add some integration-y test coverage?  but i would vote for doing that separately from getting this fix merged.

Screen Shot 2022-03-31 at 6 15 19 PM

andrewjbtw:
I was suspicious of unpublish too but hadn't investigated that since it's hard to track. When we created it, I asked (maybe only in conversation) that it apply only to the Purl, not Stacks. If it's deleting files, I can file a new issue around that.

mjgiarlo:
@andrewjbtw 💬 
> I was suspicious of unpublish too but hadn't investigated that since it's hard to track. When we created it, I asked (maybe only in conversation) that it apply only to the Purl, not Stacks. If it's deleting files, I can file a new issue around that.

AFAICT, we have a couple different unpublish functions in DSA, unfortunately. 

Neither the publish service nor the publish job clean up files---they only instruct purl-fetcher to destroy its database records related to the given druid. 

The metadata transfer service on the other hand does trigger a stacks clean-up: https://github.com/sul-dlss/dor-services-app/blob/fc66e987742001ab4bccce52bf0c5aa1f8d0f6af/app/services/publish/metadata_transfer_service.rb#L42-L45

ndushay (approved):

ndushay (approved):

jmartin-sul (approved):

mjgiarlo:
Unholding per @andrewjbtw. We want this merged today but not deployed until Monday as part of dep updates. Merging.

PR #3800, opened 2022-04-01
Remove dor-services-client.

justinlittman:
To make it easier to test cocina models changes without cutting a new release.

local, sdr-deploy

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,6 @@ gem 'uuidtools', '~> 2.1.4'
 gem 'whenever', require: false
 
 group :development do
-  gem 'dor-services-client' # used by lib/fedora_cache.rb and lib/fedora_loader.rb
   gem 'listen', '~> 3.0.5'
   gem 'marc-vocab' # used by bin/reports/report-desc-marcgac
   gem 'rubyzip', '>= 1.0.0'
--- a/lib/fedora_cache.rb
+++ b/lib/fedora_cache.rb
@@ -157,6 +157,6 @@ class FedoraCache
   end
 
   def fetch_tags(druid)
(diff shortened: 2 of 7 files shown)


lib/fedora_cache.rb
```diff
@@ -157,6 +157,6 @@ def apo_for(rels_ext)
   end
 
   def fetch_tags(druid)
-    Dor::Services::Client.object(druid).administrative_tags.list
+    AdministrativeTags.for(identifier: druid)
```
mjgiarlo, line 160: Hmm, I'm curious why you use the service class here and use the HTTP API in `lib/fedora_leader.rb`

Gemfile
```diff
@@ -46,7 +46,6 @@ gem 'uuidtools', '~> 2.1.4'
 gem 'whenever', require: false
 
 group :development do
-  gem 'dor-services-client' # used by lib/fedora_cache.rb and lib/fedora_loader.rb
```
ndushay, line 49: I was always a bit uncomfy with the app requiring its own client

lib/fedora_cache.rb
justinlittman: This invokes the implementation in `lib/fedora_loader.rb` (which is a monkeypatch of `AdministrativeTags`).

ndushay (approved):
I'm good with this as long as it was tested on the bin/*** scripts.

Gemfile
justinlittman: It was a cheat.

lib/fedora_cache.rb
mjgiarlo: Aha! Right, now I see. Perfect.

PR #3808, opened 2022-04-04
Add report for single-entry parallelValues

### LINKED ISSUE #3795: Report on parallelValue in Cocina
Report to identify objects where the value array for the parallelValue property contains fewer than 2 entries. Desired output is druid and the JSON Path to the property.

mjgiarlo:

Fixes #3795

To accomplish this, modify the Report class so it can map Fedora content to Cocina and report on that.

CI and tested the new report and an existing report on sdr-deploy


edsu (approved):
🆒 to be able to get the Cocina!

PR #3809, opened 2022-04-04
TitleBuilder should be called with an array of titles

jcoyne:
Otherwise it raises a deprecation warning about the argument provided


ndushay (approved):
Are we sure this is the only place this needs a fix?

jcoyne:
This fixed the only deprecation emitted during testing. 🤷

ndushay:
I did a search in github for org:sul-dlss and didn't find any more.

PR #3816, opened 2022-04-05
Report on invalid EDTF in MODS dates

### LINKED ISSUE #3797: Report on invalid and complex edtf dates in MODS
This report should identify any date value (dateCreated, dateIssued, dateCaptured, dateValid, dateModified, copyrightDate, dateOther) that has an encoding attribute with the value edtf but does not match one of the patterns below. The desired output is the druid and the date value.

Same as w3cdtf:
Year:
YYYY (eg 1997)
Year and month:
YYYY-MM (eg 1997-07)
Complete date:
YYYY-MM-DD (eg 1997-07-16)
Complete date plus hours and minutes:
YYYY-MM-DDThh:mmTZD (eg 1997-07-16T19:20+01:00)
Complete date plus hours, minutes and seconds:
YYYY-MM-DDThh:mm:ssTZD (eg 1997-07-16T19:20:30+01:00)
Complete date plus hours, minutes, seconds and a decimal fraction of a
second
YYYY-MM-DDThh:mm:ss.sTZD (eg 1997-07-16T19:20:30.45+01:00)

Additional
-YYYY (eg -3999)

edsu:

As part of migration we needed a report for what look like invalid EDTF values in MODS records.

Closes #3797

Ran the report on sdr-deply against the full set of druids.txt.

### DIFF
--- a/bin/reports/report-desc-edtf
+++ b/bin/reports/report-desc-edtf
@@ -24,18 +24,10 @@ formats = [
 ]
 
 Report.new(name: 'desc-edtf', dsids: ['descMetadata']).run do |ng_xml|
-  errors = []
+union_xpath_selector = date_els.map { |el| "//mods:#{el}[@encoding='edtf']" }.join(' | ')
+errors = ng_xml.xpath(union_xpath_selector, mods: MODS_NS).filter_map do |el|
+  el.content unless formats.any? { |format| format.match?(el.content) }
+end
 
-  date_els.each do |el_name|
-    ng_xml.xpath("//mods:#{el_name}[@encoding='edtf']", mods: MODS_NS).each do |el|
-      # at least one of the formats must match or else we've found an error
-      errors.push(el.content) unless formats.map { |format| format.match? el.content }.any?
-    end
-  end
-
-  if errors.empty?
-    false
-  else
-    errors.join(';')
-  end
+errors.join(';').presence
 end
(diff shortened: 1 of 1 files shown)


bin/reports/report-desc-edtf
```diff
@@ -0,0 +1,41 @@
...
+      # at least one of the formats must match or else we've found an error
+      errors.push(el.content) unless formats.map { |format| format.match? el.content }.any?
+    end
+  end
+
+  if errors.empty?
+    false
+  else
+    errors.join(';')
+  end
```
mjgiarlo, line 40: I'm thinking about performance vs. readability, and wondering whether considering something like this is an improvement or not, or maybe even a wash?

```suggestion
union_xpath_selector = date_els.map { |el| "//mods:#{el}[@encoding='edtf']" }.join(' | ')
errors = ng_xml.xpath(union_xpath_selector, mods: MODS_NS).filter_map do |el|
  el.content unless formats.any? { |format| format.match?(el.content) }
end

errors.join(';').presence
```

edsu: I'm sure that's faster -- I'm all for saving some electricity.

edsu: PS. I really like what you did here:

```ruby
el.content unless formats.any? { |format| format.match?(el.content) }
```

I didn't know that `any?` could take a proc! It ready a lot better that way.

mjgiarlo (approved):
Rubocop is grumpy but this looks good to me once Rubocop is pleased. Thanks for considering my suggestion!

PR #3833, opened 2022-04-07
Add spec for multipart duplicate name

arcadiafalcone:
Handle pattern found in data

### DIFF
--- a/spec/services/cocina/mapping/descriptive/mods/name_spec.rb
+++ b/spec/services/cocina/mapping/descriptive/mods/name_spec.rb
@@ -2337,6 +2337,69 @@ RSpec.describe 'MODS name <--> cocina mappings' do
     end
   end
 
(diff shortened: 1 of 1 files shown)


spec/services/cocina/mapping/descriptive/mods/name_spec.rb
```diff
@@ -197,7 +197,7 @@
             {
               name: [
                 {
-                  parallelValue: [
+                  groupedValue: [
```
ndushay, line 200: what does this have to do with?

arcadiafalcone: A mistake - it should be changed back to parallelValue.

ndushay: will do.

ndushay (approved):

PR #3834, opened 2022-04-07
Add report for specific identifier type values

### LINKED ISSUE #3811: Report on objects with specific identifier type values in MODS
Desired output: a list of item and APO druids for objects with an identifier element at any level with a type attribute that is one of the following:
apis
naf
swets
vintage
Wikidata Item

If the output can also include which term was matched, that would be helpful.

mjgiarlo:

Fixes #3811

CI & sdr-deploy

### DIFF
--- a/bin/reports/report-desc-identifier_specific_types
+++ b/bin/reports/report-desc-identifier_specific_types
@@ -1,0 +1,23 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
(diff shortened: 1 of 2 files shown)


jcoyne (approved):
Can you let us know where the output is located?

mjgiarlo:
@jcoyne I always paste the output in a comment on the issue linked to the PR.

PR #3851, opened 2022-04-11
Bump to cocina-models 0.72.0

### LINKED ISSUE #3767: Set refresh: true on first catalogLink when mapping from fedora
See https://github.com/sul-dlss/cocina-models/pull/373

Blocked until cocina-models 0.71.0 is released

mjgiarlo:

Fixes #3767

This commit bumps to the latest cocina-models release which includes a new set of semantic validators. One of them replaces the DSA implementation of the ValidateDarkService. Another adds catalog link validation which is why this commit changes the mapping from Fedora to Cocina for catalog links.

Ran all tests locally against local cocina-models bump. (CI will fail until the cocina-models 0.72.0 is cut for real)


app/services/cocina/from_fedora/identification.rb
```diff
@@ -54,11 +56,19 @@ def doi
         value.text.delete_prefix('https://doi.org/')
       end
 
+      def refreshable?
+        AdministrativeTags.for(identifier: fedora_object.pid).exclude?(DO_NOT_REFRESH_TAG)
```
justinlittman, line 60: Should this be assigned to a class variable to prevent repeated lookups (i.e., one for each catkey)?

justinlittman (approved):

Did you need to add the `name` check back into cocina models for this?

mjgiarlo:
@justinlittman 💬  
> Did you need to add the `name` check back into cocina models for this?

I did, yeah. It's in 0.72.0.

app/services/cocina/from_fedora/identification.rb
mjgiarlo: @justinlittman Got it.

PR #3855, opened 2022-04-12
Updated w3cdtf report

### LINKED ISSUE #3848: Updated W3CDTF date report with additional valid patterns
Desired output: druid, APO, catkey, and date value.

This report should return records that have any date element (dateCreated, dateIssued, dateCaptured, dateValid, dateModified, copyrightDate, dateOther) that has an encoding attribute with the value w3cdtf but does not match a valid w3cdtf pattern.

Valid patterns: sul-dlss/cocina-models#392

Previous report: #3796

lwrubel:
Resolves #3848. 

Report run on sdr-deploy against `druids.txt`. Report is attached to the issue. 8016 druids identified with invalid date patterns. 

### DIFF
--- a/bin/reports/report-desc-w3cdtf-updated
+++ b/bin/reports/report-desc-w3cdtf-updated
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative '../../config/environment'
+require_relative '../../lib/report'
+
+date_elements = %w[
+  dateCreated
+  dateIssued
+  dateCaptured
+  dateValid
+  dateModified
+  copyrightDate
+  dateOther
+].freeze
+
+formats = [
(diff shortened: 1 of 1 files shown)


edsu (approved):
Is there a reason to not overlay this report over the old one?

lwrubel:
@edsu I suppose that is a question for @justinlittman and @arcadiafalcone? I was assuming there was some value in having both reports, but maybe that's not really needed. Totally get that it's odd to have two such very similar reports here.

justinlittman:
Let's hold on merging this. https://github.com/sul-dlss/dor-services-app/pull/3857 is going to provide a DateTimeValidator that should supercede this. However, the results from running this version will be useful to testing the DateTimeValidator.

justinlittman:
closing in favor of https://github.com/sul-dlss/dor-services-app/pull/3857

PR #3872, opened 2022-04-14
update_marc_records_service: 856 |x for _set_ is no longer needed

### LINKED ISSUE #3871: Do not include "x|set" in the 856 for SW release
[This was reported to me via email from members of the maps team]

The 856 created for SearchWorks release is including a field labeled `x| set`. I think this may be a legacy of SDR once having "collections" and "sets". Since we don't have sets anymore, we should not include the `x | set` data.

An example from https://searchworks.stanford.edu/view/14057069
```
856 4 1
u| https://purl.stanford.edu/ng733sv9246 x| SDR-PURL x| item x| barcode:36105228500315 x| file:ng733sv9246%2Fng733sv9246_0001.jp2 x| collection:vt578hf4454::Conrad Collection on Dutch Waterways x| set:vt578hf4454::Conrad Collection on Dutch Waterways x| rights:world
```

In the example, the `x|set` comes right after the `x|collection` and has the same data. This seems to be having an odd effect on the SW display, which is inserting the link to the collection record twice on the left-hand sidebar. It should only be providing the "This item belongs to a collection" link.

ndushay:

Per Andrew, when we export an 856 to Symphony, we no longer need |x for SET;  set objects are passé, and the information in |x for "set" is exactly the same as the information in |x for "collection"

Fixes #3871

### DIFF
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -50,9 +50,8 @@ module Dor
     # Subfield x #4 (optional): the barcode if known (<identityMetadata><otherId name="barcode">, recorded as barcode:barcode-value
     # Subfield x #5 (optional): the file-id to be used as thumb if available, recorded as file:file-id-value
     # Subfield x #6..n (optional): Collection(s) this object is a member of, recorded as collection:druid-value:ckey-value:title
-    # Subfield x #7..n (optional): Set(s) this object is a member of, recorded as set:druid-value:ckey-value:title
-    # Subfield x #8..n (optional): label and part sort keys for the member
-    # Subfield x #9..n (optional): High-level rights summary
+    # Subfield x #7..n (optional): label and part sort keys for the member
+    # Subfield x #8..n (optional): High-level rights summary
     def generate_symphony_records
       return [] unless ckeys?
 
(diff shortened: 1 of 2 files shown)


app/models/dor/update_marc_record_service.rb
```diff
@@ -160,18 +159,6 @@ def get_x2_collection_info
       collection_info
     end
 
-    # returns the constituent information subfields if exists
-    # @return [String] the constituent information druid-value:catkey-value:title format
-    def get_x2_constituent_info
```
ndushay, line 165: despite using "constituent" in the method name, this is NOT related to virtual objects, and if you chase all the code down, it exactly repeats the information for subfield x for collection

app/models/dor/update_marc_record_service.rb
```diff
@@ -229,12 +216,6 @@ def released_for
       ::ReleaseTags.for(cocina_object: @cocina_object)
     end
 
-    def dor_items_for_constituents
```
ndushay, line 232: despite using "constituents" in the method name, this seemingly has nothing to do with virtual objects.  Compare this code with `get_x2_collectrion_info` on line 147

PR #3880, opened 2022-04-15
Align purls and external identifiers.

justinlittman:
Next cocina models release is going to require that purls and external identifiers match.

### DIFF
--- a/spec/dor/update_marc_record_service_spec.rb
+++ b/spec/dor/update_marc_record_service_spec.rb
@@ -9,6 +9,7 @@ RSpec.describe Dor::UpdateMarcRecordService do
   let(:druid) { 'druid:bc123dg9393' }
   let(:bare_druid) { druid.delete_prefix('druid:') }
   let(:collection_druid) { 'druid:cc111cc1111' }
+  let(:collection_bare_druid) { collection_druid.delete_prefix('druid:') }
   let(:dro_object_label) { 'A generic label' }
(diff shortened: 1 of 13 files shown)


spec/dor/update_marc_record_service_spec.rb
```diff
@@ -9,6 +9,7 @@
   let(:druid) { 'druid:bc123dg9393' }
   let(:bare_druid) { druid.delete_prefix('druid:') }
   let(:collection_druid) { 'druid:cc111cc1111' }
+  let(:collection_bare_druid) { collection_druid.delete_prefix('druid:') }
```
ndushay, line 12: maybe make a ticket for druid-tools to have a method for bare_druid?  this is scattered everywhere in our code.

ndushay (approved):
suggestion for a separate ticket;  maybe not worth it - dunno.

spec/dor/update_marc_record_service_spec.rb
justinlittman: I feel like there are probably existing helper methods for this, but never seemed worth it compared to just `.delete_prefix('druid:')`, which doesn't require anything to remember. But I'm not opposed either.

mjgiarlo: I believe @jcoyne added a druid model to... Argo? dsa? for this. Might be worth extracting that to either druid-tools or cocina-models?

PR #3885, opened 2022-04-18
Adds script for validating cocina models changes.

justinlittman:
Allow testing cocina models changes.

QA

### DIFF
--- a/bin/validate-cocina
+++ b/bin/validate-cocina
@@ -1,0 +1,65 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Validates cocina model changes by loading each object into cocina to determine if an error is raised.
+# This should be run on sdr-deploy since it requires using a version of cocina models that is not the latest release.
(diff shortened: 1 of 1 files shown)


justinlittman:
Note that I will be adding documentation on this to the cocina models README in separate PR.

justinlittman:
https://github.com/sul-dlss/cocina-models/pull/421

bin/validate-cocina
```diff
@@ -0,0 +1,65 @@
...
+    advance: num_for_progress_advance(count),
+    total: count
+  )
+end
+
+def validate(clazz, sample_size, writer)
+  objs = if sample_size
+           clazz.limit(sample_size)
+         else
+           clazz.all
```
edsu, line 33: Is this an iterator?

edsu (approved):

bin/validate-cocina
justinlittman: Yes, and it is lazy.

PR #3887, opened 2022-04-18
Create a report for valid-catkeys

jcoyne:

Report if the value was valid or not.


edsu (approved):
It looks like this only validates the first catkey?

jcoyne:
Yes.

PR #3888, opened 2022-04-18
log missing druids during migration

### LINKED ISSUE #3882: bin/migrate should log missing druids as errors
It would be useful to be able to notice missing druids when grepping the log for errors afterwards.

peetucket:

Fixes #3882 - log missing druids during migration

It was not


peetucket (commented):
question: what does `ar_exists?(druid)` check ?  it is ActiveRecord, Fedora?

jcoyne:
I think that `ar` == `ActiveRecord`

jcoyne (approved):

PR #3896, opened 2022-04-18
add versions/is_open endpoint

ndushay:

For sul-dlss/argo/issues/3493, we need a way to check if a versionWF is in the "open" state for an object.  We can't use the opposite of "openable?" because "is_open" may be used for objects that are not yet accessioned -- they have only gotten as far as the `open` step in the `versionWF`.

I named it `is_open` rather than `open?` because in a URL, the question mark would need to be encoded and it didn't seem worth it.

This adds an endpoint and we already had the internal method to check.  Whee!

specs

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -49,6 +49,12 @@ class VersionsController < ApplicationController
     render build_error('Unable to check if openable due to preservation client error', e, status: :internal_server_error)
   end
 
+  def open?
+    render plain: VersionService.open?(@cocina_object)
+  rescue Dor::WorkflowException => e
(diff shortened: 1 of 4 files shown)


jcoyne:
@ndushay this endpoint is calling workflow service.  Would it be less risky and faster for argo to just call the workflow service?  I believe that Argo currently does this very same query.

ndushay:
@jcoyne - thanks!  great steer. closing this.

> @ndushay this endpoint is calling workflow service. Would it be less risky and faster for argo to just call the workflow service? I believe that Argo currently does this very same query here: https://github.com/sul-dlss/argo/blob/bfc7cde41feeaa23f3c481d282cc0b783d5ac30b/app/services/state_service.rb#L58

PR #3899, opened 2022-04-19
Remove update_descriptive feature flag

jcoyne:

This is no longer used in the code anywhere


edsu (approved):
Should this be removed from [spec/services/cocina/object_updater_spec.rb](https://github.com/sul-dlss/dor-services-app/blob/bb6d2e3394927ee9a729e8e24bead73bb9b88327/spec/services/cocina/object_updater_spec.rb#L33) as well?

jcoyne:
@edsu good catch, thank you.

PR #3912, opened 2022-04-20
fix some typos I discovered on the way to something else

ndushay:

I corrected these typos on two separate PRs that got closed without merging.

### DIFF
--- a/spec/models/object_version_spec.rb
+++ b/spec/models/object_version_spec.rb
@@ -77,7 +77,7 @@ RSpec.describe ObjectVersion, type: :model do
     end
   end
 
-  describe '#update_version' do
+  describe '#update_current_version' do
     context 'when description and significance not provided' do
       before do
         described_class.create(druid: druid, version: 1, tag: '1.0.0')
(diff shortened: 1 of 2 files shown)


spec/models/object_version_spec.rb
```diff
@@ -77,7 +77,7 @@
     end
   end
 
-  describe '#update_version' do
+  describe '#update_current_version' do
```
ndushay, line 80: probably method was renamed at some point, but this wasn't updated.

spec/models/object_version_spec.rb
```diff
@@ -106,7 +106,7 @@
 
       it 'does not update' do
         described_class.update_current_version(druid: druid, significance: :minor)
-        expect(described_class.find_by(druid: druid, version: 1).version).to eq(1)
+        expect(described_class.current_version(druid).version).to eq(1)
```
ndushay, line 109: this proves the current_version is still 1, which is a better test.

spec/requests/versions_spec.rb
```diff
@@ -146,7 +146,7 @@
         allow(VersionService).to receive(:can_open?).and_return(false)
       end
 
-      it 'returns true' do
+      it 'returns false' do
```
ndushay, line 149: fix copy-pasta.

jcoyne (approved):

PR #3914, opened 2022-04-21
Unopaque locks.

justinlittman:
Opaque locks are difficult to troubleshoot. Since this is an internal API opaque locks are unnecessary.

### DIFF
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -184,7 +184,8 @@ class CocinaObjectStore
   end
 
   def fedora_lock_for(fedora_object)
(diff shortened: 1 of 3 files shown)


jcoyne (commented):
Can you explain a scenario where this make it easier to troubleshoot? All we need to know is that you have a key and it either is or is not the expected key. Can you add a log message that says "Returning <druid> with <key>" and "Updating provided <key1> but expected <key2>" That would give clarity, right?

My concern is that people will start gleaning extra value from the lock value.

justinlittman:
Scenario: I was just troubleshooting https://app.honeybadger.io/projects/49894/faults/85052392. It would have been helpful to relate the locks reported in the alert to the lock version stored in the database.

jcoyne (approved):

PR #3915, opened 2022-04-21
Changes version open endpoint to return cocina object.

justinlittman:
In some uses of the endpoint, the client intends to further update the cocina object. To do this, it needs a cocina object with the updated version AND the etag lock for the current version.

Unit

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -70,4 +70,11 @@ class ApplicationController < ActionController::API
   def load_cocina_object
     @cocina_object = CocinaObjectStore.find(params[:object_id] || params[:id])
   end
+
+  # Adds headers from the cocina object.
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -70,4 +70,11 @@ def http_auth_header
   def load_cocina_object
     @cocina_object = CocinaObjectStore.find(params[:object_id] || params[:id])
   end
+
+  # Adds headers from the cocina object.
+  def add_cocina_headers(cocina_object)
```
jcoyne, line 75: Can this be named add_headers? I don't think the `cocina` part is adding anything. Properly these are HTTP headers.

jcoyne (changes requested):

jcoyne (approved):

PR #3919, opened 2022-04-22
Add db not null constraints for tag and description fields for object version

### LINKED ISSUE #3905: Require object version tag and description fields in db
Currently, the tag and description fields are optional for the `object_versions` table. They should be `NOT NULL`.

As part of the migration, populate existing null fields:
* For tag, it should be "\<version\>.0.0", e.g., "4.0.0" for version 4.
* For description is should be "Version \<tag\>", e.g., "Version 4.1.0" for tag "4.1.0".

justinlittman:
closes #3905

Enforce cleaner data.

### DIFF
--- a/app/models/object_version.rb
+++ b/app/models/object_version.rb
@@ -6,22 +6,24 @@ class ObjectVersion < ApplicationRecord
   # @param [String] druid
   # @param [Symbol] significance which part of the version tag to increment
   #  :major, :minor, :admin (see VersionTag#increment)
-  # @param [String] description optional text describing version change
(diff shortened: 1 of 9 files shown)


app/models/object_version.rb
```diff
@@ -8,20 +8,22 @@ class ObjectVersion < ApplicationRecord
   #  :major, :minor, :admin (see VersionTag#increment)
   # @param [String] description optional text describing version change
```
lwrubel, line 9: Update params to make description and significance not optional? (Here and in other updated methods below)

PR #3921, opened 2022-04-22
require description and significance parameters for VersionsController.create

### LINKED ISSUE #3906: Require parameters for versions create endpoint
Currently, a number of parameters are permitted: (https://github.com/sul-dlss/dor-services-app/blob/main/app/controllers/versions_controller.rb#L71-L78)

The description (string)  and significance (String enum:  'major', 'minor', 'patch') parameters should be required.

ndushay:

This is part of a series of tickets to ensure that object versions, moving forward, will always have a description and a tag.

Fixes #3906

specs, ran all integration tests on qa

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -62,7 +62,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorResponse'          
+                $ref: '#/components/schemas/ErrorResponse'
       parameters:
         - name: barcode
           in: query
@@ -91,7 +91,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorResponse'          
+                $ref: '#/components/schemas/ErrorResponse'
       parameters:
         - name: barcode
           in: query
@@ -198,7 +198,7 @@ paths:
             enum:
               - accessionWF
               - releaseWF
-            example: releaseWF
+          example: releaseWF
         - name: lane-id
           in: query
           description: Lane for prioritizing the work
@@ -237,7 +237,6 @@ paths:
             enum:
               - accessionWF
               - assemblyWF
-            example: assemblyWF
         - name: significance
           in: query
           description: the significance of the version change (if versioning is needed)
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -27,6 +27,7 @@ Layout/LineLength:
   Exclude:
     - 'bin/remediate-stanford-placeTerm'
     - 'spec/services/cocina/mapping/descriptive/h2_datacite/subject_h2_datacite_spec.rb'
+    - 'spec/requests/versions_spec.rb'
```
jcoyne, line 30: Would you mind moving this exception to the file itself?  I don't think we want to exclude the whole file from this rule, just the line that is long.

openapi.yml
```diff
@@ -1142,6 +1142,37 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+        - name: description
+          in: query
+          description: short description for object version being opened
+          required: true
+          schema:
+            type: string
+            example: 'Adding thumbnail'
```
justinlittman, line 1151: Should example be up a level (i.e., under description)?

spec/requests/versions_spec.rb
```diff
@@ -109,15 +109,16 @@
 
       it 'opens a new object version when posted to' do
         post '/v1/objects/druid:mx123qw2323/versions',
-             headers: { 'Authorization' => "Bearer #{jwt}" }
+             params: open_params.to_json,
+             headers: { 'Authorization' => "Bearer #{jwt}", 'Content-Type' => 'application/json' }
```
justinlittman, line 113: Why the Content-Type?

justinlittman (approved):

spec/requests/versions_spec.rb
ndushay: because the params are coming in as json;  this matched an existing spec that passed.

.rubocop.yml
ndushay: sure thing.

openapi.yml
ndushay: `example:` always seems to be at the same level as `type:` so I believe this is correct.

ndushay: this also matches the params defined for accession ...

ndushay: huh - can go either way.

PR #3922, opened 2022-04-22
Extract the Fedora coupled persistence methods to their own module

jcoyne:

This will make them easier to remove when the time comes. (fewer merge conflicts)

test suite


justinlittman (changes requested):
I'd like to hold this until after we've deployed for migration.

ndushay (approved):

justinlittman (approved):

PR #3928, opened 2022-04-25
Add spec for uniform title with parallel corporate author

arcadiafalcone:
Roundtrip validation review showed data loss for this pattern


jcoyne:
@justinlittman @ndushay should we consider porting this to cocina-models?

justinlittman (approved):

justinlittman:
Yes, I will once I have the fix too.

PR #3937, opened 2022-04-26
Version open endpoint

justinlittman:
So that openapi reflects the expected params.


app/controllers/versions_controller.rb
```diff
@@ -73,25 +73,27 @@ def build_error(msg, err, status: :unprocessable_entity)
...
       :start_accession,
-      :user_name,
-      :version_num
+      :user_name
     ).to_h.symbolize_keys
+    boolean_param(new_params, :start_accession)
+  end
+
+  def boolean_param(param_hash, param)
+    param_hash[param] = param_hash[param] == 'true' if param_hash.key?(param)
```
jcoyne, line 96: How about using `ActiveRecord::Type::Boolean.new.deserialize(param_hash[param])`

jcoyne (changes requested):

app/controllers/versions_controller.rb
jcoyne: Doesn't committee already do this for us? https://github.com/interagent/committee/blob/0ca9c8f777572bf699c5067824cbcaebbbd68d27/lib/committee/schema_validator/hyper_schema/string_params_coercer.rb#L66-L71

jcoyne (approved):

openapi.yml
```diff
@@ -1086,6 +1086,30 @@ paths:
             default: true
           required: false
           description: Indicates if accessionWF should be started after closing the version
+        - name: significance
+          in: query
+          description: the significance of the version change
+          required: false
```
ndushay, line 1092: true?

openapi.yml
```diff
@@ -1086,6 +1086,30 @@ paths:
...
+            type: string
+            enum:
+              - minor
+              - major
+              - admin
+            example: admin
+        - name: description
+          in: query
+          description: the description of the version change
+          required: false
```
ndushay, line 1103: true?

ndushay (commented):
Two possible mistakes in openapi.yml?

PR #3945, opened 2022-04-26
strip metadata before publishing embargo release message to queue

peetucket:

Fixes sul-dlss/happy-heron#2230

We need to strip the metadata from the cocina model before publishing to the queue, since the consumer (currently H2) does not expect it to be there.

Updated specs to verify metadata is stripped when publishing the message.

### DIFF
--- a/app/services/notifications/embargo_lifted.rb
+++ b/app/services/notifications/embargo_lifted.rb
@@ -13,7 +13,7 @@ module Notifications
     end
 
     def initialize(model:, channel:)
-      @model = model
+      @model = Cocina::Models.without_metadata(model)
       @channel = channel
     end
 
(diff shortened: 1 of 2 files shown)


ndushay (commented):
this looks ok, but I'd like @justinlittman to review.  Also - are there any other consumers besides H2?   Etd?  Argo?

peetucket:
> this looks ok, but I'd like @justinlittman to review. Also - are there any other consumers besides H2? Etd? Argo?

I'm not sure, though since this is stripping the metadata on the publishing side (as opposed to having H2 do it on other end), this should restore the message format to what it was previously, thus allowing other consumers to consume the message as they did before

justinlittman (approved):

PR #3947, opened 2022-04-27
Rename active record factories.

justinlittman:
To avoid naming conflicts when start using cocina model factories.

### DIFF
--- a/spec/factories/admin_policies.rb
+++ b/spec/factories/admin_policies.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :admin_policy do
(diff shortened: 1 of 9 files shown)


spec/factories/admin_policies.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :admin_policy do
+  factory :ar_admin_policy, class: 'AdminPolicy' do
```
mjgiarlo, line 4: How would you feel about `persisted_` vs `ar_`?

mjgiarlo (approved):
I retract my prior comment. This is a better approach.

PR #3954, opened 2022-04-28
[HOLD] Remove VersionMigrationService

jcoyne:
This step is not required once the migration is complete.

Integration tests


justinlittman (changes requested):
It also is here: https://github.com/sul-dlss/dor-services-app/blob/main/lib/fedora_migrator.rb#L80

jcoyne:
Fixed

justinlittman (approved):

PR #3956, opened 2022-04-29
Fix 856 generation for parallel title.

justinlittman:
* Clean-up existing spec.
* Handle parallel titles so that release to SearchWorks works correctly.

**Note that this PR contains 2 commits: The first is to clean the existing spec. The second is the change for this ticket. It might be helpful to look at them separately.**

Unit

### DIFF
--- a/app/models/dor/update_marc_record_service.rb
+++ b/app/models/dor/update_marc_record_service.rb
@@ -160,21 +160,9 @@ module Dor
     end
 
     def get_x2_part_info
-      title_info = @cocina_object.description&.title&.first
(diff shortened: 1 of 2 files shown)


justinlittman:
@andrewjbtw Who can review these specs? https://github.com/sul-dlss/dor-services-app/blob/9b6ddfd6e14e14649fdec1877b018b9a7e2b5982/spec/dor/update_marc_record_service_spec.rb#L504-L847

jcoyne (approved):

arcadiafalcone (approved):

PR #3994, opened 2022-05-04
Refactor catkeys/barcode fetching method

peetucket:

After working on #3993 (modifying how barcode/catkeys are refreshed), I thought it would be wise to remove the duplication of the logic I modified and place in a central location as well.  

Added some tests too.

Question: is there a reason the refresh action in the controller goes digging for a barcode while the cocina object store class does not?  Should they work the same way (i.e. by moving that bit of logic out of the controller and into the single "identifiers" method)?

New tests

### DIFF
--- a/app/controllers/metadata_refresh_controller.rb
+++ b/app/controllers/metadata_refresh_controller.rb
@@ -23,9 +23,7 @@ class MetadataRefreshController < ApplicationController
   private
 
   def identifiers
-    return [] unless @cocina_object.identification&.catalogLinks
-
-    @identifiers ||= @cocina_object.identification.catalogLinks.filter_map { |clink| "catkey:#{clink.catalogRecordId}" if clink.catalog == 'symphony' && clink.refresh }.tap do |id|
+    RefreshMetadataAction.identifiers(cocina_object: @cocina_object).tap do |id|
       # Not all Cocina objects have identification metadata that includes barcodes (e.g., collections)
       next unless @cocina_object.identification.try(:barcode)
 
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -322,10 +322,6 @@ class CocinaObjectStore
     cocina_request_object.new(label: label, description: description_props)
   end
 
-  def catkeys_for(cocina_request_object)
-    cocina_request_object.identification&.catalogLinks&.filter_map { |clink| "catkey:#{clink.catalogRecordId}" if clink.catalog == 'symphony' && clink.refresh }
-  end
-
   # Converts from Cocina::Models::RequestDRO|RequestCollection|RequestAdminPolicy to Cocina::Models::DRO|Collection||AdminPolicy
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/CyclomaticComplexity
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -309,8 +309,7 @@ class CocinaObjectStore
   def sync_from_symphony(cocina_request_object, druid)
     return cocina_request_object if cocina_request_object.admin_policy?
 
-    catkeys = catkeys_for(cocina_request_object)
-
+    catkeys = RefreshMetadataAction.identifiers(cocina_object: cocina_request_object)
     return cocina_request_object if catkeys.blank?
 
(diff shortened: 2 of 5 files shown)


spec/services/refresh_metadata_action_spec.rb
```diff
@@ -4,73 +4,114 @@
```
peetucket, line 4: the existing tests were not modified, the file just got restructured a bit since I added "describe" blocks and moved some of the "lets" around... I really only added new tests for the new method added to `RefreshMetadataAction`

app/controllers/metadata_refresh_controller.rb
```diff
@@ -23,9 +23,7 @@ def refresh
   private
 
   def identifiers
-    return [] unless @cocina_object.identification&.catalogLinks
-
-    @identifiers ||= @cocina_object.identification.catalogLinks.filter_map { |clink| "catkey:#{clink.catalogRecordId}" if clink.catalog == 'symphony' && clink.refresh }.tap do |id|
+    RefreshMetadataAction.identifiers(cocina_object: @cocina_object).tap do |id|
       # Not all Cocina objects have identification metadata that includes barcodes (e.g., collections)
       next unless @cocina_object.identification.try(:barcode)
```
peetucket, line 29: should this bit looking for a barcode in the identifier block move to the new `RefreshMetadataAction.identifiers` method?  it if so, this would allow us to further simply the controller code to mimic what is done in `CocinaObjectStore`

app/services/cocina_object_store.rb
```diff
@@ -323,10 +322,6 @@ def sync_from_symphony(cocina_request_object, druid)
     cocina_request_object.new(label: label, description: description_props)
   end
 
-  def catkeys_for(cocina_request_object)
-    cocina_request_object.identification&.catalogLinks&.filter_map { |clink| "catkey:#{clink.catalogRecordId}" if clink.catalog == 'symphony' && clink.refresh }
-  end
-
   # Converts from Cocina::Models::RequestDRO|RequestCollection|RequestAdminPolicy to Cocina::Models::DRO|Collection||AdminPolicy
```
peetucket, line 325: this method gets moved to `RefreshMetadataAction` so we can use it here and from the controller code

spec/services/refresh_metadata_action_spec.rb
```diff
@@ -4,73 +4,114 @@
...
+        expect(refresh.failure?).to be(true)
+      end
     end
   end
 
-  context 'when Descriptive.props returns nil' do
-    before do
-      allow(Cocina::Models::Mapping::FromMods::Description).to receive(:props).and_return(nil)
+  describe '#identifiers' do
+    subject(:indentifiers) { described_class.identifiers(cocina_object: cocina_object) }
```
peetucket, line 80: test the new method's ability to get identifiers

app/services/refresh_metadata_action.rb
```diff
@@ -12,6 +12,12 @@ def self.run(identifiers:, cocina_object:, druid:)
     new(identifiers: identifiers, cocina_object: cocina_object, druid: druid).run
   end
 
+  # @param [Cocina::Models::DRO|Collection|APO|Agreement|RequestDRO|RequestCollection|RequestAPO|RequestAgreement] cocina_object
+  # @return [Array] Refreshable identifiers in the object
+  def self.identifiers(cocina_object:)
+    cocina_object.identification&.catalogLinks&.filter_map { |clink| "catkey:#{clink.catalogRecordId}" if clink.catalog == 'symphony' && clink.refresh }
```
justinlittman, line 18: Identification will always have catalogLinks array, so I think this can just be `cocina_object.identification.catalogLinks.filter_map`.

peetucket: ok, this was a copy from the existing way of doing it...was assuming it was guarded against for some reason

is it possible for `catalogLinks` to be nil (instead of empty array) though?

justinlittman: I'm fairly sure not. In the past all of this guarding was necessary, but not with current cocina models.

justinlittman (approved):

peetucket:
And based on a test I just had to update after the guard was removed, is it possible collections doesn't have the `catalogLinks` ?

justinlittman:
The test that you had to update was with a double, not an actual cocina object to not too surprised.

I'm not sure why there is the different behavior looking for a barcode.

PR #4009, opened 2022-05-09
Remove ContentMetadataNormalizer

jcoyne:
and the tests that use it

Removing dead code now that we are off of Fedora.


spec/services/cocina/mapping/structural/dro_structural_spec.rb
```diff
@@ -1,362 +0,0 @@
-# frozen_string_literal: true
```
justinlittman, line 1: This and virtual_object_spec test ContentMetadataGenerator, which is still in use so I don't think they can be deleted. `dro_structural_mapping_helper` probably needs to be refactored to only test from cocina to fedora and `Cocina::FromFedora::DroStructural` needs to be removed.

justinlittman (changes requested):
This tests the

PR #4015, opened 2022-05-09
Remove unnecessary rescue blocks

jcoyne:

We don't have Mapping errors anymore.

CI

### DIFF
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -17,8 +17,6 @@ class CocinaObjectStore
   # Retrieves a Cocina object from the datastore.
   # @param [String] druid
   # @return [Cocina::Models::DROWithMetadata, Cocina::Models::CollectionWithMetadata, Cocina::Models::AdminPolicyWithMetadata] cocina_object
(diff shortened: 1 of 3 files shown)


app/services/cocina_object_store.rb
```diff
@@ -18,7 +18,6 @@ class StaleLockError < CocinaObjectStoreError; end
   # @param [String] druid
   # @return [Cocina::Models::DROWithMetadata, Cocina::Models::CollectionWithMetadata, Cocina::Models::AdminPolicyWithMetadata] cocina_object
   # @raise [SolrConnectionError] raised when cannot connect to Solr. This error will no longer be raised when Fedora is removed.
```
lwrubel, line 20: Should the docs about raise SolrConnectionError also be removed?

lwrubel (approved):
Approved, left a question

app/services/cocina_object_store.rb
lwrubel: (Or the description changed)

jcoyne: Yes, nice catch. I've removed it.

PR #4030, opened 2022-05-10
[refactor] Move DataErrorNotifier up the root namespace

jcoyne:
There no longer is need of a ToFedora module and this class was used outside of the module anyway

ci


app/services/data_error_notifier.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Notifies when a data error is encountered. Notification is performed with Honeybadger.
+class DataErrorNotifier
+  # In addition, it distinguishes between warnings and errors, even if both are notified the same.
```
justinlittman, line 5: Delete L5-6

justinlittman (approved):
Approve pending handling of comment.

app/services/data_error_notifier.rb
jcoyne: Thank you. Done.

PR #4046, opened 2022-05-11
Remove process tag

### LINKED ISSUE #4045: Stop creating Process : Content Type tags on registration.
These are no longer necessary.

jcoyne:
We now model everything that process tags handled in Cocina.

Fixes #4045

ci


app/services/goobi_service.rb
```diff
@@ -11,6 +11,31 @@ class ServerError < StandardError; end
                           Faraday::TimeoutError,
                           Faraday::RetriableResponse].freeze
 
+  # Formats the content type value as Goobi expects
+  class ContentType
+    # TODO: add Software
```
ndushay, line 16: I don't get this comment - can it be made more meaningful or removed?

ndushay (commented):
Was this a thing that was needed by Goobi or some such?  And we no longer need them because ...?

Is this something that can be tested with Goobi somehow, or did you do that already?

app/services/goobi_service.rb
jcoyne: I just moved it from where it was before. I thought it meant that software should be added to the hash at some point.

jcoyne:
@ndushay this was how the registration service used to tell what content type to create. Now that is all in cocina. Andrew said we're free to remove this.

ndushay (approved):

PR #4050, opened 2022-05-11
Rails passes a hash of options to #to_xml, not kwargs

jcoyne:
This PR makes this differentiation, because Ruby 3 is not as tolerant of this ambiguity

CI

### DIFF
--- a/app/services/publish/public_desc_metadata_service.rb
+++ b/app/services/publish/public_desc_metadata_service.rb
@@ -17,8 +17,9 @@ module Publish
     end
 
     # @return [String] Public descriptive medatada XML
-    def to_xml(include_access_conditions: true, prefixes: nil, template: nil)
-      ng_xml(include_access_conditions: include_access_conditions).to_xml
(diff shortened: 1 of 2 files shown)


app/services/publish/public_desc_metadata_service.rb
```diff
@@ -17,8 +17,9 @@ def doc
     end
 
     # @return [String] Public descriptive medatada XML
-    def to_xml(include_access_conditions: true, prefixes: nil, template: nil)
-      ng_xml(include_access_conditions: include_access_conditions).to_xml
+    # @params [Hash] _opts ({}) Rails sends args when rendering XML but we ignore them
+    def to_xml(_opts = {})
+      ng_xml(include_access_conditions: true).to_xml
```
mjgiarlo, line 22: Before this change, we could call `#to_xml(include_access_conditions: false` and that'd get passed to `ng_xml`. Do we not need that ability to vary the value?

mjgiarlo (approved):

PR #4051, opened 2022-05-11
Remove Hydrus test

jcoyne:
There no longer are Hydrus objects.


ndushay (approved):
poof!  more code gone!

PR #4052, opened 2022-05-11
[hold] Remove unused DateTimeValidator

jcoyne:
It's not used.


justinlittman (changes requested):
It will be used for upcoming DateTime validation (but will probably need to move to cocina models). Can we leave for now?

ndushay:
there are plans afoot for date validation; does this imply this PR should be closed?

ndushay:
or maybe this should be closed because it's now in cocina-models??

justinlittman:
It hasn't been moved yet; it is still in DSA.

ndushay:
It's possible we want to keep the regex code, or change the reports to use the regexps in cocina-models ...

mjgiarlo:
Done in #4091. I believe this can be closed now, @jcoyne

ndushay:
closing this as Mike took care of it in #4091

PR #4053, opened 2022-05-11
Refactor ShelvingService

jcoyne:
Previously both ShelvableFilesStager and ShelvingService needed to be aware of ContentMetadata. Now that concern is wholly within ShelvingService


justinlittman (approved):

ndushay:
conflict!!!!

PR #4055, opened 2022-05-11
Update to rails 6.1 to support ruby 3.1.2

### LINKED ISSUE #4036: Upgrade to ruby 3.1

aaron-collier:

Closes #4036 by updating to rails 6.1 in order to support ruby 3.1.2

This passes all tests when run against both ruby 3.0.3 and 3.1.2


jcoyne (changes requested):
Did you run `bin/rails app:update`?

aaron-collier:
@jcoyne good catch, I forgot about that step after `bundle update`. Walking through it now.

jcoyne (approved):

PR #4061, opened 2022-05-13
Pass user_name to the delete service

### LINKED ISSUE #4013: Allow the purge endpoint to take the username as a parameter
Put this name in the event that is created after purging.

aaron-collier:

Closes #4013 this passes a user_name to the delete service in order to record in the event.

If there is more to this ticket, please comment and I'm happy to address.

Updated unit tests.


justinlittman (commented):
Don't forget to add to openapi.yml.

jcoyne (approved):

PR #4079, opened 2022-06-06
Rake task to find objects missing from the index and output a list of…

aaron-collier:
… druids

Part of [dor-indexing-app/843](https://github.com/sul-dlss/dor_indexing_app/issues/843)

This initial rake task will be used to produce a report to determine the extent of missing druids in the index. It is possible that this rake task will be relatively temporary, but depending on the number of missing items, we may want to keep it for recurring reporting as further work on 843 continues.

### DIFF
--- a/lib/tasks/missing_druids.rake
+++ b/lib/tasks/missing_druids.rake
@@ -8,8 +8,8 @@ namespace :missing_druids do
     druids_from_solr = []
 
     loop do
-      results = SolrService.query('*:*', fl: 'id', rows: 10000, start: druids_from_solr.length, sort: 'id asc')
-      break if results.empty?
+      results = SolrService.query('id:*', fl: 'id', rows: 10_000_000, sort: 'id asc', wt: 'csv')
+      break unless results.empty?
 
       results.each { |r| druids_from_solr << r['id'] }
       sleep(0.5)
--- a/lib/tasks/missing_druids.rake
+++ b/lib/tasks/missing_druids.rake
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+namespace :missing_druids do
+  desc 'Find unindexed druids'
+  task unindexed_objects: :environment do
+    druids = []
+
+    models = [AdminPolicy, Collection, Dro]
+
+    models.each do |model|
+      model.find_each do |object|
+        current_druid = object.external_identifier
+        result = SolrService.query("id:#{current_druid}")
+        next unless result.empty?
+
+        puts "#{current_druid} (#{model}) not found in SOLR"
+        druids << current_druid
+      end
+    end
+
+    File.open('missing_druids.txt', 'w') do |file|
+      druids.each do |druid|
+        file.write("#{druid}\n")
+      end
+    end
+  end
+end
--- a/lib/tasks/missing_druids.rake
+++ b/lib/tasks/missing_druids.rake
@@ -7,23 +7,16 @@ namespace :missing_druids do
     druids_from_db = []
     druids_from_solr = []
 
-    loop do
-      results = SolrService.query('id:*', fl: 'id', rows: 10_000_000, sort: 'id asc', wt: 'csv')
-      break unless results.empty?
-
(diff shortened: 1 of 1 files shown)


edsu (approved):

jcoyne (changes requested):

lib/tasks/missing_druids.rake
```diff
@@ -0,0 +1,27 @@
...
+
+namespace :missing_druids do
+  desc 'Find unindexed druids'
+  task unindexed_objects: :environment do
+    druids = []
+
+    models = [AdminPolicy, Collection, Dro]
+
+    models.each do |model|
+      model.find_each do |object|
```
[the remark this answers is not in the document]
aaron-collier: @jcoyne good point, I've updated this to query SOLR first, then diff the results from the db per discussion.

lib/tasks/missing_druids.rake
```diff
@@ -0,0 +1,33 @@
...
+
+namespace :missing_druids do
+  desc 'Find unindexed druids'
+  task unindexed_objects: :environment do
+    models = [AdminPolicy, Collection, Dro]
+    druids_from_db = []
+    druids_from_solr = []
+
+    loop do
+      results = SolrService.query('*:*', fl: 'id', rows: 10000, start: druids_from_solr.length, sort: 'id asc')
```
ndushay, line 11: if you set `wt: 'csv'` and ask for 10 million, you should get all the druids in a very compact response, if you can wait long enough for the response (without losing the connection).  You should also see if the query `id:*` speeds it up at all, so Solr returns all docs with ids (which should be all docs.

lib/tasks/missing_druids.rake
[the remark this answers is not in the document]
ndushay: I *think* what JCoyne is suggesting is that you can use the list of druids to access each object by id, and determine which ones aren't there???    Or I could be confused.

ndushay (commented):
added some possibly useless info ...

lib/tasks/missing_druids.rake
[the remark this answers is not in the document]
aaron-collier: @ndushay we discussed in a slack thread getting a list of each (db, solr) then diffing the results.

lib/tasks/missing_druids.rake
aaron-collier: That's a good idea to look into if this sticks around, but since initially we're trying to get a first list, I would think this is good for now. But if not, let me know.

ndushay:
I'm okay running with this as is for now;  hopefully it won't be horribly non-performant.   

I would expect we will need to run the same thing (get DSA druids, get Solr druids, compare entire lists) as a cron in the future, so maybe another ticket for doing that which can include paying more attention to performance, after this is done ....

aaron-collier:
@ndushay yeah, this being a cron-able task is what it was initially intended for, but discussion this morning suggested that we wanted an initial report. Perhaps I misunderstood.

lib/tasks/missing_druids.rake
jcoyne: I'm with @ndushay.  Doing 1 request instead of 400 is the simplest thing and we should try it first.

lib/tasks/missing_druids.rake
```diff
@@ -0,0 +1,33 @@
...
+      results = SolrService.query('id:*', fl: 'id', rows: 10_000_000, wt: 'csv')
+      break unless results.empty?
+
+      results.each { |r| druids_from_solr << r['id'] }
+      sleep(0.5)
+    end
+    puts "Retrieved #{druids_from_solr.length} druids from SOLR"
+
+    models.each do |model|
+      model.find_each do |object|
```
jcoyne, line 20: If you do `model.all.pluck(:external_identifier)` here, you don't need this inner loop.

aaron-collier: oh, that's cool, I didn't realize pluck would be that helpful there. On it.

ndushay (approved):

PR #4082, opened 2022-06-06
Remove invalid iso8601 pattern

jcoyne:

This is not a valid ISO8601 date.

### DIFF
--- a/app/validators/date_time_validator.rb
+++ b/app/validators/date_time_validator.rb
@@ -30,7 +30,6 @@ class DateTimeValidator
   ].freeze
 
   ISO8601_FORMATS = COMMON_FORMATS + [
-    Regexp.new("^#{YEAR}#{MONTH}--$"),
     Regexp.new("^#{YEAR}#{MONTH}#{DAY}$"),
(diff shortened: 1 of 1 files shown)


jcoyne:
@arcadiafalcone can we remove this as an iso8601 format. I can't find this format in the spec.

app/validators/date_time_validator.rb
```diff
@@ -30,7 +30,6 @@ class DateTimeValidator
   ].freeze
 
   ISO8601_FORMATS = COMMON_FORMATS + [
-    Regexp.new("^#{YEAR}#{MONTH}--$"),
```
ndushay, line 33: @arcadiafalcone is year-month without the following dashes a valid pattern, or something like that?  

If not, merge at will.

ndushay (approved):
question for Arcadia on ISO8601 in our data (ISO8601 according to which version of the spec ...)

ndushay:
Oh no!  failing tests!

ndushay:
closing this in favor of #4091 work

PR #4083, opened 2022-06-06
Consolodate regexes

jcoyne:

Searching through a list of regexes takes a long time. Instead consolidate where possible

ci

### DIFF
--- a/app/validators/date_time_validator.rb
+++ b/app/validators/date_time_validator.rb
@@ -31,11 +31,7 @@ class DateTimeValidator
 
   ISO8601_FORMATS = COMMON_FORMATS + [
     Regexp.new("^#{YEAR}#{MONTH}--$"),
-    Regexp.new("^#{YEAR}#{MONTH}#{DAY}$"),
-    Regexp.new("^#{YEAR}#{MONTH}#{DAY}T#{HOUR}#{MINUTE}$"),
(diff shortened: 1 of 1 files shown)


app/validators/date_time_validator.rb
```diff
@@ -31,11 +31,7 @@ class DateTimeValidator
 
   ISO8601_FORMATS = COMMON_FORMATS + [
     Regexp.new("^#{YEAR}#{MONTH}--$"),
-    Regexp.new("^#{YEAR}#{MONTH}#{DAY}$"),
-    Regexp.new("^#{YEAR}#{MONTH}#{DAY}T#{HOUR}#{MINUTE}$"),
-    Regexp.new("^#{YEAR}#{MONTH}#{DAY}T#{HOUR}#{MINUTE}#{SECOND}$"),
-    Regexp.new("^#{YEAR}#{MONTH}#{DAY}#{HOUR}#{MINUTE}$"),
-    Regexp.new("^#{YEAR}#{MONTH}#{DAY}#{HOUR}#{MINUTE}#{SECOND}$")
+    Regexp.new("^#{YEAR}#{MONTH}#{DAY}T?(#{HOUR}#{MINUTE}|#{HOUR}#{MINUTE}#{SECOND})?$"),
```
ndushay, line 34: I find the other way a LOT easier to grok, only because of the number of nested parens.  Is performance a concern?

Also, there shouldn't be a T without something following, no?  I thik your regexp allows T on its own.  (QED)

ndushay (changes requested):

Issue #4072, opened 2022-06-01
Report on invalid ISO8601 dates in Cocina description


arcadiafalcone:
This is to check remediation progress for invalid ISO 8601 dates. For valid patterns, see #3849.

Desired output: druid, collection druid, catkey, date value(s).

Check for values in description..date..value with description..date.encoding.code iso8601.

ndushay:
How one might do this:

https://www.postgresql.org/docs/current/functions-json.html

```
dor_services=# select external_identifier from dros where jsonb_path_query_first(description, 'strict $.**.parallelContributor.size() > 0') = 'true' limit 1;
 external_identifier
---------------------
(0 rows)
```

![image.png](https://images.zenhubusercontent.com/5d89272aae87b600015057e7/54882009-8917-412d-8b4b-93b55e3344af
)

![image.png](https://images.zenhubusercontent.com/5d89272aae87b600015057e7/a9cab2f2-46e5-463a-a192-9471e704cac4)

ndushay:
See https://github.com/sul-dlss/dor-services-app/issues/4073#issuecomment-1146289667

jcoyne:
It looks like there are 1.4million objects that contain iso8601 dates that must be checked:
```
Dro.where("jsonb_path_exists(description, '$.**.date.encoding.code ? (@ ==  \"iso8601\")')").count
   (235522.8ms)  SELECT COUNT(*) FROM "dros" WHERE (jsonb_path_exists(description, '$.**.date.encoding.code ? (@ ==  "iso8601")'))
 => 1396795
```

jcoyne:
@mjgiarlo still running/tuning the report

mjgiarlo:
The report code has been written and merged to DSA. @jcoyne feel free to close the issue once you've handed the report off to @arcadiafalcone (apologies if you already did and I missed it)

ndushay:
I'm closing this;  feel free to re-open if needed.

ndushay:
I'm closing this;  feel free to re-open if needed.

jcoyne:
@arcadiafalcone here is the final report:
[invalid-iso8601-v4.csv](https://github.com/sul-dlss/dor-services-app/files/8862734/invalid-iso8601-v4.csv)

mjgiarlo:
@arcadiafalcone I re-ran the report after we changed it to find more bad data. Attached:
[...]

PR #4091, opened 2022-06-07
Add reports for edtf and w3cdtf validation

### LINKED ISSUE #4071: Report on invalid EDTF dates in Cocina description
This is to check remediation progress for invalid EDTF dates. For valid date patterns, see #3797.

Desired output: druid, collection druid, catkey, date value(s).

Check for values in description..date..value with description..date.encoding.code edtf.

### LINKED ISSUE #4073: Report on invalid W3CDTF dates in Cocina description
This is to check remediation progress for invalid W3CDTF dates. For valid date patterns see #3848.

Desired output: druid, collection druid, catkey, date value(s).

Check for values in description..date..value with description..date.encoding.code w3cdtf.

mjgiarlo:

Connects to #4071
Connects to #4073

Also brings the ISO8601 date report into alignment with the new reports and dumps bespoke date validation in favor of using upstream libraries.

Ran all three reports on stage


app/reports/bad_iso8601_dates.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def collection_id
+    dro.structural['isMemberOf'].first
+  end
+
+  def catkey
+    dro.identification['catalogLinks'].find { |link| link['catalog'] == 'symphony' }&.fetch('catalogRecordId')
+  end
+end
```
mjgiarlo, line 46: I refactored and renamed this report, @jcoyne, to align with the other two date reports and to accommodate a couple changes:

1. Report on all bad date values, not just the first
1. Report on values that are nested in structured and parallel and grouped values

ndushay (commented):
So this addresses #4071, #4072, #4072 and #4052?

How does it cope with Arcadia's desire for what may be a subset of the allowed regex values in the etdf gem, for example?

How will this relate to the date validation code going into cocina when someone picks up sul-dlss/cocina-models#483?   I see you remove that code here.  Also, we're going to do date validation via cocina-models so it seems prudent to have these reports run through the same validation code (I know,  sul-dlss/cocina-models#483 didn't exist when you picked up this ticket ...)

Beyond those questions, the code here looks fine to me.  Confer with JCoyne and Arcadia, I guess.

mjgiarlo:
@ndushay 
> So this addresses #4071, #4072, #4072 and #4052?

And #4073, yeah. I wasn't aware of #4052 when I started this PR, and I can remove those bits if folks prefer; I don't need it for these reports. I'm flexible.
 
> How does it cope with Arcadia's desire for what may be a subset of the allowed regex values in the etdf gem, for example?

This is the first I've heard of it. Can you or @arcadiafalcone point me at more information?
 
> How will this relate to the date validation code going into cocina when someone picks up [sul-dlss/cocina-models#483](https://github.com/sul-dlss/cocina-models/issues/483)? I see you remove that code here ...

Once it's in Cocina and that's deployed, entering bad dates will no longer be possible, but we may still have data to remediate. I'd say once all three reports run clean, we can ditch these reports and the changes in this commit.
 
> Beyond those questions, the code here looks fine to me. Confer with JCoyne and Arcadia, I guess.

cc: @jcoyne @arcadiafalcone 😄

app/reports/bad_iso8601_dates.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def initialize(dro:)
+    @dro = dro
+  end
+
+  def report
+    bad_values = path.on(dro.description.to_json).filter_map do |date_value|
+      DateTime.iso8601(date_value)
+      nil
+    rescue ArgumentError
```
jcoyne, line 22: Shouldn't this be a `Date::Error`?

jcoyne (changes requested):

app/reports/bad_iso8601_dates.rb
mjgiarlo: Yes, copypasta. Thank you!

jcoyne (approved):

ndushay:
@mjgiarlo can you easily determine which date patterns are allowed by this code?   The regexp in the date validator class (which will be migrating to cocina models)  should reflect what @arcadiafalcone wanted.  I don't know if it's a subset or not ... it's just that we've got more of a black box on valid dates with this implementation.

jcoyne:
@ndushay a Regex is not sufficient for validating dates (unless it's monstrously complex).  Consider leap years, Sept 31, etc.

mjgiarlo:
@ndushay @jcoyne I chatted with Arcadia and she is on board with using upstream validators ([comment](https://github.com/sul-dlss/cocina-models/issues/392#issuecomment-1150440283)). @ndushay can you re-review?

ndushay (approved):

PR #4104, opened 2022-06-10
[refactor] Extract UpdateObjectService from CocinaObjectStore

jcoyne:
Now CocinaObjectStore has fewer dependencies (e.g. EventFactory) and is only concerned with operations on the datastore.

CI


app/services/update_object_service.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# This handles all of the business logic around updating an object.
+# This includes:
+#   sending a rabbitMQ notification
+#   logging an event
+class UpdateObjectService
+  # TODO: This method should be renamed to `update'.
```
mjgiarlo, line 8: I think this TODO can go byebye now

mjgiarlo (approved):
Approve pending consideration of comment

app/services/update_object_service.rb
jcoyne: Thank you. I removed it.

PR #4112, opened 2022-06-13
Avoid passing the deprecated --without flag to bundler

jcoyne:
Set the environment variable instead

n/a


Dockerfile
```diff
@@ -15,6 +15,7 @@ ENV SETTINGS__SSL__KEY_PASS="${SETTINGS__SSL__KEY_PASS}"
 # dor-services-app cannot respond to this second request, so the indexing call times out.
 # This probably wouldn't be a problem in Rails 6, but ActiveFedora is preventing that upgrade.
 ENV RAILS_ENV=production
+ENV BUNDLER_WITHOUT development test
```
mjgiarlo, line 18: Should there be an `=` in there? The dockerfile reference discourages this syntax, apparently: https://docs.docker.com/engine/reference/builder/

jcoyne: Thank you. I fixed it.

mjgiarlo (approved):

PR #4137, opened 2022-07-05
Add rake task for unindexed druids.

justinlittman:
Make it easier to index unindexed druids.

QA


jcoyne (approved):
Should this be the task the cron runs? That is, why bother with missing_druids.txt?  Why not just index them right away?

justinlittman:
I'm inclined to see how this plays out and if the audit uncovers code problems. So far we've seen this in qa and stage, not prod so it might just be a one-time issue.

Issue #4143, opened 2022-07-12
Report on EDTF dates containing 'x' or 'X'


arcadiafalcone:
Report on all (any) dates within a cocina description with `encoding: "edtf"` and a value containing the character upper case `X` or lower case `x`. Output data should include druid, collection, catkey, and value. Concatenate multiple values for the same record with a semicolon.

Analysis for https://github.com/sul-dlss/mods/issues/54

See also #4071 for related report.

ndushay:
@arcadiafalcone I found NO instances of dates encoded as edtf with either 'x' or 'X'.     I checked my reports by looking for '1' instead, and that got output.

ndushay:
I'm going to close this ticket;  if it is still needed, please re-open it.

PR #4145, opened 2022-07-12
reports for edtf dates with x or X

ndushay:

Part of #4143, this is the code for the requested reports:  edtf dates that have lower case x  and another report for upper case X.

### DIFF
--- a/app/reports/edtf_lower_x_dates.rb
+++ b/app/reports/edtf_lower_x_dates.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+# Invoke via:
+# bin/rails r -e production "EdtfLowerXDates.report"
+class EdtfLowerXDates
+  def self.report
+    puts "item_druid,collection_druid,catkey,values\n" # rubocop:disable Rails/Output
+
+    Dro.where("jsonb_path_exists(description, '$.**.date.encoding.code ? (@ ==  \"edtf\")')").find_each do |dro|
+      new(dro:).report
+    end
+  end
+
+  def initialize(dro:)
+    @dro = dro
+  end
+
+  def report
+    matching_values = path.on(dro.description.to_json).uniq.filter_map do |date_value|
+      date_value if date_value.include?('x')
+    end
+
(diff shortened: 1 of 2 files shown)


app/reports/edtf_lower_x_dates.rb
```diff
@@ -0,0 +1,57 @@
...
+      new(dro:).report
+    end
+  end
+
+  def initialize(dro:)
+    @dro = dro
+  end
+
+  REGEX_OPTS = Regexp::IGNORECASE | Regexp::MULTILINE
+  DECADE_4CHAR_REGEX = Regexp.new('(^|\D)\d{3}[u\-?x]($|\D)', REGEX_OPTS)
```
mjgiarlo, line 19: Hmm, where is this used?

app/reports/edtf_upper_x_dates.rb
```diff
@@ -0,0 +1,57 @@
...
+      new(dro:).report
+    end
+  end
+
+  def initialize(dro:)
+    @dro = dro
+  end
+
+  REGEX_OPTS = Regexp::IGNORECASE | Regexp::MULTILINE
+  DECADE_4CHAR_REGEX = Regexp.new('(^|\D)\d{3}[u\-?x]($|\D)', REGEX_OPTS)
```
mjgiarlo, line 19: Same question.

mjgiarlo (commented):
Have you run the report yet? You may want to run it in prod and make sure it does what you expect and then mark this PR as ready for review or unheld or whatever. (At least, I've found this useful in the past as I inevitably get stuff wrong that I don't notice until I run it...)

app/reports/edtf_lower_x_dates.rb
ndushay: whoops - cruft.

ndushay:
I have run a report with '1' substituted for 'x' with `.limit(10)` and got output, so I'm confident these work.

mjgiarlo (approved):

PR #4154, opened 2022-07-21
Add a report for invalid location uris

### LINKED ISSUE #4152: Report on location URIs in descriptive metadata
Criteria: values of `event.location.uri` that do not begin with `http://` or `https://` or which end with `.html`.

Output: object druid, object catkey, collection druid, value of `uri`, value of sibling `code`, value of sibling `value`

jcoyne:
Fixes #4152

report here https://gist.github.com/jcoyne/c671e5223885cd9fcb702be7ff313cf2

### DIFF
--- a/app/reports/invalid_location_uris.rb
+++ b/app/reports/invalid_location_uris.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+# Invoke via:
+# bin/rails r -e production "InvalidLocationUris.report"
+class InvalidLocationUris
+  JSON_PATH = '$.**.event.location.uri'
+  SQL = <<~SQL.squish.freeze.squish
+    SELECT (jsonb_path_query_array(description, '#{JSON_PATH} ? (@ like_regex "^.*\.html$")') ||
+            jsonb_path_query_array(description, '#{JSON_PATH} ? (@ like_regex "^(?!https?://).*$")')) ->> 0 as value,
+           external_identifier,
+           jsonb_path_query(identification, '$.catalogLinks.catalogRecordId') ->> 0 as catkey,
+           jsonb_path_query(structural, '$.isMemberOf') ->> 0 as collection_id
+           FROM "dros" WHERE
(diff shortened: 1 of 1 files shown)


app/reports/invalid_location_uris.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+# Invoke via:
+# bin/rails r -e production "InvalidLocationUris.report"
+class InvalidLocationUris
+  JSON_PATH = '$.**.event.location.uri'
+  SQL = <<~SQL.squish.freeze.squish
```
mjgiarlo, line 7: yo dawg i heard u like squish?

app/reports/invalid_location_uris.rb
```diff
@@ -0,0 +1,34 @@
...
+    SELECT (jsonb_path_query_array(description, '#{JSON_PATH} ? (@ like_regex "^.*\.html$")') ||
+            jsonb_path_query_array(description, '#{JSON_PATH} ? (@ like_regex "^(?!https?://).*$")')) ->> 0 as contrib,
```
mjgiarlo, line 9: How do you feel about updating this binding name relative to what's in the JSON PATH above?

app/reports/invalid_location_uris.rb
jcoyne: Rubocop does this. I agree it seems dumb.

app/reports/invalid_location_uris.rb
jcoyne: updated to the generic `value`

app/reports/invalid_location_uris.rb
mjgiarlo: @jcoyne Odd. Which cop? Seems strange to squish an already squished frozen string.

jcoyne: If you run with `-A` this is what you get 🤷

aaron-collier (approved):

PR #4156, opened 2022-07-23
Only return symphony catatalogRecordIds

jcoyne:

This excludes others like 'previous symphony'


app/reports/invalid_location_uris.rb
```diff
@@ -25,7 +25,7 @@ def self.rows(sql)
     result = ActiveRecord::Base.connection.execute(sql)
 
     result.to_a.map do |row|
-      [row['external_identifier'], row['catkey'], row['collection_id'], row['uri'], row['code'], row['value']].join(',')
+      [row['external_identifier'], row['catkey'], row['collection_id'], row['uri'], row['code'], "\"#{row['value']}\""].join(',')
```
jcoyne, line 28: Need quotes here, because sometimes the values themselves have a comma, which breaks the CSV.

justinlittman (approved):

PR #4163, opened 2022-08-01
Remove unused dor-rights-auth gem

jcoyne:
We no longer use this in DSA


mjgiarlo (approved):
IT'S BEAUTIFUL 😭

PR #4169, opened 2022-08-03
rubocop Style/CaseLikeIf enabled as bug was fixed

ndushay:

https://github.com/rubocop-hq/rubocop/issues/8508 was fixed in August 2020.

### DIFF
--- a/app/models/symphony_reader.rb
+++ b/app/models/symphony_reader.rb
@@ -70,16 +70,16 @@ class SymphonyReader
   def symphony_response(url)
     resp = client.get(url)
 
-    if resp.status == 200
+    case resp.status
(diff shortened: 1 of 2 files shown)


app/models/symphony_reader.rb
```diff
@@ -69,12 +69,11 @@ def fetch_marc_response
 
   def symphony_response(url)
     resp = client.get(url)
+    raise NotFound, "Record not found in Symphony. Catkey: #{catkey}. API call: #{url}" if resp.status == 404
```
jcoyne, line 72: this would have been a good place to use `case resp.status ...`

jcoyne (approved):

Issue #4184, opened 2022-08-16
Report request for contributor role URIs


arcadiafalcone:
Report is for values of `contributor.role.uri` anywhere in the descriptive metadata that do not begin with any of the following (for http or https) or do not have at least two characters after the matched URL pattern (i.e., after the final slash):
```
http(s)://id.loc.gov/vocabulary/relators/
http(s)://id.loc.gov/performanceMediums/
http(s)://nomisma.org/id/
```
Report output should include object druid, catkey, collection druid, **collection name** (new addition not in previous reports), and the value. If a record returns multiple matches, the values should be concatenated on one line with a semicolon.

I created test object vz332dg7306 which has 3 incorrect role URIs.

jcoyne:
@arcadiafalcone can you explain what you mean by "at least two characters after the matched pattern"?

arcadiafalcone:
The pattern is the matched URL, for example https://id.loc.gov/marcrelators/. https://id.loc.gov/marcrelators/12 would not be in the report, but https://id.loc.gov/marcrelators/1 would. I've clarified the language above.

jcoyne:
@arcadiafalcone this one has 3.3M rows. Mostly because of uris like `http://id.loc.gov/vocabulary/relators/pbl` for example: https://argo.stanford.edu/view/jc162vw4899  Was this a mistake?

arcadiafalcone:
The URL was incorrect - the spec has been updated above.

jcoyne:
@arcadiafalcone it looks like the updated spec would still consider uris like `http://id.loc.gov/vocabulary/relators/pbl` invalid.  Can you confirm this is what you want?

arcadiafalcone:
Updated again for real correctness.

arcadiafalcone:
I could have sworn I double checked that one, but you are correct.

jcoyne:
Here's the report. It looks like there's a lot of Getty AAT uris:
https://gist.github.com/jcoyne/eea476052ac69b5d590a6cc259e727bc

PR #4190, opened 2022-08-19
Update report for invalid roles

jcoyne:

Ref #4184


justinlittman (approved):

ndushay (approved):
This is approved, so if it's now ready to be merged, please remove HOLD from the title and go for it!

PR #4197, opened 2022-08-23
When a collection title is updated, reindex the members

### LINKED ISSUE #4196: When a collection name is updated, reindex all items
When a collection name is changed, send a update message for each item in the collection to RabbitMQ.
This will cause the item to be reindexed, so that the item will show in the correct collection name facet.

jcoyne:

Fixes #4196 

Test suite

### DIFF
--- a/app/services/member_service.rb
+++ b/app/services/member_service.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# Finds the members of a collection by using Solr
+class MemberService
+  # @param [String] druid the identifier of the collection
+  # @param [Boolean] only_published when true, restrict to only published items
+  # @return [Array<Hash<String,String>>] the members of this collection
+  def self.for(druid, only_published: false)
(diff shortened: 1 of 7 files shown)


app/services/member_service.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Finds the members of a collection by using Solr
+class MemberService
+  # @param [String] druid the identifier of the collection
+  # @param [Boolean] only_published when true, restrict to only published items
+  # @return [Array<Hash<String,String>>] the members of this collection
+  def self.for(druid, only_published: false)
+    query = "is_member_of_collection_ssim:\"info:fedora/#{druid}\""
```
justinlittman, line 9: Why query Solr and not the DB?

jcoyne: I'm just reusing the code that was already in use.  I just moved it out of a controller and put it in a place to facilitate this reuse.

justinlittman (approved):

PR #4216, opened 2022-09-01
report invalid_subject_source_codes adjustments

### LINKED ISSUE #4214: Refine invalid subject source code report
Updates to the report in #4208 

- Do not include `source` when it has the ancestor `valueLanguage`
- Add classification source codes to valid list:
```
anscr
bcl
bcmc
bisacsh
bkl
blsrissc
cadocs
clc
ddc
dopaed
farl
ifzs
kktb
kssb
lcc
loovs
moys
msc
ncsclt
njb
njb/9
nlm
rpb
rswk
rvk
sbb
sdnb
sfb
ssgn
sswd
stub
sudocs
udc
zdbs
```

ndushay:

Fixes #4214 (which Arcadia already closed)

### DIFF
--- a/app/reports/invalid_subject_source_codes.rb
+++ b/app/reports/invalid_subject_source_codes.rb
@@ -14,6 +14,7 @@ class InvalidSubjectSourceCodes
   # > arrays when using the lax mode. To avoid surprising results, we recommend
   # > using the .** accessor only in the strict mode.
   JSON_PATH = 'strict $.**.subject.**.source.code'
+  OMIT_JSON_PATH = 'strict $.**.subject.**.valueLanguage.**.source.code'
   VALID_CODES = %w[
     aat
     abne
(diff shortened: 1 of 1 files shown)


app/reports/invalid_subject_source_codes.rb
```diff
@@ -14,6 +14,7 @@ class InvalidSubjectSourceCodes
   # > arrays when using the lax mode. To avoid surprising results, we recommend
   # > using the .** accessor only in the strict mode.
   JSON_PATH = 'strict $.**.subject.**.source.code'
+  OMIT_JSON_PATH = 'strict $.**.subject.**.valueLanguage.**.source.code'
```
ndushay, line 17: I tried to figure out how to do a path that could use the `**` wildcard with an "Except NOT these" and couldn't find a way quickly so I did it with comparisons below.  Maybe it's possible, maybe not.  Don't care much.

app/reports/invalid_subject_source_codes.rb
```diff
@@ -129,17 +159,21 @@ def self.rows(sql)
...
-        id,
-        rows.first['catkey'],
-        collection_druid,
-        "\"#{collection_name}\"",
-        rows.map { |row| row['value'] }.join(';')
-      ].join(',')
-    end
+        omit_values = rows.filter_map { |row| row['omit_value'] }.uniq
+        keep_values = rows.filter_map { |row| row['value'] unless omit_values.include?(row['value']) }
+        next if keep_values.empty?
```
ndushay, line 167: This is the approach I came up with to filter out the unwanted subject source codes -- do the filtering here in this way.  Working code wins.  Not even sure there is a better way.

justinlittman (approved):

PR #4222, opened 2022-09-02
Provide validation for file ids that are not formatted like a URI

jcoyne:

This will allow us to find places where we have non-standard URI creation

CI

### DIFF
--- a/app/services/cocina/id_generator.rb
+++ b/app/services/cocina/id_generator.rb
@@ -28,6 +28,10 @@ module Cocina
       "#{ID_NAMESPACE}/file/#{druid}-#{resource_id}/#{file_id}"
     end
 
+    def self.valid_file_id?(identifier)
+      identifier.start_with?("#{ID_NAMESPACE}/file/")
+    end
+
     private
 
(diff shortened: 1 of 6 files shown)


justinlittman:
Why not fix the problem and make it cocina validation? It is clearer to have all model validation in a single location.

jcoyne:
@justinlittman we use the cocina models in sdr-api that use encoded identifiers, not URIs.

justinlittman:
Ah, right.

app/services/cocina/id_generator.rb
```diff
@@ -28,6 +28,10 @@ def generate_or_existing_file_id
       "#{ID_NAMESPACE}/file/#{druid}-#{resource_id}/#{file_id}"
     end
```
ndushay, line 30: Please consider adding a comment to bring the content of the PR discussion into the code.  Something like

```
# this validation is useful for (e.g. sdr-api that uses) encoded identifiers, not URIs
```

mjgiarlo (approved):

PR #4224, opened 2022-09-08
[HOLD] Republish collection members when collection is (re-)published

### LINKED ISSUE #4195: Republish all objects in a collection when a collection is changed
When a collection is published,
then republish all of the published items that are not "opened" in that collection.

Ref sul-dlss/happy-heron#2354

> updating the collection name on the Purl requires updating the public XML, which mean republishing the XML for each item in a collection after its collection’s name has been updated. There’s no way to avoid having to do this because the Purls are static and there’s no rolling process to pull in new data. So what we’d need to do every time a collection name is updated would be to republish (but not necessarily re-release) every item in the collection (except for Registered items and dark items since they have no Purl). This is generally safe to do for items that already have Purls except in the case of items that are “Opened” – these might be in the process of being edited and could contain metadata not ready to be published.

mjgiarlo:

Fixes #4195

This commit changes the `Publish::MetadataTransferService` such that when a collection is (re-)published, its members are (re-)published. Not all members, though: only those members that are 1) already published, and 2) not in the `Opened` state. It does this work using the existing `MemberService` which is extended here to add a param excluding items in the `Opened` state.

CI. This needs testing in a deployed environment. Will work with Andrew on that.

### DIFF
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -15,6 +15,7 @@ module Publish
 
     # Appends contentMetadata file resources from the source objects to this object
     def publish
+      republish_members!
       return unpublish unless discoverable?
 
       # Retrieve release tags from identityMetadata and all collections this item is a member of
@@ -48,6 +49,16 @@ module Publish
       cocina_object.access.view != 'dark'
     end
 
+    def republish_members!
+      return unless cocina_object&.collection?
+
+      Array.wrap(
+        MemberService.for(cocina_object.externalIdentifier, exclude_opened: true, only_published: true)
+      ).each do |member|
+        self.class.publish(CocinaObjectStore.find(member['id']))
+      end
+    end
+
     # Create a file inside the content directory under the stacks.local_document_cache_root
     # @param [String] content The contents of the file to be created
     # @param [String] filename The name of the file to be created
@@ -77,7 +88,7 @@ module Publish
     end
 
     def purl_services_url
-      raise 'You have not configured perl-fetcher (Settings.purl_services_url).' unless Settings.purl_services_url
+      raise 'You have not configured purl-fetcher (Settings.purl_services_url).' unless Settings.purl_services_url
 
       "#{Settings.purl_services_url}/purls/#{cocina_object.externalIdentifier.delete_prefix('druid:')}"
     end
--- a/spec/services/publish/metadata_transfer_service_spec.rb
+++ b/spec/services/publish/metadata_transfer_service_spec.rb
@@ -4,9 +4,7 @@ require 'rails_helper'
 
 RSpec.describe Publish::MetadataTransferService do
   let(:druid) { 'bc123df4567' }
-
   let(:access) { {} }
-
   let(:cocina_object) do
     build(:dro, id: "druid:#{druid}").new(
       access:,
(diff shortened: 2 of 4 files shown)


app/services/publish/metadata_transfer_service.rb
```diff
@@ -15,6 +15,7 @@ def initialize(cocina_object)
 
     # Appends contentMetadata file resources from the source objects to this object
     def publish
+      republish_members!
```
mjgiarlo, line 18: I assume we want to republish members when a collection is published and when it is unpublished, hence the new logic living before we branch off.

app/services/publish/metadata_transfer_service.rb
```diff
@@ -48,6 +49,16 @@ def discoverable?
       cocina_object.access.view != 'dark'
     end
 
+    def republish_members!
+      return unless cocina_object&.collection?
+
+      Array.wrap(
+        MemberService.for(cocina_object.externalIdentifier, exclude_opened: true, only_published: true)
```
mjgiarlo, line 56: I assume we only want to republish non-opened members that are already published? 🤷🏻‍♂️

app/services/publish/metadata_transfer_service.rb
```diff
@@ -77,7 +88,7 @@ def publish_delete_on_success
     end
 
     def purl_services_url
-      raise 'You have not configured perl-fetcher (Settings.purl_services_url).' unless Settings.purl_services_url
+      raise 'You have not configured purl-fetcher (Settings.purl_services_url).' unless Settings.purl_services_url
```
mjgiarlo, line 91: Typo.

spec/services/publish/metadata_transfer_service_spec.rb
```diff
@@ -69,7 +118,7 @@
       end
     end
 
-    context 'copies to the document cache' do
+    describe 'copies to the document cache' do
```
mjgiarlo, line 121: More idiomatic RSpec usage.

spec/services/publish/metadata_transfer_service_spec.rb
```diff
@@ -116,24 +165,14 @@
...
-          #                                type: Cocina::Models::ObjectType.collection,
-          #                                label: 'some collection object',
-          #                                version: 1,
-          #                                description: {
-          #                                  title: [{ value: 'Constituent label &amp; A Special character' }],
-          #                                  purl: 'https://purl.stanford.edu/xh235dd9059'
-          #                                },
-          #                                identification: { sourceId: 'sul:123' },
-          #                                access: { view: 'world' },
-          #                                administrative: { hasAdminPolicy: 'druid:fg890hx1234' })
```
mjgiarlo, line 129: Smells like cruft.

spec/services/publish/metadata_transfer_service_spec.rb
```diff
@@ -172,7 +211,7 @@
       end
 
       it 'writes empty notification file' do
-        expect { notify }.to raise_error 'You have not configured perl-fetcher (Settings.purl_services_url).'
+        expect { notify }.to raise_error 'You have not configured purl-fetcher (Settings.purl_services_url).'
```
mjgiarlo, line 214: Typo.

app/services/publish/metadata_transfer_service.rb
jcoyne: @andrewjbtw can you answer this question?  When a collection is unpublished, do republish the members?

andrewjbtw: As long as the members are published, republish the members even if the collection itself is not published. It's an unusual state but the Purl will show the collection name even if the collection object isn't published, so that title should still be updated.

mjgiarlo: OK, great, thanks!

app/services/publish/metadata_transfer_service.rb
andrewjbtw: Yes - we should apply this to items that have already been published, which means they show the old name. Non-published items are either Registered or their rights are dark (but I don't think we need to exclude that case since the publish step handles dark rights appropriately already?)

PR #4226, opened 2022-09-12
new report: edtf problem dates; more or less in general, and when structuredValue

### LINKED ISSUE #4218: Update report for invalid EDTF dates to include structuredValue
The report should catch the following data patterns.

1. structuredValue where entire date is EDTF encoded and either or both of the values is not valid EDTF.
```
{ event: [ { date: [ { structuredValue : [ { value: "123", type: "start"}, { value: "456", type: "end"} ], encoding: { code: "edtf" } ] } ] }
```

2. structuredValue where only one of the values has encoding indicated as EDTF, *even if* that date is valid EDTF, so that I can remediate those to apply the encoding to the entire date range.
```
{ event: [ { date: [ { structuredValue : [ { value: "0123", type: "start", encoding: { code: "edtf" } }, { value: "0456", type: "end" } ] } ] }
```

Test record vz332dg7306 should produce two report results.

ndushay:

Fixes #4218  - hopefully finding all remaining problem children for edtf dates.  But perhaps not.

I ran the report successfully on sdr-infra.

### DIFF
--- a/app/reports/invalid_edtf_dates.rb
+++ b/app/reports/invalid_edtf_dates.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+# Invoke via:
+# bin/rails r -e production "InvalidEdtfDates.report"
+class InvalidEdtfDates
+  # NOTE: Prefer strict JSON querying over lax when using the `.**` operator, per
+  #       https://www.postgresql.org/docs/14/functions-json.html#STRICT-AND-LAX-MODES
+  #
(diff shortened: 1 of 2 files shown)


peetucket:
Thanks for doing this report.  I presume the report results picked up some known bad dates at least?

peetucket (approved):

ndushay:
@peetucket the reports found about 400 edtf date problems in 2 collections - both papyrus collections.

PR #4231, opened 2022-09-13
improve invalid date reports

ndushay:

- to clarify why there are 2 different invalid_edtf_date reports
- to replace the bad_xxx_date reports with a different implementation that runs much more efficiently
- to trap for invalid values for date encodings that were missed in the bad_xxx_date reports


app/reports/bad_edtf_dates.rb
```diff
@@ -1,54 +0,0 @@
-# frozen_string_literal: true
-
-# Invoke via:
-# bin/rails r -e production "BadEdtfDates.report"
-class BadEdtfDates
-  def self.report
-    puts "item_druid,collection_druid,catkey,invalid_values\n"
-
-    Dro.where("jsonb_path_exists(description, '$.**.date.encoding.code ? (@ ==  \"edtf\")')").find_each do |dro|
```
ndushay, line 9: the missing values were because "encoding" is not always a direct child of "date" -- it can be inside a structuredValue, for example.

ndushay, line 9: this implementation approach takes much longer to execute than the one in the invalid_xxx_reports (which I shamelessly cribbed from other invalid_xxx reports)

justinlittman (approved):

Issue #4242, opened 2022-09-22
reports on nesting of structural Value and parallelValue


ndushay:
We are interested in learning more about the shape of our data -- how deeply nested is it?  

Reports desired
- [ ] structuredValue within structuredValue 
- [ ] structuredValue within parallelValue
- [ ] parallelValue within structuredValue
- [ ] parallelValue within parallelValue

Report should include:

- druid, 
- catkey if there is one,
- collection druid, 
- collection name, 
- parent property of the outer elements if possible (e.g. "contributor->name" if there is a structuredValue within a parallelValue inside contributor name.)

This is all with an eye to:  are there validations we can employ to restrict the levels of nesting - to keep our descriptive data rich but not arbitrarily complex?

justinlittman:
@ndushay Just trying to wrap my head around the context for this: Why do we care how deeply nested it is? Why would we want to restrict this? Who/what do we think might produce deeply nested content? What is the problem that we're trying to solve?

ndushay:
- structuredValue contains structuredValue
  - collections
  - dros 
- structuredValue contains parallelValue
  - collections
  - dros 
 - parallelValue contains parallelValue
  - collections
  - dros 
- parallelValue contains parallelValue
  - collections
  - dros

ndushay:
> @ndushay Just trying to wrap my head around the context for this: Why do we care how deeply nested it is? Why would we want to restrict this? Who/what do we think might produce deeply nested content? What is the problem that we're trying to solve?

@justinlittman:  I firmly believe that the arbitrary nesting levels in our descriptive metadata "smell."   Writing reports to find values when there can be arbitrary nesting is difficult.  Writing the mappings to and from mods were difficult.   I think, ultimately, it would be great to have validations that restrict structuredValue and parallelValue nesting levels as much as Arcadia thinks is practical.

These reports are about finding out where we are now, which can help us determine what is practical.

ndushay:
ew! 
![image](https://user-images.githubusercontent.com/96775/192385813-e4a9aa01-e565-4c93-8b2e-0932bc6698cb.png)

ndushay:
- structuredValue contains structuredValue
   - collections [collectionsStructuredStructured.csv](https://github.com/sul-dlss/dor-services-app/files/9636423/collectionsStructuredStructured.csv) - 68 lines, including header
[...]

PR #4247, opened 2022-09-26
add reports property_contains_property for dros, collections

ndushay:

- To create a general tool that can be used by devs or @arcadiafalcone for this purpose.
- For #4242

### DIFF
--- a/app/reports/property_contains_property_collections.rb
+++ b/app/reports/property_contains_property_collections.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+# Report on properties containing another specified property
+
+# Invoke via:
+# bin/rails r -e production "PropertyContainsPropertyCollections.report"
+class PropertyContainsPropertyCollections
+  # NOTE: Prefer strict JSON querying over lax when using the `.**` operator, per
+  #       https://www.postgresql.org/docs/14/functions-json.html#STRICT-AND-LAX-MODES
+  #
+  # > The .** accessor can lead to surprising results when using the lax mode.
+  # > ... This happens because the .** accessor selects both the segments array
+  # > and each of its elements, while the .HR accessor automatically unwraps
+  # > arrays when using the lax mode. To avoid surprising results, we recommend
+  # > using the .** accessor only in the strict mode.
+  # {1 TO LAST} is needed if you're looking for a property inside the same property
+  #   (e.g. structuredValue inside structuredValue), as .** includes itself
+  #   see table 8.25 here https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING
+  JSON_PATH = 'strict $.**.parallelValue ? (exists(@.**{1 TO LAST}.structuredValue ? (@.size() > 0)))'
+  SQL = <<~SQL.squish.freeze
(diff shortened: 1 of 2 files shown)


app/reports/property_contains_property_collections.rb
```diff
@@ -0,0 +1,42 @@
...
+class PropertyContainsPropertyCollections
+  # NOTE: Prefer strict JSON querying over lax when using the `.**` operator, per
+  #       https://www.postgresql.org/docs/14/functions-json.html#STRICT-AND-LAX-MODES
+  #
+  # > The .** accessor can lead to surprising results when using the lax mode.
+  # > ... This happens because the .** accessor selects both the segments array
+  # > and each of its elements, while the .HR accessor automatically unwraps
+  # > arrays when using the lax mode. To avoid surprising results, we recommend
+  # > using the .** accessor only in the strict mode.
+  JSON_PATH = 'strict $.**.parallelValue ? (exists(@.**{1 TO LAST}.structuredValue ? (@.size() > 0)))'
```
mjgiarlo, line 16: So... give me a list of parallel values that have one or more structured values within them?

mjgiarlo: And if that's right... is [collectionsParallelStructured.csv](https://github.com/sul-dlss/dor-services-app/files/9649313/collectionsParallelStructured.csv) the related report?

ndushay: > So... give me a list of parallel values that have one or more structured values within them?

^^ is correct.

I pipe the output to a specified csv file;   so that `[collectionsParallelStructured.csv](https://github.com/sul-dlss/dor-services-app/files/9649313/collectionsParallelStructured.csv)` IS the report run for this file as it appears here, but you can imagine how easy it is to change the properties shown to get other configurations.

mjgiarlo: Ah, I see. Thank you!

mjgiarlo, line 16: @ndushay I wonder if this path is more or less equivalent to `strict $.**.parallelValue.**.structuredValue ? (@.size() > 0)` ❓ FWIW, when I change the path to this shorter, simpler one, it returns exactly the same results (ran against collections, not DROs). Not sure if that means they're equivalent or just equivalent enough given our current collection metadata usage.

mjgiarlo: @ndushay as in... it under-reports or over-reports? (Bear with me! I'm not trying to nitpick what you've already done so much as understand how we're using these paths going forward, as someone who likes to pick up these report tickets.) I'm seeing that path return a 69-line CSV.

ndushay: `.**` overreports because it reports a single structuredValue, indistinguishable from the structuredValue inside a structuredValue.   Hence the `1 TO LAST` thing - it's a zero based index.

ndushay: I added a comment.

mjgiarlo (approved):

app/reports/property_contains_property_collections.rb
mjgiarlo: thx!

PR #4271, opened 2022-10-17
Change workspace cleanup to async.

justinlittman:
refs https://github.com/sul-dlss/argo/issues/2752

Avoid timeouts

Unit

### DIFF
--- a/app/services/cleanup_service.rb
+++ b/app/services/cleanup_service.rb
@@ -22,19 +22,11 @@ class CleanupService
   # @param [String] druid The identifier for the object whose data is to be removed
   # @return [void] remove copy of the data that was exported to preservation core
   def self.cleanup_export(druid)
-    id = druid.split(':').last
+    id = druid.delete_prefix('druid:')
     bag_dir = File.join(Settings.cleanup.local_export_home, id)
-    remove_branch(bag_dir)
+    FileUtils.rm_rf(bag_dir)
     tarfile = "#{bag_dir}.tar"
-    remove_branch(tarfile)
+    FileUtils.rm_f(tarfile)
   end
   private_class_method :cleanup_export
-
-  # @param [Pathname,String] pathname The full path of the branch to be removed
-  # @return [void] Remove the specified directory and all its children
-  def self.remove_branch(pathname)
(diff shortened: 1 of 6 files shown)


app/services/cleanup_service.rb
```diff
@@ -22,19 +22,11 @@ def self.cleanup_workspace_content(druid, base)
...
     bag_dir = File.join(Settings.cleanup.local_export_home, id)
-    remove_branch(bag_dir)
+    FileUtils.rm_rf(bag_dir)
     tarfile = "#{bag_dir}.tar"
-    remove_branch(tarfile)
+    FileUtils.rm_f(tarfile)
   end
   private_class_method :cleanup_export
-
-  # @param [Pathname,String] pathname The full path of the branch to be removed
```
justinlittman, line 33: With the original code, when invoked from a job, an error was being thrown on `pathname.rmtree` claiming that the file does not exist. Logging confirmed that `pathname.exist?` returns true.

My theory is that the Reset Workspace workflow step (prior to Clean Workspace) is deleting the directory. However, when Clean Workspace was changed to a job, it is now performed on a different server (the worker server) and there is an NFS sync latency. (Thus, this service thinks that the directory exists, but when it goes to delete the directory, it doesn't exist.)

This changes side-steps the problem by using a forgiving delete.

spec/jobs/cleanup_job_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+require 'rails_helper'
+
+RSpec.describe CleanupJob, type: :job do
+  subject(:perform) do
+    described_class.perform_now(druid:, background_job_result: result, workflow:)
+  end
+
+  let(:druid) { 'druid:mk420bs7601' }
+  let(:result) { create(:background_job_result) }
+  # let(:item) { instance_double(Cocina::Models::DRO) }
```
lwrubel, line 12: Cruft?

lwrubel (approved):
Approved, noting some possible cruft.

spec/jobs/cleanup_job_spec.rb
justinlittman: Yup. Thanks.

PR #4275, opened 2022-10-27
Adds new rubocops.

justinlittman:
All the rubos.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -88,6 +88,12 @@ RSpec/ExpectActual:
 RSpec/NestedGroups:
   Max: 5
 
+Style/RedundantStringEscape:
+  Enabled: true
+  Exclude:
+    # Does not play well with raw SQL.
+    - 'app/reports/**'
+
 Style/WordArray:
   MinSize: 3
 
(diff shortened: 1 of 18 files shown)


app/reports/invalid_form_uris.rb
```diff
@@ -5,14 +5,14 @@
 class InvalidFormUris
   JSON_PATH = '$.**.form.uri'
   SQL = <<~SQL.squish.freeze
-    SELECT (jsonb_path_query_array(description, '#{JSON_PATH} ? (@ like_regex "^.*\.html$")') ||
```
ndushay, line 8: actually, I think the period here is meant to be a real period, not an anything char.

app/reports/invalid_form_uris.rb
```diff
@@ -5,14 +5,14 @@
...
+    SELECT (jsonb_path_query_array(description, '#{JSON_PATH} ? (@ like_regex "^.*.html$")') ||
             jsonb_path_query_array(description, '#{JSON_PATH} ? (@ like_regex "^(?!https?://).*$")')) ->> 0 as value,
            external_identifier,
            jsonb_path_query(identification, '$.catalogLinks[*] ? (@.catalog == "symphony").catalogRecordId') ->> 0 as catkey,
            jsonb_path_query(structural, '$.isMemberOf') ->> 0 as collection_id
            FROM "dros" WHERE
            (jsonb_path_exists(description, '#{JSON_PATH} ? (@ like_regex "^(?!https?://).*$")') OR
-           jsonb_path_exists(description, '#{JSON_PATH} ? (@ like_regex "^.*\.html$")'))
```
ndushay, line 15: ditto here...

app/reports/invalid_language_uris.rb
```diff
@@ -7,18 +7,18 @@ class InvalidLanguageUris
...
             jsonb_path_query_array(description, '#{JSON_PATH2} ? (@ like_regex "^(?!https?://).*$")')) ->> 0 as contrib,
            external_identifier,
            jsonb_path_query(identification, '$.catalogLinks[*] ? (@.catalog == "symphony").catalogRecordId') ->> 0 as catkey,
            jsonb_path_query(structural, '$.isMemberOf') ->> 0 as collection_id
            FROM "dros" WHERE
            (jsonb_path_exists(description, '#{JSON_PATH1} ? (@ like_regex "^(?!https?://).*$")') OR
-           jsonb_path_exists(description, '#{JSON_PATH1} ? (@ like_regex "^.*\.html$")') OR
+           jsonb_path_exists(description, '#{JSON_PATH1} ? (@ like_regex "^.*.html$")') OR
            jsonb_path_exists(description, '#{JSON_PATH2} ? (@ like_regex "^(?!https?://).*$")') OR
-           jsonb_path_exists(description, '#{JSON_PATH2} ? (@ like_regex "^.*\.html$")')
```
ndushay, line 21: and here

ndushay (changes requested):
I think the "\." in these cases are intentional - it should be a period, not an anything char?

justinlittman:
Great catch @ndushay! I excluded that rule for all reports.

ndushay (approved):

PR #4317, opened 2023-01-26
don't add collection info to 856 when collection not released to searchworks

### LINKED ISSUE #4297: When releasing to SearchWorks using an 856 field, do not include the collection if the collection has not been released
This item has a catkey and is part of an SDR collection: https://searchworks.stanford.edu/view/14283268

The item page in SearchWorks looks ok but something is wrong with the display when the record is shown in SearchWorks search results:

https://searchworks.stanford.edu/?search_field=search&q=14283268

The link to the "Acquisitions Serials" collection should not be shown there because that collection has not been released to SearchWorks, only items have been released. The link is also broken. 

Even though this is a SearchWorks display issues, I'm suspicious that the SW display is being driven by SDR data, specifically, the data that SDR puts in the MARC record.

In the MARC record, the collection name is visible in the most recent 856 fields. Each field represents a different issue and the late 2022 issues all show the collection name
```
856 4 1 z| Available to Stanford-affiliated users. u| https://purl.stanford.edu/mv522rq4275 x| SDR-PURL x| item x| collection:gk894yk3598::Acquisitions Serials x| label:No. 150, Aug/Sept 2022 x| sort:150 x| rights:group=stanford
```
Earlier issues do not have the collection name in the 856:
```
856 4 1 z| Available to Stanford-affiliated users. u| https://purl.stanford.edu/nc476cj7267 x| SDR-PURL x| item x| label:No. 149, Jun/Jul 2022 x| sort:149 x| rights:group=stanford
```
Since the collection druid is not released, the collection name should not be part of the 856 data. SearchWorks should not even "know" that the collection exists.

**Additional information**
The fact that the collection wasn't included in these releases until the second half of 2022 makes me wonder if there was previously a check performed on SW releases that determined whether an item was in a collection that had been released or a collection that had not been released, and then included the collection only if the collection had been released.

There's no guarantee that removing the collection name will actually fix the SW display, but it still seems to be the right thing to do.

peetucket:

Fixes #4297 - do not add collection info to 856 if collection not released to searchworks

I don't believe we checked for this in the past (but perhaps it is an uncommon occurrence).  The code used to be in `app/models/dor/update_marc_record_service.rb` before it was moved sometime mid-2022, but digging back to spring 2022, the code looks basically the same as it does now.

~~[HOLD] until we know this answer to this question: do we skip if the collection has *no* release tags (also implies not released), or only skip if the collection explicitly has a "false" release tag (which is what the example druid in the ticket has, item: https://argo.stanford.edu/view/druid:mv522rq4275, associated collection: https://argo.stanford.edu/view/druid:gk894yk3598)~~  validated

Updated tests

### DIFF
--- a/app/services/update_marc_record_service.rb
+++ b/app/services/update_marc_record_service.rb
@@ -60,7 +60,7 @@ class UpdateMarcRecordService
     # now add the current ckey
     if @cocina_object.identification.catalogLinks.find { |link| link.catalog == 'symphony' }.present?
       catalog_record_id = @cocina_object.identification.catalogLinks.find { |link| link.catalog == 'symphony' }.catalogRecordId
-      records << (released_to_searchworks? ? new_856_record(catalog_record_id) : get_identifier(catalog_record_id))
+      records << (released_to_searchworks?(@cocina_object) ? new_856_record(catalog_record_id) : get_identifier(catalog_record_id))
     end
 
     records
@@ -150,6 +150,8 @@ class UpdateMarcRecordService
 
     collections.each do |collection_druid|
       collection = CocinaObjectStore.find(collection_druid)
+      next unless released_to_searchworks?(collection)
+
       catkey = collection.identification&.catalogLinks&.find { |link| link.catalog == 'symphony' }
       collection_info += "|xcollection:#{collection.externalIdentifier.sub('druid:',
                                                                            '')}:#{catkey&.catalogRecordId}:#{Cocina::Models::Builders::TitleBuilder.build(collection.description.title)}"
@@ -191,16 +193,13 @@ class UpdateMarcRecordService
     BORN_DIGITAL_APOS.include? @cocina_object.administrative.hasAdminPolicy
   end
 
-  def released_to_searchworks?
+  def released_to_searchworks?(cocina_object)
+    released_for = ::ReleaseTags.for(cocina_object:)
     rel = released_for.transform_keys { |key| key.to_s.upcase } # upcase all release tags to make the check case insensitive
     rel.dig('SEARCHWORKS', 'release').presence || false
   end
 
   private
-
-  def released_for
-    ::ReleaseTags.for(cocina_object: @cocina_object)
-  end
 
   # adapted from mods_display
   def parts_delimiter(elements)
--- a/spec/services/update_marc_record_service_spec.rb
+++ b/spec/services/update_marc_record_service_spec.rb
@@ -506,12 +506,42 @@ RSpec.describe UpdateMarcRecordService do
       end
     end
 
-    context 'when an object with a collection' do
+    context 'when an object with a collection has no release tags' do
       let(:cocina_object) do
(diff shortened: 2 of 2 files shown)


app/services/update_marc_record_service.rb
```diff
@@ -191,17 +193,14 @@ def born_digital?
...
     rel.dig('SEARCHWORKS', 'release').presence || false
   end
 
   private
 
-  def released_for
-    ::ReleaseTags.for(cocina_object: @cocina_object)
-  end
-
   # adapted from mods_display
```
peetucket, line 204: method only used in one spot, method above, so just combining for simplicity

app/services/update_marc_record_service.rb
```diff
@@ -191,17 +193,14 @@ def born_digital?
     BORN_DIGITAL_APOS.include? @cocina_object.administrative.hasAdminPolicy
   end
 
-  def released_to_searchworks?
+  def released_to_searchworks?(cocina_object)
+    released_for = ::ReleaseTags.for(cocina_object:)
     rel = released_for.transform_keys { |key| key.to_s.upcase } # upcase all release tags to make the check case insensitive
     rel.dig('SEARCHWORKS', 'release').presence || false
   end
```
peetucket, line 201: modifying this method so we can check both the object itself, and any collections it is in

mjgiarlo (approved):

spec/services/update_marc_record_service_spec.rb
```diff
@@ -506,13 +506,43 @@
       end
     end
 
-    context 'when an object with a collection' do
+    context 'when an object with a collection has no release tags' do
       let(:cocina_object) do
```
peetucket, line 510: default test setup mocks *no* release tags

mjgiarlo:
@peetucket 💬 
> Question: do we skip if the collection has _no_ release tags (also implies not released), or only skip if the collection explicitly has a "false" release tag (which is what the example druid in the ticket has, item: [argo.stanford.edu/view/druid:mv522rq4275](https://argo.stanford.edu/view/druid:mv522rq4275), associated collection: [argo.stanford.edu/view/druid:gk894yk3598](https://argo.stanford.edu/view/druid:gk894yk3598))

Do we need this question answered before merging the PR?

peetucket:
> @peetucket 💬
> 
> > Question: do we skip if the collection has _no_ release tags (also implies not released), or only skip if the collection explicitly has a "false" release tag (which is what the example druid in the ticket has, item: [argo.stanford.edu/view/druid:mv522rq4275](https://argo.stanford.edu/view/druid:mv522rq4275), associated collection: [argo.stanford.edu/view/druid:gk894yk3598](https://argo.stanford.edu/view/druid:gk894yk3598))
> 
> Do we need this question answered before merging the PR?

Yes, I'll ask Andrew at standup.  I'll mark the PR as [HOLD] until we know.  If the answer is that we only skip if explicit "false" release (which I think is unlikely), then this PR will need some changes.

PR #4319, opened 2023-01-26
Add honeybadger check-ins to cron jobs

### LINKED ISSUE #4293: use Honeybadger checkins to ensure crons are running
See https://docs.honeybadger.io/guides/check-ins/

This approach would be a fine way to ensure our cron jobs are running. Note similar work in prescat: PR https://github.com/sul-dlss/preservation_catalog/pull/2075

mjgiarlo:

Fixes #4293

This commit uses honeybadger check-ins to ensure regularly scheduled cron jobs are running, like the approach already being used in prescat and H2.

CI

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -21,14 +21,25 @@
 
 # Learn more: http://github.com/javan/whenever
 
+# Load rubyconfig gem so that we have access to env-specific settings
+require 'config'
+
+Config.load_and_set_settings(Config.setting_files('config', 'production'))
+
+# These define jobs that checkin with Honeybadger.
+# If changing the schedule of one of these jobs, also update at https://app.honeybadger.io/projects/50568/check_ins
+job_type :rake, "cd :path && :environment_variable=:environment bundle exec rake :task --silent :output && curl 'https://api.honeybadger.io/v1/check_in/:check_in"
+
 # Suppress warnings to avoid unnecessary cron emails.
 env 'RUBYOPT', '-W0'
 
 every :day, at: '2:16am' do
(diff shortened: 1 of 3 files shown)


aaron-collier (approved):

config/schedule.rb
```diff
@@ -21,14 +21,25 @@
...
 # Learn more: http://github.com/javan/whenever
 
+# Load rubyconfig gem so that we have access to env-specific settings
+require 'config'
+
+Config.load_and_set_settings(Config.setting_files('config', 'production'))
+
+# These define jobs that checkin with Honeybadger.
+# If changing the schedule of one of these jobs, also update at https://app.honeybadger.io/projects/50568/check_ins
+job_type :rake, "cd :path && :environment_variable=:environment bundle exec rake :task --silent :output && curl 'https://api.honeybadger.io/v1/check_in/:check_in"
```
lwrubel, line 31: This is already closed and merged, @mjgiarlo, but wondering--are the quotation marks alright in this one (right in front of `https`)? Was comparing to https://github.com/sul-dlss/preservation_catalog/blob/5fb451cdf117430eb6ec0f3c33689d1cb639aaf0/config/schedule.rb#L14. I haven't worked with these before but I think I remember some questions about singe/double quotes coming up before.

mjgiarlo: Good eye! I was cribbing from an PR that had a bug in it. 😄

mjgiarlo: @lwrubel https://github.com/sul-dlss/dor-services-app/pull/4320

Issue #4327, opened 2023-01-30
Remediate Cocina with barcode only, no catkey


arcadiafalcone:
FOLIO HRIDs will be based on Symphony catkeys, not barcodes, so the catkeys must be added to the Cocina in order to also add the HRID.
 @andrewjbtw identified 581 such objects: ~~https://stanford.enterprise.slack.com/files/WLPPXB54G/F04LRQRRQGP/only-barcodes-with-title.csv~~  (arcadia supplied report below in a comment).

The action for this ticket is to take Arcadia's list below and determine the ckeys for each barcode.  Output of druid, barcode, catkey would be helpful;  ask Arcadia if she wants more than that.   If it gets more complicated, you might be able to leverage the "reports" approach to find other info, like collection or collection title or whatever.

Read comments for more info.

arcadiafalcone:
Blocks #4326

justinlittman:
* The slack link doesn't work.
* Can you give hints as to how this is supposed to happen?

arcadiafalcone:
[only-barcodes-with-title.csv](https://github.com/sul-dlss/dor-services-app/files/10677237/only-barcodes-with-title.csv)

I was thinking to take the barcodes and use them to look up the corresponding catkey in Symphony, then add those catkeys to the identification metadata in the objects. This might be possible via Symphony reporting, but I'm not familiar enough with that to have more details.

ndushay:
The code to get a ckey from a barcode is in `SymphonyReader.fetch_catkey`

andrewjbtw:
Maybe I'm extra cautious but I think it would be good to check that the catkey retrieved from the barcode matches the existing title, just in case there was some data issue where a barcode was changed.

And since I suggested it, I am also willing to go through the list. I could do it from a report with: druid, barcode, ckey, title from Symphony, title currently in SDR

peetucket:
I used the `SymphonyReader` class to lookup the catkeys for all barcodes attached, and then used the druid to lookup the SDR label in the cocina object.  The resulting report is attached here in this comment and includes the original columns, plus a `catkey` column and an `sdr_label` column.

There was one barcode which could not be found in symphony: `36105006174440` and thus no catkey was returned.

Question:
- In this report, I am outputing the `label` attribute of the cocina object instead of digging into the descriptive metadata in cocina.  I am not sure if this is producing the expected output.
[...]

PR #4351, opened 2023-02-07
[WIP] add rake task to lookup catkey from barcode and add to cocina

peetucket:

To deal with remediation identified in #4327 

Not sure if we need this merged if it's only a one-time occurrence

1. Report rake task run against actual prod data
2. Update rake task run against two objects in QA to verify opening/closing/updating works as expected

### DIFF
--- a/lib/tasks/missing_catkeys.rake
+++ b/lib/tasks/missing_catkeys.rake
@@ -1,0 +1,118 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+# rubocop:disable Metrics/BlockLength
+# see https://github.com/sul-dlss/dor-services-app/issues/4327
+namespace :catkeys do
+  desc 'Lookup catkey from barcode and create a new output file that includes the catkey'
+  # requires input file of CSV with three columns with lowercase headers: druid,barcode,title
+  # writes out a new CSV file with the same columsn but adds the catkey if available
+  # rake catkeys:lookup['tmp/only-barcodes-with-title.csv']
+  task :lookup, %i[input_file] => :environment do |_t, args|
+    input_file = args[:input_file]
+    raise 'input_file not specified' unless input_file
+
+    output_file = input_file.gsub('.csv', '_output.csv')
+
+    rows = CSV.parse(File.read(input_file), headers: true)
+    num_rows = rows.size
+
+    puts "Input file: #{input_file}"
+    puts "Output file: #{output_file}"
+    puts "Num rows: #{num_rows}"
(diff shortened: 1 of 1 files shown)


ndushay:
Not sure it will be a one off, a "few off" or what.  there may be more of these to do before Folio cutover.  

@arcadiafalcone @andrewjbtw - do you think we'll need ckey from barcode from Symphony for a bit yet?

I suppose it also depends on how this will appear in the SDR writ large going forward -- part of Argo registration?  Do we have a ticket for that?   Will we ever be doing it for more than one record at a time?

lib/tasks/missing_catkeys.rake
```diff
@@ -0,0 +1,96 @@
...
+      catkey = row['catkey']
+      puts "#{i + 1} of #{num_rows} : #{druid}"
+      begin
+        cocina_object = CocinaObjectStore.find(druid_with_prefix)
+        catkey_exists = cocina_object.identification&.catalogLinks&.any? { |link| link.catalog == 'symphony' && link.catalogRecordId == catkey }
+        if catkey_exists
+          num_catkey_exists += 1
+        else
+          updated_cocina_object = cocina_object.dup
+          catalog_link = Cocina::Models::CatalogLink.new(catalog: 'symphony', refresh: true, catalogRecordId: catkey)
```
justinlittman, line 80: Are you sure refresh should be true?

mjgiarlo:
If we're not sure this will be a one-off (or "few-off" 😃), I'd be inclined to leave this up as a draft PR and if we wind up re-running it two more times, that might be evidence that it's time to add this to DSA. If we do go that route, my code review in summary would be "move most / all of the logic out of the rake tasks into one or more service classes, so that we can test it more easily, and minimize the amount of hard-to-test code in the rake tasks themselves."

justinlittman:
Concur with @mjgiarlo.

peetucket:
Added commit that changes a publish after update to an open/close, which kicks off accessioning and therefore publishes as part of this.  Tested with a couple objects on QA and it works as expected.

lib/tasks/missing_catkeys.rake
arcadiafalcone: Refresh should be set to true because in this case we are adding catkeys to objects that don't already have them, so we know that this is the primary catkey and that they're not using the digital serials model.

mjgiarlo:
@peetucket Do you think we can close this now that we're on Folio?

peetucket:
Seems like it's not useful anymore, so yes.

Issue #4349, opened 2023-02-07
allow for bulk update of existing cocina data


ndushay:
create a framework for being able to do this - to bulk update a bunch of existing cocina data.

ndushay:
@justinlittman I think this ticket about having a way to do bulk updates of cocina data is done, so I am closing it.  Whee!

PR #4368, opened 2023-02-13
Adds support for performing cocina migrations.

justinlittman:
refs #4349

Provide framework for performing cocina data migrations.

sdr-infra

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,3 +162,27 @@ Missing druids can be indexed with:
 ```
 RAILS_ENV=production bundle exec rake missing_druids:index_unindexed_objects
 ```
+
+## Data migrations / bulk remediations
+`bin/migrate-cocina` provides a framework for data migrations and bulk remediations. It supports optional versioning and publishing of objects after migration.
+
+```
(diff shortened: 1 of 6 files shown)


ndushay (commented):
Given that we are likely going to be migrating cocina data from time to time, it seems sensible to write much of the code in a ?? service ?? class and have the rake task be skinny.  This would allow for test code, no?

justinlittman:
With this approach, the logic of the migration is in the Migrator::* class, which can be tested. The mechanics of the migration are in the executable script.

peetucket (approved):

ndushay (commented):
I'm sorry I didn't get to reviewing this yesterday.

A few thoughts, belatedly:
1.  "exemplar" is a better name for the test class you wrote;  "test" is an overloaded term
2.  For base class, add the comments that are obious to you:
- "override this method for ..."
3.  have specs for what happens when you do and don't override "publish" and "version" and "version_description"

Would you like me to put up a quick PR for these?

justinlittman:
Sure, since the proposed changes are probably clearer to you.

PR #4389, opened 2023-02-17
extract and separate classes for generating marc and writing symphony files

### LINKED ISSUE #4377: Extract services from UpdateMarcRecordService
Currently, the UpdateMarcRecordService (1) generates a MARC record and (2) delivers it to Symphony. 

In preparation of delivering to Folio, extract the following classes from UpdateMarcRecordService:
1. MarcGenerator, which generates the MARC record.
2. SymphonyWriter, which delivers a MARC record to Symphony.

UpdateMarcRecordService (and no other class) should invoke these new services.

peetucket:

Fixes #4377 - extract code into separate classes that are responsible for creating marc and writing symphony, so that presumably later on we can use the same marc generation code but a separate mechanism for sending the data to Folio

Of note:
- some refactoring was done to put the responsibility for determining if any marc 856 records exist into just the marc 856 generation class ... therefore we do not need to dig into the cocina object at all in the `UpdateMarc856RecordService` class anymore (e.g. to look to see if there are ckeys)
- removed some code duplication by creating method for fetching ckeys (either current or previous) and then calling these new methods when needed
- renamed/updated references in this code from "marc" to make it clear it is "marc 856", including the script and settings 
- verified that the settings changed here are *not* in any shared_configs, so this change is sufficient

~~HOLD for a question about if we want to generate records in the case where there are *only* previous ckeys and no current ckeys:~~
~~- the current code does not write out any marc records if there are no current ckeys (e.g. if there is a record that has ONLY "previous symphony" ckeys, this record will not have a record written to disk): https://github.com/sul-dlss/dor-services-app/blob/main/app/services/update_marc_record_service.rb#L22-L30~~
~~- however, the current code also tests that we can generate blank rows for this scenario: https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/update_marc_record_service_spec.rb#L250-L275   .... these will never be written to disk (unless there is also a current ckey)~~

~~The refactored code here will not work in this same way (as I was hoping to not have inspect the ckeys again to decide whether to write a file or not).  In other words, `MarcGenerator.create` will happily create a records when only previous ckeys are available (the same test ported over to the new spec I created) just as it does now, but my refactored `UpdateMarcRecordService` class will write any records it returns (again, instead of checking the ckeys in cocina like the current code does here: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/update_marc_record_service.rb#L22-L30)~~

~~If we want it work like the current code does, then I'd suggest the best course of action is to not generate those blank previous ckey records unless there is also a current ckey.  This would just update the logic a bit in the new `MarcGenerator` class (here: https://github.com/sul-dlss/dor-services-app/pull/4389/files#diff-434e9b20dbee7e77c8ac8d04d6f87ba9c987d7881120601be382a3ccb03d3a4fR38), and then update the tests to match.  So we only need to write a file if marc actually gets generated, and the new `MarcGenerator` class contains all the logic that decides if this is going to happen.  Simpler and cleaner in my view.  The only reason NOT to do this is if we ever wanted to generate marc for some other reason for records that only have previous ckeys (?)~~

Updated specs


app/controllers/objects_controller.rb
```diff
@@ -129,7 +129,7 @@ def unpublish
   end
 
   def update_marc_record
-    UpdateMarcRecordService.new(@cocina_object, thumbnail_service: ThumbnailService.new(@cocina_object)).update
+    UpdateMarcRecordService.update(@cocina_object, thumbnail_service: ThumbnailService.new(@cocina_object))
     head :created
```
peetucket, line 133: Updating this reference to match how we call it elsewhere in the code (`app/services/constituent_service.rb`)

app/services/marc_generator.rb
```diff
@@ -0,0 +1,231 @@
+# frozen_string_literal: true
```
peetucket, line 1: A new class containing all of the logic that starts with a cocina object and ends with a Marc record (extracted from https://github.com/sul-dlss/dor-services-app/blob/main/app/services/update_marc_record_service.rb)

app/services/symphony_writer.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
peetucket, line 1: A new class that takes a marc record and writes it to disk (extracted from https://github.com/sul-dlss/dor-services-app/blob/main/app/services/update_marc_record_service.rb)

app/services/update_marc_record_service.rb
```diff
@@ -1,249 +1,21 @@
 # frozen_string_literal: true
```
peetucket, line 2: This class is now greatly simplified ... it simply calls the new classes to generate a marc record and then write it out to disk.

app/services/marc_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+  #     else use the value of the <contentMetadata> "type" attribute if present, e.g., image, book, file
+  #     else use the value "citation"
+  # Subfield x #4 (optional): the barcode if known (<identityMetadata><otherId name="barcode">, recorded as barcode:barcode-value
+  # Subfield x #5 (optional): the file-id to be used as thumb if available, recorded as file:file-id-value
+  # Subfield x #6..n (optional): Collection(s) this object is a member of, recorded as collection:druid-value:ckey-value:title
+  # Subfield x #7..n (optional): label and part sort keys for the member
+  # Subfield x #8..n (optional): High-level rights summary
+  def create
+    return [] if ckeys.empty? && previous_ckeys.empty?
+
```
peetucket, line 39: no need to do anything if we have no ckeys (same functionally as we have currently, just refactored a bit to avoid some code duplication)

spec/services/update_marc_record_service_spec.rb
```diff
@@ -5,1059 +5,42 @@
 RSpec.describe UpdateMarcRecordService do
   subject(:umrs) { described_class.new(cocina_object, thumbnail_service:) }
```
peetucket, line 7: most of this spec was moved to the specs for the new classes created

this spec now just tests that we call those classes as expected

justinlittman (commented):
I have an additional request: The code calls things "marc" when it is really just 856 fields. This makes the code confusing (in fact, I misunderstood it when I wrote this ticket). Could this be refactored to make clear when a marc and when 856 fields? (For example, SymphonyWriter doesn't write MARC records, it writes 856 fields.)

If you'd prefer this be changed to a separate ticket / PR that's fine though I think it might be just as easy to do here.

peetucket:
> I have an additional request: The code calls things "marc" when it is really just 856 fields. This makes the code confusing (in fact, I misunderstood it when I wrote this ticket). Could this be refactored to make clear when a marc and when 856 fields? (For example, SymphonyWriter doesn't write MARC records, it writes 856 fields.)
> 
> If you'd prefer this be changed to a separate ticket / PR that's fine though I think it might be just as easy to do here.

I can do this here, but will double-check to make sure I understand which case is which

peetucket:
(1) Renamed classes/comments to indicate it is stub 856 for symphony transfer and not a full marc record
(2) made all methods that do not need to be public in the marc 856 generation private, and then updated tests to match.

justinlittman (approved):

PR #4391, opened 2023-02-21
Adds feature for syncing catalog links.

### LINKED ISSUE #4376: Keep Folio CatalogLinks in sync with Symphony CatalogLinks
When a Cocina object is persisted using the UpdateObjectService or the CreateObjectService, the Folio CatalogLinks should be synced with the Symphony CatalogLinks. This action should be controlled by its own feature flag.

To do this, delete all existing Folio CatalogLinks (`catalog` = "folio") and for every Symphony CatalogLink (`catalog` = "symphony") create a corresponding Folio CatalogLink for which the value of `refresh` is the same (true or false), and the `catalogRecordId` is the same except prefixed with an "a".

Previous Symphony CatalogLinks do not need to be synced.

justinlittman:
closes #4376

Allow Symphony and Folio catalog links to peacefully co-exist during the transition.

Unit

### DIFF
--- a/app/services/catalog/add_folio_catalog_links_service.rb
+++ b/app/services/catalog/add_folio_catalog_links_service.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module Catalog
+  # Adds Folio catalog links to cocina object based on symphony catalog links.
+  # This is for use prior to switch over from folio. After that this should be
+  # removed.
+  class AddFolioCatalogLinksService
+    def self.add(cocina_object)
+      new(cocina_object).add
+    end
+
+    def initialize(cocina_object)
+      @cocina_object = cocina_object
+    end
+
+    def add
+      return cocina_object if cocina_object.admin_policy?
+      return cocina_object if cocina_object.identification.catalogLinks.blank?
+
+      cocina_object.new(identification: cocina_object.identification.new(catalogLinks: synced_catalog_links))
+    end
+
+    private
+
+    attr_reader :cocina_object
+
+    # The following catkeys are valid but are from Lane Medical Library records
+    # that will not be migrated. They should remain in the record for now but
+    # no HRIDs should be generated based on them.
+    LANE_NOT_MIGRATING = %w[
+      10872078
+      10906003
+      11574734
+      11718696
+      12208745
+      12307427
+      12718359
+      12811199
+      13638174
+      11827181
+      12311317
+    ].freeze
+
+    def synced_catalog_links
+      catalog_links = []
+      cocina_object.identification.catalogLinks.each do |catalog_link|
+        next if non_lane_folio_link?(catalog_link)
+
+        if migrate_catalog_link?(catalog_link)
+          catalog_links << Cocina::Models::FolioCatalogLink.new(catalog: 'folio', catalogRecordId: "a#{catalog_link.catalogRecordId}",
+                                                                refresh: catalog_link.refresh)
+        end
+        catalog_links << catalog_link
+      end
+      catalog_links
+    end
+
+    def non_lane_folio_link?(catalog_link)
+      # Keep Lane folio catalog links (start with L)
+      catalog_link.catalog == 'folio' && !catalog_link.catalogRecordId.start_with?('L')
+    end
+
+    def migrate_catalog_link?(catalog_link)
+      catalog_link.catalog == 'symphony' && LANE_NOT_MIGRATING.exclude?(catalog_link.catalogRecordId)
+    end
+  end
+end
--- a/app/services/create_object_service.rb
+++ b/app/services/create_object_service.rb
@@ -36,6 +36,7 @@ class CreateObjectService
     updated_cocina_request_object = add_description(updated_cocina_request_object)
     cocina_object = cocina_from_request(updated_cocina_request_object, druid, assign_doi)
     cocina_object = assign_doi(cocina_object) if assign_doi
+    cocina_object = Catalog::AddFolioCatalogLinksService.add(cocina_object) if Settings.enabled_features.sync_cataloglinks
     cocina_object_with_metadata = CocinaObjectStore.store(cocina_object, skip_lock: true)
     add_project_tag(druid, cocina_request_object)
     # This creates version 1.0.0 (Initial Version)
--- a/app/services/update_object_service.rb
+++ b/app/services/update_object_service.rb
@@ -30,6 +30,8 @@ class UpdateObjectService
     # If this is a collection and the title has changed, then reindex the children.
     update_items = need_to_update_members?
 
+    @cocina_object = Catalog::AddFolioCatalogLinksService.add(cocina_object) if Settings.enabled_features.sync_cataloglinks
+
     # Only update if already exists in PG (i.e., added by create or migration).
     cocina_object_with_metadata = CocinaObjectStore.store(cocina_object, skip_lock:)
 
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -6,6 +6,7 @@ enabled_features:
   postgres: false
   file_hierarchy_validation: true
   read_folio: false
+  sync_cataloglinks: false
 
 # Ur Admin Policy
 ur_admin_policy:
--- a/spec/services/catalog/add_folio_catalog_links_service_spec.rb
+++ b/spec/services/catalog/add_folio_catalog_links_service_spec.rb
@@ -1,0 +1,114 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Catalog::AddFolioCatalogLinksService do
+  subject(:synced_object) { described_class.new(cocina_object).add }
+
+  context 'when the object does not have catalog links (AdminPolicy)' do
+    let(:cocina_object) { build(:admin_policy) }
+
+    it 'does not change the object' do
+      expect(synced_object).to eq cocina_object
+    end
+  end
+
+  context 'when the catalog links are empty' do
+    let(:cocina_object) { build(:dro) }
+
+    it 'does not change the object' do
+      expect(synced_object).to eq cocina_object
+    end
+  end
+
+  context 'when there are catalog links' do
+    let(:cocina_object) do
+      build(:dro).new(identification: {
+                        sourceId: 'sul:1234',
+                        catalogLinks: [
+                          { catalog: 'symphony', catalogRecordId: '12345', refresh: true },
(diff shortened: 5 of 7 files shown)


justinlittman:
Blocked by https://github.com/sul-dlss/cocina-models/issues/574

app/services/catalog_link_sync_service.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+# Syncs Symphony catalog links with Folio catalog links
```
ndushay, line 3: add a comment that this should only be in place BEFORE Folio is in place as ILS in production?

app/services/catalog_link_sync_service.rb
```diff
@@ -0,0 +1,63 @@
...
+      if migrate_catalog_link?(catalog_link)
+        catalog_links << Cocina::Models::FolioCatalogLink.new(catalog: 'folio', catalogRecordId: "a#{catalog_link.catalogRecordId}",
+                                                              refresh: catalog_link.refresh)
+      end
+      catalog_links << catalog_link
+    end
+    catalog_links
+  end
+
+  def drop_catalog_link?(catalog_link)
```
ndushay, line 55: I think this method would be more clearly named as "keep_existing_folio_link?"

app/services/catalog_link_sync_service.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def synced_catalog_links
+    catalog_links = []
+    cocina_object.identification.catalogLinks.each do |catalog_link|
+      next if drop_catalog_link?(catalog_link)
+
+      if migrate_catalog_link?(catalog_link)
+        catalog_links << Cocina::Models::FolioCatalogLink.new(catalog: 'folio', catalogRecordId: "a#{catalog_link.catalogRecordId}",
+                                                              refresh: catalog_link.refresh)
+      end
```
ndushay, line 49: so if we have a symphony link, and it's not in the skip list .. then we add a folio link.  But if a record is updated 2 or 3 times, won't we be adding a folio link each time?   

If I understand the code, I *think* this will be ok for objects with multiple catkeys, tho.

app/services/catalog_link_sync_service.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+# Syncs Symphony catalog links with Folio catalog links
+class CatalogLinkSyncService
```
ndushay, line 4: Actualy, this is a service to add missing folio links;  it is not exactly "sync" -- there is no true mirroring of links happening or desired.  Would it be better named "AddFolioCatalogLinks" ?

app/services/catalog_link_sync_service.rb
```diff
@@ -0,0 +1,63 @@
...
+  def self.sync(cocina_object)
+    new(cocina_object).sync
+  end
+
+  def initialize(cocina_object)
+    @cocina_object = cocina_object
+  end
+
+  def sync
```
ndushay, line 13: better name:  add_folio_catalog_links?

app/services/catalog_link_sync_service.rb
```diff
@@ -0,0 +1,63 @@
...
+    12208745
+    12307427
+    12718359
+    12811199
+    13638174
+    11827181
+    12311317
+  ].freeze
+
+  def synced_catalog_links
```
ndushay, line 41: "populate_folio_links" ?

spec/services/catalog_link_sync_service_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  context 'when the catalog links are empty' do
+    let(:cocina_object) { build(:dro) }
+
+    it 'does not change the object' do
+      expect(synced_object).to eq cocina_object
+    end
+  end
+
+  context 'when there are catalog links' do
```
ndushay, line 24: I think there are more cases -

- when record is updated mutiple times (starting without folio link;  starting with folio link)
it will likely be clearer to test these wrinkles with separate tests.

- how about a dro that only has "previous symphony"
- how about a dro that only has "folio" (????) or "previous folio" ???
- collection objects?
- is it correct that agreement and adminPolicy objects are never in Symphony?

any other wrinkles?   Ask @andrewjbtw and @arcadiafalcone ?

ndushay (changes requested):
I think the approach of having create and update objects do this is sound.  

I remain uncomfortable with "sync" as the name for this, as it is really only adding folio identifiers, and only some of the time. 

I have concerns about multiple updates to the same object adding extra folio identifiers, and I also have concerns as to whether we could end up with multiple folio identifiers with refresh=true in those circumstances.

Therefore, even after covering those cases with tests, I think this should be tested by @andrewjbtw and/or Arcadia  to try to trip it up it in qa or stage before this is merged.

app/services/catalog_link_sync_service.rb
justinlittman: Done.

app/services/catalog_link_sync_service.rb
justinlittman: Makes sense. Done.

spec/services/catalog_link_sync_service_spec.rb
justinlittman: I've added tests for (1) when there is no existing Folio link and (2) a collection.

Updating multiple times is not distinct from the existing tests.
Handling of "previous symphony" is covered by the existing tests.
It's not clear what the proper handling of folio-only is because that is not an intended state prior to switchover to Folio.
Agreements are just DROs, so no special consideration required.
I don't know whether AdminPolicies are ever in Symphony. However, in cocina they cannot have catalog links.

app/services/catalog_link_sync_service.rb
justinlittman: It passes judgment on any sort of catalog link, so I don't think adding `folio` is accurate.

It could be renamed to `keep_existing ...`, but then its use becomes `unless ...` instead of `if ...`. The negative has to go somewhere, I'm not sure it matters where. If you feel strongly, let me know.

app/services/catalog_link_sync_service.rb
justinlittman: Existing folio links are dropped (`next if drop_catalog_link?(catalog_link)`), so the previous folio link will be dropped and a new folio link will be added.

app/services/catalog_link_sync_service.rb
justinlittman: I just changed it to `add` because `AddFolioCatalogLinksService.add` seems clear enough.

app/services/catalog/add_folio_catalog_links_service.rb
```diff
@@ -0,0 +1,67 @@
...
+        if migrate_catalog_link?(catalog_link)
+          catalog_links << Cocina::Models::FolioCatalogLink.new(catalog: 'folio', catalogRecordId: "a#{catalog_link.catalogRecordId}",
+                                                                refresh: catalog_link.refresh)
+        end
+        catalog_links << catalog_link
+      end
+      catalog_links
+    end
+
+    def drop_catalog_link?(catalog_link)
```
ndushay, line 58: this method should be named `non_lane_folio_link?` - because that is what the logic says

ndushay (changes requested):
Thanks for the refactors.  I have another, better/clearer suggestion, and another possible hole in the logic.  And a quibble in spec description for your consideration.

app/services/catalog/add_folio_catalog_links_service.rb
```diff
@@ -0,0 +1,67 @@
...
+      12811199
+      13638174
+      11827181
+      12311317
+    ].freeze
+
+    def synced_catalog_links
+      catalog_links = []
+      cocina_object.identification.catalogLinks.each do |catalog_link|
+        next if drop_catalog_link?(catalog_link)
```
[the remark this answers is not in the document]
justinlittman: Skipping it means it is dropped, so mission accomplished. Unless I'm missing something?

arcadiafalcone (approved):

arcadiafalcone:
Everything I tested passed.

Not tested:
-collection records (pre-existing bug)
-release workflow

Testing notes:
br092xk8968 - existing ckey, opened (accessioned)
-refresh from source - adds FOLIO HRID with same refresh value
-change catkey - updates FOLIO HRID, no previous_catkey created
-new version, restore previous catkey - updates FOLIO HRID, no previous_catkey created
-change refresh status, both directions - updates FOLIO HRID refresh status
-remove catkey - removes FOLIO HRID, previous_catkey created
-re-add same catkey - adds FOLIO HRID, catkey
-replace with different catkey - adds FOLIO HRID, no previous_catkey created (but would be duplicate)
-add second catkey - adds second FOLIO HRID with same refresh status
-reverse refresh status on two catkeys - updates FOLIO refresh status
-remove second catkey - removes second FOLIO catkey
-FOLIO + symphony + previous symphony versions without error
-remove catkey when previous_catkey present - removes FOLIO HRID
-add catkey when only previous_catkey present - adds FOLIO HRID
-add second catkey and change refresh status - adds FOLIO HRID and updates refresh status

bc941dh2269 - no catkey, accessioned
-add catkey to object that never had catkey - adds FOLIO HRID and refresh status
-add forbidden Lane catkey to object with other catkey - message of successful update, catkey added, FOLIO HRID not added
-remove additional catkey from object with forbidden Lane catkey - message of successful update, catkey becomes previous

bn325xc3747 - no catkey, registered
-add catkey - adds FOLIO HRID and refresh status

ft043bp0584 - registered with catkey, form
-adds FOLIO HRID and refresh status
-pulls in descriptive metadata

nd981qw5082, qn756gy8851 - registered with catkey, tab-delimited
-adds FOLIO HRID and refresh status
-pulls in descriptive metadata
-tracking sheets use catkey label and value

hb877zh8366, qp882tk2097 - registered with catkey, CSV file upload (bulk action)
-adds FOLIO HRID and refresh status
-pulls in descriptive metadata

fm516px0309 - registered with forbidden Lane catkey, form
-adds catkey only
-pulls in descriptive metadata

bc544pb9701 - virtual object
-refreshed from source - FOLIO HRID and refresh status added


Bulk actions
-edit catkeys and barcodes - OK
-refresh MODS from catkey - OK
-edit catkeys and barcodes from CSV - OK
-register druids - OK

ndushay (approved):

ndushay:
approved - merge at will (take out the HOLD in the title, please) and ensure this won't be on in prod before we intend

PR #4394, opened 2023-02-22
remove unused catkey lookup endpoint

### LINKED ISSUE #4392: Remove catkey lookup support
There is currently an endpoint (MarcXmlController.catkey) to look up catkey given a barcode. This endpoint is not currently used and so should be removed.

* Remove MarcXmlController.catkey endpoint.
* Make SymphonyReader.catkey private.

Note: In the future, it may be desirable to look up an HRID given a barcode during registration in Argo. However, this is directly support by Folio Client and does not require a DSA endpoint.

Also note: This is low priority and not blocking other work.

peetucket:

Fixes #4392 - remove catkey endpoint that is not used

Next, head to sul-dlss/dor-services-client#331 for the removal of the method from dor-serivces-client

Update tests to reflect new private method


justinlittman (changes requested):
Just remembered: please remove from OpenApi.yml.

peetucket:
> Just remembered: please remove from OpenApi.yml.

I knew I was forgetting something...

justinlittman (approved):

Issue #4388, opened 2023-02-16
Add HRIDs for selected Lane objects with catkeys


arcadiafalcone:
[lane_hrids.csv](https://github.com/sul-dlss/dor-services-app/files/10771847/lane_hrids.csv)

These are the Lane objects now in Symphony with a Symphony assigned catkey that will need to have the Folio instance HRID based on the Voyager ID for the Lane record, prefixed with "L".

These SDR objects all have a single catkey that is currently for the Lane record as represented in Symphony, but in production, these records will have a new instance HRID in Folio based on the Voyager ID (not this catkey). 

The action for this ticket is to, for each object

1.  open a version
2. change the cocina for these objects - in identification, a new `catalogLink` for these objects should be created with a Folio instance HRID and the value based on the Voyager id with "L" (capital L) as a prefix. Refresh should be set to false.
3. close the version and accession

CatalogLink example
```
"identification": {
        "catalogLinks": [
              {
                    "catalog": "folio",
                    "refresh": true,
                    "catalogRecordId": "L235413"
               }
        ],
        "sourceId": "lane:H154H_F61"
    }
```

Background:

Once upon a time, Voyager records were imported into Symphony and assigned catkeys in Symphony, with the Voyager identifier being preserved in the marc bib as the 001 field with a value starting "LL" and there is a 596 field with LANE-MED.

When we switch to Folio, the Lane Voyager data will be loaded into Folio separately from the Symphony data, and the Folio instance ids for this load into Folio will insert an "L" prefix in front of the native Voyager identifier as the new Folio instance HRID for the existing Voyager record (e.g. L34532542).

ndushay:
So is the action for this ticket to take the Lane objects with druids, and ensure that we assign the Folio instance HRID (corresponding to the ckey) to the cocina?

How is this different from assigning Folio instance HRIDs to other SDR objects with ckeys?

Are there Lane objects with ckeys we should NOT assign Folio instance HRIDs to?   We will likely need a way to know how to skip those (and not skip these)

- lines 40-48 don't have druids, so we ain't gonna have cocina for 'em.

justinlittman:
Remediation complete.

PR #4400, opened 2023-02-24
Adds migrator for adding lane HRID.

justinlittman:
refs #4388

See ticket for context.

Unit, QA

### DIFF
--- a/app/services/migrators/add_lane_hrid.rb
+++ b/app/services/migrators/add_lane_hrid.rb
@@ -1,0 +1,95 @@
+# frozen_string_literal: true
+
+module Migrators
+  # Adds Lane HRIDs.
+  # See https://github.com/sul-dlss/dor-services-app/issues/4388 for context.
+  class AddLaneHrid < Base
+    def self.druids
+      HRID_MAP.keys
+    end
+
+    def migrate?
+      HRID_MAP.key?(obj.external_identifier) && !has_hrid?
+    end
+
+    def migrate
+      catalog_links << { 'catalogRecordId' => hrid, 'catalog' => 'folio', 'refresh' => true }
+    end
+
+    def version?
+      true
+    end
+
+    def version_description
+      'Add Lane HRID'
+    end
+
+    private
+
+    def hrid
+      @hrid ||= HRID_MAP.fetch(obj.external_identifier)
+    end
+
+    def has_hrid?
+      catalog_links.any? { |link| link['catalogRecordId'] == hrid }
+    end
+
+    def catalog_links
+      @catalog_links ||= obj.identification['catalogLinks'] ||= []
+    end
+
+    HRID_MAP = {
+      'druid:xx702qn0170' => 'L79041',
+      'druid:wc042cz6208' => 'L114671',
+      'druid:gt417zw6352' => 'L35131',
+      'druid:rr941xv3570' => 'L35162',
+      'druid:hs702zb1215' => 'L35167',
+      'druid:rt260gd2393' => 'L118281',
+      'druid:by152np7439' => 'L235413',
+      'druid:bm895yk7224' => 'L28378',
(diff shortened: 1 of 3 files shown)


app/services/migrators/add_lane_hrid.rb
```diff
@@ -0,0 +1,95 @@
...
+    def hrid
+      @hrid ||= HRID_MAP.fetch(obj.external_identifier)
+    end
+
+    def has_hrid?
+      catalog_links.any? { |link| link['catalogRecordId'] == hrid }
+    end
+
+    def catalog_links
+      @catalog_links ||= obj.identification['catalogLinks'] ||= []
```
mjgiarlo, line 38: Could/should the second `||=` be `||` instead, or is the desired effect to mutate the in-memory `obj` to contain an empty array of catalog links?

app/services/migrators/add_lane_hrid.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    HRID_MAP = {
```
mjgiarlo, line 41: If this will change, we could consider stuffing these values in config vs. hard-coding, but it's no big deal.

app/services/migrators/add_lane_hrid.rb
```diff
@@ -0,0 +1,95 @@
...
+      'druid:qw952fj1420' => 'L28378',
+      'druid:rc082fp0541' => 'L28378',
+      'druid:sk549jr0359' => 'L28378',
+      'druid:ts582jh7602' => 'L28378',
+      'druid:vj723gg3770' => 'L28378',
+      'druid:vm714zs7289' => 'L28378',
+      'druid:vv024th7904' => 'L28378',
+      'druid:wf990sm5026' => 'L28378',
+      'druid:wh261jh3899' => 'L28378',
+      'druid:wk032ss9426' => 'L28378',
```
mjgiarlo, line 76: All these druids map to the same Lane ID? Huh.

bin/migrate-cocina
```diff
@@ -78,11 +78,11 @@ def perform_migrate(migrator_class:, obj:, mode:)
   return [obj.id, obj.external_identifier, 'SKIPPED'] unless migrator.migrate?
 
   migrator.migrate
-  updated_cocina_object = obj.to_cocina # This validates the cocina object
+  updated_cocina_object = obj.to_cocina_with_metadata # This validates the cocina object
   raise 'Cannot version' if migrator.version? && !(VersionService.can_open?(updated_cocina_object) || VersionService.open?(updated_cocina_object))
 
   if mode == :migrate
-    UpdateObjectService.update(updated_cocina_object, skip_lock: true)
+    UpdateObjectService.update(updated_cocina_object)
```
mjgiarlo, line 85: 🙌🏻

mjgiarlo, line 85: 🙌🏻

app/services/migrators/add_lane_hrid.rb
justinlittman: Mutate the in-memory `obj`.

app/services/migrators/add_lane_hrid.rb
justinlittman: These aren't expected to change, but can revisit if this assumption is wrong.

app/services/migrators/add_lane_hrid.rb
justinlittman: This is what was in the spreadsheet. Can you verify @arcadiafalcone ?

app/services/migrators/add_lane_hrid.rb
mjgiarlo: 👍🏻

arcadiafalcone:
That is correct - it's issues of a journal.

mjgiarlo (approved):

Issue #4398, opened 2023-02-23
Write Folio equivalent to SymphonyReader


justinlittman:
FolioReader should have the same public interface as SymphonyReader, except it should use "folio_instance_hrid" instead of "catkey". It should use the FolioClient gem for reading from Folio.

It should perform the following transformations on the retrieved MARC record:
* Delete the existing 001 and 003 (each tag should appear in the record only once).
* Add a 001 tag with the HRID at the top of the record.
* Add a 003 tag with the value FOLIO under the 001.
See https://github.com/sul-dlss/dor-services-app/issues/4395. Also, see the existing code in SymphonyReader for mutating 001 and 003. And check with @arcadiafalcone to verify that implementation is correct.

peetucket:
Are there some example equivalent records that can be pulled from both symphony and folio to see if the marc format lines up?  Looks like the JSON formatting is different in the couple examples I have looked at so far, but this could be due to the random record I found in okapi-test

PR #4401, opened 2023-02-24
service class for reading MARC data from Folio

jmartin-sul:

connects with #4398

per that ticket, adds a `FolioReader` class with "the same public interface as SymphonyReader, except it should use `folio_instance_hrid` instead of catkey`. It should use the `FolioClient` gem for reading from Folio".

the ticket's a bit ambiguous about this, but the `FolioReader` class in this PR also lacks the option for providing a barcode (instead of hrid/catkey) for lookup.  but we might want to add that, esp since the folio_client has a method for getting the hrid given a barcode?

this PR also doesn't implement any of the transformations that the ticket suggests carrying over from `SymphonyReader`.

cc @peetucket and @justinlittman 

unit tests


justinlittman (approved):

jmartin-sul:
merging with @justinlittman's approval (was looking for short-circuit `||` approval, not `&&`, but no way to indicate that in the GH UI 😆)

PR #4404, opened 2023-02-24
Migration code tweaks

ndushay:

Belated tweaks to migrators as I wasn't able to review PR #4400 in time.

unit tests.


bin/clean-druid-list
```diff
@@ -1,34 +0,0 @@
-#!/usr/bin/env ruby
```
justinlittman, line 1: I wouldn't get rid of this. Still a useful utility.

bin/generate-druid-list
```diff
@@ -1,44 +0,0 @@
-#!/usr/bin/env ruby
```
justinlittman, line 1: I wouldn't get rid of this. Still a useful utility.

app/services/migrators/invalid_lane_catkey.rb
```diff
@@ -22,21 +22,21 @@ def version?
     end
 
     def version_description
-      'Invalidate Lane catkeys'
+      'Change Invalid Lane catkey to "previous symphony" identifier'
```
justinlittman, line 25: Did you check with Arcadia on this change?

justinlittman (changes requested):
There is also a tweak needed in README.

bin/clean-druid-list
ndushay: whoops - mistake.

bin/generate-druid-list
ndushay: whoops mistake

app/services/migrators/invalid_lane_catkey.rb
ndushay: I have now:

image

ndushay:
@justinlittman thanks for feedback.  I think I've addressed it all.

PR #4409, opened 2023-02-27
update comments to reflect actual errors raised

### LINKED ISSUE #4399: Cleanup SymphonyReader exceptions
SymphonyReader defines some exceptions. References to those exceptions are in various parts of the DSA codebase. At this point, all handling of these exceptions should be performed by MarcService.

For this ticket, remove all SymphonyReader exceptions error handling that is sprinkled erroneously through DSA (except in MarcService).

peetucket:

Fixes #4399 

My reading of the code is that we are not actually raising `SymphonyReader` exceptions anywhere other than the `MarcService` where they are re-raised as `MarcService` exception.  There were a number of code comments where this was incorrect, so I changed those, and fixed a mocked test to reflect reality.  Not sure there are actual code changes needed.

CI

### DIFF
--- a/app/services/create_object_service.rb
+++ b/app/services/create_object_service.rb
@@ -14,7 +14,8 @@ class CreateObjectService
   # @param [#create] event_factory creates events
   # @param [#call] id_minter assigns identifiers. You can provide your own minter if you want to use a specific druid for an item.
   # @return [Cocina::Models::DROWithMetadata,Cocina::Models::CollectionWithMetadata,Cocina::Models::AdminPolicyWithMetadata]
-  # @raises [Catalog::MarcService::MarcServiceError] if cannot refresh descMetadata from source
+  # @raise [Catalog::MarcService::MarcServiceError::CatalogRecordNotFoundError] if catalog identifer not found when refreshing descMetadata
+  # @raise [Catalog::MarcService::MarcServiceError::CatalogResponseError] if other error occurred refreshing descMetadata from catalog source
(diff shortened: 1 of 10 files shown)


app/services/create_object_service.rb
```diff
@@ -14,7 +14,7 @@ class CreateObjectService
   # @param [#create] event_factory creates events
   # @param [#call] id_minter assigns identifiers. You can provide your own minter if you want to use a specific druid for an item.
   # @return [Cocina::Models::DROWithMetadata,Cocina::Models::CollectionWithMetadata,Cocina::Models::AdminPolicyWithMetadata]
-  # @raises [Catalog::SymphonyReader::ResponseError] if symphony connection failed
+  # @raises [Catalog::MarcService::MarcServiceError] if cannot refresh descMetadata from source
```
ndushay, line 17: thanks for the better clarification.    Note that @raise is the correct Yard tag:  https://rubydoc.info/gems/yard/file/docs/Tags.md 

would it be more correct to do anything nuanced between CatalogResponseError and CatalogRecordNotFoundError ?   (ok to punt on this.

ndushay (approved):
approving but would like yard tag fixed before merge.

app/services/create_object_service.rb
peetucket: 1. swapped `@raises` to `@raise` everywhere
2. added another comment to clarify the two different exceptions that can be raised and when

justinlittman (approved):

PR #4413, opened 2023-02-27
[HOLD] add HB notification for updating marc as part of virtual object creation

ndushay:
HOLD - as it may become clear in testing on stage/qa that we don't need this HB notification in production.

Andrew and I are spelunking on whether or not virtual object manipulation
- actually calls out to update marc records with 856s now, and under which circumstances
- whether it needs to, or if this could be covered by the publish step in the releaseWF.

tagging @mjgiarlo as he may remember more information from 2019 (!) to help us with this.

### DIFF
--- a/app/services/constituent_service.rb
+++ b/app/services/constituent_service.rb
@@ -75,6 +75,9 @@ class ConstituentService
 
       next unless cocina_item.identification&.catalogLinks&.any? { |link| link.catalog == 'symphony' }
 
+      msg = "ConstituentService is about to call Catalog::UpdateMarc856RecordService for #{constituent_druid} in #{virtual_object_druid}"
+      Honeybadger.notify("[DEBUG NOTIFICATION] #{msg}")
+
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@ndushay 💬 
> Andrew and I are spelunking on whether or not virtual object manipulation
> 
> * actually calls out to update marc records with 856s now, and under which circumstances

naïvely 👀 the code, it seems that every constituent object with a Symphony catalog link is updated whenever it's added to a virtual object.

> * whether it needs to, or if this could be covered by the publish step in the releaseWF.

That's a better question and I'm honestly not sure.

andrewjbtw:
> publish step in the releaseWF

And/or the `update-marc` step.

lwrubel:
Confirmed by @andrewjbtw we need this.

andrewjbtw:
Elaborating: I'm not sure how to hit this path with the Argo UI, so the notice will confirm I've actually figured that out.

PR #4414, opened 2023-02-27
Validate Folio IDs in API more strictly

mjgiarlo:

Lean on the API spec, Committee, and existing schemas to get automatic validation of Folio IDs at the API layer.

CI


ndushay (approved):

justinlittman:
Thanks @mjgiarlo !

PR #4415, opened 2023-02-28
Bump to cocina-models 0.89.0 release and API spec

mjgiarlo:

Keep pace with cocina-models releases

Will be tested as part of cocina-models rollout


justinlittman (approved):

justinlittman:
Both approved but not merged.

PR #4423, opened 2023-03-03
Refactor 856 marc serialization into hash

### LINKED ISSUE #4419: Refactor Marc856Generator
In the earlier refactoring work, I didn't quite get the requirements for responsibilities right. The Marc856Generator should return an abstract representation of the 856s (this can be anything -- hash/arrays, MARC:Fields, struct), rather than a Symphony specific serialization.

The Symphony specific serialization should be moved into the SymphonyWriter.

The motivation is to make it easier to introduce a Folio writer, which will merge the 856s into `MARC:Record`s.

aaron-collier:

Closes #4419 

### DIFF
--- a/app/services/catalog/marc_856_generator.rb
+++ b/app/services/catalog/marc_856_generator.rb
@@ -19,6 +19,22 @@ module Catalog
     end
 
     # @return [Hash] all data required to create stub 856 records for this object
+    # catkey: the catalog key that associates a DOR object with a specific Symphony record.
+    # druid: the druid
+    # .856. 41 or 40 (depending on APO)
+    # Subfield u (required): the full Purl URL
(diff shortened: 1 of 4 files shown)


justinlittman (changes requested):
This doesn't quite get right the refactoring I was intending. I was hoping for a representation of the 856, not the data used to generate an 856. For example:
{
  indicator1: '4',
  indicator2: '1',
  subfieldu: ['https://purl.stanford.edu/...'],
  subfieldx: ['file:bc123dg9393%2Fwt183gy6220_00_0001.jp2', 'rights:group=stanford']
}

Sorry I wasn't clearer on that.

aaron-collier:
@justinlittman no, it was clear, it felt a little strange to me to include those values and took the refactor a bit further. Happy to back it up a bit though :-)

justinlittman (approved):

PR #4429, opened 2023-03-10
Revert "Refactor 856 marc serialization into hash"

ndushay:
Reverts sul-dlss/dor-services-app#4423

Turning the Marc 856 into a hash created problems -- 
- the output contained "\" where previously there were none
- it's unclear if this change will have deleterious effects on records with previous symphony ckeys  (this is presumably to zero out 856s from incorrect ckeys)
- it may be that the Symphony code that reads the 856s has nuances beyond just adding 856s to a record.  Sometimes they are removed.  Sometimes they are updated.

We are backing this out for now;  we need to understand the goals better before we do this

### DIFF
--- a/app/services/catalog/marc_856_generator.rb
+++ b/app/services/catalog/marc_856_generator.rb
@@ -18,7 +18,7 @@ module Catalog
       @thumbnail_service = thumbnail_service
     end
 
-    # @return [Hash] all data required to create stub 856 records for this object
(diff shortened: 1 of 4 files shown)


ndushay:
@justinlittman - this is reverting #4419;   Andrew found some problems while testing something else.

aaron-collier (approved):

PR #4432, opened 2023-03-10
Refactor of marc856generator to serialize 856 string

aaron-collier:

This refactors `marc856generator` to return an array of hashed data used to create stub 856 records and moves all symphony specific serialization into `SymphonyWriter` 

Example output with 2 previous catkey values:
```
        [
          {
            catalog_record_id: '123',
            druid: bare_druid
          },
          {
            catalog_record_id: '456',
            druid: bare_druid
          },
          {
            catalog_record_id: '8832162',
            druid: bare_druid,
            indicators: '41',
            subfields: [
              { code: 'u', value: "https://purl.stanford.edu/#{bare_druid}" },
              { code: 'x', value: 'SDR-PURL' },
              { code: 'x', value: 'item' },
              { code: 'x', value: "file:#{bare_druid}%2Fwt183gy6220_00_0001.jp2" },
              { code: 'x', value: "collection:#{collection_bare_druid}:8832162:Collection label & A Special character" },
              { code: 'x', value: 'rights:world' }
            ]
          }
        ]
```

### DIFF
--- a/app/services/catalog/marc_856_generator.rb
+++ b/app/services/catalog/marc_856_generator.rb
@@ -2,7 +2,7 @@
 
 # rubocop:disable Metrics/ClassLength
 module Catalog
-  # Creates a stub MARC 856 field (currently for transferring to symphony only) given a cocina object.
+  # Creates a hash of identifiers and MARC 856 field data given a cocina object.
   class Marc856Generator
     # objects goverened by these APOs (ETD and EEMs) will get indicator 2 = 0, else 1
     BORN_DIGITAL_APOS = %w[druid:bx911tp9024 druid:jj305hm5259].freeze
@@ -18,33 +18,33 @@ module Catalog
       @thumbnail_service = thumbnail_service
     end
 
-    # @return [Array] all stub 856 records for this object (this is not a fully valid 856, but specific for transfer to symphony)
-    # catkey: the catalog key that associates a DOR object with a specific Symphony record.
+    # @return [Array<Hash>] of data required to update 856 fields for purls pertaining to this object
+    # catalog_record_id: the catalog key that associates a DOR object with a specific Symphony record.
     # druid: the druid
-    # .856. 41 or 40 (depending on APO)
-    # Subfield u (required): the full Purl URL
-    # Subfield x #1 (required): The string SDR-PURL as a marker to identify 856 entries managed through DOR
-    # Subfield x #2 (required): Object type (<identityMetadata><objectType>) - item, collection,
-    #     (future types of sets to describe other aggregations like albums, atlases, etc)
-    # Subfield x #3 (required): The display type of the object.
-    #     use an explicit display type from the object if present (<identityMetadata><displayType>)
-    #     else use the value of the <contentMetadata> "type" attribute if present, e.g., image, book, file
-    #     else use the value "citation"
-    # Subfield x #4 (optional): the barcode if known (<identityMetadata><otherId name="barcode">, recorded as barcode:barcode-value
-    # Subfield x #5 (optional): the file-id to be used as thumb if available, recorded as file:file-id-value
-    # Subfield x #6..n (optional): Collection(s) this object is a member of, recorded as collection:druid-value:ckey-value:title
-    # Subfield x #7..n (optional): label and part sort keys for the member
-    # Subfield x #8..n (optional): High-level rights summary
+    # indicators: 40 or 41 (depending on APO)
+    # subfields: (ordered)
+    # u - access:
+    # z - purl: the full Purl URL
+    # x - sdr_purl_marker (required): The string SDR-PURL as a marker to identify 856 entries managed through DOR
+    # x - object_type (required): Object type (<identityMetadata><objectType>) - item, collection,
+    #                         (future types of sets to describe other aggregations like albums, atlases, etc)
+    # x - barcode (optional): the barcode if known (<identityMetadata><otherId name="barcode">, recorded as barcode:barcode-value
+    # x - thumbnail (optional): the file-id to be used as thumb if available, recorded as file:file-id-value
+    # x - collections (optional): Collection(s) this object is a member of, formatted as: druid:catkey:label
+    # x - parts-label (optional):
+    # x - parts-sort (optional):
+    # x- rights (optional): High-level rights summary
     def create
-      return [] if ckeys.empty? && previous_ckeys.empty?
+      return [] if catalog_record_ids.empty? && previous_catalog_record_ids.empty?
 
       # first create "blank" records for any previous catkeys
-      records = previous_ckeys.map { |previous_catkey| new_identifier_record(previous_catkey) }
+      records = previous_catalog_record_ids.map { |previous_catkey| new_identifier_record(previous_catkey) }
 
-      # now add the current ckey
-      unless ckeys.empty?
-        catalog_record_id = ckeys.first
-        records << (released_to_searchworks?(@cocina_object) ? new_856_record(catalog_record_id) : new_identifier_record(catalog_record_id))
+      unless catalog_record_ids.empty?
+        catalog_record_id = catalog_record_ids.first
+        record = new_identifier_record(catalog_record_id)
+        record.merge!(new_856_data) if released_to_searchworks?(@cocina_object)
+        records << record
       end
 
       records
@@ -52,90 +52,112 @@ module Catalog
 
     private
 
-    # NOTE: this is a stub 856 record which is used to communicate with symphony (not a properly formatted 856 field nor a marc record)
-    def new_856_record(ckey)
-      new856 = "#{new_identifier_record(ckey)}#{get_856_cons} #{get_1st_indicator}#{get_2nd_indicator}#{get_z_field}#{get_u_field}#{get_x1_sdrpurl_marker}|x#{get_object_type_from_uri}"
-      new856 += "|xbarcode:#{@cocina_object.identification.barcode}" if @cocina_object.identification.respond_to?(:barcode) && @cocina_object.identification.barcode
-      new856 += "|xfile:#{thumb}" unless thumb.nil?
-      new856 += get_x2_collection_info unless get_x2_collection_info.nil?
-      new856 += get_x2_part_info unless get_x2_part_info.nil?
-      new856 += get_x2_rights_info unless get_x2_rights_info.nil?
-      new856
+    # NOTE: this is the data to go in an 856 field (not a properly formatted 856 field nor a marc record)
+    def new_856_data
+      {
+        indicators:,
+        subfields: [
+          subfield_z_access,
+          subfield_u_purl,
+          subfield_x_sdr_purl_marker,
+          subfield_x_object_type,
+          subfield_x_barcode,
+          subfield_x_thumbnail,
+          subfield_x_collections,
+          subfield_x_parts,
+          subfield_x_rights
+        ].compact.flatten
+      }
     end
 
-    def new_identifier_record(ckey)
-      "#{ckey}\t#{@druid_id}\t"
+    def new_identifier_record(catalog_record_id)
+      {
+        catalog_record_id:,
+        druid: @druid_id
+      }
     end
 
-    # This should only be reached for dro and collection objects
-    def get_object_type_from_uri
-      return 'item' if @cocina_object.dro?
-      return 'collection' if @cocina_object.collection?
-
-      nil
+    def indicators
+      "#{first_indicator}#{second_indicator}"
     end
 
-    # returns 856 constants
-    def get_856_cons
-      '.856.'
-    end
-
-    # returns First Indicator for HTTP (4)
-    def get_1st_indicator
+    def first_indicator
       '4'
     end
 
-    # returns Second Indicator for Version of resource
-    def get_2nd_indicator
+    def second_indicator
       born_digital? ? '0' : '1'
     end
 
-    # returns text in the z field based on permissions
-    def get_z_field
-      return '' unless @access.view == 'stanford' || (@access.respond_to?(:location) && @access.location)
-
-      '|zAvailable to Stanford-affiliated users.'
+    def born_digital?
+      BORN_DIGITAL_APOS.include? @cocina_object.administrative.hasAdminPolicy
     end
 
-    # builds the PURL uri based on the druid id
-    def get_u_field
-      "|u#{Settings.release.purl_base_url}/#{@druid_id}"
+    def subfield_z_access
+      return unless @access.view == 'stanford' || (@access.respond_to?(:location) && @access.location)
+
+      { code: 'z', value: 'Available to Stanford-affiliated users.' }
+    end
+
+    def subfield_u_purl
+      { code: 'u', value: "#{Settings.release.purl_base_url}/#{@druid_id}" }
     end
 
     # returns the SDR-PURL subfield
-    def get_x1_sdrpurl_marker
-      '|xSDR-PURL'
+    def subfield_x_sdr_purl_marker
+      { code: 'x', value: 'SDR-PURL' }
+    end
+
+    # This should only be reached for dro and collection objects
+    def subfield_x_object_type
+      return { code: 'x', value: 'item' } if @cocina_object.dro?
+      return { code: 'x', value: 'collection' } if @cocina_object.collection?
+    end
+
+    def subfield_x_barcode
+      return unless @cocina_object.identification.respond_to?(:barcode) && @cocina_object.identification.barcode
+
+      { code: 'x', value: "barcode:#{@cocina_object.identification.barcode}" }
+    end
+
+    # the @id attribute of resource/file elements including extension
+    # @return [String] thumbnail filename (nil if none found)
+    def subfield_x_thumbnail
+      return unless @thumbnail_service.thumb
+
+      { code: 'x', value: "file:#{ERB::Util.url_encode(@thumbnail_service.thumb)}" }
     end
 
     # returns the collection information subfields if exists
     # @return [String] the collection information druid-value:catkey-value:title format
-    def get_x2_collection_info
+    def subfield_x_collections
       return unless @cocina_object.respond_to?(:structural) && @cocina_object.structural
 
       collections = @cocina_object.structural.isMemberOf
-      collection_info = ''
+      collection_info = []
 
       collections.each do |collection_druid|
         collection = CocinaObjectStore.find(collection_druid)
         next unless released_to_searchworks?(collection)
 
         catkey = collection.identification&.catalogLinks&.find { |link| link.catalog == 'symphony' }
-        collection_info += "|xcollection:#{collection.externalIdentifier.sub('druid:',
-                                                                             '')}:#{catkey&.catalogRecordId}:#{Cocina::Models::Builders::TitleBuilder.build(collection.description.title)}"
+        collection_info << { code: 'x', value: "collection:#{collection.externalIdentifier.sub('druid:', '')}:#{catkey&.catalogRecordId}:#{Cocina::Models::Builders::TitleBuilder.build(collection.description.title)}" }
       end
 
       collection_info
     end
 
-    def get_x2_part_info
-      str = ''
-      str += "|xlabel:#{part_label}" if part_label.present?
-      str += "|xsort:#{part_sort}" if part_sort.present?
-      str
+    def subfield_x_parts
+      return unless part_label.present? || part_sort.present?
+
+      [].tap do |part|
+        part << { code: 'x', value: "label:#{part_label}" } if part_label.present?
+        part << { code: 'x', value: "sort:#{part_sort}" } if part_sort.present?
+      end
     end
 
-    def get_x2_rights_info
-      return get_x2_collection_rights_info unless @access.respond_to?(:download)
+    def subfield_x_rights
+      return [{ code: 'x', value: "rights:#{@access.view}" }] unless @access.respond_to?(:download)
 
       values = []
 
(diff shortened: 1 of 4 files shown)


app/services/catalog/marc_856_generator.rb
```diff
@@ -36,80 +36,77 @@ def initialize(cocina_object, thumbnail_service:)
     # Subfield x #7..n (optional): label and part sort keys for the member
     # Subfield x #8..n (optional): High-level rights summary
     def create
-      return [] if ckeys.empty? && previous_ckeys.empty?
+      return {} if ckeys.empty? && previous_ckeys.empty?
 
       # first create "blank" records for any previous catkeys
-      records = previous_ckeys.map { |previous_catkey| new_identifier_record(previous_catkey) }
+      record = {
```
ndushay, line 42: This code still mistakenly believes all the information goes in ONE record.  It does not.  Each previous_ckey currently creates a record in the array.  Then the current ckey creates a more complicated record in the array.

app/services/catalog/marc_856_generator.rb
```diff
@@ -36,80 +36,77 @@ def initialize(cocina_object, thumbnail_service:)
...
     # This should only be reached for dro and collection objects
-    def get_object_type_from_uri
-      return 'item' if @cocina_object.dro?
-      return 'collection' if @cocina_object.collection?
+    def subfield_x2
+      return '|xitem' if @cocina_object.dro?
+      return '|xcollection' if @cocina_object.collection?
 
       nil
     end
```
ndushay, line 83: this |x notation is unlikely to be understood by Folio's QuickMarc (or any other API).

app/services/catalog/marc_856_generator.rb
```diff
@@ -36,80 +36,77 @@ def initialize(cocina_object, thumbnail_service:)
...
+        subfield_x5:,
+        subfield_x6:,
+        subfield_x7:,
+        subfield_x8:
+      }
     end
 
     def new_identifier_record(ckey)
       "#{ckey}\t#{@druid_id}\t"
     end
```
ndushay, line 75: this is a Stanford hoked-up notation between the SDR and the script that loads the data on the SYmphony side.  That script is going away in favor of the SDR making direct API calls to Folio.  We hope.

app/services/catalog/symphony_writer.rb
```diff
@@ -6,18 +6,20 @@
 module Catalog
   # Writes the stub 856 record for sending to symphony
   class SymphonyWriter
```
ndushay, line 8: This class is where we would take the array of records, described in some hash format, and turn that hash into the String format Symphony is expecting in a file that is output from the SDR

ndushay (commented):
a few small notes on ...stuff we didn't fully understand before today.   Tagging @justinlittman as well.

ndushay (changes requested):
We're going to talk about this;  just making extra sure we don't merge by accient.

ndushay:
Possibly useful for later:   `/sdrpurl_transfer` is where to find the output file on a dsa box.

app/services/catalog/marc_856_generator.rb
aaron-collier: My approach here is to to include the previous catkey lines in the single hash and then in `SymphonyWriter` treat them as individual records for output. That avoids the need to do any inspection, etc. when we get to Folio and they can simply be ignored if necessary.

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,151 +18,175 @@ def initialize(cocina_object, thumbnail_service:)
...
+        indicators:,
+        access:,
+        purl:,
+        sdr_purl_marker:,
+        object_type:,
+        barcode:,
+        thumbnail:,
+        collections:,
+        parts:,
+        rights:
```
ndushay, line 71: while I love the meaningful keys here, I was thinking more like this:

```
...
indicators:
subfields [
   { x: value },
   { u: value },
   { x: nother value }
   { x: nother nother value }
]
```

The subfield char ("code") is generic marc;   the `|` is the part that was unique to the symphony script.

This also is presenting the info in an Array, which guarantees order, while a ruby Hash does not.  Order matters in Marc.

The values above could be the meaningful names you have chosen for the methods or attr with the germane data.

app/services/catalog/marc_856_generator.rb
```diff
@@ -2,7 +2,7 @@
 
 # rubocop:disable Metrics/ClassLength
 module Catalog
-  # Creates a stub MARC 856 field (currently for transferring to symphony only) given a cocina object.
+  # Creates a hash that is used to stub the MARC 856 field (currently for transferring to symphony only) given a cocina object.
```
ndushay, line 5: ditch "currenty for symphony only" clause

And now were not talking about a stub any longer -- this class is returning identifiers and then a hash of data to build a plain old marc 856 field, or no such hash to imply 856 matching druid should be removed.

@justinlittman did you want to keep the identifier information and whether or not it has an 856 associated in THIS class?  (makes sense to ... just confirming)

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,151 +18,175 @@ def initialize(cocina_object, thumbnail_service:)
...
-    # Subfield x #3 (required): The display type of the object.
-    #     use an explicit display type from the object if present (<identityMetadata><displayType>)
-    #     else use the value of the <contentMetadata> "type" attribute if present, e.g., image, book, file
-    #     else use the value "citation"
-    # Subfield x #4 (optional): the barcode if known (<identityMetadata><otherId name="barcode">, recorded as barcode:barcode-value
-    # Subfield x #5 (optional): the file-id to be used as thumb if available, recorded as file:file-id-value
-    # Subfield x #6..n (optional): Collection(s) this object is a member of, recorded as collection:druid-value:ckey-value:title
-    # Subfield x #7..n (optional): label and part sort keys for the member
-    # Subfield x #8..n (optional): High-level rights summary
+    # @return [Array] hash of all data required to create stub 856 records for this object
```
ndushay, line 21: "... required to update marc bib 856 fields to purls" ???


"@return [Array] hash" -- shouldn't this be [Array<Hash>] and then a description of the hash structure?

mjgiarlo, line 21: Clarifying this is a hash in the text is a good tweak. And I think we can also tighten the return type specification in YARD by considering `@return [Array<Hash>] ...`

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,151 +18,175 @@ def initialize(cocina_object, thumbnail_service:)
...
+    # collections [Aarray] (optional): Collection(s) this object is a member of
+    #   druid:
+    #   catkey:
+    #   label:
+    # parts (optional):
+    #   label:
+    #   sort:
+    # rights [Array] (optional): High-level rights summary
     def create
       return [] if ckeys.empty? && previous_ckeys.empty?
```
ndushay, line 42: I've forgotten where we are with this -- should this still be "ckeys" and previous_ckeys"?   And it will accommodate folio hrids in a separate ticket?  Or is this "catalog_record_ids"?   Or ?

app/services/catalog/symphony_writer.rb
```diff
@@ -27,5 +27,56 @@ def run_write_script(command)
...
+    def marc_856_subfield_x(record)
+      subfield_x = "|x#{record[:sdr_purl_marker]}"
+      subfield_x += "|x#{record[:object_type]}"
+      subfield_x += "|xbarcode:#{record[:barcode]}" if record[:barcode].present?
+      subfield_x += "|xfile:#{record[:thumbnail]}" if record[:thumbnail].present?
+      subfield_x += collections_for(record[:collections]).join
+      subfield_x += label_for(record[:parts]).to_s
+      subfield_x += sort_for(record[:parts]).to_s
+      subfield_x + "|xrights:#{record[:rights]}"
+    end
```
ndushay, line 62: so switching to the Array for subfields changes this stuff to something like:

```
marc_856 = ".856.#{record[:indicators]}"
record[:subfields].each { |subfield| marc_856 << "|#{subfield.first.key}#{subfield.first.value}" }

^^ thus, all the marc 856 logic is in the data structure, and the writer is a very stupid class that just takes the strings and turns them into the desired output.   Happy to talk through this if I'm not making sense.
```

ndushay (changes requested):
Im happy to pair or talk through my thinking on the 856 generator vs. the symphony writer -- what logic lives where and what the data structure needs to be.

app/services/catalog/marc_856_generator.rb
aaron-collier: I have the hash layed out below this comment, maybe a better format?

app/services/catalog/symphony_writer.rb
aaron-collier: As discussed, I did not like the lack of explictness of the data when following that pattern (which is where this PR started) but can adopt it if that's what we prefer.

app/services/catalog/marc_856_generator.rb
aaron-collier: The problem I ran into with this proposed format is for fields that can have multiples, like collections and rights, when do the `|xcollection:` and `|xrights:` get serialized? If we follow this pattern it would need to be something like:

```
subfields: [
...
  {  x: `|xcollection:ckey_1:druid_1:title_1|xcollection:ckey_2:druid_2:title_2` }
]
```
so now we have the symphony specific pipe embedded in the data. With the approach I took, that is handled in the writer.

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
       @thumbnail_service = thumbnail_service
     end
 
-    # @return [Array] all stub 856 records for this object (this is not a fully valid 856, but specific for transfer to symphony)
-    # catkey: the catalog key that associates a DOR object with a specific Symphony record.
+    # @return [Array<Hash>] of data required to create stub 856 records for this object
+    # catalog_reocrd_id: the catalog key that associates a DOR object with a specific Symphony record.
```
ndushay, line 22: typo

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
       @thumbnail_service = thumbnail_service
     end
 
-    # @return [Array] all stub 856 records for this object (this is not a fully valid 856, but specific for transfer to symphony)
-    # catkey: the catalog key that associates a DOR object with a specific Symphony record.
+    # @return [Array<Hash>] of data required to create stub 856 records for this object
```
ndushay, line 21: "... of data required to update 856 fields for purls pertaining to this object" ???

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
+    # indicators: 40 or 41 (depending on APO)
+    # subfields: (ordered)
+    # u - access:
+    # z - purl: the full Purl URL
+    # x - sdr_purl_marker (required): The string SDR-PURL as a marker to identify 856 entries managed through DOR
+    # x - object_type (required): Object type (<identityMetadata><objectType>) - item, collection,
+    #                         (future types of sets to describe other aggregations like albums, atlases, etc)
+    # x - barcode (optional): the barcode if known (<identityMetadata><otherId name="barcode">, recorded as barcode:barcode-value
+    # x - thumbnail (optional): the file-id to be used as thumb if available, recorded as file:file-id-value
+    # x - collections (optional): Collection(s) this object is a member of formatted as: druid:catkey:label
```
ndushay, line 33: "... is a member of, formatted ..."

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
-    # NOTE: this is a stub 856 record which is used to communicate with symphony (not a properly formatted 856 field nor a marc record)
-    def new_856_record(ckey)
-      new856 = "#{new_identifier_record(ckey)}#{get_856_cons} #{get_1st_indicator}#{get_2nd_indicator}#{get_z_field}#{get_u_field}#{get_x1_sdrpurl_marker}|x#{get_object_type_from_uri}"
-      new856 += "|xbarcode:#{@cocina_object.identification.barcode}" if @cocina_object.identification.respond_to?(:barcode) && @cocina_object.identification.barcode
-      new856 += "|xfile:#{thumb}" unless thumb.nil?
-      new856 += get_x2_collection_info unless get_x2_collection_info.nil?
-      new856 += get_x2_part_info unless get_x2_part_info.nil?
-      new856 += get_x2_rights_info unless get_x2_rights_info.nil?
-      new856
+    # NOTE: this is the data to serialize a stub 856 record (not a properly formatted 856 field nor a marc record)
```
ndushay, line 55: "... the data to go in an 856 field (not a ...."  ?

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
+          access,
+          purl,
+          sdr_purl_marker,
+          object_type,
+          barcode,
+          thumbnail,
+          collections,
+          parts,
+          rights
+        ].compact.flatten
```
aaron-collier, line 69: Note that this `compact` and `flatten` removes any `nil` values returned from the list of methods, and then flattens the ones that return arrays like `collections` and `rights`

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
+      }
     end
 
     def new_identifier_record(ckey)
```
ndushay, line 73: i would rename the param, too.

ndushay (commented):
changes to 856 looking good;  a few small quibbles.   I also float the idea that you migh somehow indicate that, for example, "access" is returning the hash for subfield z.  Maybe `access_subfield_z`  or `subfield_z_access` or maybe you have something better.

spec/services/catalog/symphony_writer_spec.rb
```diff
@@ -22,28 +27,85 @@
...
+              { code: 'x', value: 'barcode:36105216275185' },
+              { code: 'x', value: "file:#{bare_druid}%2Fwt183gy6220_00_0001.jp2" },
+              { code: 'x', value: "collection:#{collection_bare_druid}:8832162:Collection label & A Special character" },
+              { code: 'x', value: nil },
+              { code: 'x', value: 'rights:world' }
+            ]
+          }
+        ]
+      end
+      let(:marc856) do
```
ndushay, line 50: maybe call this marc856file or something?

spec/services/catalog/symphony_writer_spec.rb
```diff
@@ -22,28 +27,85 @@
...
+              { code: 'x', value: "file:#{bare_druid}%2Fwt183gy6220_00_0001.jp2" },
+              { code: 'x', value: "collection:#{collection_bare_druid}:8832162:Collection label & A Special character" },
+              { code: 'x', value: "label:we call this a 'part'" },
+              { code: 'x', value: 'sort:123' },
+              { code: 'x', value: 'rights:world' }
+            ]
+          }
+        ]
+      end
+      let(:marc856) do
```
ndushay, line 92: maybe call this marc856file or something?

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
+    # x - parts-label (optional):
+    # x - parts-sort (optional):
+    # x- rights (optional): High-level rights summary
     def create
-      return [] if ckeys.empty? && previous_ckeys.empty?
+      return [] if catalog_record_ids.empty? && previous_catalog_record_ids.empty?
 
       # first create "blank" records for any previous catkeys
-      records = previous_ckeys.map { |previous_catkey| new_identifier_record(previous_catkey) }
+      records = previous_catalog_record_ids.map { |previous_catkey| new_identifier_record(previous_catkey) }
```
ndushay, line 41: ugh - missed this before - previous_catkey

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
-    # NOTE: this is a stub 856 record which is used to communicate with symphony (not a properly formatted 856 field nor a marc record)
-    def new_856_record(ckey)
-      new856 = "#{new_identifier_record(ckey)}#{get_856_cons} #{get_1st_indicator}#{get_2nd_indicator}#{get_z_field}#{get_u_field}#{get_x1_sdrpurl_marker}|x#{get_object_type_from_uri}"
-      new856 += "|xbarcode:#{@cocina_object.identification.barcode}" if @cocina_object.identification.respond_to?(:barcode) && @cocina_object.identification.barcode
-      new856 += "|xfile:#{thumb}" unless thumb.nil?
-      new856 += get_x2_collection_info unless get_x2_collection_info.nil?
-      new856 += get_x2_part_info unless get_x2_part_info.nil?
-      new856 += get_x2_rights_info unless get_x2_rights_info.nil?
-      new856
+    # NOTE: this is the data to go in an 856 field (not a (not a properly formatted 856 field nor a marc record)
```
ndushay, line 55: not a not a

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
+    # indicators: 40 or 41 (depending on APO)
+    # subfields: (ordered)
+    # u - access:
+    # z - purl: the full Purl URL
+    # x - sdr_purl_marker (required): The string SDR-PURL as a marker to identify 856 entries managed through DOR
+    # x - object_type (required): Object type (<identityMetadata><objectType>) - item, collection,
+    #                         (future types of sets to describe other aggregations like albums, atlases, etc)
+    # x - barcode (optional): the barcode if known (<identityMetadata><otherId name="barcode">, recorded as barcode:barcode-value
+    # x - thumbnail (optional): the file-id to be used as thumb if available, recorded as file:file-id-value
+    # x - collections (optional): Collection(s) this object is a member of, formatted as: druid:catkey:label
```
ndushay, line 33: catkey

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
+    # the @id attribute of resource/file elements including extension
+    # @return [String] thumbnail filename (nil if none found)
+    def subfield_x_thumbnail
+      return unless @thumbnail_service.thumb
+
+      { code: 'x', value: "file:#{ERB::Util.url_encode(@thumbnail_service.thumb)}" }
     end
 
     # returns the collection information subfields if exists
     # @return [String] the collection information druid-value:catkey-value:title format
```
ndushay, line 132: catkey

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
 
       collections = @cocina_object.structural.isMemberOf
-      collection_info = ''
+      collection_info = []
 
       collections.each do |collection_druid|
         collection = CocinaObjectStore.find(collection_druid)
         next unless released_to_searchworks?(collection)
 
         catkey = collection.identification&.catalogLinks&.find { |link| link.catalog == 'symphony' }
```
ndushay, line 143: catkey

app/services/catalog/marc_856_generator.rb
```diff
@@ -18,124 +18,146 @@ def initialize(cocina_object, thumbnail_service:)
...
-        collection_info += "|xcollection:#{collection.externalIdentifier.sub('druid:',
-                                                                             '')}:#{catkey&.catalogRecordId}:#{Cocina::Models::Builders::TitleBuilder.build(collection.description.title)}"
+        collection_info << { code: 'x', value: "collection:#{collection.externalIdentifier.sub('druid:', '')}:#{catkey&.catalogRecordId}:#{Cocina::Models::Builders::TitleBuilder.build(collection.description.title)}" }
```
ndushay, line 144: catkey

app/services/catalog/marc_856_generator.rb
```diff
@@ -177,30 +191,24 @@ def parts_delimiter(elements)
...
 
-    def previous_ckeys
-      @previous_ckeys ||= fetch_ckeys(current: false)
+    def previous_catalog_record_ids
+      @previous_catalog_record_ids ||= fetch_catalog_record_ids(current: false)
     end
 
     # List of current or previous ckeys for the cocina object (depending on parameter passed)
     # @param current [boolean] if you want the current or previous ckeys
     # @return [Array] previous or current catkeys for the object in an array, empty array if none exist
```
ndushay, line 204: catkey

spec/services/catalog/marc856_generator_spec.rb
```diff
@@ -165,7 +165,7 @@
 
     context "when the druid object doesn't have catkey or previous catkeys" do
```
ndushay, line 166: look for catkey and ckey in the spec, too.

ndushay (approved):
Approving, modulo:
- look for catkey and ckey in the 856 code and spec (can leave for symphony writer)
- typo in comment line 55

having Andrew (?) try it in test.

justinlittman (commented):
This PR gets us most of the way there. However, there is still some Symphony specific logic:
* Determining which records have 865s to be deleted and added.
* The inclusion of the catkey in subfield x collections. (Missed that before.)

That being said, I'd suggest that this is good enough (=huge improvement and step in the right direction) for this PR. I'd suggest that we make the minor tweaks recommended by Naomi, but leave all remaining catkey references as is. (Just renaming them to catalog record id doesn't solve the problem.) I'll address them in a follow-up PR that removes the last bits of Symphony-specific logic.

Also, makes sense to delay testing until the follow-up PR.

app/services/catalog/marc_856_generator.rb
justinlittman: Fixed.

spec/services/catalog/symphony_writer_spec.rb
justinlittman: Fixed.

justinlittman:
For the sake of expediency, I made the tweaks and squished the PR.

PR #4448, opened 2023-03-17
[WIP] T4426 folio writer

lwrubel:


app/services/catalog/folio_writer.rb
```diff
@@ -0,0 +1,78 @@
...
+    def previous_catalog_record_ids
+      @previous_catalog_record_ids ||= fetch_catalog_record_ids(current: false)
+    end
+
+    # List of current or previous catalog record ids for the cocina object (depending on parameter passed)
+    # @param current [boolean] if you want the current or previous catalog record ids
+    # @return [Array] previous or current catalog_record_ids for the object in an array, empty array if none exist
+    def fetch_catalog_record_ids(current:)
+      ckey_type = current ? 'folio' : 'previous folio'
+      @cocina_object.identification.catalogLinks.select { |link| link.catalog == ckey_type }.map(&:catalogRecordId)
```
mjgiarlo, line 61: I wonder if we should consider renaming this variable so it's not coupled to the previous implementation details or if there's value in keeping this name.

mjgiarlo, line 61: What would you think about using `filter_map` in lieu of `select.map` to get down to a single iteration?

```suggestion
      ckey_type = current ? 'folio' : 'previous folio'
      @cocina_object.identification.catalogLinks.filter_map { |link| link.catalogRecordId if link.catalog == ckey_type }
```

app/services/catalog/folio_writer.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def released_to_searchworks?(cocina_object)
+      released_for = ::ReleaseTags.for(cocina_object:)
+      rel = released_for.transform_keys { |key| key.to_s.upcase } # upcase all release tags to make the check case insensitive
+      rel.dig('SEARCHWORKS', 'release').presence || false
```
mjgiarlo, line 67: Could we go without the `|| false` here and use the resulting `nil` from `#presence` as the falsey return value, or do we want to avoid a `nil` in this context?

app/services/catalog/folio_writer.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module Catalog
+  # Updates the FOLIO MARC record's 856 fields
+  class FolioWriter
+    def self.save(cocina_object:, marc_856_data:)
+      new(cocina_object:, marc_856_data:).save
```
mjgiarlo, line 7: This looks uncovered but it may not be worth adding a unit test.

app/services/catalog/folio_writer.rb
lwrubel: Yes, that's more concise and makes sense. Have updated that.

app/services/catalog/folio_writer.rb
lwrubel: I was thinking this would be easier to test once this class is hooked into `UpdateMarc856RecordService` in the next ticket, since that service will be using this method. But maybe I'm missing something about testing this and it's better to do here?

app/services/catalog/folio_writer.rb
lwrubel: Good idea. Have gone ahead and renamed this to `catalog_record_id_type`. A little long, but still readable, I think.

app/services/catalog/folio_writer.rb
lwrubel: I *think* that is fine, but since this is copied from `SymphonyWriter` and this class is the alternative to that for the moment, would prefer to keep it as is for simplicity in comparing for now.

mjgiarlo (approved):

app/services/catalog/folio_writer.rb
mjgiarlo: kk

PR #4455, opened 2023-03-22
ignore nfs files when building file list for preservation

### LINKED ISSUE #4450: Do not include files with a filename pattern like '.nfsXXXXX' in the bag exported to preservation
We have been running into issues where files with names like '.nfs00000000097b892a00000325' are treated as if they are files that should be deposited into preservation. This means that the files are placed in the bag created at `/dor/export/{druid}' and included in bagit manifests:

```
85e2549c08830e85e89b12868f624d42 data/content/sc0487_1995_040_B6_TalkingStories_08001.tif
8ff225843f70a7f53e58b8dcc325a09a data/content/sc0487_1995_040_B6_TalkingStories_08001.pdf
0aaf3be17363afa532425b43c45aa5ae data/content/sc0487_1995_040_B6_TalkingStories_08001.xml
b183e88ca9dde028db4b3119e47343e3 data/content/sc0487_1995_040_B6_TalkingStories_08002.tif
0902b1a08e574773eea8cb11d53fb5cf data/content/sc0487_1995_040_B6_TalkingStories_08002.pdf
11c1ddd11d6dc5bf2b5cfd6d3bface0c data/content/sc0487_1995_040_B6_TalkingStories_08002.xml
ea4e211a07cc9237d431c5c3238fbbea data/content/kf609yz9250.pdf
8c9f3c14f896418dd6c329c426ac34ee data/metadata/.nfs00000000097b892a00000325  <----------- the problem file
5964ffd2ced0b8196db013019236cefe data/metadata/cocina.json
d1be5acdcf11313bf91fa86a4e0b515c data/metadata/contentMetadata.xml
234ce46cec2f1d61034e7c3e38981879 data/metadata/versionMetadata.xml
```
This appears to be causing the `transfer-bag` step to fail, though the error message says only
>Robots::SdrRepo::PreservationIngest::ItemError: Error transferring bag (via dor_services@dor-services-worker-prod-a.stanford.edu) for druid:kf609yz9250: Transfering bag for druid:kf609yz9250 to preservation failed. STDOUT =

Based on which items this affects - Goobi deposits - I think what is happening is this:

1. Goobi sends files to SDR, including a "stubContentMetadata.xml" file that SDR uses to create structural metadata
2. The stubContentMetadata.xml file is placed in the "metadata" folder for the druid
3. Workflow step takes the stubContentMetadata.xml file and generates Cocina
4. Workflow step deletes the stubContentMetadata.xml file as it has served its purpose
5. The NFS "lazy-delete" function leaves behind a '.nfsXXXXX' file in the metadata folder ([see lazy-delete FAQ](https://nfs.sourceforge.net/#faq_d2))
6. Workflow step creates the bag to be exported to preservation
7. The bag created in /dor/export/{druid} includes all files in the metadata folder, including '.nfsXXX'
8. `transfer-object` sends the bag to preservation
9. But the `.nfsXXX` file finally gets deleted
10. `transfer-object` fails because a file has disappeared

Based on prior Slack discussion, we think the simplest approach to solving this at this point is to exclude the `.nfsXXX` file from the bag and thus from the export to preservation: https://stanfordlib.slack.com/archives/C04D8DJDRJ9/p1677862646686569

It's not clear how to prevent the '.nfsXXX' files from being created. Potentially that would be solved by changing how Goobi sends data to SDR so that it doesn't create a stubContentMetadata.xml file in the first place.

peetucket:

Fixes #4450 - ignore .nfs files from metadata directory when building bag for preservation ingest

This changed `FileInventoryBuilder` is used by the `PreservationIngestService` to prepare the bag for sending to preservation, so I believe this change will exclude the .nfs files.  See https://github.com/sul-dlss/dor-services-app/blob/main/app/services/preservation_ingest_service.rb#L20-L22

Added an example `.nfs` file to the existing fixtures used by the spec I changed, and then verified that the new tests which ensure it's not there fail before the code fix and pass after it.

### DIFF
--- a/app/services/preserve/file_inventory_builder.rb
+++ b/app/services/preserve/file_inventory_builder.rb
@@ -54,7 +54,7 @@ module Preserve
 
       # reject any files in the metadata folder that start with .nfs followed by digits (e.g. .nfsNNNNNN)
       #  see https://github.com/sul-dlss/dor-services-app/issues/4450
-      metadata_file_group.signature_hash.reject! { |_, file_sig| file_sig.paths.any?(/\A(.nfs)\d+\z/) }
+      metadata_file_group.signature_hash.reject! { |_, file_sig| file_sig.paths.any?(/\A\.nfs\d+\z/) }
       metadata_file_group
     end
   end
--- a/app/services/preserve/file_inventory_builder.rb
+++ b/app/services/preserve/file_inventory_builder.rb
@@ -48,7 +48,14 @@ module Preserve
 
     # @return [Moab::FileGroup] Traverse the metadata directory and generate a metadata group
     def metadata_file_group
-      Moab::FileGroup.new(group_id: 'metadata').group_from_directory(metadata_dir)
+      metadata_file_group = Moab::FileGroup.new(group_id: 'metadata').group_from_directory(metadata_dir)
+
+      return unless metadata_file_group # this can be nil if there are no metadata files
+
+      # reject any files in the metadata folder that start with .nfs followed by digits (e.g. .nfsNNNNNN)
+      #  see https://github.com/sul-dlss/dor-services-app/issues/4450
+      metadata_file_group.signature_hash.reject! { |_, file_sig| file_sig.paths.any?(/\A(.nfs)\d+\z/) }
+      metadata_file_group
     end
   end
 end
--- a/spec/fixtures/workspace/ab/123/cd/4567/ab123cd4567/metadata/.nfs123456
+++ b/spec/fixtures/workspace/ab/123/cd/4567/ab123cd4567/metadata/.nfs123456
@@ -1,0 +1,1 @@
+File to be ignored
--- a/spec/services/preserve/file_inventory_builder_spec.rb
+++ b/spec/services/preserve/file_inventory_builder_spec.rb
@@ -37,6 +37,16 @@ RSpec.describe Preserve::FileInventoryBuilder do
       expect(result).to be_instance_of Moab::FileInventory
       expect(result.groups.size).to eq 2
     end
+
+    it 'includes the correct content files from contentMetadata' do
+      expect(result.groups[0].file_count).to eq(4) # four content files
+      expect(result.groups[0].path_list).to eq(['00000001.jp2', '00000001.jp2', '1.html', '2.html'])
+    end
+
+    it 'excludes .nfs files from inventory of metadata files' do
+      expect(result.groups[1].file_count).to eq(2) # two metadata files (excludes the .nfs file!)
+      expect(result.groups[1].path_list).to eq(['contentMetadata.xml', 'descMetadata.xml'])
+    end
   end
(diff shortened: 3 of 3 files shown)


app/services/preserve/file_inventory_builder.rb
```diff
@@ -48,7 +48,14 @@ def content_metadata_pathname
...
     # @return [Moab::FileGroup] Traverse the metadata directory and generate a metadata group
     def metadata_file_group
-      Moab::FileGroup.new(group_id: 'metadata').group_from_directory(metadata_dir)
+      metadata_file_group = Moab::FileGroup.new(group_id: 'metadata').group_from_directory(metadata_dir)
+
+      return unless metadata_file_group # this can be nil if there are no metadata files
+
+      # reject any files in the metadata folder that start with .nfs followed by digits (e.g. .nfsNNNNNN)
+      #  see https://github.com/sul-dlss/dor-services-app/issues/4450
+      metadata_file_group.signature_hash.reject! { |_, file_sig| file_sig.paths.any?(/\A(.nfs)\d+\z/) }
```
ndushay, line 57: https://github.com/sul-dlss/moab-versioning/blob/main/lib/moab/file_signature.rb has no `paths` method, but it does have a `path` method -- where does `.paths` come from?

you could save a line by doing select:

```
metadata_file_group.signature_hash.select! { |_, file_sig| file_sig.paths.none?(/\A(.nfs)\d+\z/) }
```

would return the correct group, I think.

ndushay (approved):
one suggestion for your consideration.

app/services/preserve/file_inventory_builder.rb
mjgiarlo: I'm guessing `Moab::FileManifestation#paths` is the method being used here.

peetucket: I like this idea, however `.select!` returns either self (if changes were made) or nil (if no changes were made). Also, if it returned self, it would return the `signature_hash` we are operating on, and not the `metadata_file_group`.  So I think your suggestion would work fine, but would still require us to explicitly return the `metadata_file_group` at the end of the method anyway.

peetucket: > I'm guessing `Moab::FileManifestation#paths` is the method being used here.

correct

mjgiarlo: Two questions about the regex:

1. Did you find you needed the parens? I'm not clear on what they're getting for us here.
2. Won't `/\A(.nfs)\d+\z/` also match a filename `anfs123`?

Sooo, would `/\A\.nfs\d+\z/` be just as good and a little tighter in terms of what it'll match?

peetucket: Thanks mike, you are 100% correct ... your regex makes more sense (and fixes my problem with the `.`).  Changed

PR #4458, opened 2023-03-23
[refactor] Remove indirection around ReleaseTags

jcoyne:

There is no need for this to be in two separate classes, one will do

CI


app/services/release_tags.rb
```diff
@@ -4,9 +4,154 @@
...
-    IdentityMetadata.for(cocina_object).released_for({})
+    new(cocina_object).released_for
   end
+
+  def initialize(cocina_object)
+    @cocina_object = cocina_object
+  end
+
+  # Determine projects in which an item is released
+  # @param [Hash{String => Boolean}] the released hash to add tags to
```
mjgiarlo, line 19: This method doesn't look like it takes any params. Cruft?

app/services/release_tags.rb
```diff
@@ -4,9 +4,153 @@
...
+  # @return [Hash] a hash of all tags
+  def release_tags_for_item_and_all_governing_sets
+    return_tags = release_tags # this objects initial release tags
+
+    return return_tags unless cocina_object.dro? # no need to continue if this is a collection, since they don't nest anymore
+
+    # now go through any collections it is a member of and add them
+    cocina_object.structural.isMemberOf.each do |collection_druid|
+      collection_tags = self.class.for(cocina_object: CocinaObjectStore.find(collection_druid))
+      return_tags = combine_two_release_tag_hashes(return_tags, collection_tags)
```
mjgiarlo, line 55: This is an elegant and simple way to build up return tags. 🍻

app/services/release_tags.rb
```diff
@@ -4,9 +4,153 @@
...
+    administrative_tags = AdministrativeTags.for(identifier: cocina_object.externalIdentifier) # Get admin tags once here and pass them down
+
+    # We now have the keys for all potential releases, we need to check the tags: the most recent timestamp with an explicit true or false wins.
+    # In a nil case, the lack of an explicit false tag we do nothing.
+    # Don't bother checking if already added to the release hash, they were added due to a self tag so that has won
+    (potential_applicable_release_tags.keys - released_hash.keys).each do |key|
+      latest_tag = latest_applicable_release_tag_in_array(potential_applicable_release_tags[key], administrative_tags)
+      next if latest_tag.nil? # Otherwise, we have a valid tag, record it
+
+      released_hash[key] = { 'release' => latest_tag['release'] }
```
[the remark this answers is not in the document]
jcoyne: This is just a copy of what used to be in IdentityMetadata https://github.com/sul-dlss/dor-services-app/pull/4458/files#diff-3750e58611ec2de32c3ddf15eed23439a0bef2b70dfcaf003209d80b5b493af5L19  Maybe it was wrong before? I haven't changed it.

mjgiarlo:
@jcoyne Since this appears to be a straight-up refactor, feel free to merge when 💚

PR #4464, opened 2023-03-23
Fix bug with generating thumbnail links for virtual objects

### LINKED ISSUE #4456: 856: When releasing a virtual object, the 856 should include the filename of the first JP2 of the first constituent
This ticket applies to the content of the 856 and applies to both Symphony and Folio. 

When a virtual object is released, the 856 should include the filename for the first image in the object. This is used as a thumbnail in various displays. For virtual objects, all images are contained in the constituent druids. The virtual object is just an aggregate of the druids it contains.

This means that when creating an 856 for a virtual object, the filename that goes into the 856 is taken from the first constituent.

For example, [in this prod object](https://searchworks.stanford.edu/view/10448742), the 856 contains a reference to
- the virtual object (the purl) 
- the first image file in the first constituent (`file:kq126jw7402%2F1592A.jp2`)
```
 u| http://purl.stanford.edu/cp799hh4428 x| SDR-PURL x| item x| file:kq126jw7402%2F1592A.jp2 x| collection:xh235dd9059::David Rumsey Map Collection at Stanford University Libraries
```

However, after the 856 refactor, the filename is missing from the 856. Only the constituent druid is being included:
```
‡u https://sul-purl-stage.stanford.edu/jj557rm0778 ‡x SDR-PURL ‡x item ‡x file:druid%3Acm619nj6426 ‡x rights:world 
```
In this case `file:druid%3Acm619nj6426` is correct in that it references the first constituent, it's just missing the filename.

mjgiarlo:

Fixes #4456

This bug, as far as I can tell, has been around since we cocinized the thumbnail service which prior to this commit returned the druid of the first member of a virtual object as its thumbnail link. Instead, pass this druid back to the thumbnail service to get the expected thumbnail link (the bare druid followed by the filename of the first jp2 of the first member).

- [x] CI
- [x] QA/stage

### DIFF
--- a/app/services/thumbnail_service.rb
+++ b/app/services/thumbnail_service.rb
@@ -24,8 +24,8 @@ class ThumbnailService
       end
     end
 
-    return if object.structural.hasMemberOrders.empty?
+    return if object.structural.hasMemberOrders.empty? || object.structural.hasMemberOrders.first.members.empty?
 
-    object.structural.hasMemberOrders.first.members.first
+    self.class.new(CocinaObjectStore.find(object.structural.hasMemberOrders.first.members.first)).thumb
   end
 end
--- a/spec/services/thumbnail_service_spec.rb
+++ b/spec/services/thumbnail_service_spec.rb
@@ -19,7 +19,6 @@ RSpec.describe ThumbnailService do
 
     context 'for an item' do
       let(:druid) { 'druid:bc123df4567' }
-
       let(:object) { build(:dro, id: druid).new(structural:) }
 
       context 'with no structural metadata' do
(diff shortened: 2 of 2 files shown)


app/services/thumbnail_service.rb
```diff
@@ -26,6 +26,6 @@ def thumb
 
     return if object.structural.hasMemberOrders.empty?
 
-    object.structural.hasMemberOrders.first.members.first
+    self.class.new(CocinaObjectStore.find(object.structural.hasMemberOrders.first.members.first)).thumb
```
mjgiarlo, line 29: This is the bugfix.

aaron-collier (approved):

PR #4467, opened 2023-03-23
no need to memoize results only used once

peetucket:

This issue sul-dlss/common-accessioning#1043 discusses intermittent problems with the goobi service sending data during registration.  I can't think of a reason for this (particularly as the tags used to determine the workflow name we send to goobi are in local database and so there should be nothing like a race condition for when they are available).

I did notice that we memoize certain values in the goobi service, including the value that apparently sometimes gets set as the default incorrectly, for no apparent reason (they are used only once per instance).  I don't have a reason that this would explain the intermittency, but just putting this up for discussion/reaction.

Existing tests

### DIFF
--- a/app/services/goobi_service.rb
+++ b/app/services/goobi_service.rb
@@ -87,7 +87,7 @@ class GoobiService
   # @return [String] first goobi workflow tag value if one exists (default from config if none)
   def goobi_workflow_name
     dpg_workflow_tag_id = 'DPG : Workflow : '
-    content_tag = AdministrativeTags.for(identifier: cocina_obj.externalIdentifier).select { |tag| tag.include?(dpg_workflow_tag_id) }
+    content_tag = AdministrativeTags.for(identifier: cocina_obj.externalIdentifier).select { |tag| tag.start_with?(dpg_workflow_tag_id) }
     content_tag.empty? ? Settings.goobi.default_goobi_workflow_name : content_tag[0].split(':').last.strip
   end
 
@@ -132,7 +132,7 @@ class GoobiService
   # @return [boolean]
   def goobi_ocr_tag_present?
     dpg_goobi_ocr_tag = 'DPG : OCR : TRUE'
-    AdministrativeTags.for(identifier: cocina_obj.externalIdentifier).any? { |tag| tag.casecmp(dpg_goobi_ocr_tag).zero? } # case insensitive compare
+    AdministrativeTags.for(identifier: cocina_obj.externalIdentifier).any? { |tag| tag.casecmp?(dpg_goobi_ocr_tag) } # case insensitive compare
   end
 
   # returns the first collection_id the object is contained in (if any)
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

mjgiarlo:
LGTM. Feel free to try it out to see if it makes things better?

app/services/goobi_service.rb
```diff
@@ -86,11 +86,9 @@ def xml_request
   # returns the name of the goobiworkflow in the object by examining the objects tags
   # @return [String] first goobi workflow tag value if one exists (default from config if none)
   def goobi_workflow_name
-    @goobi_workflow_name ||= begin
-      dpg_workflow_tag_id = 'DPG : Workflow : '
-      content_tag = AdministrativeTags.for(identifier: cocina_obj.externalIdentifier).select { |tag| tag.include?(dpg_workflow_tag_id) }
-      content_tag.empty? ? Settings.goobi.default_goobi_workflow_name : content_tag[0].split(':').last.strip
-    end
+    dpg_workflow_tag_id = 'DPG : Workflow : '
+    content_tag = AdministrativeTags.for(identifier: cocina_obj.externalIdentifier).select { |tag| tag.include?(dpg_workflow_tag_id) }
```
jcoyne, line 90:
```suggestion
    content_tag = AdministrativeTags.for(identifier: cocina_obj.externalIdentifier).select { |tag| tag.start_with?(dpg_workflow_tag_id) }
```

peetucket: Oh that's interesting -- possibly an extra space or other innocuous issue?

jcoyne: I just noticed these potential changes that I think make it more readable, but this is what the code had before.

peetucket: actually still doesn't explain the retry fixing it (unless they also edited the tag too and didn't mention it)

andrewjbtw:
Deploying is worth a try. There are a couple in this situation at the moment where they seem to have all the right tags
https://argo.stanford.edu/catalog?f%5Bwf_wps_ssim%5D%5B%5D=goobiWF%3Agoobi-notify%3Aerror

The users who reported this are familiar with the process of adding missing tags, so these are all cases where no editing was needed, just a retry. I tried to reproduce the issue in stage by registering 100 items in a batch to the goobiWF but it went fine.

andrewjbtw:
I've wondered if moving to sidekiq had an effect on this, maybe by sending in more requests in a short period than Goobi could handle.

peetucket:
> I've wondered if moving to sidekiq had an effect on this, maybe by sending in more requests in a short period than Goobi could handle.

Possible I guess the problem is on the Goobi side, though that error they are returning would be an odd one (as if we sent the wrong data to them).  Would be nice to inspect their app logs right after it happens (not sure where they are on the goobi server though and the one object i looked at errored out a couple weeks ago now)

peetucket:
Propose we just merge and all it go to prod next week and see if this makes any difference.

PR #4473, opened 2023-03-23
Export release tags to the public cocina

jcoyne:
The public cocina release tags will only contain the latest release tag for each project (same as the public XML)

- [x] CI
- [x] stage

### DIFF
--- a/app/services/catalog/folio_writer.rb
+++ b/app/services/catalog/folio_writer.rb
@@ -42,7 +42,7 @@ module Catalog
       retry_connection do
         FolioClient.edit_marc_json(hrid: catalog_record_id) do |marc_json|
           marc_json['fields'].reject! { |field| (field['tag'] == '856') && (field['content'].include? purl_subfield) }
-          marc_json['fields'] << marc_856_field if released_to_searchworks?(cocina_object)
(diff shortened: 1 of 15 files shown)


app/services/catalog/folio_writer.rb
```diff
@@ -62,9 +62,7 @@ def fetch_catalog_record_ids(current:)
     end
 
     def released_to_searchworks?(cocina_object)
-      released_for = ::ReleaseTags.for(cocina_object:)
-      rel = released_for.transform_keys { |key| key.to_s.upcase } # upcase all release tags to make the check case insensitive
-      rel.dig('SEARCHWORKS', 'release').presence || false
+      ReleaseTags.released_to_searchworks?(cocina_object:)
     end
```
mjgiarlo, line 66: Do you think it's still valuable to have `FolioWriter#released_to_searchworks?` defined? (Same question for the two classes below as well.)

mjgiarlo:
@andrewjbtw If you have time, would you like to test this in stage?

app/services/catalog/folio_writer.rb
jcoyne: I don't care one way or t'other

mjgiarlo: K. No need to go the extra step, then.

mjgiarlo (approved):

Issue #4454, opened 2023-03-22
856/Folio: when releasing more than one druid linked to the same Folio HRID, only one job actually succeeds


andrewjbtw:
The digitized serials workflow involves associating multiple druids with a single catalog record. When releasing more than one druid at the same time (via bulk action) to the same catalog record, only one of the jobs succeeds. The other hits an error.

Example: these two druids are associated with the same catalog record
https://argo-qa.stanford.edu/catalog?f%5Bexploded_tag_ssim%5D%5B%5D=856+%3A+serial&f%5Btag_ssim%5D%5B%5D=Project+%3A+856

To release at the same time:

- Click on the "bulk action" button to the right above the search result list
<img width="866" alt="Screenshot 2023-03-22 at 10 53 35 AM" src="https://user-images.githubusercontent.com/8975583/226994891-c076429d-166b-430c-98b6-eed2e5b99f02.png">

- Choose "new bulk action"
- Choose "manage release" (this is the top of the list, so should be selected already
- Keep "release this object" selected
- Choose "Populate from previous search" to select both druids
- Submit the job
- Go back to the search
- Wait for a while
- Reload the search results

- See that the second item has a 500 error
<img width="1183" alt="Screenshot 2023-03-22 at 10 57 09 AM" src="https://user-images.githubusercontent.com/8975583/226995535-8c6ec9bc-221c-42df-9b27-414915682835.png">

- See HB errors as well

It looks like this is happening for both "release" and "unrelease". I'm guessing this is because only one edit can be made at a time.

andrewjbtw:
This has stopped working. If I release the two items in the example in the ticket, only one succeeds. The other one works if I wait and then re-release. I see the same problem with un-releasing. Only one job succeeds.

mjgiarlo:
@andrewjbtw Well, that wasn't supposed to happen. 😆 

I'll re-open and check it out after lunch. Should I use the same test objects above?

mjgiarlo:
@andrewjbtw I tried the release bulk action on the above two test objects and it seemed to work fine (releaseWF in both completed). Should I use diff test objects? Or run it _n_ times in a row?

andrewjbtw:
Let me try it with just those two. I was trying it with a larger batch of unrelated records and it never worked that way.

andrewjbtw:
[...]

PR #4474, opened 2023-03-23
[HOLD] Retry Folio edit operations a few times 

mjgiarlo:

Fixes https://github.com/sul-dlss/dor-services-app/issues/4454

If multiple Folio edits are requested serially, at least two different exceptions may be raised:

* a Faraday timeout
* a StandardError emitted from FolioClient

This is documented in the FolioClient gem: https://github.com/sul-dlss/folio_client/blob/181ee68c8879030734f5ad6a639904145c4ffc06/lib/folio_client/records_editor.rb#L19-L24

Both of these may be safely retried a small number of times, and both conditions go away within seconds. I can be convinced that we ought to transparently handle this in the gem, but I'm not sure we'll be encountered this outside the DSA FolioWriter so maybe this is a fine location for now.

- [x] CI
- [ ] QA

### DIFF
--- a/spec/services/catalog/folio_writer_spec.rb
+++ b/spec/services/catalog/folio_writer_spec.rb
@@ -98,6 +98,51 @@ RSpec.describe Catalog::FolioWriter do
         expect(FolioClient).to have_received(:edit_marc_json).with(hrid:)
         expect(folio_response_json).to eq(updated_marc_json)
       end
+
+      context "when Folio operation raises less than #{described_class::MAX_TRIES} times" do
+        before do
+          client_responses = [nil, nil, folio_response_json]
+          allow(FolioClient).to receive(:edit_marc_json) do |_, &block|
+            response = client_responses.shift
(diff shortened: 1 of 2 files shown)


spec/services/catalog/folio_writer_spec.rb
```diff
@@ -98,6 +98,51 @@
...
+        end
+
+        it 'logs a warning message and notifies Honeybadger' do
+          folio_writer.save
+          expect(FolioClient).to have_received(:edit_marc_json).with(hrid:).exactly(4).times
+          expect(Rails.logger).to have_received(:warn).exactly(4).times
+          expect(Honeybadger).to have_received(:notify)
+            .with(
+              'Error retrying Folio edit operations',
+              backtrace: instance_of(Array),
```
mjgiarlo, line 138: We really don't care to test what's in the backtrace.

aaron-collier (approved):

Issue #4453, opened 2023-03-22
856/Folio: Un-release should remove 856 from MARC record


andrewjbtw:
When un-releasing a druid, the 856 corresponding to that druid should be removed from the MARC record. I tested with https://argo-qa.stanford.edu/view/yj417jr1159 by

- release to SW 
- unrelease to SW

The workflows report success but the 856 remains in the record ([folio link](https://folio-test.stanford.edu/inventory/view/9afe9b02-1e4b-5ba8-9842-0214ca9d63d1?qindex=hrid&query=a999)):
```
‡z Available to Stanford-affiliated users. ‡u https://sul-purl-stage.stanford.edu/yj417jr1159 ‡x SDR-PURL ‡x item ‡x file:yj417jr1159%2Fpage_0001.jp2 ‡x rights:group=stanford 
```

To reproduce:

- Go to Argo item page for item with an HRID
- Release first
  - Click "manage release" button
  - Choose "release this object"
  - Wait for process to complete and verify 856 is in the MARC
- Then unrelease
  - Go back to Argo page
  - Click "manage release" button
  - Choose "Do not release this object"
  - Wait for process to finish and check the MARC again

edsu:
I happened to notice that releasing again (after un-release) adds a duplicate 856?

andrewjbtw:
That is, multiple 856s are fine as long as they refer to different druids.

edsu:
If I'm reading [this test](https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/catalog/folio_writer_spec.rb#L148-L167) correctly it looks like a MARC record will not get a duplicate 856 field if the release tags are set up correctly? So I guess they are not being set correctly when something is Unreleased?

edsu:
Unless I'm reading it wrong I think this test ensures that FolioWriter will remove any 856 fields for the same DRUID, assuming that the correct ReleaseTags are used? 

This seems to imply to me, that perhaps the Cocina object hasn't been updated such that the ReleaseTags are correct? Still digging...

edsu:
Maybe I'm not understanding the DSA API correctly. When I watch the DSA Apache logs during the `release` action I can see DSA gets a POST to `/v1/objects/{id}/publish`. But when I `un-release` I don't see a POST to `/v1/objects/{id}/unpublish` and see a POST to `/v1/objects/{id}/publish` instead?

DSA HTTP traffic when releasing to SearchWorks:

```
[...]

PR #4475, opened 2023-03-24
Explicit unrelease

edsu:

For peace of mind this is just making the unrelease a bit more explicit.

Refs #4453

Unit tests.

### DIFF
--- a/spec/services/catalog/folio_writer_spec.rb
+++ b/spec/services/catalog/folio_writer_spec.rb
@@ -145,7 +145,7 @@ RSpec.describe Catalog::FolioWriter do
       end
     end
 
-    context 'when a single catalog record id that has not been released to Searchworks' do
+    context 'when a single catalog record id that is unreleased' do
       let(:identification) do
         {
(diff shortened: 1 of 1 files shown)


spec/services/catalog/folio_writer_spec.rb
```diff
@@ -157,7 +157,7 @@
           ]
         }
       end
-      let(:release_data) { {} }
+      let(:release_data) { { 'Searchworks' => { 'release' => false } } }
```
mjgiarlo, line 160: Is it worth breaking this into two tests: one with explicit and one with implicit?

edsu: Does it make sense the way I did it with nested context?

mjgiarlo (approved):

spec/services/catalog/folio_writer_spec.rb
mjgiarlo: I think so, yeah. :)

PR #4478, opened 2023-03-27
Upgrade to Sidekiq7

justinlittman:
6 + 1 = 7

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -144,7 +144,7 @@ class ObjectsController < ApplicationController
                                      'Datacite requires that this object have creators and a datacite extension with resourceTypeGeneral')
     end
 
-    UpdateDoiMetadataJob.perform_later(@cocina_object)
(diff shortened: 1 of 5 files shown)


app/controllers/objects_controller.rb
```diff
@@ -144,7 +144,7 @@ def update_doi_metadata
                                      'Datacite requires that this object have creators and a datacite extension with resourceTypeGeneral')
     end
 
-    UpdateDoiMetadataJob.perform_later(@cocina_object)
+    UpdateDoiMetadataJob.perform_later(Cocina::Models.without_metadata(@cocina_object).to_json)
```
edsu, line 147: Were these changes meant for the upgrade of Sidekiq?

edsu (approved):

app/controllers/objects_controller.rb
justinlittman: Yes, this version is more insistent on serialization-safe job paramters.

Issue #3298, opened 2022-02-03
Remediate ETD structural cocina `FileSet.type` that use Cocina::Models::Vocab::Resources.main-original, .main-augmented, .supplement, .permission


ndushay:
Per discussion with infra team and Andrew on 2/3/2022, we have agreed that the following ETD specific Resource types are deprecated.  We want to remediate any SDR objects with structural metadata that uses:

- `Cocina::Models::Vocab::Resources.main-original`
- `Cocina::Models::Vocab::Resources.main-augmented`
- `Cocina::Models::Vocab::Resources.supplement`
- `Cocina::Models::Vocab::Resources.permissions`

as the `FileSet` type to use `Cocina::Models::Vocab::Resources.file` for `FileSet` type instead.

**BE CAREFUL:  we are speaking of the `FileSet` level, _not `File`_ and `Cocina::Models::Vocab::Resources.file`, _`not Cocina::Models::Vocab.file`_**

In addition, if the "label" property _at the `FileSet` level_ is blank for any of these cocina `FileSet` objects, it should be assigned as follows:

- `Cocina::Models::Vocab::Resources.main-original` -> `FileSet` label should be `"Body of dissertation (as submitted)"`
- `Cocina::Models::Vocab::Resources.main-augmented` -> `FileSet` label should be `"Body of dissertation"`
- `Cocina::Models::Vocab::Resources.supplement` -> `FileSet` label should be `"supplemental file"`
- `Cocina::Models::Vocab::Resources.permissions` -> `FileSet` label should be `"permission file"`

This will match the practice assumed when PR https://github.com/sul-dlss/hydra_etd/pull/1167 was deployed in early Feb, 2022.

Note that it is expected that all the affected objects will be:
- ETD
- accessioned before May 2021 (exact cutoff date likely in early May)

ndushay:
NOTE:  spoke to both Andrew and Cathy today and all concurred on:

- publish:  yes
- version:  no

ndushay:
This remediation has been done in qa, stage and prod using the migrator in PR #4480.

PR #4480, opened 2023-03-29
migrator for remediating old etd fileset types and labels

ndushay:
NOTE: spoke to both Andrew and Cathy today and all concurred on: 
- publish: yes
- version: no

Part of #3298
Part of sul-dlss/cocina-models/issues/302

To shrink the FileSetType vocabulary in cocina-models to not have special case vocabulary for ETDs ... especially since when we updated the ETD app not to use Fedora3, we stopped using the special cocina vocabulary.  :-P

migrator spec and dry-run.   Also there will be validation run with migration.


app/services/migrators/etd_fileset_vocab.rb
```diff
@@ -0,0 +1,8279 @@
...
+  class EtdFilesetVocab < Base # rubocop:disable Metrics/ClassLength
+    # A migrator may provide a list of druids to be migrated (optional).
+    def self.druids
+      DRUIDS
+    end
+
+    # A migrator must implement a migrate? method that returns true if the SDR object should be migrated.
+    def migrate?
+      DRUIDS.include?(ar_cocina_object.external_identifier) &&
+        FILESET_TYPES_TO_MIGRATE.intersect?(ar_cocina_object[:structural]['contains'].pluck('type'))
```
justinlittman, line 15: `ar_cocina_object.structural` instead of `ar_cocina_object[:structural]`?

app/services/migrators/etd_fileset_vocab.rb
```diff
@@ -0,0 +1,8279 @@
...
+    # A migrator must implement a migrate method that migrates (mutates) the ActiveRecord cocina object.
+    def migrate
+      Array(ar_cocina_object[:structural]['contains']).each do |fileset|
+        next unless FILESET_TYPES_TO_MIGRATE.include?(fileset['type'])
+
+        oldtype = fileset['type']
+        fileset['type'] = Cocina::Models::FileSetType.file
+        next if fileset['label'].present?
+
+        case oldtype
```
justinlittman, line 27:
```
fileset['label'] = case ...
```

justinlittman (commented):
Super minor suggestions.

app/services/migrators/etd_fileset_vocab.rb
ndushay: I thought I tried that.  'contains' certainly needed to be as indicate ...  Will try.

justinlittman (approved):

app/services/migrators/etd_fileset_vocab.rb
ndushay: d'oh - great suggestion.

app/services/migrators/etd_fileset_vocab.rb
ndushay: yeppers!  `.structural` works but `.structural.contains` does not work, at least for my specs.

justinlittman: That makes sense: `structural` is an active record attribute; `contains` is a field in a JSON object.

PR #4491, opened 2023-04-03
spawn jobs for released members

peetucket:

fixes sul-dlss/argo#3972

I believe this should spawn a series of independent publish jobs for each member of a collection, setting them to the low priority queue, instead of trying to iterate and re-publish each collection member as part of the collection publishing.  This should allow the collection publish itself to complete quickly, while the collection members publish over time as the jobs complete.

~~HOLD - tested on stage and need to pass in the workflow param...fixing PR now~~

Updated the test
Tested on stage with a large collection and watched jobs spawn and re-publish happen

### DIFF
--- a/spec/services/publish/metadata_transfer_service_spec.rb
+++ b/spec/services/publish/metadata_transfer_service_spec.rb
@@ -244,9 +244,15 @@ RSpec.describe Publish::MetadataTransferService do
       allow(described_class).to receive(:new).and_return(service)
     end
 
-    it 'calls publish on a new instance' do
+    it 'calls publish on a new instance with the default workflow' do
       described_class.publish(cocina_object)
       expect(described_class).to have_received(:new).with(cocina_object, workflow: 'accessionWF')
+      expect(service).to have_received(:publish)
+    end
+
+    it 'calls publish on a new instance with a specific workflow' do
+      described_class.publish(cocina_object, workflow: 'releaseWF')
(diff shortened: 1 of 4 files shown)


justinlittman (approved):

peetucket:
@andrewjbtw do we have an easy way to verifying this change on stage or qa (i.e. a collection with many members that could be republished?)

andrewjbtw:
The collection "5000 items, approx, give or take" has, well, approximately 5000 items, give or take: https://argo-stage.stanford.edu/view/druid:tv031gp2794

spec/services/publish/metadata_transfer_service_spec.rb
```diff
@@ -235,4 +236,24 @@
...
+
+    before do
+      allow(described_class).to receive(:new).and_return(service)
+    end
+
+    it 'calls publish on a new instance with the default workflow' do
+      described_class.publish(cocina_object)
+      expect(described_class).to have_received(:new).with(cocina_object, workflow: 'accessionWF')
+      expect(service).to have_received(:publish)
+    end
```
peetucket, line 251: aha thanks... i basically had this exact same test but was using `to receive(:publish)` instead of `have_received(:publish)` and this wasn't working, because the double needed to be checked ... I was using the wrong expectation

peetucket:
So @justinlittman I had to add the workflow param to this PR to get it to work after you approved (this was a problem when i tested on stage).  In case you want to verify again before I merge.  It did work on stage though (at least my collection publish did).

Issue #4359, opened 2023-02-09
[SPIKE] Perform content metadata diffing as part of shelving


mjgiarlo:
CONTEXT: work is mostly done with an approach detailed in #4492, which has been reviewed by John.  What's next is to respond/review his comments, which is likely involving improved test coverage.

# Background

Currently, DSA uses the prescat HTTP API in order to [generate content diffs when shelving](https://github.com/sul-dlss/dor-services-app/blob/main/app/services/shelving_service.rb#L40-L64). These diffs are used by:

* [`ShelvableFilesStager`](https://github.com/sul-dlss/dor-services-app/blob/main/app/services/shelvable_files_stager.rb) to figure out which files need staging and which ones need to be copied over from preservation.
* [`DigitalStacksService`](https://github.com/sul-dlss/dor-services-app/blob/main/app/services/digital_stacks_service.rb) to shelve files needing shelving, to remove files from the stacks that are no longer present or shelvable, and to rename files on the stacks that have had their filenames changed.

# Problem Statement

For items with very many files (~5K-10K+), this operation on the prescat side is extremely slow: it can take over ten minutes, possibly up to thirty minutes or more, to complete the content diffing[^1]. On top of that:

* There is nothing about this content diffing that is really about the Moab file format or versioning. The only Moab versioning piece is figuring out which content metadata to compare against.
* Content diffing is only needed in one place across all of SDR and it's right here in DSA for shelving.

# Spike Proposal

Given that, we propose a spike cutting out the middle person. For content diffing, DSA should:

1. Ask prescat for the most recent content metadata XML[^2]: `Preservation::Client.objects.metadata(druid: 'bc123df4567', filepath: 'contentMetadata.xml')`
2. Convert the content metadata XML to a Cocina structural JSON[^3]
3. Produce a diff between the two Cocina structural JSON objects sufficient for what `ShelvableFilesStager` and `DigitalStacksService` need

[^1]: And this is **after** a greater than 2x performance boost was added to the code responsible for the diffing in the `moab-versioning` gem.
[^2]: Even though we'd rather compare Cocina JSON structures to figure out the diff, not all items in SDR have Cocina JSON in preservation, so instead of handling both items with Cocina JSON and items without request and work with the content metadata XML.
[...]

PR #4492, opened 2023-04-03
content metadata diffing in DSA job instead of API call

peetucket:

A fix for #4359 -- perform contentMetadata diffing in place in DSA instead of farming it out to preservation client.  For very large objects, this diffing takes a long time, during which the call to preservation client times out, and then the entire shelving job errors out.  This change will not be a performance improvement, but it moves all of the long running work directly into the DSA job, so there shouldn't be any timing out of a web service call.  Basically we only use preservation client to fetch the previous versions contentMetadataXML and then do all of the same diffing that would have happened in preservation client using the code that already exists in the moab-versioning gem, but directly in the shelving job in DSA.

Possible follow on work:
- confirm if anyone else uses the content Metadata diffing API endpoint in preservation-catalog: https://github.com/sul-dlss/preservation_catalog/blob/main/app/controllers/objects_controller.rb#L89-L108, and if not, remove it and the associated preservation client methods: `Preservation::Client.objects.content_inventory_diff` and `Preservation::Client.objects.shelve_content_diff`

Quick check seems to suggest there are no other consumers of those preservation-client methods (both of which call the same preservation catalog endpoint):
- https://github.com/search?q=org%3Asul-dlss+content_inventory_diff&type=code
- https://github.com/search?q=org%3Asul-dlss+shelve_content_diff&type=code

- Updated specs
- Integration tests on stage


app/services/shelving_service.rb
```diff
@@ -35,32 +39,29 @@ def shelve
...
-    @preserve_diff ||= begin
-      inventory_diff = Preservation::Client.objects.content_inventory_diff(druid: cocina_object.externalIdentifier, content_metadata:, subset: 'preserve')
-      inventory_diff.group_difference('content')
-    end
-  rescue Preservation::Client::Error => e
+  def content_diff(subset)
+    new_inventory = Stanford::ContentInventory.new.inventory_from_cm(content_metadata, druid, subset)
+    inventory_diff = Moab::FileInventoryDifference.new.compare(base_inventory(subset), new_inventory)
+    inventory_diff.group_difference('content')
+  rescue StandardError => e
```
peetucket, line 53: essentially just compare the previous version XML to the current version cocina converted XML using the same method that is currently used in the preservation client call

app/services/shelving_service.rb
```diff
@@ -35,32 +39,29 @@ def shelve
...
-    raise ShelvingService::ShelvingError, e
+  def base_inventory(subset)
+    cm_from_pres = Preservation::Client.objects.metadata(druid:, filepath: 'contentMetadata.xml')
+    Stanford::ContentInventory.new.inventory_from_cm(cm_from_pres, druid, subset)
+  rescue Moab::ObjectNotFoundException, Moab::FileNotFoundException
+    # Create a skeletal FileInventory object, containing no file entries
+    storage_object = Moab::StorageObject.new(druid, 'dummy')
+    base_version = Moab::StorageObjectVersion.new(storage_object, 0)
+    base_version.file_inventory('version')
   end
```
peetucket, line 65: basically what is happening in https://github.com/sul-dlss/moab-versioning/blob/main/lib/stanford/storage_services.rb#L22-L47

try to get the previous version of the contentMetadata from preservation, if this is a v0 object (not yet accessioned), we create a skeletal version

app/services/shelving_service.rb
```diff
@@ -35,32 +39,29 @@ def shelve
...
+  # @param [String] subset: 'shelve', 'preserve', 'publish', or 'all' .... filters file diffs
   # @raise [ShelvingService::ShelvingError] if something went wrong.
-  def preserve_diff
-    @preserve_diff ||= begin
-      inventory_diff = Preservation::Client.objects.content_inventory_diff(druid: cocina_object.externalIdentifier, content_metadata:, subset: 'preserve')
-      inventory_diff.group_difference('content')
-    end
-  rescue Preservation::Client::Error => e
+  def content_diff(subset)
+    new_inventory = Stanford::ContentInventory.new.inventory_from_cm(content_metadata, druid, subset)
```
peetucket, line 50: `content_metadata` is the current versions (converted from cocina, using an existing method on this class)

app/services/shelving_service.rb
```diff
@@ -35,32 +39,29 @@ def shelve
 
   private
 
-  attr_reader :cocina_object
+  attr_reader :cocina_object, :druid
```
peetucket, line 43: we use druid all over the place in this class, set it in the initializer and create an attr_reader

spec/services/shelving_service_spec.rb
```diff
@@ -60,8 +60,6 @@
     allow(WorkflowClientFactory).to receive(:build).and_return(workflow_client)
     allow(Settings.stacks).to receive_messages(local_stacks_root: stacks_root, local_workspace_root: workspace_root)
     allow(Cocina::ToXml::ContentMetadataGenerator).to receive(:generate).and_return(content_metadata)
-    allow(Preservation::Client.objects).to receive(:shelve_content_diff).and_return(mock_shelve_diff)
-    allow(Preservation::Client.objects).to receive(:content_inventory_diff).and_return(content_inventory_diff)
     allow(ShelvableFilesStager).to receive(:stage)
```
peetucket, line 63: need to add some mocking of the calls to fetch contentMetadata XML that we now do instead of this

spec/services/shelving_service_spec.rb
```diff
@@ -47,10 +47,10 @@
     XML
   end
 
+  let(:previous_content_metadata) { content_metadata }
```
jmartin-sul, line 50: since this PR is kind of reimplementing a moab-versioning comparison method, and since the result of that comparison is important to the diff that determines what gets shelved, i wonder if you'd be open to doing something more integration-y in the spec.  e.g., we mock intercept `ShelvingService#content_metadata` and return the translated-from-cocina XML (since presumably that translation is tested elsewhere), and then similar for the content metadata XML from pres client; and then expect a particular chunk of diff XML from the various moab-versioning calls?

probably the easiest and safest way i can think of for doing that is ripping off an example or two from [the moab-versioning spec for the `compare_cm_to_version` method](https://github.com/sul-dlss/moab-versioning/blob/6d0f51a2d2300e62faed203f39253bee60c4dbad/spec/unit_tests/stanford/storage_services_spec.rb).  another way might be to find an example druid and take the content metadata XML from two successive versions, and run those through the current pres cat implementation.  my gut feeling is that the latter might be slightly more work, but i'm not confident about that.

jmartin-sul (changes requested):
i'm a little wary of pulling some of the meat of the diffing logic from moab-versioning (via pres cat) into DSA, without also pulling in some more thorough test code than what DSA had before when it wasn't concerned with which moab-versioning methods need to be invoked to produce the diff we want.  so i was hoping you'd consider augmenting the tests to dig into that a bit more.  but also definitely interested to hear opinions contrary to that, or to hear if i'm misinterpreting the tests that are there now.

i guess one other option, assuming my read is correct that this PR is pulling in some of the logic of `compare_cm_to_version`, might be to do this sort of refactoring so it can take an XML string for both the base and new versions, but in moab-versioning, where it could be more easily tested.  upside of that seems to be maybe less duplicative test code across our portfolio (assuming there's agreement that my testing worry is warranted).  downside is the minor pain of cutting a new moab-versioning release, and extending a codebase that no one is eager to maintain.

if i'm not off-base with this review, i'd be interested in pairing on augmenting tests in this PR, or moab-versioning refactoring to support this PR's higher-level goal but with a little less responsibility handed over to DSA, or whatever other ideas.

jmartin-sul (commented):
got mentioned during standup, but just wanted to note explicitly that i think the overall approach in this PR makes a lot of sense, and my concerns were all around in the weeds details and testing thoroughness.  sorry, forgot to say that in my initial review.

app/services/shelving_service.rb
peetucket: Yes, I believe everything you said here is accurate.  I will add a code comment to better explain what is meant by `current version` content metadata, something like `SDR object's content metadata before being sent to preservation, i.e. either initial version if not yet accessioned, or version in the process of being accessioned.`

app/services/shelving_service.rb
```diff
@@ -35,32 +39,29 @@ def shelve
...
-  # @raise [Preservation::Client::Error] if bad response from preservation catalog.
-  # @raise [ConfigurationError] if missing local workspace root.
-  # @raise [ShelvingService::ShelvingError] if something went wrong.
-  def shelve_diff
-    @shelve_diff ||= Preservation::Client.objects.shelve_content_diff(druid: cocina_object.externalIdentifier, content_metadata:)
-  rescue Preservation::Client::Error => e
-    raise ShelvingService::ShelvingError, e
+  def base_inventory(subset)
+    cm_from_pres = Preservation::Client.objects.metadata(druid:, filepath: 'contentMetadata.xml')
+    Stanford::ContentInventory.new.inventory_from_cm(cm_from_pres, druid, subset)
```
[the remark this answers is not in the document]
peetucket: Yes, you are correct, my code here is basically copy-pasta from the code in moab-versioning you reference.  My reading of the `inventory_from_cm` method being called is that the version can be nil, and just gets left out of the inventory being generated: https://github.com/sul-dlss/moab-versioning/blob/main/lib/stanford/content_inventory.rb#L20-L30

I was assuming it would not be important for the diffing we are doing here, since it relies on grouping of XML nodes and not a version number, but I'm certainly not an expert.  If we wanted to be extra cautious, we could probably make a second preservation-client call to fetch the version number right before this line, and then pass it on (as is currently done).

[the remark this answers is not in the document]
peetucket: Something like this: 

```
version = begin
  Preservation::Client.objects.current_version(druid)
 rescue Preservation::Client::NotFoundError
   nil
end
```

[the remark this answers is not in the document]
peetucket: actually, it's even easier, as we already have that rescue block in this method, right below, just as it is in moab-versioning.  just added a commit to fetch the version and pass it on

app/services/shelving_service.rb
```diff
@@ -35,32 +39,29 @@ def shelve
...
   # @raise [ShelvingService::ShelvingError] if something went wrong.
-  def preserve_diff
-    @preserve_diff ||= begin
-      inventory_diff = Preservation::Client.objects.content_inventory_diff(druid: cocina_object.externalIdentifier, content_metadata:, subset: 'preserve')
-      inventory_diff.group_difference('content')
-    end
-  rescue Preservation::Client::Error => e
+  def content_diff(subset)
+    new_inventory = Stanford::ContentInventory.new.inventory_from_cm(content_metadata, druid, subset)
+    inventory_diff = Moab::FileInventoryDifference.new.compare(base_inventory(subset), new_inventory)
```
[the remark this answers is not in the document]
peetucket: oh yes, good catch... i think you are correct...adding in

spec/services/shelving_service_spec.rb
peetucket: Yes, additional testing from moab-versioning would be good.  I'll have a look.

peetucket:
Added some specs from https://github.com/sul-dlss/moab-versioning/blob/main/spec/unit_tests/stanford/storage_services_spec.rb#L58

Had to make some adjustments to (1) the how the two sets of contentMetadata being diffed are passed in (basically just did some different mocking to match the class in DSA and (2) slightly adjust the expectated XML being returned, because we actually getting the embedded `fileGroupDifference` back and not the full `fileInventoryDifference` node back in this case.  The test being cribbed from in moab-versioning is testing the response before it gets some additional parsing applied to it in preservation-client here: and here: https://github.com/sul-dlss/preservation-client/blob/main/lib/preservation/client/objects.rb#L28 and https://github.com/sul-dlss/preservation-client/blob/main/lib/preservation/client/objects.rb#L39 (both of these bits of code and now pulled into the `content_diff` method being tested in DSA here.

spec/services/shelving_service_spec.rb
```diff
@@ -118,4 +138,261 @@
...
+            <subset change="deleted" count="2">
+              <file change="deleted" basisPath="intro-1.jpg" otherPath="">
+                <fileSignature size="41981" md5="915c0305bf50c55143f1506295dc122c" sha1="60448956fbe069979fce6a6e55dba4ce1f915178" sha256=""/>
+              </file>
+              <file change="deleted" basisPath="intro-2.jpg" otherPath="">
+                <fileSignature size="39850" md5="77f1a4efdcea6a476505df9b9fba82a7" sha1="a49ae3f3771d99ceea13ec825c9c2b73fc1a9915" sha256=""/>
+              </file>
+            </subset>
+            <subset change="added" count="0"/>
+        </fileInventoryDifference>
```
jmartin-sul, line 189:
```suggestion
          </fileGroupDifference>
```

doesn't seem to make a difference in the test result (helpful nokogiri xml normalization making up for it?).  but i think this closing tag is slightly off from what it should be?

jmartin-sul (approved):
this looks great, thanks for the additional testing, and sorry for the slow turnaround!

i have one nitpicky suggestion, but this looks good to me overall.

random (_very_) slight mis-giving: i see an additional test here for the diff `with preserve subset`, which i don't see in moab-versioning.  i added that test to moab-versioning, and it fails for me locally, but looking at the diff, i'm having trouble seeing much in the way of salient differences, and i'm having a hard time seeing the underlying cause for the weird little differences that are jumping out at me.  would you mind pairing on that for a bit just to give me a sanity check to make sure i'm not doing that totally incorrectly on the moab-versioning side?

other stray thoughts:
1. seems like it'd be good to rebase this and integration test it before merging?  and maybe hand over to @andrewjbtw to do a quick manual test with something that's somewhat complex and/or large, just for additional peace of mind?
2. has all of the targeted file update/partial staging work been merged?  if not, maybe we should wait on merging and deploying this till that's all settled, just to keep those unentangled, in case either has bugs that need troubleshooting?  also want to be mindful of creating painful rebase work for you, so i also don't want to continue holding this up any longer than need be.
3. i wouldn't hold this PR up over it, but i think it'd be nice to add comments both to the `content_diff` code in this PR and to `compare_cm_to_version` in moab-versioning, indicating that the former is a re-constitution of the latter, and that there's some opportunity for refactoring in moab-versioning to reduce code duplication.
4. might be nice to add your new test to moab-versioning (unless it's there already and i'm missing it).

spec/services/shelving_service_spec.rb
peetucket: Thanks - fixed this.

peetucket:
I rebased from main and force pushed.  I also added a code comment here and a similar code comment in moab-versioning: https://github.com/sul-dlss/moab-versioning/pull/300

Running integration tests is a good idea before merging.  I can coordinate this.

jmartin-sul:
> I rebased from main and force pushed. I also added a code comment here and a similar code comment in moab-versioning: [sul-dlss/moab-versioning#300](https://github.com/sul-dlss/moab-versioning/pull/300)
> 
> Running integration tests is a good idea before merging. I can coordinate this.

thanks on all of those things!

peetucket:
All integration tests passed.

PR #4499, opened 2023-04-18
Add a report for descriptive metadata

edsu:

Generate a report of the shape of JSON in descriptive metadata. The shape is represented as a set of JSON paths which have either a string or number value, and the number of times they appear in the data:

```csv
path,count
.purl,4452
.title[].value,4321
.form[].type,2163
.form[].value,1810
.form[].source.value,1358
.note[].value,1319
.event[].date[].type,825
.event[].date[].encoding.code,629
.event[].date[].qualifier,2
.event[].date[].structuredValue[].type,18
.event[].date[].structuredValue[].value,18
```

The hope is that it might help in evaluating how we might want to further normalize and constrain descriptive metadata so that it is easier to edit.

Using sdr-infra.stanford.edu against QA (running now on Prod).

### DIFF
--- a/app/reports/descriptive_shape.rb
+++ b/app/reports/descriptive_shape.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+# Generate a report on the "shape" of Cocina descriptive metadata:
+#
+# bin/rails r -e production "DescriptiveShape.report"
+#
+# This will output a CSV file of JSON paths that contain a string or number value:
+#
+# path,count
+# .purl,4452
+# .title[].value,4321
+# .form[].type,2163
+# .form[].value,1810
+# .form[].source.value,1358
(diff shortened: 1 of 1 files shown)


app/reports/descriptive_shape.rb
```diff
@@ -0,0 +1,92 @@
...
+  def trace_hash(obj, path)
+    obj.each do |key, value|
+      trace(value, "#{path}.#{key}")
+    end
+  end
+
+  def save
+    File.write('descriptive-shape.json') do |fh|
+      fh.write(JSON.pretty_generate(@shape))
+    end
```
mjgiarlo, line 83: I *think* you can do this without the block?

```suggestion
    File.write('descriptive-shape.json', JSON.pretty_generate(@shape))
```

PR #4510, opened 2023-05-16
add report for all versions

### LINKED ISSUE #4509: report: all version descriptions for all versions of all druids
Not sure if this is possible or possible but very difficult, but I would like a report of all version descriptions for all versions of all druids. I think these four fields in a CSV or TSV would be sufficient:

- druid
- version number (as a whole number)
- tag (or whatever the pseudo-semantic version number is called, e.g. 2.1.0
- version description

For various threads of SDR planning, especially around versioning, I would like to do some analysis on existing data. I'm especially interested in how many versions are:

- clearly system-described (usually via bulk action)
- are metadata bulk updates, which usually get a system-made description
- were obvious fixes to errors ("fix version mismatch")
- show some indication of user attempts to signify that some versions are more significant than others (i.e. "release 2.0")

peetucket:

Fixes #4509 - Add a report for exporting all version info metadata to CSV file

By running this branch on sdr-infra

### DIFF
--- a/app/reports/version_list.rb
+++ b/app/reports/version_list.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'csv'
-
 # Exports metadata about all versions  in SDR
 #  https://github.com/sul-dlss/dor-services-app/issues/4509
 # Invoke via:
@@ -13,7 +11,7 @@ class VersionList
     CSV.open(output_file, 'w') do |csv|
       csv << %w[druid version tag description]
 
-      num_versions = ObjectVersion.all.count
+      num_versions = ObjectVersion.count
       puts "Found #{num_versions} object versions"
       i = 0
       ObjectVersion.find_each do |object_version|
--- a/app/reports/version_list.rb
+++ b/app/reports/version_list.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+# Exports metadata about all versions  in SDR
+#  https://github.com/sul-dlss/dor-services-app/issues/4509
+# Invoke via:
+# bin/rails r -e production "VersionList.report"
+class VersionList
+  def self.report
+    output_file = 'tmp/version_list_report.csv'
+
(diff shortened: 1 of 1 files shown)


app/reports/version_list.rb
```diff
@@ -0,0 +1,27 @@
...
+    output_file = 'tmp/version_list_report.csv'
+
+    CSV.open(output_file, 'w') do |csv|
+      csv << %w[druid version tag description]
+
+      num_versions = ObjectVersion.all.count
+      puts "Found #{num_versions} object versions"
+      i = 0
+      ObjectVersion.find_each do |object_version|
+        i += 1
```
peetucket, line 20: the `find_each` ensures we do this in batches (1000 by default) so memory consumption is reasonable

app/reports/version_list.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'csv'
```
mjgiarlo, line 3: Did you find you needed this or is it copypasta? I'd be surprised if we need this manual require. Rails should do this automatically.

app/reports/version_list.rb
```diff
@@ -0,0 +1,27 @@
...
+# Invoke via:
+# bin/rails r -e production "VersionList.report"
+class VersionList
+  def self.report
+    output_file = 'tmp/version_list_report.csv'
+
+    CSV.open(output_file, 'w') do |csv|
+      csv << %w[druid version tag description]
+
+      num_versions = ObjectVersion.all.count
```
mjgiarlo, line 16: I don't think the `.all` is needed here.

app/reports/version_list.rb
peetucket: Removed and still works.

app/reports/version_list.rb
peetucket: Removed

mjgiarlo (approved):
Thanks for considering my review!

PR #4512, opened 2023-05-19
do not update publish step in workflow service when pubishing items in collection

### LINKED ISSUE #4511: release-publish step should not be called on items that have not been released
Under some circumstances, the release-publish job from the releaseWF can be called on druids that have never been released. Since druids that have never been released have never gone through a releaseWF, this generates 404 errors: https://app.honeybadger.io/projects/50568/faults/58644364

I think the cause of this problem traces back to the logic in place for collection publishing. When a collection is republished, we automatically republish all the items in that collection. This ensures that the public XML/Cocina for the items gets updated when a collection name changes.

Usually that works this way:

- collection is published (either through accessionWF or the "publish" button in Argo)
- items within the collection are published

But when a collection is released, the releaseWF calls its own publish job ("release-publish")

- "release-publish" probably follows the same logic as accessionWF "publish" and triggers item publishes
- but it probably triggers "release-publish" on the tiems
- however, a collection can contain nonreleased items, so there is no 'release-publish" step for these items
- when a collection is released to "collection", it will automatically release the items, so there's no error because a releaseWF will be created for those items if there isn't one already
- but when a collection is released to "self", items that have not been released should not be released. But they should be republished. This leads to the an error where release-publish is called without releaseWF existing.

One possible approach: if an item in a collection should be republished but not released, use accessionWF publish to republish.

peetucket:

Fixes #4511: when re-publishing all items as part of a collection publish (which can happen during collection release), we do *not* need to update a workflow step when done.  The fact that a publish occurs is already noted in the Event history, and attempting to update a workflow step can cause errors such as this: https://app.honeybadger.io/projects/50568/faults/58644364  This occurs when the publish occurs for items that were never released, as part of a collection release (which triggers all member items to be published as well).  Currently we try to notify the workflow service that the `release-publish` step is completed, which is both unnecessary (as the publish step was not triggered by a workflow in that object) and will cause problems if that particular item has never been through releaseWF before.

This code simply short-circuits this workflow update when an item publish occurs as part as a collection publish (as opposed to when it is triggered from a workflow in that item itself).

Updated specs, and on -stage too.

### DIFF
--- a/app/jobs/publish_job.rb
+++ b/app/jobs/publish_job.rb
@@ -8,7 +8,7 @@ class PublishJob < ApplicationJob
   # @param [String] druid the identifier of the item to be published
   # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
   # @param [String] workflow Which workflow should this be reported to?
-  def perform(druid:, background_job_result:, workflow:)
+  def perform(druid:, background_job_result:, workflow:, log_success: true)
     background_job_result.processing!
     workflow_process = workflow == 'releaseWF' ? 'release-publish' : 'publish'
     begin
@@ -17,6 +17,8 @@ class PublishJob < ApplicationJob
       Publish::MetadataTransferService.publish(cocina_object, workflow:)
       EventFactory.create(druid:, event_type: 'publishing_complete', data: { background_job_result_id: background_job_result.id })
     end
+
+    return unless log_success
 
     LogSuccessJob.perform_later(druid:,
                                 background_job_result:,
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -53,7 +53,7 @@ module Publish
       Array.wrap(
         MemberService.for(cocina_object.externalIdentifier, exclude_opened: true, only_published: true)
       ).each do |member|
-        PublishJob.set(queue: :publish_low).perform_later(druid: member['id'], background_job_result: BackgroundJobResult.create, workflow:)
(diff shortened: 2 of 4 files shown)


app/jobs/publish_job.rb
```diff
@@ -18,6 +18,8 @@ def perform(druid:, background_job_result:, workflow:)
       EventFactory.create(druid:, event_type: 'publishing_complete', data: { background_job_result_id: background_job_result.id })
     end
 
+    return unless log_success
+
     LogSuccessJob.perform_later(druid:,
                                 background_job_result:,
                                 workflow:,
```
peetucket, line 25: This job just updates the workflow service step (it's a bit of a misnomer, should really be `LogCompleteWorkflowStep` or something like that)

app/services/publish/metadata_transfer_service.rb
```diff
@@ -53,7 +53,7 @@ def republish_members!
       Array.wrap(
         MemberService.for(cocina_object.externalIdentifier, exclude_opened: true, only_published: true)
       ).each do |member|
-        PublishJob.set(queue: :publish_low).perform_later(druid: member['id'], background_job_result: BackgroundJobResult.create, workflow:)
+        PublishJob.set(queue: :publish_low).perform_later(druid: member['id'], background_job_result: BackgroundJobResult.create, workflow:, log_success: false)
       end
     end
```
peetucket, line 58: This `PublishJob` is now spawned for all items in a collection when collection is published.  This was a recent change ... it used to just iterate over all items and publish them immediately, which took a long time for big collections, and caused the collection publish job to time out.

justinlittman (approved):

PR #4516, opened 2023-05-24
[HOLD] only add last releaseTag to a collection object in public XML

### LINKED ISSUE #4515: Only publish the latest "release tag" to the public XML/Cocina
**Describe the bug**

At some point in the last year, SDR has started publishing more than one SearchWorks "release" tag to the public XML. This is causing some releases to Searchworks to fail because the Access systems expect to see only one tag.

Example: https://purl.stanford.edu/xh235dd9059.xml

In this object's public XML, there are multiple release tags:
```xml
<releaseData>
    <release to="Searchworks">false</release>
    <release to="Searchworks">true</release>
    <release to="Searchworks">true</release>
    <release to="Searchworks">true</release>
    <release to="Searchworks">true</release>
    <release to="Searchworks">true</release>
    <release to="Searchworks">true</release>
  </releaseData>
```
This object should appear in SW but it does not because one of those release tags is set to "false". In the [object history in Argo](https://argo.stanford.edu/view/xh235dd9059), you can see that the object has been sent through the release process multiple times. 

Screenshot 2023-05-23 at 9 41 21 AM

The Cocina also includes multiple release tags. The earliest release tag is the tag that was set to false, indicating that many years ago the object was withdrawn from SW. However, all of the later tags are set to "true", showing that the object should have been released.

The problem seems to be that SDR should be publishing only the most recent release tag, not the full set of historical release tags. According to Chris Beer, as long as purl-fetcher receives a "false" release tag somewhere in the list, whether or not it's the last value, it will interpret that as an instruction to keep the object out of SW.

**Expected behavior**

When an item or collection is released to or withdrawn from SearchWorks or EarthWorks, a `<release>` tag is inserted into its public XML:

The tag should look like this when an item is released:
```xml
<releaseData>
    <release to="Searchworks">true</release>
</releaseData>
```
When an item has been un-released (withdrawn) from SW, the tag looks like this:
```xml
<releaseData>
    <release to="Searchworks">true</release>
</releaseData>
```
There should never be more than one release tag in the public XML for SearchWorks and one release tag for EarthWorks. So an item or collection public XML can have:

- no release tag
- one SW release tag
- one EW release tag
- both one SW release tag and one EW release tag

**Additional context**

I happen to have a copy of the public XML for this object from April 2022, just before the Fedora migration. In that copy, there is only a single release tag for SW, which confirms that the publishing behavior changed at some point in the last year.

peetucket:

[HOLD] to verify collection publishing works as expected (test in stage) and isn't missing additional info neded

Fixes #4515 - collection publishing should use same logic for building administrative public XML as we do for items

Question: previously we were just returning the entire collection cocina, which is different than we do for items.   Unclear if this approach is removing public XML data for collections that should otherwise be published (though all current tests pass still).

Added unit tests

### DIFF
--- a/app/services/publish/public_cocina_service.rb
+++ b/app/services/publish/public_cocina_service.rb
@@ -15,12 +15,13 @@ module Publish
     # remove any file_set that doesn't have at least one published file
     # remove partOfProject (similar to how we remove tags from identityMetadata)
     def build
-      return cocina unless cocina.dro? || cocina.collection?
-
-      return cocina.new(administrative: build_administrative) if cocina.collection?
-
-      cocina.new(structural: build_structural,
-                 administrative: build_administrative)
+      if cocina.dro?
+        cocina.new(structural: build_structural, administrative: build_administrative)
+      elsif cocina.collection?
+        cocina.new(administrative: build_administrative)
+      else
+        cocina
+      end
     end
 
     private
--- a/app/services/publish/public_cocina_service.rb
+++ b/app/services/publish/public_cocina_service.rb
@@ -20,7 +20,7 @@ module Publish
       elsif cocina.collection?
         cocina.new(administrative: build_administrative)
       else
-        cocina
+        raise "unexpected call to PublicCocinaService.build for #{cocina.externalIdentifier}"
       end
     end
 
--- a/app/services/publish/public_cocina_service.rb
+++ b/app/services/publish/public_cocina_service.rb
@@ -15,7 +15,9 @@ module Publish
     # remove any file_set that doesn't have at least one published file
     # remove partOfProject (similar to how we remove tags from identityMetadata)
     def build
-      return cocina unless cocina.dro?
+      return cocina unless cocina.dro? || cocina.collection?
+
+      return cocina.new(administrative: build_administrative) if cocina.collection?
 
       cocina.new(structural: build_structural,
                  administrative: build_administrative)
--- a/spec/services/publish/public_xml_service_spec.rb
+++ b/spec/services/publish/public_xml_service_spec.rb
@@ -102,7 +102,7 @@ RSpec.describe Publish::PublicXmlService do
       end
     end
 
-    context 'with an problematic location code' do
+    context 'with a problematic location code' do
       let(:cocina_object) do
         build(:dro, id: 'druid:bc123df4567').new(
           access: {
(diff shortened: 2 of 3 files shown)


spec/services/publish/public_xml_service_spec.rb
```diff
@@ -302,6 +302,30 @@
...
+        end
+
+        it 'includes only the latest releaseData element from release tags for each target' do
+          releases = ng_xml.xpath('/publicObject/releaseData/release')
+          expect(releases.size).to eq 2
+          expect(releases.map(&:inner_text)).to eq %w[true true]
+          expect(releases.pluck('to')).to eq %w[Searchworks Some_special_place]
+        end
+      end
     end
```
peetucket, line 329: new test to verify item release tags work as expected in public XML (no change needed ... this currently works)

spec/services/publish/public_xml_service_spec.rb
```diff
@@ -326,6 +350,52 @@
...
```
peetucket, line 399: new test to verify colllection release tags work as expected in public XML (this last test I added failed until I made the code change above)

app/services/publish/public_cocina_service.rb
```diff
@@ -15,7 +15,9 @@ def initialize(cocina)
     # remove any file_set that doesn't have at least one published file
     # remove partOfProject (similar to how we remove tags from identityMetadata)
     def build
-      return cocina unless cocina.dro?
+      return cocina unless cocina.dro? || cocina.collection?
+
+      return cocina.new(administrative: build_administrative) if cocina.collection?
```
peetucket, line 21: this is the change...instead of returning all cocina for a collection, we will use the `build_administrative` method, which correctly builds the release tags based on dates, just like it does for items)

The question is: does this mean there is cocina published for collections now that will not get published with this change, and if so, we need to add it here.

app/services/publish/public_cocina_service.rb
```diff
@@ -15,7 +15,9 @@ def initialize(cocina)
...
 
       cocina.new(structural: build_structural,
                  administrative: build_administrative)
```
mjgiarlo, line 23: Given we have multiple checks in a couple different spots (lines 18 and 20), I wonder if we want to consider a different structure now, *e.g.*:

```suggestion
      if cocina_object.dro?
        cocina.new(structural: build_structural, administrative: build_administrative)
      elsif cocina_object.collection?
        cocina.new(administrative: build_administrative)
      else
        cocina
      end
```

app/services/publish/public_cocina_service.rb
mjgiarlo: > The question is: does this mean there is cocina published for collections now that will not get published with this change, and if so, we need to add it here.

Is this something we could ask Andrew to help us test in QA or stage?

mjgiarlo:
> Mike, I took your suggested change (had to fix a typo) but then saw there was an untested case (something other than item or collection) and realized we really should never be publishing anything other than an item or collection anyway (e.g. APO). So I decided to raise an error in this case (which again, should never happen if the rest of the system is coded correctly), and then test for that.

Sounds good to me, Peter. I'll approve and you can ship it once you feel it's been sufficiently put through its paces.

mjgiarlo (approved):

PR #4524, opened 2023-06-22
increase sidekiq latency threshold and put in config

peetucket:

Sidekiq latency can grow large if a lot of content is accessioned all at once (as the queue will grow faster than it's being worked on for a while).  This is OK and we don't need to be alerted, so increase the failure level much higher (to 15 minutes).  Also put in config so we can adjust easily and per environment if needed.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -112,6 +112,11 @@ datacite:
   prefix: '10.80343'
   host: fake.datacite.example.com
 
+sidekiq:
+  latency_threshold: 900 # number of seconds above which okcomputer check for sidekiq latency fails
+                         # latency is expected to be temporarily big if lots of content is accessioned all at once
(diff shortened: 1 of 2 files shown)


config/settings.yml
```diff
@@ -112,6 +112,11 @@ datacite:
   prefix: '10.80343'
   host: fake.datacite.example.com
 
+sidekiq:
+  latency_threshold: 900 # number of seconds above which okcomputer check for sidekiq latency fails
+                         # latency is expected to be temporarily big if lots of content is accessioned all at once
+                         # see https://github.com/sidekiq/sidekiq/wiki/Monitoring#monitoring-queue-latency
```
jmartin-sul, line 118: i ❤️ the sidekiq docs.  thorough but concise.

jmartin-sul (approved):

PR #4534, opened 2023-07-16
Adds create/update orcid work job.

### LINKED ISSUE #4532: Create create/update orcid work job
This can be patterned after https://github.com/sul-dlss/dor-services-app/blob/main/app/jobs/update_doi_metadata_job.rb.

It should:
1. Check if any orcids are found in the cocina descriptive. (While there are infinite places this can be, for initial purposes it is good enough to look in the places that H2 might place them. See the H2 to cocina mapping in H2 codebase.) Return if no orcids are found.
2. For each orcid, use of Mais Orcid client to check (1) if a Stanford user with write permissions and (2) get the access token. Return if no Stanford users with write permission.
3. Map the Cocina item to Orcid work metadata and generate a checksum. (See #4531 for creating this mapper.)
4. Check the orcid model to see if there is an existing put code for the cocina item and the orcid id with a different checksum. (Creating the active record model for this is in the scope of this ticket.)
5. If there is an existing put code with a different checksum, use the Orcid client to perform an update of orcid work using the put code. If there is not an existing put code, perform a create of orcid work and record the resulting put code and checksum.

justinlittman:
closes #4532

Unit, QA

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,9 +11,11 @@ gem 'datacite', '~> 0.3.0'
 gem 'dor-workflow-client', '~> 5.0'
 gem 'druid-tools', '~> 2.2'
 gem 'folio_client', '~> 0.8'
+gem 'mais_orcid_client'
 gem 'marc'
 gem 'moab-versioning', '~> 6.0', require: 'moab/stanford'
 gem 'preservation-client', '~> 6.0'
+gem 'sul_orcid_client', '~> 0.3'
 # Pinning stanford-mods since >=3 breaks dor-services.
 gem 'stanford-mods', '~> 2.6'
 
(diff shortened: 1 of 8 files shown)


ndushay (approved):
this looks ok to me but I didn't look at it super closely.

Curious about storing orcid stuff (for ONE of our apps) in a table in dsa - wondering if there are any alternatives worth considering.

justinlittman:
@ndushay Good wondering. There was a discussion about whether this feature should be implemented specific to H2 or more generally, so that it could be used with items created by other apps. The decision made by the stakeholders was for the more general solution, as it was imagined that it might be used with other items in the future, e.g., ETDs.

PR #4551, opened 2023-08-12
adding comments pertaining to folio cutover

ndushay:

to make code clearer for post-symphony hires

comments only

### DIFF
--- a/app/services/create_object_service.rb
+++ b/app/services/create_object_service.rb
@@ -3,7 +3,7 @@
 # This handles all of the business logic around registering an object.
 # This includes:
 #   Minting a druid identifier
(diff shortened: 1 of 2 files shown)


app/services/create_object_service.rb
```diff
@@ -3,7 +3,7 @@
 # This handles all of the business logic around registering an object.
 # This includes:
 #   Minting a druid identifier
-#   Importing metadata from Symphony if there is a catkey
+#   Importing metadata from ILS catalog if there is a catalogRecordId
```
justinlittman, line 6: Same, suggest "Folio" instead of "ILS catalog".

justinlittman (approved):

Issue #4549, opened 2023-08-11
ensure soup to nuts Goobi processing will work with Folio


ndushay:
Goobi processing:

There are at least TWO parts to this ticket: 
- one for infrastructure team developers
- one for DPG to ensure Goobi will work with record identifiers with the folio prefixes of "a" or "in" or whatever.

In the goobiWF, the goobi-notify step triggers dor-services-app (via objects_controller) to call the GoobiService class, which produces xml like:

```xml
        <stanfordCreationRequest>
            <objectId>#{cocina_obj.externalIdentifier}</objectId>
            <objectType>item</objectType>
            <sourceID>#{source_id&.encode(xml: :text)}</sourceID>
            <title>#{title_or_label.encode(xml: :text)}</title>
<snip/>
            <catkey>#{catkey}</catkey>
<snip/>
        </stanfordCreationRequest>
``` 

@peetucket, @calavano:  I assume changing the XML element name will require Goobi company to do something to their software, so we want to keep the element name but use a folio_instance_hrid as the value.

## Infrastructure Developers:

- [x] GoobiService catkey value is becomes folio_instance_hrid  when we switch to Folio. - addressed by #4555
- [x] We need to know what to do when there is no ILS (talk to Tony C, to Peter, to ???) - should not be a problem I don't think, we don't ever talk to the ILS directly, only look into cocina

## DPG Folks

We need to be certain that:
- [ ] downstream use of "catkey" will keep working (Tony C is on this)
    - [ ] when DPG does searches in Goobi
    - [x] if Goobi happens to use it anywhere when it tries to kick off assemblyWF. (@calavano says: "I'm fairly certain that it only uses the druid at that point.")

ndushay:
From @calavano on PR #4555:

> @peetucket I can confirm that our only use of it in Goobi is informational. We might want to change the label in Goobi at some point, but I feel that it would be fine to move forward.

closing this ticket.

PR #4555, opened 2023-08-15
send either catkey or folio_instance_hrid to goobi depending on what is active

peetucket:

Part of #4549 

When we switch to Folio, we will want to send folio_instance_hrids to the Goobi system instead of catkeys.  Thus the XML we construct needs to use our ILS feature flag to decide what to send.

Note that the name of the XML element we send to Goobi will continue to be "catkey", as changing this will require a code change on the Goobi end (which is out of our control and which the contractor would have to complete).

Two notes:
- The call that Goobi makes at the end of the process is not affected by the ILS change, as it is simply a call to start accessioning to the `/v1/objects/druid:DRUID_HERE/accession` endpoint of DSA.  There is no reference to catkey at all at this point. See https://github.com/sul-dlss/DevOpsDocs/blob/master/projects/goobi/operations-concerns.md for more details.
- We should verify on QA or stage that we are allowed to send folio formatted IDs to Goobi without an issue (e.g. "a1234" or "in1234").  We need to verify their code accepts non numeric characters in the "catkey" field in the XML we send.

- Added a new unit test.
- Deployed to QA (which has folio turned on) and verified that I was able to register a new object with a folio instance HRID ("a12345"), and see this sent over to Goobi.  It is still labeled as "catkey" in the Goobi UI (to be expected), but the value was sent correctly, shows correctly, and didn't cause any errors (see screenshot below).  Object: https://argo-qa.stanford.edu/view/druid:xw490jn2818 and you can find it in Goobi stage (which is what DSA/Argo QA is connected to) by searching for the druid. https://goobi-stage-a.stanford.edu/goobi/uii/process_all.xhtml?dswid=5693

Screen Shot 2023-08-16 at 3 32 03 PM

### DIFF
--- a/app/services/goobi_service.rb
+++ b/app/services/goobi_service.rb
@@ -71,7 +71,7 @@ class GoobiService
             <title>#{title_or_label.encode(xml: :text)}</title>
             <contentType>#{content_type}</contentType>
             <project>#{project_name.encode(xml: :text)}</project>
-            <catkey>#{catkey}</catkey>
+            <catkey>#{catalog_id}</catkey>
             <barcode>#{barcode}</barcode>
             <collectionId>#{collection_id}</collectionId>
             <collectionName>#{collection_name.encode(xml: :text)}</collectionName>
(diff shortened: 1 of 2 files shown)


ndushay (approved):
LGTM!

approving modulo it working on stage

peetucket:
@calavano I was able to register an object with folio instance HRID on our QA environment and see it sent over to Goobi.  It shows in the metadata as "catkey" on the Goobi side.  I doubt if it it's actually used anywhere other than informational purposes, so I believe this is really all that is needed?

calavano:
@peetucket I can confirm that our only use of it in Goobi is informational. We might want to change the label in Goobi at some point, but I feel that it would be fine to move forward.

PR #4556, opened 2023-08-15
Make update_marc_record a job

### LINKED ISSUE #4454: 856/Folio: when releasing more than one druid linked to the same Folio HRID, only one job actually succeeds

lwrubel:
Resolves #4454 to update FOLIO marc record 856 fields in the releaseWF, without concurrent updates. Will require deploying along with puppet and shared_configs PRs. 

Do not merge until coordinated with related shared_configs, common_accessioning, puppet PR reviews.

Tested in QA with 100 druids released to same FOLIO record.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -131,7 +131,7 @@ class ObjectsController < ApplicationController
   # Called by common-accessioning
   def update_marc_record
     result = BackgroundJobResult.create
-    UpdateMarcJob.perform_later(Cocina::Models.without_metadata(@cocina_object).to_json, background_job_result: result)
+    UpdateMarcJob.perform_later(druid: params[:id], background_job_result: result)
     head :accepted
   end
 
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -128,9 +128,11 @@ class ObjectsController < ApplicationController
     head :accepted, location: result
   end
 
+  # Called by common-accessioning
   def update_marc_record
-    Catalog::UpdateMarc856RecordService.update(@cocina_object, thumbnail_service: ThumbnailService.new(@cocina_object))
-    head :created
+    result = BackgroundJobResult.create
+    UpdateMarcJob.perform_later(Cocina::Models.without_metadata(@cocina_object).to_json, background_job_result: result)
+    head :accepted
   end
 
   # Called by the robots.
--- a/app/jobs/update_marc_job.rb
+++ b/app/jobs/update_marc_job.rb
@@ -6,9 +6,8 @@ class UpdateMarcJob < ApplicationJob
   queue_as :update_marc
 
   # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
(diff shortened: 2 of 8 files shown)


ndushay (commented):
we need the "if Settings.folio" (or whatever) block somewhere here, I think, as we need DSA to be set to symphony through the week of no ILS available, and then change the setting in shared configs to point all to Folio.

In terms of code to make this work, it looks great to me.  Thank you for all the hard work!

lwrubel:
@ndushay the feature flag for Folio is in `Catalog::UpdateMarc856RecordService`, which I didn't need to touch for this PR.

https://github.com/sul-dlss/dor-services-app/blob/cf5f7ca21bcb9a26008cfa6feb290496c783aa6b/app/services/catalog/update_marc_856_record_service.rb#L18

The LogFailure error message that references Folio is the only reference to Folio not behind the feature flag, but we're a couple of days from that not mattering, so I didn't include logic there.

ndushay (approved):

app/controllers/objects_controller.rb
```diff
@@ -128,9 +128,11 @@ def unpublish
     head :accepted, location: result
   end
 
+  # Called by common-accessioning
   def update_marc_record
-    Catalog::UpdateMarc856RecordService.update(@cocina_object, thumbnail_service: ThumbnailService.new(@cocina_object))
-    head :created
+    result = BackgroundJobResult.create
+    UpdateMarcJob.perform_later(Cocina::Models.without_metadata(@cocina_object).to_json, background_job_result: result)
```
justinlittman, line 134: I'm wondering if for large items, including the full JSON in the job payload is going to be too large.

lwrubel: @justinlittman I changed the job to take the druid as an argument instead and look up the cocina from within the UpdateMarcJob. It seems like DSC will be sending the druid as a param? I wasn't sure how to confirm that. Let me know if you think this is sufficient, and I'll deploy to QA again to test.

justinlittman (approved):

PR #4557, opened 2023-08-16
Adds GraphQL endpoint.

justinlittman:
Allow for speedier retrieval of large cocina objects.


app/graphql/types/mutation_type.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# This was generated by graphql gem.
+module Types
+  class MutationType < Types::BaseObject
+    # TODO: remove me
```
ndushay, line 6: cruft?   Refers to this whole file, maybe?

app/graphql/types/node_type.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# This was generated by graphql gem.
+module Types
+  module NodeType
+    include Types::BaseInterface
+    # Add the `id` field
```
ndushay, line 7: is this a todo, or what the next line does?  I'm unclear.

ndushay (commented):
A couple of inline comments ...

Also wondering if, given all the type classes added and so on, do the specs adequately exercise the added code?

app/graphql/types/mutation_type.rb
justinlittman: These were examples created by the generator. I added a comment to clarify and commented out the examples.

app/graphql/types/node_type.rb
justinlittman: It was created by the generator, so I don't know either. Hopefully, if a future dev needs to touch this code, it will be clearer.

justinlittman:
I understand your concern about testing the code added by the generator. We test the code that we use; the rest is scaffolding for features that we aren't currently using. If in the future, we were to use those features then we would certainly need additional tests. I'm also reluctant to remove files created by the generator even if they're not necessary as it might make life difficult for our future selves. The approach I took was to clearly indicate which files were created by the generator.

app/graphql/types/mutation_type.rb
ndushay: examples not looking commented out here ...

ndushay (approved):

justinlittman:
@ndushay Apparently I have to commit the changes, not just type them. Fixed.

PR #4558, opened 2023-08-17
Avoid loading cocina object to check existence.

justinlittman:
Performance improvement, in particular, from Argo.

Unit


app/services/cocina_object_store.rb
```diff
@@ -39,10 +39,18 @@ def self.find_by_source_id(source_id)
   # Determine if an object exists in the datastore.
   # @param [String] druid
   # @return [boolean] true if object exists
+  # @raise [CocinaObjectNotFoundError] raised when the requested Cocina object is not found.
   def self.exists?(druid)
     new.exists?(druid)
   end
 
+  # Checks if an object exists in the datastore.
```
ndushay, line 47: can it be clearer how this differs from the previous method?  I missed it.  "does not raise" or whatever where appropriate.  Same for below in the instance method.

ndushay (approved):
Nice work!  One request for clearer comments distinguishing `exists?` from `exists!` for the bleary eyed such as myself.

app/services/cocina_object_store.rb
justinlittman: Sure. Done!

app/services/cocina_object_store.rb
```diff
@@ -36,13 +36,21 @@ def self.find_by_source_id(source_id)
     new.find_by_source_id(source_id)
   end
 
-  # Determine if an object exists in the datastore.
+  # Determine if an object exists in the datastore, returning a boolean.
   # @param [String] druid
   # @return [boolean] true if object exists
+  # @raise [CocinaObjectNotFoundError] raised when the requested Cocina object is not found.
```
mjgiarlo, line 42: Should this go on `.exists!` instead of `.exists?` ?

app/services/cocina_object_store.rb
```diff
@@ -36,13 +36,21 @@ def self.find_by_source_id(source_id)
...
   def self.exists?(druid)
     new.exists?(druid)
   end
 
+  # Checks if an object exists in the datastore, raising if it does not exist.
```
ndushay, line 47: does this raise, or the other one?

app/services/cocina_object_store.rb
ndushay, line 42: i think this is on the wrong method?

justinlittman: Yes. Fixed.

justinlittman: It is. Fixed.

mjgiarlo (commented):
Have a question that is probably easy to answer.

app/services/cocina_object_store.rb
```diff
@@ -36,13 +36,21 @@ def self.find_by_source_id(source_id)
...
+  # @param [String] druid
+  # @return [boolean] true if object exists
+  # @raise [CocinaObjectNotFoundError] raised when the requested Cocina object is not found.
+  def self.exists!(druid)
+    new.exists!(druid)
+  end
```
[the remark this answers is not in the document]
justinlittman: That's it; one raises, the other doesn't.

The reason for the new raising existence check is for use in the places in which code was using `.find` as a raising existence check. Using `.find` is expensive because it actually retrieves the db object and instantiates the Cocina object, even though the Cocina object is not actually used. (For a large item, that is ~3.5 seconds.)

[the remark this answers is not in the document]
mjgiarlo: @justinlittman So `.exists!` will raise and the exception will bubble up before `.find` can be called, avoiding the DB load and instantiation? Sounds good.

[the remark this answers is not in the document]
justinlittman: No, `.exists!` is a replacement for `.find` in these cases. The actual cocina object isn't needed.

PR #4560, opened 2023-08-18
Add committee configuration to remove warnings.

justinlittman:
Avoid future committee configuration problems.

Before:
```
bundle exec rspec spec/requests/authorization_spec.rb
[DEPRECATION] Committee: please set parameter_overwite_by_rails_rule = false because we'll change default value to "true" in next major version.
[DEPRECATION] Committee: please set parameter_overwite_by_rails_rule = false because we'll change default value to "true" in next major version.
```

After:
```
bundle exec rspec spec/requests/authorization_spec.rb
```

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -52,7 +52,8 @@ module DorServices
       error_class: JSONAPIError,
       accept_request_filter: accept_proc,
       parse_response_by_content_type: false,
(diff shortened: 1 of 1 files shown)


ndushay (approved):
bless you, my child.

PR #4564, opened 2023-08-21
No longer update marc 856 from ConstituentService.

### LINKED ISSUE #4548: No longer update marc 856 from ConstituentService
After migration to FOLIO, remove guard clause concerning Symphony catalogLinks and the call to `Catalog::UpdateMarc856Service`. Rely on releaseWF for updating FOLIO in the unusual case that there is a catalog link in a constituent.

https://github.com/sul-dlss/dor-services-app/blob/5a0f5671a6b59e1ce07256dfb7910627ea307a7a/app/services/constituent_service.rb#L76

For background, see: https://github.com/sul-dlss/dor-services-app/issues/4445

justinlittman:
closes #4548

See ticket.

Unit

### DIFF
--- a/app/services/constituent_service.rb
+++ b/app/services/constituent_service.rb
@@ -72,10 +72,6 @@ class ConstituentService
     constituent_druids.each do |constituent_druid|
       cocina_item = CocinaObjectStore.find(constituent_druid)
       Publish::MetadataTransferService.publish(cocina_item)
-
-      next unless cocina_item.identification&.catalogLinks&.any? { |link| link.catalog == 'symphony' }
-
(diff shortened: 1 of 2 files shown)


justinlittman:
This PR is to be held until after folio migration.

ndushay:
I'm not clear if this is something we should do when Folio is "ready", or if it's something we should do after we configure and turn on Folio and shake it for a bit.

Also, should this be a real PR but on Hold, so it gets reviews?

lwrubel:
This is part of extracting Symphony-related code from SDR after we're live on FOLIO. So it is not needed for go-live. Can be merged at some point after we're firmly on FOLIO.

lwrubel (approved):

PR #4567, opened 2023-08-23
Add rake tasks to support seeding.

justinlittman:
Provide an approach to allow seeding APOs / Admin Policies for QA/stage reset.

QA

### DIFF
--- a/app/validators/cocina/apo_existence_validator.rb
+++ b/app/validators/cocina/apo_existence_validator.rb
@@ -11,6 +11,9 @@ module Cocina
 
     # @return [Boolean] false if the APO is not in the repository
     def valid?
+      # Skip if ur admin policy. This helps with bootstrapping.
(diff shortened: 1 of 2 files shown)


mjgiarlo (commented):
😍

jcoyne:
The conventional place for this in a rails app is `db/seed.rb` Then when you call `bin/rake db:seed` it's loaded.

app/validators/cocina/apo_existence_validator.rb
```diff
@@ -23,6 +26,6 @@ def valid?
 
     private
 
-    attr_reader :apo_id
+    attr_reader :apo_id, :druid
```
jcoyne, line 29: Why is this change needed?

justinlittman: It's not. Remnants of the code path not taken. Removed.

mjgiarlo (approved):

PR #4571, opened 2023-08-24
replace symphony check with folio check, if feature flag enabled

### LINKED ISSUE #4570: okcomputer Symphony check:  change to Folio check?
Do we want an okcomputer Folio check? (ask in infra slack channel?)

If so, replace Symphony check with Folio check - create a PR, mark it HOLD and get it approved.

If not, remove Symphony check.

peetucket:

Fixes #4570 - replace symphony with folio check

For discussion with team, re: do we want external checks like this or are separate nagios alerts sufficient?  

See https://stanfordlib.slack.com/archives/C09M7P91R/p1692899929140329

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -20,10 +20,14 @@ class CustomAppVersionCheck < OkComputer::AppVersionCheck
   end
 end
 
-class SymphonyCheck < OkComputer::HttpCheck
-  def perform_request
-    Timeout.timeout(request_timeout) do
-      Catalog::SymphonyReader.client.get(url.to_s)
+class FolioCheck < OkComputer::Check
+  def check
+    if Settings.enabled_features.read_folio
+      Timeout.timeout(5) do
+        Catalog::FolioReader.to_marc(barcode: '12345')
+      end
+    else
+      mark_message 'folio disabled'
     end
   end
 end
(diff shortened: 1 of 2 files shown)


config/initializers/okcomputer.rb
```diff
@@ -20,10 +20,14 @@ def version_from_version_file
...
 end
 
-class SymphonyCheck < OkComputer::HttpCheck
-  def perform_request
-    Timeout.timeout(request_timeout) do
-      Catalog::SymphonyReader.client.get(url.to_s)
+class FolioCheck < OkComputer::Check
+  def check
+    if Settings.enabled_features.read_folio
+      Timeout.timeout(5) do
```
ndushay, line 26: if request_timeout is set somewhere, should we use it here?

ndushay (approved):

config/initializers/okcomputer.rb
peetucket: `request_timeout` in the symphony checker was the default 5 seconds from the `OkComputer::HttpCheck` class we were subclassing before (but not for this check, where we instead use the class method to run the check instead of just hitting a URL like we did for Folio)

we could configure that timeout somewhere, but I think it's just used in one spot (here)

justinlittman (approved):

PR #4578, opened 2023-08-30
[SPIKE] Decouple DSA from Solr

### LINKED ISSUE #4459: Decouple from Solr
This is an "SDR evolution" ticket, the intent of which is to reduce dependencies within DSA. (HT to @jcoyne for the idea!)

As far as we know, Solr is used within DSA for:

* Reports (could move to Argo which is more dependent on Solr);
* [Embargo release](https://github.com/sul-dlss/dor-services-app/blob/f51bbcea710b7612f251a3922c5164ec69ba39aa/app/services/embargo_release_service.rb)
* [Collection members](https://github.com/sul-dlss/dor-services-app/blob/f51bbcea710b7612f251a3922c5164ec69ba39aa/app/services/member_service.rb)
* [Virtual objects](https://github.com/sul-dlss/dor-services-app/blob/f51bbcea710b7612f251a3922c5164ec69ba39aa/app/services/virtual_object.rb)

Since DSA already sits atop the source of truth for Cocina (Postgres), and it's queryable, DSA can get this information directly from Postgres without needing to consult Solr.

mjgiarlo:
# Why was this change made?

Fixes #4459

This is a spike commit towards an SDR Evolution the team has been batting around for a while now, namely severing DSA's dependency on Solr. The spike largely replaces Solr queries with direct DB queries, and for most use cases this works just fine. The key word here is "most..."

* The Solr queries have been replaced with DB queries that reach into JSONB columns which results in table scans. I tested all of these queries in stage with large-ish, but not prod-huge, data sets (~25K records) and most of them perform fine. That said, we might want to test this with prod-like data and do some benchmarking to determine if we want to index more of the JSONB data.
* A notable performance outlier is `MemberService.for` which needs to make a single Workflow API call for *each* member of a virtual object. These are impressively slow for a virtual object with a few thousand members, taking over a minute to complete.

Another question we'd need to answer to take this work forward is what to do about `bin/generate-druid-list`, which allows a user to issue Solr queries directly, and `lib/tasks/missing_druids.rake`, which compares what's in the DSA DB and what's in Solr to determine if any objects need (re-)indexing. Are these still useful? If so, could they live elsewhere or could we solve these problems in a different way? If the answer is no, we may not want to proceed with this decoupling.

**NOTE:** Since this is a spike meant to generate discussion, I have not yet bothered with deal with changing the tests (or caring about linting). That will naturally come later if we decide the idea and implementation has merit.

# How was this change tested?

N/A, see above.


app/reports/apo_catkey.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'csv'
 
-# Find items that are goverened by the provided APO and then return all catkeys and refresh status.
+# Find items that are governed by the provided APO and then return all catkeys and refresh status.
```
mjgiarlo, line 5: Typo.

app/models/dro.rb
```diff
@@ -5,6 +5,11 @@ class Dro < RepositoryRecord
   validates :content_type, :access, :administrative, :description,
             :identification, :structural, presence: true
 
+  scope :has_admin_policy, ->(admin_policy_druid) { where("administrative ->> 'hasAdminPolicy' = '#{admin_policy_druid}'").select(:external_identifier).order(:external_identifier) }
+  scope :in_virtual_objects, ->(member_druid) { where("structural #> '{hasMemberOrders,0}' -> 'members' ? :druid", druid: member_druid) }
+  scope :members_of_collection, ->(collection_druid) { where("structural -> 'isMemberOf' ? :druid", druid: collection_druid).select(:external_identifier, :version, :content_type) }
+  scope :embargoed_and_releaseable, -> { where("(access -> 'embargo' ->> 'releaseDate')::timestamp <= ?", Time.zone.now).select(:external_identifier) }
```
mjgiarlo, line 11: If nothing else, I learned a bit more about how to do fancy JSONB queries with our crazy nested JSON.

app/services/delete_service.rb
```diff
@@ -4,7 +4,6 @@
 class DeleteService
   # Tries to remove any existence of the object in our systems
   #   Does the following:
-  #   - Removes item from Fedora/Solr
```
mjgiarlo, line 7: Cruft.

app/services/member_service.rb
```diff
@@ -1,19 +1,37 @@
...
+      WorkflowClientFactory.build.status(druid: member.external_identifier, version: member.version).display_simplified == 'Opened'
+    end
+  end
+
+  def self.select_published_members(members, only_published)
+    return members unless only_published
+
+    members.select do |member|
+      WorkflowClientFactory.build.lifecycle(druid: member.external_identifier, milestone_name: 'published', version: member.version).present?
+    end
```
mjgiarlo, line 35: This is the sluggish part.

bin/generate-druid-list
```diff
@@ -4,6 +4,8 @@
 require_relative '../config/environment'
 require 'optparse'
 
+# TODO: Figure out if we still want this or not, given how tightly coupled this functionality is to Solr
```
mjgiarlo, line 7: See comment.

lib/tasks/missing_druids.rake
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# TODO: Figure out if we still want this or not, given how tightly coupled this functionality is to Solr
```
mjgiarlo, line 3: See comment.

PR #4580, opened 2023-09-01
Allow a previous folio hrid record to be missing.

### LINKED ISSUE #4579: Release/unrelease from SW should be able to handle cases where MARC record does not exist in FOLIO
**Context**

This issue applies to items that have 

- FOLIO instance HRIDs

and/or

- previous Folio instance HRIDs

**Describe the bug**

When an SDR item or collection that is linked to a MARC record is unreleased from SW, SDR sends an update to FOLIO to remove any 856 that might correspond to the SDR druid from the relevant FOLIO record(s). If no FOLIO instance exists with the corresponding HRID(s) then the releaseWF throws an error:

`update-marc : problem with update-marc (BackgroundJob: 659738): {:errors=>[{:title=>"Update MARC error", :detail=>"No matching instance found for a33"}]}`

This came up in testing after the FOLIO launch: https://argo-stage.stanford.edu/catalog?f%5Bwf_wps_ssim%5D%5B%5D=releaseWF%3Aupdate-marc%3Aerror

The specific use case that triggered this is running releaseWF on objects (items or collections) that have "previous FOLIO HRIDs" that do not match FOLIO instances. Whenever the releaseWF runs, it tries to run *unrelease* any "previous FOLIO HRIDs" listed in the object. This is intended to "clean" up any FOLIO records that should no longer be linked to SDR items.

This means that if any "previous FOLIO instance HRIDs" listed in an object don't exist in FOLIO, the releaseWF will fail - even if the *current* FOLIO Instance HRID exists and is entered correctly.

**User impact**

Users cannot fix linkages between SDR objects and FOLIO Instance HRIDs if the releaseWF fails in this situation. A few ways that SDR objects end up being linked to FOLIO Instance HRIDs that do not exist:

- The FOLIO instance could have been withdrawn from FOLIO after the object was accessioned into SDR
- The FOLIO instance HRID could have been mistyped when entered in Argo

Both of those situations are not uncommon, especially when some materials get re-cataloged after being accessioned into SDR. Both situations result in SDR storing "previous FOLIO HRIDs" after the metadata is updated to the correct FOLIO instance HRID via Argo.

**Expected behavior**

When *unreleasing*, do not stop the releaseWF if a FOLIO record does not exist. Because *unrelease* only removes metadata, it should be ok if there's no FOLIO record. The overall goal, getting the SDR item out of SW, would be achieved by the FOLIO record not existing. We could have an HB alert in this situation if we want to know when it happens, as long as that doesn't stop the releaseWF.

Conversely, the error should be kept for SDR releases that add SDR metadata to a FOLIO instance. In that case, if the FOLIO record does not exist, then we can't achieve the goal of getting the SDR content into SW. That should continue to be an error that shows up in Argo as a workflow error.

justinlittman:
closes #4579

Since previous records were not necessarily migrated, they may be missing.

Unit, Andrew

### DIFF
--- a/app/services/catalog/folio_writer.rb
+++ b/app/services/catalog/folio_writer.rb
@@ -20,7 +20,7 @@ module Catalog
       return if catalog_record_ids.empty? && previous_catalog_record_ids.empty?
 
       # remove 856 for previous catkeys
-      previous_catalog_record_ids.each { |previous_id| delete_previous_ids(catalog_record_id: previous_id) }
+      previous_catalog_record_ids.each { |previous_id| delete_previous_ids(catalog_record_id: previous_id, ignore_not_found: true) }
 
       # replace 856 for current catkeys
       catalog_record_ids.each { |catalog_record_id| update_current_ids(catalog_record_id:) }
@@ -30,7 +30,7 @@ module Catalog
 
     attr_reader :marc_856_data, :cocina_object
 
-    def delete_previous_ids(catalog_record_id:)
+    def delete_previous_ids(catalog_record_id:, ignore_not_found: false)
       FolioClient.edit_marc_json(hrid: catalog_record_id) do |marc_json|
         marc_json['fields'].reject! { |field| (field['tag'] == '856') && (field['content'].include? purl_subfield) }
       end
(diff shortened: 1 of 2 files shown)


lwrubel (approved):

peetucket (approved):

peetucket:
HOLD can be removed?

andrewjbtw:
I've tested and this can be unheld. There was some weird behavior with one test druid in stage but it turns out to be a data problem. All my tests with other druids/HRIDs were successful.

PR #4582, opened 2023-09-01
Add rake task to register with known druids

### LINKED ISSUE #4581: Rake task for registration with existing druid
As part of seeding the Canonical objects, it is necessary to be able to register items with a druid already assigned.

For this ticket, create a rake task that will perform such a registration. It should read a spreadsheet similar to the existing registration form.

For an approach to creating items with an assigned druid, see https://github.com/sul-dlss/dor-services-app/pull/4567

aaron-collier:

Fixes #4581 

Note that `RegistrationCsvConverter` is largely a direct copy out of Argo for use here (and it's test). 

### DIFF
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -17,17 +17,15 @@ class RegistrationCsvConverter
 
   # @param [String] csv_string CSV string
   # @return [Array<Result>] a list of registration requests suitable for passing off to dor-services-client
-  def self.convert(csv_string:, params: {})
-    new(csv_string:, params:).convert
+  def self.convert(csv_string:)
+    new(csv_string:).convert
   end
 
-  attr_reader :csv_string, :params
+  attr_reader :csv_string
 
   # @param [String] csv_string CSV string
-  # @param [Hash] params that can be used instead of CSV columns. Keys are same as column headers.
-  def initialize(csv_string:, params:)
+  def initialize(csv_string:)
     @csv_string = csv_string
-    @params = params
   end
 
   # @return [Result] an array of dry-monad results
@@ -55,7 +53,7 @@ class RegistrationCsvConverter
   def convert_row(row)
     model = Cocina::Models::RequestDRO.new(model_params(row))
     Success(model:,
-            workflow: params[:initial_workflow] || row.fetch('initial_workflow'),
+            workflow: row.fetch('initial_workflow'),
             tags: tags(row))
   rescue Cocina::Models::ValidationError => e
     Failure(e)
@@ -63,11 +61,11 @@ class RegistrationCsvConverter
 
   def model_params(row)
     model_params = {
-      type: dro_type(params[:content_type] || row.fetch('content_type')),
+      type: dro_type(row.fetch('content_type')),
       version: 1,
       label: row['label'],
       administrative: {
-        hasAdminPolicy: params[:administrative_policy_object] || row.fetch('administrative_policy_object')
+        hasAdminPolicy: row.fetch('administrative_policy_object')
       },
       identification: {
         sourceId: row.fetch('source_id'),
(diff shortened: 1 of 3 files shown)


lib/tasks/seed.rake
```diff
@@ -42,4 +42,23 @@ namespace :seed do # rubocop:disable Metrics/BlockLength
...
+    results.each do |parse_result|
+      puts parse_result[:cocina_request_object]
+      parse_result[:cocina_request_object].either(lambda { |value|
+        CreateObjectService.create(value[:model], id_minter: -> { parse_result[:druid] })
+        client = WorkflowClientFactory.build
+        client.create_workflow_by_name(druid, value[:workflow], version: value[:model][:version])
+      },
+                                                  ->(error) { log_error(error, bulk_action:, log:) })
+    end
+  rescue Errno::ENOENT => e
```
justinlittman, line 61: What is this for?

app/services/registration_csv_converter.rb
```diff
@@ -0,0 +1,144 @@
...
+  # @param [String] csv_string CSV string
+  # @return [Array<Result>] a list of registration requests suitable for passing off to dor-services-client
+  def self.convert(csv_string:, params: {})
+    new(csv_string:, params:).convert
+  end
+
+  attr_reader :csv_string, :params
+
+  # @param [String] csv_string CSV string
+  # @param [Hash] params that can be used instead of CSV columns. Keys are same as column headers.
```
justinlittman, line 27: Why not get rid of the params if we're always passing `{}`?

lib/tasks/seed.rake
```diff
@@ -42,4 +42,23 @@ namespace :seed do # rubocop:disable Metrics/BlockLength
...
+    input_file = args[:input_file] || 'registration.csv'
+    puts "Registering objects from #{input_file}"
+    results = RegistrationCsvConverter.convert(csv_string: File.read(input_file), params: {})
+    results.each do |parse_result|
+      puts parse_result[:cocina_request_object]
+      parse_result[:cocina_request_object].either(lambda { |value|
+        CreateObjectService.create(value[:model], id_minter: -> { parse_result[:druid] })
+        client = WorkflowClientFactory.build
+        client.create_workflow_by_name(druid, value[:workflow], version: value[:model][:version])
+      },
```
justinlittman, line 58: Do we need to create tags?

lib/tasks/seed.rake
```diff
@@ -42,4 +42,23 @@ namespace :seed do # rubocop:disable Metrics/BlockLength
...
+                                                  ->(error) { log_error(error, bulk_action:, log:) })
```
justinlittman, line 59: What provides log_error()?

lib/tasks/seed.rake
aaron-collier: I should probably just drop that since this is a rake task. It catches if the input file is not found. But that will be obvious for a command line tool.

lib/tasks/seed.rake
aaron-collier: Yes, I think so, I meant to add that. Done.

lib/tasks/seed.rake
aaron-collier: Copypasta, replaced. However, I'm open to this not logging but reporting to the user. Hmm.

aaron-collier:
I believe the coverage drop here is for coverage of the big switch statement in RegistrationCsvConverter. The test was copied from Argo and would just switch the type between tests, so seems a bit excessive to try to do anything with coverage there, but happy to do it if we prefer avoiding that drop.

lib/tasks/seed.rake
```diff
@@ -42,4 +42,23 @@ namespace :seed do # rubocop:disable Metrics/BlockLength
...
+    results.each do |parse_result|
+      parse_result[:cocina_request_object].either(lambda { |value|
+        CreateObjectService.create(value[:model], id_minter: -> { parse_result[:druid] })
+
+        value[:tags].map { |tag| AdministrativeTags.create(identifier: parse_result[:druid], tags: tag) }
+
+        client = WorkflowClientFactory.build
+        client.create_workflow_by_name(druid, value[:workflow], version: value[:model][:version])
+      },
+                                                  ->(error) { Rails.logger.error(error) })
```
justinlittman, line 61: Since this is a rake task, can we also send the error to stdout?

justinlittman (changes requested):
What happens if this is run with an object that already exists?

lib/tasks/seed.rake
aaron-collier: Sure - just `puts`-ing it now if that's sufficient.

aaron-collier:
@justinlittman now capturing cocina and pg based duplication errors.

justinlittman (approved):

Issue #4586, opened 2023-09-07
Some SDR objects with catkeys do NOT have current folio_instance_hrid, only "previous folio"


ndushay:
We stumbled onto this bug when looking for SDR objects that were not ETDs that were recently released to SW.  

We do not know how many SDR objects are affected.  Here are 2 of them:

For https://argo.stanford.edu/items/druid:gw145sz1930.json

![image](https://github.com/sul-dlss/dor-services-app/assets/96775/64f5949d-ddf2-42a0-8772-ba5f959cacc1)

For https://argo.stanford.edu/items/druid:bq617py5227.json

![image](https://github.com/sul-dlss/dor-services-app/assets/96775/f51851b0-1cc0-4950-8f7a-f313a81fb5b6)

Here is one that was released to SearchWorks recently that was NOT affected:

In https://argo.stanford.edu/items/druid:gw739nt6189.json

![image](https://github.com/sul-dlss/dor-services-app/assets/96775/c69b9801-e1d6-42c0-97ea-f66079d2c36e)

peetucket:
An attempt to write a report to find the problems: https://github.com/sul-dlss/dor-services-app/pull/4588

justinlittman:
Are we sure that the folio HRIDs weren't just deleted by a user?

justinlittman:
Looks like Jessica made changes about 3 days ago.

peetucket:
Seeing that from version history?  It says "Remove collection catkey" which seems like maybe we have some process that isn't working as expected.  Seems odd to remove the single folio ID.

I'm still twiddling with a report to see what items may have this ... not sure I have the queries quite right yet.

justinlittman:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa0dCIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e166433b88ed0f3ddef84ebd21516d61330ada80/image.png)

peetucket:
Interesting.  I got the report running and found these with previous folios, all except one has no current folio, so presumably we can verify these are the ones that were deliberately removed:

And druid:wx466ty0174 is the one from yesterday that Geoff fixed!

````
item_druid,folios,previous_folios
druid:xq195jn0278,,a4085058
druid:wx466ty0174,in00000002962,in00000002965
druid:kc972mz7533,,a4085693
[...]

PR #4588, opened 2023-09-07
Add new report for reporting on previous folio ID

peetucket:

To help understand the scope of the problem found in #4586 -- which objects have a previous folio ID but not a current folio ID.  Finds all objects with previous folio ID and lists the current ones too so you can find any that are blank.

Ran on sdr-infra

### DIFF
--- a/app/reports/catalog_links_previous_folio_only.rb
+++ b/app/reports/catalog_links_previous_folio_only.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# Find items that have identification with catalogLink(s) that have "previous folio"
+
+# Invoke via:
+# bin/rails r -e production "CatalogLinksPreviousFolioOnly.report"
+class CatalogLinksPreviousFolioOnly
+  JSONB_FOLIO_LINKS_PATH = '$.catalogLinks[*] ? (@.catalog == "folio" && @.refresh == true)'
+  JSONB_PREVIOUS_FOLIO_CLINK_PATH = '$.catalogLinks[*] ? (@.catalog == "previous folio")'
+
+  SQL = <<~SQL.squish.freeze
+    SELECT dros.external_identifier as item_druid,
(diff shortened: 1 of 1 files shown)


app/reports/catalog_links_previous_folio_only.rb
```diff
@@ -0,0 +1,39 @@
...
+class CatalogLinksPreviousFolioOnly
+  JSONB_FOLIO_LINKS_PATH = '$.catalogLinks[*] ? (@.catalog == "folio" && @.refresh == true)'
+  JSONB_PREVIOUS_FOLIO_CLINK_PATH = '$.catalogLinks[*] ? (@.catalog == "previous folio")'
+
+  SQL = <<~SQL.squish.freeze
+    SELECT dros.external_identifier as item_druid,
+           jsonb_path_query_array(dros.identification, '#{JSONB_FOLIO_LINKS_PATH}.catalogRecordId') as folios,
+           jsonb_path_query_array(dros.identification, '#{JSONB_PREVIOUS_FOLIO_CLINK_PATH}.catalogRecordId') as previous_folios
+           FROM "dros" WHERE
+           jsonb_path_exists(dros.identification, '#{JSONB_PREVIOUS_FOLIO_CLINK_PATH}')
```
ndushay, line 16: huh - I would have thought we needed a report on "previous folio" that have no current "folio" but whatevs.

ndushay (approved):

app/reports/catalog_links_previous_folio_only.rb
peetucket: Yeah, I was trying to write a query that did that, but was having troubles.  I think it should be possible though.  In the end for expediency I settled for a report that just found anything with a previous folio ID (which is what this does), and then list them all (with any current folio IDs too).  Easier to filter for those of interest.  It's a small output now, but will grow over time.

Issue #4595, opened 2023-09-22
DSA Post Folio cutover cleanup


ndushay:
This issue is to create PR(s) for removing ILS cutover code.

Please do NOT do this in one giant PR. I recommend making multiple small PRs. This ticket can also be broken up into multiple tickets -- there is a lot here.

- [ ] remove code such as `if Settings.catalog_record_id == 'symphony'`  
- [x] remove SymphonyChecker and such classes
- [x] remove MarcComparisonService and such classes
- [ ] PublicXmlService can stop exposing catkeys?
- [x] specs involving "symphony" or "catkey" should be changed to Folio
- [ ] update the README
- [ ] search on, case insensitive, e.g.:
    - [ ] `catalogRecordId`
    - [ ] `catalog_record_id`
    - [ ] `ckey`
    - [ ] `catkey`
    - [ ] `symphony`
    - [ ] `ils`
    - [ ] ...
- [x] no longer use `/sdrpurl_transfer' - take it out of settings.yml in repo and in all shared configs, and tell ops they can remove the mount from DSA apps?

andrewjbtw:
This item should be verified with the Access team. I think they're still applying logic where they read catkey or FOLIO hrid and I don't know that catkey can be removed without an issue.

>PublicXmlService can stop exposing catkeys?

PR #4610, opened 2023-10-04
Remove out of date reports that reference symphony

aaron-collier:

Part of #4595 


ndushay (changes requested):
The "catkey" in these reports isn't hard to remove or change to folio_instance_hrid. 

Please do NOT remove any reports starting "property"
I would also keep some others.  (coming to this comment soon.)

aaron-collier:
In that case I would suggest closing this PR. It doesn't seem a valuable use of time to tease through them. I'm curious how used they are as I wasn't able to find my evidence in history that any report has been run since cutover (or really at all recently) but I could have looked in the wrong location.

ndushay:
> In that case I would suggest closing this PR. It doesn't seem a valuable use of time to tease through them. I'm curious how used they are as I wasn't able to find my evidence in history that any report has been run since cutover (or really at all recently) but I could have looked in the wrong location.

I think for those of us running on infra, the output can be anywhere we pipe it too.  Plus some folks run in their individual directories and some run from dor-services-app directory.

Happy NOT to poke through inidividual reports.  dunno how to be clear that the "property" ones are designed for Arcadia self-service.

Dunno how many reports we'll need in our new non-fedora world, either.  But I expect some.

PR #4625, opened 2023-10-12
move backup_path to settings

peetucket:

Follow on from #4598 incorporating suggestions from John.

1. Moves the backup_path to settings to make it easily configurable and then stop passing it around as much
2. Add some additional output spacing

Updates specs


ndushay (approved):
LGTM;  @jmartin-sul ?

PR #4627, opened 2023-10-13
Adds unpreserved, unmatches files report.

### LINKED ISSUE #4626: Report for unpreserved files that are not matched with preserved files
I would like a report for files that are on Stacks that are:

- not preserved
- not connected to preserved files (i.e. not JP2s that correspond to TIFFS)

I am not sure it's possible to get this exactly in a single query. 

These are the criteria that I think will get closest to that answer:

- files where "shelve = yes" and "preserve = no"
- that are contained in filesets that do not contain any files that have "preserve = yes"

If that query is not possible, I would still benefit from this modified query:

- files where "shelve = yes" and "preserve = no"
- that are in their own fileset (with no other file in the same fileset)

This is related to https://github.com/sul-dlss/common-accessioning/issues/853 My goal is to find files that probably should be in preservation but are stored only on Stacks. There have been cases where the preserve setting was mistakenly set to "no" when it should have been "yes".

justinlittman:
closes #4626

Per Andrew

### DIFF
--- a/app/reports/unpreserved_unmatched_files.rb
+++ b/app/reports/unpreserved_unmatched_files.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+# Generates a report of unpreserved files that are not matched with preserved files.
+# That is, files where "shelve = yes" and "preserve = no" that are contained in
+# filesets that do not contain any files that have "preserve = yes"
+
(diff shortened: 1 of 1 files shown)


justinlittman:
Report delivered to and accepted by Andrew.

lwrubel (approved):

PR #4630, opened 2023-10-17
Fix flattening in bin/validate-cocina

justinlittman:
The script wasn't working correctly.

sdr-infra


ndushay (approved):
approving but unclear on how this works ... was it under- or over- flattening?   Is it gonna look at the whole beast?

justinlittman:
It was over-flattening. (By default `.flatten` is recursive.)

PR #4635, opened 2023-10-18
Mark workflow step completed after actually performing reset workspace.

### LINKED ISSUE #4634: [dor-services-app/prod] Errno::ENOENT: No such file or directory @ rb_file_s_rename - (/dor/workspace/vt/466/zt/7148/vt466zt7148, /dor/workspace/vt/466/zt/7148/vt466zt7148_v1)
## Backtrace

    line 23 of [PROJECT_ROOT]/app/services/reset_workspace_service.rb: reset_workspace_druid_tree
    line 10 of [PROJECT_ROOT]/app/services/reset_workspace_service.rb: reset
    line 10 of [PROJECT_ROOT]/app/jobs/reset_workspace_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/97140442)

justinlittman:
closes #4634

The reset_workspace step processing is done here in DSA by a job class, so that job class should updated the workflow status to "completed" ... since it is doing the work!

Unit, stage

### DIFF
--- a/app/jobs/reset_workspace_job.rb
+++ b/app/jobs/reset_workspace_job.rb
@@ -2,14 +2,33 @@
 
 # Invokes the ResetWorkspaceService
 class ResetWorkspaceJob < ApplicationJob
-  queue_as :low
+  queue_as :default
+
+  WORKFLOW_PROCESS = 'reset-workspace'
 
   # @param [String] druid the identifier of the object to be reset
   # @param [Integer] version the version of the object to be reset
-  def perform(druid:, version:)
-    ResetWorkspaceService.reset(druid:, version:)
-  rescue ResetWorkspaceService::DirectoryAlreadyExists
-    # We're trapping errors and doing nothing, because the belief is that these indicate
-    # this API has already been called and completed.
+  # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
+  # @param [String] workflow Which workflow should this be reported to?
+  def perform(druid:, version:, background_job_result:, workflow:)
+    background_job_result.processing!
+
+    begin
+      ResetWorkspaceService.reset(druid:, version:)
+    rescue ResetWorkspaceService::DirectoryAlreadyExists
+      # We're trapping errors and doing nothing, because the belief is that these indicate
+      # this API has already been called and completed.
+    rescue StandardError => e
+      return LogFailureJob.perform_later(druid:,
+                                         background_job_result:,
+                                         workflow:,
+                                         workflow_process: WORKFLOW_PROCESS,
+                                         output: { errors: [{ title: 'Unable to reset workspace', detail: e.message }] })
+    end
+
+    LogSuccessJob.perform_later(druid:,
+                                workflow:,
+                                background_job_result:,
+                                workflow_process: WORKFLOW_PROCESS)
   end
 end
--- a/spec/jobs/reset_workspace_job_spec.rb
+++ b/spec/jobs/reset_workspace_job_spec.rb
@@ -4,20 +4,30 @@ require 'rails_helper'
 
 RSpec.describe ResetWorkspaceJob do
   subject(:perform) do
-    described_class.perform_now(druid:, version:)
+    described_class.perform_now(druid:, version:, background_job_result: result, workflow:)
   end
 
   let(:druid) { 'druid:mk420bs7601' }
   let(:version) { 1 }
+  let(:result) { create(:background_job_result) }
+  let(:workflow) { 'accessionWF' }
+
+  before do
+    allow(result).to receive(:processing!)
+    allow(LogSuccessJob).to receive(:perform_later)
+  end
 
   context 'with no errors' do
     before do
       allow(ResetWorkspaceService).to receive(:reset)
     end
 
-    it 'invokes the ResetService' do
+    it 'invokes the ResetService and logs success' do
       perform
+      expect(result).to have_received(:processing!).once
       expect(ResetWorkspaceService).to have_received(:reset).with(druid:, version:).once
+      expect(LogSuccessJob).to have_received(:perform_later)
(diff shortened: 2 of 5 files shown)


app/jobs/reset_workspace_job.rb
```diff
@@ -2,14 +2,33 @@
...
   # @param [String] druid the identifier of the object to be reset
   # @param [Integer] version the version of the object to be reset
-  def perform(druid:, version:)
-    ResetWorkspaceService.reset(druid:, version:)
-  rescue ResetWorkspaceService::DirectoryAlreadyExists
-    # We're trapping errors and doing nothing, because the belief is that these indicate
-    # this API has already been called and completed.
+  # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
+  # @param [String] workflow Which workflow should this be reported to?
+  def perform(druid:, version:, background_job_result:, workflow:)
```
ndushay, line 13: won't changing this signature kind of bork any objects currently in flight that would call this reset?   Objects that errored and are going through sidekiq or whatever?

I mean, yes, we need to be able to change these signatures ... but is there a deployment gotcha to be aware of?

ndushay (commented):
one inline question.  I confess I'm not fully grokking this.

common-accessioning accessionWF reset_workspace is a robot that calls the DSA reset_workspace_job via DSC workspace.reset method ...

common-accessioning robot needs to be a noop that completes, and then DSA background job can do the reset of the workspace as needed?   I don't get how this solves the problem.

Can you break it down for my brain??

app/jobs/reset_workspace_job.rb
```diff
@@ -2,14 +2,33 @@
...
+                                         background_job_result:,
+                                         workflow:,
+                                         workflow_process: WORKFLOW_PROCESS,
+                                         output: { errors: [{ title: 'Unable to reset workspace', detail: e.message }] })
+    end
+
+    LogSuccessJob.perform_later(druid:,
+                                workflow:,
+                                background_job_result:,
+                                workflow_process: WORKFLOW_PROCESS)
```
ndushay, line 32: and this marks the workflow step as "completed", yes?

spec/jobs/reset_workspace_job_spec.rb
```diff
@@ -26,9 +36,26 @@
       allow(ResetWorkspaceService).to receive(:reset).and_raise(ResetWorkspaceService::DirectoryAlreadyExists)
     end
 
-    it 'ignores' do
+    it 'ignores and logs success' do
```
ndushay, line 39: "logs makes me think it's writing to a rails log ... this is also updating the workflow status, yes?

spec/requests/reset_workspace_spec.rb
```diff
@@ -4,17 +4,21 @@
...
     it 'is successful' do
-      post "/v1/objects/#{druid}/workspace/reset", headers: { 'Authorization' => "Bearer #{jwt}" }
-      expect(ResetWorkspaceJob).to have_received(:perform_later).with(druid:, version: 2)
-      expect(response).to be_no_content
+      post "/v1/objects/#{druid}/workspace/reset?workflow=accessionWF&lane-id=low",
+           headers: { 'Authorization' => "Bearer #{jwt}" }
+      expect(ResetWorkspaceJob).to have_received(:set).with(queue: 'low')
+      expect(job).to have_received(:perform_later)
+        .with(druid:, version: 2, background_job_result: BackgroundJobResult, workflow: 'accessionWF')
+      expect(response).to have_http_status(:created)
```
ndushay, line 21: would it be stupid to show that something is sent to workflow service with "completed" as status?

ndushay (commented):
Is common accessioning the only consumer of the reset_workspace_job?

app/jobs/reset_workspace_job.rb
justinlittman: For the moments between common accessioning being deployed and DSA being deployed, errors might occur. If that does occur, they should be easy to remediate. FR (@mjgiarlo) should definitely be given a head's up.

app/jobs/reset_workspace_job.rb
justinlittman: Yes.

spec/jobs/reset_workspace_job_spec.rb
justinlittman: Saying "logs success" because it invokes the LogSuccessJob.

spec/requests/reset_workspace_spec.rb
justinlittman: This is just testing that the correct job is invoked; however, the job is mocked since testing the job is handled by the job's specs.  The job's specs in turn test that a LogSuccessJob is created. I'm assuming that the specs for the LogSuccessJob test that the workflow service is called correctly. And all of this is verified at a high level by the integration tests.

By my reckoning, it's covered but open to further discussion.

justinlittman:
> common-accessioning accessionWF reset_workspace is a robot that calls the DSA reset_workspace_job via DSC workspace.reset method ...

> common-accessioning robot needs to be a noop that completes, and then DSA background job can do the reset of the workspace as needed? I don't get how this solves the problem.

Before, reset_workspace robot would report the reset workspace workflow step as complete as soon as it requested that DSA reset the workspace. The workflow would then immediately continue, even though DSA was resetting the workspace asynchronously (in a job). This means that the following step could be performed before the reset workspace step was actually completed, especially if there was an error.

After, the reset_workspace robot does not report the reset workspace workflow step as completed. Rather, the DSA reset workspace job reports the workflow step as completed when it is actually completed.

(Previously, the DSA reset workspace endpoint operated synchronously; however, sometimes it is slow, resulting in timeouts. This work should have been done when resetting the workspace was changed to a job, but I screwed it up.)

ndushay (approved):
Thanks for the explanations!

PR #4652, opened 2023-12-05
Enhance events with display label report.

### LINKED ISSUE #4649: Report request: event with displayLabel
Select for report: digital objects that have a value in cocina event.displayLabel (top-level event only).

Include in output: object purl, object title, displayLabel value, catalog ID if present, collection name

If multiple instances of event have a value in displayLabel, either include all on separate lines or concatenate - whichever is easier.

Output should be in CSV format.

justinlittman:
closes #4649

Per Arcadia

### DIFF
--- a/app/reports/property_events_with_display_label.rb
+++ b/app/reports/property_events_with_display_label.rb
@@ -5,16 +5,36 @@
 # bin/rails r -e production "PropertyEventsWithDisplayLabel.report"
 #
 class PropertyEventsWithDisplayLabel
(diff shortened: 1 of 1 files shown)


jmartin-sul (approved):

jmartin-sul:
whoops, should've looked at the ticket before reviewing/merging...

easy enough to construct after the fact from the druid, but it seems like the report doesn't include object PURL, though the ticket asks for that?

PR #4667, opened 2024-01-05
[HOLD] Retrieve members of collection from db.

justinlittman:
Remove dependency on Solr.

Unit


app/controllers/members_controller.rb
```diff
@@ -2,8 +2,8 @@
 
 # Responds to queries about the members of a collection
 class MembersController < ApplicationController
-  # Return the published members of this collection
```
justinlittman, line 5: This endpoint is being changed to return ALL members, not just published. This is because this is a synchronous endpoint, but limiting to published members might be slow (since it requires hitting workflow service for each item).

The single consumer of this endpoint (common accessioning) will need to limit to published members itself.

app/services/member_service.rb
```diff
@@ -5,15 +5,45 @@ class MemberService
   # @param [String] druid the identifier of the collection
   # @param [Boolean] only_published when true, restrict to only published items
   # @param [Boolean] exclude_opened when true, exclude opened items
-  # @return [Array<Hash<String,String>>] the members of this collection
+  # @return [Array<String>] the druids for the members of this collection
```
justinlittman, line 8: Changed to only return an array of druids.

db/migrate/20240104210953_add_member_of_collection_index_to_dros.rb
```diff
@@ -0,0 +1,5 @@
+class AddMemberOfCollectionIndexToDros < ActiveRecord::Migration[7.1]
+  def change
+    add_index :dros, "(structural->'isMemberOf')", using: 'gin'
```
justinlittman, line 3: This adds an index that makes the json query fast.

openapi.yml
```diff
@@ -665,23 +665,11 @@ paths:
         '200':
           description: OK
           content:
-            application/json:
+            application/json:
```
justinlittman, line 668: This changes the endpoint to just return an array of druids.

Note that the previous endpoint definition seemed to allow for collections to be collection members; they can't -- only DROs can.

aaron-collier: I'm only seeing this change as some end-of-line whitespace, is that expected?

justinlittman: Sorry, my comment is outdated. The relevant openapi changes are L679-680.

app/views/members/index.json.jbuilder
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 json.members @members do |member|
-  json.externalIdentifier member['id']
-  json.type member['objectType_ssim']&.first
+  json.externalIdentifier member.external_identifier
+  json.version member.version
```
mjgiarlo, line 5: Can you say more about why we don't need the object type in here now? I think I grok adding the version: because common-accessioning will need to figure out if the object version is published or not (confirm/deny/amend?)

mjgiarlo: Having reviewed the DSC and C-A PRs, I now grok this. Nevermind!

mjgiarlo (approved):

PR #4676, opened 2024-01-16
Adds rolling indexer

justinlittman:
Better indexer performance since removes API calls to DSA.

QA

### DIFF
--- a/config/deploy/prod.rb
+++ b/config/deploy/prod.rb
@@ -3,6 +3,6 @@
 server 'dor-services-app-prod-a.stanford.edu', user: 'dor_services', roles: %w[web app]
 server 'dor-services-app-prod-b.stanford.edu', user: 'dor_services', roles: %w[web app]
 server 'dor-services-worker-prod-a.stanford.edu', user: 'dor_services', roles: %w[app worker scheduler]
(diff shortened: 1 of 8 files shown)


config/deploy/prod.rb
```diff
@@ -3,6 +3,6 @@
 server 'dor-services-app-prod-a.stanford.edu', user: 'dor_services', roles: %w[web app]
 server 'dor-services-app-prod-b.stanford.edu', user: 'dor_services', roles: %w[web app]
 server 'dor-services-worker-prod-a.stanford.edu', user: 'dor_services', roles: %w[app worker scheduler]
-server 'dor-services-worker-prod-b.stanford.edu', user: 'dor_services', roles: %w[app worker]
+server 'dor-services-worker-prod-b.stanford.edu', user: 'dor_services', roles: %w[app worker rolling_indexer]
```
ndushay, line 6: should this be on -a for consistency?  should qa and stage be on -b for consistency?  Does naomi want consistency?

ndushay (approved):
comments for your consideration.

config/deploy/prod.rb
justinlittman: I put it on -b, since -a had scheduler role. Probably makes no difference, but thought I would spread the additional load around.

PR #4709, opened 2024-03-11
Normalize administrative tag labels to remove leading/trailing whitespace

mjgiarlo:
# Why was this change made?

Fixes sul-dlss/argo#4327

The regular expression used in the DSA codebase and at the API layer (as expressed in `openapi.yml`) allows the user to enter both leading and trailing whitespace for administrative tags, and we don't want this whitespace persisted. The solution in this commit adds normalization at the DB model layer, removing the bad whitespace.

I opted for this solution over a validation-based solution (tightening the various regular expressions) because that may make remediation of currently persisted bad data harder.

We could consider a follow-up step that tweaks the regexes, but also, that's arguably of limited value once this commit is merged and bad data is remediated. Why? More work than it's worth, because we'd likely want to couple that change with user messaging to give them more guidance about how to enter a valid admin tag.

# How was this change tested?

CI

### DIFF
--- a/app/models/administrative_tag.rb
+++ b/app/models/administrative_tag.rb
@@ -2,7 +2,5 @@
 
 # Administrative tags for an object
 class AdministrativeTag < ApplicationRecord
-  VALID_TAG_PATTERN = /\A.+( : .+)+\z/
-
   belongs_to :tag_label
 end
--- a/app/models/tag_label.rb
+++ b/app/models/tag_label.rb
@@ -3,14 +3,12 @@
 # This stores the tag text. This text could be applied to more than one object by an AdministrativeTag
 class TagLabel < ApplicationRecord
   VALID_TAG_PATTERN = /\A.+( : .+)+\z/
-  has_many :administrative_tags, dependent: :destroy
 
-  # rubocop:disable Rails/I18nLocaleTexts
+  has_many :administrative_tags, dependent: :destroy
+  normalizes :tag, with: ->(tag) { tag.squish }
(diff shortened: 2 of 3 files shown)


app/models/administrative_tag.rb
```diff
@@ -2,7 +2,5 @@
 
 # Administrative tags for an object
 class AdministrativeTag < ApplicationRecord
-  VALID_TAG_PATTERN = /\A.+( : .+)+\z/
```
mjgiarlo, line 5: Cruft. This regex moved to the `TagLabel` model.

app/models/tag_label.rb
```diff
@@ -3,14 +3,12 @@
 # This stores the tag text. This text could be applied to more than one object by an AdministrativeTag
 class TagLabel < ApplicationRecord
   VALID_TAG_PATTERN = /\A.+( : .+)+\z/
-  has_many :administrative_tags, dependent: :destroy
```
mjgiarlo, line 6: Put the class-level macros in alpha order.

app/models/tag_label.rb
```diff
@@ -3,14 +3,12 @@
...
 
-  # rubocop:disable Rails/I18nLocaleTexts
+  has_many :administrative_tags, dependent: :destroy
+  normalizes :tag, with: ->(tag) { tag.squish }
```
mjgiarlo, line 8: Here's the new code that fixes the related issue.

justinlittman (approved):

PR #4723, opened 2024-03-20
Create a active_record model for release tags

### LINKED ISSUE #4716: Implement ReleaseTag active record model
This should be consistent with cocina release tag model: https://github.com/sul-dlss/cocina-models/blob/main/openapi.yml#L1523-L1550

Initially, a ReleaseTag references an item via druid field (similar to AdministrativeTags). Once the Object model is implemented, the druid will be replaced with an association to the Object to ensure referential integrity.

jcoyne:

Fixes #4716

### DIFF
--- a/app/models/release_tag.rb
+++ b/app/models/release_tag.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class ReleaseTag < ApplicationRecord
+end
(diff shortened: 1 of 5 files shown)


justinlittman:
@jcoyne Approve pending resolving conflict.

PR #4725, opened 2024-03-21
Add an API for release tags

### LINKED ISSUE #4714: Create ReleaseTags API endpoints and new release tags service
The endpoint and service should support adding and getting release tags for DROs and Collections. The initial implementation should read and write to the cocina of Dro and Collection active record objects. (That is, this is just an abstraction around current functionality for release tags.)

Note:
* Release tags are added, but not updated or deleted.
* Getting release tags should be the release tags for the item only, not the item and its collection. (Getting release tags for an item and its collection is handled by the existing ReleaseTags service.)
* There is an existing ReleaseTags service that munges release tags for an item and its collection. It is up to dev to determine how to handle naming of the new and existing service, but they should remain separate services.

jcoyne:
Fixes #4714

CI

### DIFF
--- a/app/controllers/release_tags_controller.rb
+++ b/app/controllers/release_tags_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+# Provides an API for release tags
+class ReleaseTagsController < ApplicationController
+  before_action :load_cocina_object, only: [:index, :create]
+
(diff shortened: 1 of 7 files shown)


justinlittman (changes requested):
Can you please abstract the release tag actions (get, add) behind a service? This will lay the groundwork for switching the persistence of the release tags.

PR #4737, opened 2024-03-26
Adds default order for release tags.

### LINKED ISSUE #4736: Order ReleaseTags by date ascending
This is to replicate ordering in cocina, where new release tags were appended.

justinlittman:
closes #4736

To ensure that ordering correct when querying release tags.

Unit

### DIFF
--- a/app/models/release_tag.rb
+++ b/app/models/release_tag.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 
+# Model for release tags, which indicate that an item or collection should be released.
 class ReleaseTag < ApplicationRecord
+  default_scope { order(created_at: :asc) }
 end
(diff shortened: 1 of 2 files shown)


app/models/release_tag.rb
```diff
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 
+# Model for release tags, which indicate that an item or collection should be released.
 class ReleaseTag < ApplicationRecord
+  default_scope { order(created_at: :asc) }
```
jcoyne, line 5: Can you explain here why we need an order? I'm always cautious of using default scope, because it's not always clear if we can change it safely. Can we add explicit order on the API, if that is the place this is mainly used?

jcoyne (changes requested):

justinlittman:
@jcoyne In answering your question, I came up with a different approach for migrating from cocina release tags to active record release tags that didn't require default ordering. (That is, the normal default ordering would be correct.)

PR #4741, opened 2024-03-26
Delete ReleaseTag model objects when deleting an item.

### LINKED ISSUE #4740: Delete ReleaseTags in DeleteService
Similar to other activerecord objects.

justinlittman:
closes #4740

Cleanup, similar to other active record models.

Unit

### DIFF
--- a/app/services/delete_service.rb
+++ b/app/services/delete_service.rb
@@ -54,6 +54,7 @@ class DeleteService
     AdministrativeTags.destroy_all(identifier: druid)
     ObjectVersion.where(druid:).destroy_all
     Event.where(druid:).destroy_all
+    ReleaseTag.where(druid:).destroy_all
   end
 
   def druid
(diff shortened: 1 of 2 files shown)


app/services/delete_service.rb
```diff
@@ -54,6 +54,7 @@ def delete_from_dor
     AdministrativeTags.destroy_all(identifier: druid)
     ObjectVersion.where(druid:).destroy_all
     Event.where(druid:).destroy_all
+    ReleaseTag.where(druid:).destroy_all
```
jcoyne, line 57: Can/should this method be wrapped in a transaction?

jcoyne (changes requested):

app/services/delete_service.rb
justinlittman: I don't think we should. Since all of the other delete actions aren't transactional, it increases the likelihood that an item might be left in a condition in which all of the files, workflows, etc. have been deleted but the Dro/Collection/AdminPolicy has not. That state is likely to cause problems for a variety of systems, e.g., indexing.

If it is not in a transaction and something goes wrong with deleting, say ReleaseTags, it might leave orphaned ReleaseTags but at least the Dro/Collection/AdminPolicy has been deleted. That is not likely to cause problems for other systems.

Once Dro/Collection/AdminPolicy have been consolidated into Objects we can consider RI for AdministrativeTags, ObjectVersions, Events, and ReleaseTags.

PR #4752, opened 2024-03-27
Update dor-indexing and other dependencies

aaron-collier:

Bump Dependencies to pick up dor-indexing 1.5.0

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -164,6 +164,13 @@ GEM
       capistrano-shared_configs
     docile (1.4.0)
     domain_name (0.6.20240107)
+    dor-services-client (14.4.0)
+      activesupport (>= 4.2, < 8)
+      cocina-models (~> 0.95.1)
+      deprecation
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -164,16 +164,24 @@ GEM
       capistrano-shared_configs
     docile (1.4.0)
     domain_name (0.6.20240107)
+    dor-services-client (14.4.0)
```
jcoyne, line 167: This is when we start wondering if we should just move all the indexing into DSA.

justinlittman (changes requested):
Needs injection of configured Dor Services Client.

aaron-collier:
@justinlittman Done (I believe) - this will need an update to staged_configs, I'll put up PRs for that now as well to go with these.

mjgiarlo:
Superseded by https://github.com/sul-dlss/dor-services-app/pull/4756

Issue #4757, opened 2024-03-28
Remove code from ReleaseTags service that reads to or writes to cocina


justinlittman:

justinlittman:
Blocked until migration is complete (#4749), Dor Indexing work is complete (https://github.com/sul-dlss/dor_indexing/issues/41), and we are confident that release tags are working as expected.

justinlittman:
Please hold PR for @andrewjbtw approval.

PR #4761, opened 2024-03-28
Notify honeybadger when an update is issued on a non-opened object

### LINKED ISSUE #4759: Check in UpdateObjectStore.update that item is open
If it is not open, send a Honeybadger notification.

jcoyne:

Fixes #4759


jcoyne:
Holding pending completion of release tags work (#4757).

justinlittman (approved):

PR #4762, opened 2024-03-28
Add RepositoryObject and RepositoryObjectVersion models

### LINKED ISSUE #4760: Implement RepositoryObject and RepositoryObjectVersion models
RepositoryObject
- type: dro, admin_policy, or collection
- external_identifier: unique
- source_id: unique
- head: one-to-one with latest closed RepositoryObjectVersion; nullable
- open: one-to-one with open RepositoryObjectVersion; nullable
- object_versions: one-to-many with RepositoryObjectVersions
- lock

RepositoryObjectVersion
- version
- version description
- cocina_version (cocina)
- content_type (cocina)
- label (cocina)
- access (cocina)
- administrative (cocina)
- description (cocina)
- identification (cocina)
- structural (cocina)
- geographic (cocina)
- closed_at

Please check Dro, AdminPolicy, and Collections for any needed indexes.

jcoyne:
Fixes #4760 

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+# Models a repository object (item/DRO, collection, or admin policy)
+class RepositoryObject < ApplicationRecord
+  class VersionAlreadyOpened < StandardError; end
+  class VersionNotOpened < StandardError; end
+
+  has_many :versions,  class_name: 'RepositoryObjectVersion', dependent: :destroy, inverse_of: 'repository_object'
+
+  belongs_to :current, class_name: 'RepositoryObjectVersion', optional: true
+  belongs_to :head, class_name: 'RepositoryObjectVersion', optional: true
+  belongs_to :open, class_name: 'RepositoryObjectVersion', optional: true
+
+  enum :object_type, %i[dro admin_policy collection].index_with(&:to_s)
+
+  validates :external_identifier, :object_type, presence: true
+  validates :source_id, presence: true, if: -> { dro? }
+  validate :head_and_open_cannot_be_same_version
+  validate :current_must_be_either_head_or_open
+
+  after_create_commit :open_first_version
+  before_destroy :unset_version_relationships, prepend: true
+
+  # NOTE: This block uses metaprogramming to create the equivalent of scopes that query the RepositoryObjectVersion table using only rows that are a `current` in the RepositoryObject table
+  #
+  # So it's a more easily extensible version of:
+  #
+  # scope :currently_in_virtual_objects, ->(member_druid) { joins(:current).merge(RepositoryObjectVersion.in_virtual_objects(member_druid)) }
+  # scope :currently_members_of_collection, ->(collection_druid) { joins(:current).merge(RepositoryObjectVersion.members_of_collection(collection_druid)) }
+  class << self
+    def method_missing(method_name, ...)
+      if method_name.to_s =~ /#{current_scope_prefix}(.*)/
+        joins(:current).merge(
+          RepositoryObjectVersion.public_send(Regexp.last_match(1).to_sym, ...)
+        )
+      else
+        super
+      end
+    end
+
+    def respond_to_missing?(method_name, include_private = false)
+      method_name.to_s.start_with?(current_scope_prefix) || super
+    end
+
+    private
+
+    def current_scope_prefix
+      'currently_'
+    end
+  end
+
+  def open_version!
+    raise VersionAlreadyOpened, "Cannot open new version because one is already open: #{current.version}" if open?
+
+    RepositoryObject.transaction do
+      version_to_open = head.dup.tap { |object_version| object_version.version += 1 }
+      update!(open: version_to_open, current: version_to_open)
+    end
+  end
+
+  def close_version!
+    raise VersionNotOpened, "Cannot close version because current version is closed: #{current.version}" if closed?
+
+    RepositoryObject.transaction do
+      version_to_close = self.open.tap { |object_version| object_version.closed_at = Time.current }
+      update!(open: nil, head: version_to_close, current: version_to_close)
+    end
+  end
+
+  def open?
+    current == open
+  end
+
+  def closed?
+    current == head
+  end
+
+  private
+
+  def open_first_version
+    RepositoryObject.transaction do
+      first_version = versions.create!(version: 1, version_description: 'Initial version')
+      update!(open: first_version, current: first_version)
+    end
+  end
+
+  def unset_version_relationships
+    update(head: nil, current: nil, open: nil)
+  end
+
+  def head_and_open_cannot_be_same_version
+    return if (head.nil? && open.nil?) || head != open
+
+    errors.add(:head, 'cannot be the same version as the open version')
+  end
+
+  def current_must_be_either_head_or_open
+    return if current.nil? || current == head || current == open
+
+    errors.add(:current, 'must point at either the head version or the open version')
+  end
+end
--- a/app/models/repository_object_version.rb
+++ b/app/models/repository_object_version.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+# Models a repository object as it looked at a particular version.
+class RepositoryObjectVersion < ApplicationRecord
+  belongs_to :repository_object
+
+  scope :in_virtual_objects, ->(member_druid) { where("structural #> '{hasMemberOrders,0}' -> 'members' ? :druid", druid: member_druid) }
+  scope :members_of_collection, ->(collection_druid) { where("structural -> 'isMemberOf' ? :druid", druid: collection_druid) }
+
+  validates :version, :version_description, presence: true
+end
--- a/db/migrate/20240328141842_create_repository_objects.rb
+++ b/db/migrate/20240328141842_create_repository_objects.rb
@@ -1,0 +1,18 @@
+class CreateRepositoryObjects < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :repository_object_type, %w[dro admin_policy collection]
+
+    create_table :repository_objects do |t|
+      t.enum :object_type, enum_type: :repository_object_type, null: false
+      t.string :external_identifier, null: false
+      t.string :source_id, null: true
+      t.integer :lock
+
+      t.timestamps
+    end
+
+    add_index :repository_objects, :object_type
+    add_index :repository_objects, :external_identifier, unique: true
+    add_index :repository_objects, :source_id, unique: true
+  end
+end
--- a/db/migrate/20240328142339_create_repository_object_versions.rb
+++ b/db/migrate/20240328142339_create_repository_object_versions.rb
@@ -1,0 +1,24 @@
+class CreateRepositoryObjectVersions < ActiveRecord::Migration[7.1]
+  def change
+    create_table :repository_object_versions do |t|
+      t.references :repository_object, null: false, foreign_key: true
+      t.integer :version, null: false
+      t.string :version_description, null: false
+      t.integer :cocina_version
+      t.string :content_type
+      t.string :label
+      t.jsonb :access
+      t.jsonb :administrative
+      t.jsonb :description
+      t.jsonb :identification
+      t.jsonb :structural
+      t.jsonb :geographic
+      t.datetime :closed_at
+
+      t.timestamps
+    end
+
+    add_index :repository_object_versions, "(structural#>'{hasMemberOrders,0}'->'members')", using: 'gin'
+    add_index :repository_object_versions, "(structural->'isMemberOf')", using: 'gin'
+  end
+end
--- a/spec/factories/repository_objects.rb
+++ b/spec/factories/repository_objects.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :repository_object do
+    object_type { 'dro' }
+    external_identifier { generate(:unique_druid) }
+    source_id { "sul:#{SecureRandom.uuid}" }
+    lock { 'MyString' }
+  end
+end
--- a/db/migrate/20240328142859_add_version_refs_to_repository_objects.rb
+++ b/db/migrate/20240328142859_add_version_refs_to_repository_objects.rb
@@ -1,0 +1,7 @@
+class AddVersionRefsToRepositoryObjects < ActiveRecord::Migration[7.1]
+  def change
+    add_reference :repository_objects, :current, null: true, foreign_key: { to_table: :repository_object_versions }
+    add_reference :repository_objects, :head, null: true, foreign_key: { to_table: :repository_object_versions }
+    add_reference :repository_objects, :open, null: true, foreign_key: { to_table: :repository_object_versions }
+  end
+end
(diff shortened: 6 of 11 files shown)


db/migrate/20240328141842_create_repository_objects.rb
```diff
@@ -0,0 +1,11 @@
+class CreateRepositoryObjects < ActiveRecord::Migration[7.1]
+  def change
+    create_table :repository_objects do |t|
+      t.string :type, null: false
```
mjgiarlo, line 4: Is it worth considering backing the repository object type enum with a DB-defined enum? We *do* follow this practice in a number of places (e.g., background job result statuses in DSA), but I suspect we're inconsistent about this practice, which is why I ask.

app/models/repository_object.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Models a repository object (item, collection, or admin policy)
+class RepositoryObject < ApplicationRecord
+  has_many :repository_object_versions, dependent: :destroy
+  belongs_to :head, class_name: 'RepositoryObjectVersion', optional: true
+  belongs_to :open, class_name: 'RepositoryObjectVersion', optional: true
+
+  validates :external_identifier, :type, presence: true
```
mjgiarlo, line 9: I assume we don't benefit much from druid validation here?

spec/factories/repository_objects.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :repository_object do
+    type { '' }
```
mjgiarlo, line 5: Should we default this to DRO?

db/migrate/20240328142339_create_repository_object_versions.rb
```diff
@@ -0,0 +1,21 @@
+class CreateRepositoryObjectVersions < ActiveRecord::Migration[7.1]
+  def change
+    create_table :repository_object_versions do |t|
+      t.references :repository_object, null: false, foreign_key: true
+      t.integer :version, null: false
+      t.string :version_description
+      t.integer :cocina_version, null: false
```
justinlittman, line 7: All of the cocina attributes need to be nullable since the existing ObjectVersions that are going to be migrated don't have any cocina.

app/models/repository_object_version.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Models a repository object as it looked at a particular version.
+class RepositoryObjectVersion < ApplicationRecord
+  belongs_to :repository_object
+
+  validates :version, :label, :administrative, presence: true
```
justinlittman, line 7: Presence can't be validated, since existing ObjectVersions that are being migrated won't have them.

db/migrate/20240328141842_create_repository_objects.rb
justinlittman: Can you say more about the merits of DB-defined enums? Just more validation?

db/migrate/20240328141842_create_repository_objects.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.string :external_identifier, null: false
+      t.integer :lock
```
justinlittman, line 6: As discussed with @jcoyne, it will be necessary to have a source_id attribute (nullable, unique).

In addition, a validation should be added to check that if a DRO or Collection, source_id is present.

spec/factories/repository_objects.rb
justinlittman: I'm taking these factories as stubs. Most of the values are not helpful, so whoever uses them first will need to fix them.

db/migrate/20240328141842_create_repository_objects.rb
mjgiarlo: This is one of those rare spots where we might feel that Rails doesn't provide us enough safeguards, but I concede that preventing race conditions by moving model validations to DB constraints (e.g., for uniqueness conditions) is a different problem than stuffing enums in the DB. We've used enums in DSA, sdr-api, and H2 for some spots where we want to prevent variation (background job results, H2 work access levels), and this *may* be one of those spots. YAGNI? Maybe. But also quick and easy to add if we want to 💯 prevent winding up with bespoke repository object type values. I could go either way. Tech lead call? @justinlittman

mjgiarlo: @justinlittman If you want to see what the change would look like, here you go: [`a8c7adf` (#4762)](https://github.com/sul-dlss/dor-services-app/pull/4762/commits/a8c7adf3430d01a11554f1cdceffa0cf2fa333fe)

db/migrate/20240328141842_create_repository_objects.rb
```diff
@@ -1,7 +1,9 @@
 class CreateRepositoryObjects < ActiveRecord::Migration[7.1]
   def change
+    create_enum :repository_object_type, %w[dro admin_policy collection]
+
     create_table :repository_objects do |t|
-      t.string :type, null: false
+      t.enum :type, enum_type: :repository_object_type, null: false
```
mjgiarlo, line 6: Rails 7 added some nifty enum-related methods so we don't need raw SQL in migrations.

app/models/repository_object_version.rb
mjgiarlo: @justinlittman Which of these won't existing ObjectVersions have? All of them? Just the administrative? etc.

db/migrate/20240328141842_create_repository_objects.rb
mjgiarlo: @justinlittman Can you say more about why it should be both nullable and always present?

spec/factories/repository_objects.rb
mjgiarlo: K, fine by me.

db/migrate/20240328141842_create_repository_objects.rb
justinlittman: DB-defined enum works for me.

app/models/repository_object_version.rb
justinlittman: Of this list, only `:version`.

db/migrate/20240328141842_create_repository_objects.rb
justinlittman: AdminPolicies don't have source ids.

bergandr: I think some collections have source ids and some don't. It shouldn't be required.

andrewjbtw: [sorry, if you get an email from a weird user id, that's because I commented from the wrong github account previously]

I think source_id is optional for collections. Some may have them and some don't.

app/models/repository_object.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  after_create_commit :open_first_version
+  before_destroy :unset_version_relationships, prepend: true
+
+  # NOTE: This block uses metaprogramming to create the equivalent of scopes that query the RepositoryObjectVersion table using only rows that are a `current` in the RepositoryObject table
+  #
+  # So it's a more extensible version of:
+  #
+  # scope :currently_in_virtual_objects, ->(member_druid) { joins(:current).merge(RepositoryObjectVersion.in_virtual_objects(member_druid)) }
+  # scope :currently_members_of_collection, ->(collection_druid) { joins(:current).merge(RepositoryObjectVersion.members_of_collection(collection_druid)) }
```
mjgiarlo, line 29: Here's what these look like under the hood:

```
3.2.2 :011 > RepositoryObject.currently_members_of_collection('collectiondruid1').explain
  RepositoryObject Load (2.9ms)  SELECT "repository_objects".* FROM "repository_objects" INNER JOIN "repository_object_versions" ON "repository_object_versions"."id" = "repository_objects"."current_id" WHERE (structural -> 'isMemberOf' ? 'collectiondruid1')
 => 
EXPLAIN SELECT "repository_objects".* FROM "repository_objects" INNER JOIN "repository_object_versions" ON "repository_object_versions"."id" = "repository_objects"."current_id" WHERE (structural -> 'isMemberOf' ? 'collectiondruid1')
                                      QUERY PLAN
---------------------------------------------------------------------------------------
 Hash Join  (cost=13.33..30.29 rows=5 width=120)
   Hash Cond: (repository_objects.current_id = repository_object_versions.id)
   ->  Seq Scan on repository_objects  (cost=0.00..15.50 rows=550 width=120)
   ->  Hash  (cost=13.30..13.30 rows=2 width=8)
         ->  Seq Scan on repository_object_versions  (cost=0.00..13.30 rows=2 width=8)
               Filter: ((structural -> 'isMemberOf'::text) ? 'collectiondruid1'::text)
(6 rows)


3.2.2 :012 > RepositoryObject.currently_in_virtual_objects('testdruid1').explain
  RepositoryObject Load (6.9ms)  SELECT "repository_objects".* FROM "repository_objects" INNER JOIN "repository_object_versions" ON "repository_object_versions"."id" = "repository_objects"."current_id" WHERE (structural #> '{hasMemberOrders,0}' -> 'members' ? 'testdruid1')
 => 
EXPLAIN SELECT "repository_objects".* FROM "repository_objects" INNER JOIN "repository_object_versions" ON "repository_object_versions"."id" = "repository_objects"."current_id" WHERE (structural #> '{hasMemberOrders,0}' -> 'members' ? 'testdruid1')
                                                   QUERY PLAN
-----------------------------------------------------------------------------------------------------------------
 Hash Join  (cost=13.88..30.84 rows=5 width=120)
   Hash Cond: (repository_objects.current_id = repository_object_versions.id)
   ->  Seq Scan on repository_objects  (cost=0.00..15.50 rows=550 width=120)
   ->  Hash  (cost=13.85..13.85 rows=2 width=8)
         ->  Seq Scan on repository_object_versions  (cost=0.00..13.85 rows=2 width=8)
               Filter: (((structural #> '{hasMemberOrders,0}'::text[]) -> 'members'::text) ? 'testdruid1'::text)
(6 rows)
```

app/models/repository_object.rb
```diff
@@ -0,0 +1,102 @@
...
+  class VersionNotOpened < StandardError; end
+
+  has_many :versions,  class_name: 'RepositoryObjectVersion', dependent: :destroy, inverse_of: 'repository_object'
+
+  belongs_to :current, class_name: 'RepositoryObjectVersion', optional: true
+  belongs_to :head, class_name: 'RepositoryObjectVersion', optional: true
+  belongs_to :open, class_name: 'RepositoryObjectVersion', optional: true
+
+  validates :external_identifier, :object_type, presence: true
+  validates :external_identifier, uniqueness: true
```
justinlittman, line 15: Can we just use a db uniqueness constraint instead of incurring the runtime overhead of AR uniqueness check?

app/models/repository_object.rb
```diff
@@ -0,0 +1,102 @@
...
+  validates :source_id, uniqueness: true, allow_nil: true
```
justinlittman, line 16: Same.

app/models/repository_object.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    def respond_to_missing?(method_name, include_private = false)
+      method_name.to_s.start_with?('currently_') || super
+    end
+  end
+
+  enum :object_type, %i[dro admin_policy collection].index_with(&:to_s)
+
+  def open_version
```
justinlittman, line 48: Bangs here and below? `open_version!`

app/models/repository_object.rb
```diff
@@ -0,0 +1,102 @@
...
+    raise VersionNotOpened, "Cannot close version because current version is closed: #{current.version}" if closed?
+
+    RepositoryObject.transaction do
+      version_to_close = self.open
+      update!(open: nil, head: version_to_close, current: version_to_close)
+    end
+  end
+
+  def open?
+    current.present? && current == open
```
justinlittman, line 67: When would current not be present?

app/models/repository_object.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+  end
+
+  enum :object_type, %i[dro admin_policy collection].index_with(&:to_s)
+
+  def open_version
+    raise VersionAlreadyOpened, "Cannot open new version because one is already open: #{current.version}" if open?
+
+    RepositoryObject.transaction do
+      version_to_open = latest_version.dup.tap { |object_version| object_version.version += 1 }
```
justinlittman, line 52: Isn't latest version just the head?

app/models/repository_object_version.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Models a repository object as it looked at a particular version.
+class RepositoryObjectVersion < ApplicationRecord
+  belongs_to :repository_object
+
+  scope :in_virtual_objects, ->(member_druid) { where("structural #> '{hasMemberOrders,0}' -> 'members' ? :druid", druid: member_druid) }
+  scope :members_of_collection, ->(collection_druid) { where("structural -> 'isMemberOf' ? :druid", druid: collection_druid) }
+
```
justinlittman, line 9: Should also validate presence of "description".

db/migrate/20240328141842_create_repository_objects.rb
```diff
@@ -0,0 +1,17 @@
+class CreateRepositoryObjects < ActiveRecord::Migration[7.1]
+  def change
+    create_enum :repository_object_type, %w[dro admin_policy collection]
+
+    create_table :repository_objects do |t|
+      t.enum :object_type, enum_type: :repository_object_type, null: false
+      t.string :external_identifier, null: false
+      t.string :source_id, null: true
```
justinlittman, line 8: Unique index?

db/migrate/20240328142339_create_repository_object_versions.rb
```diff
@@ -0,0 +1,21 @@
+class CreateRepositoryObjectVersions < ActiveRecord::Migration[7.1]
+  def change
+    create_table :repository_object_versions do |t|
+      t.references :repository_object, null: false, foreign_key: true
+      t.integer :version, null: false
+      t.string :version_description
```
justinlittman, line 6: null = false.

app/models/repository_object_version.rb
justinlittman: "Initial version" can be used as a default for first version. Otherwise, I think it will need to be provided when opening a new version.

db/migrate/20240328142339_create_repository_object_versions.rb
```diff
@@ -0,0 +1,21 @@
...
+      t.integer :cocina_version
+      t.string :content_type
+      t.string :label
+      t.jsonb :access
+      t.jsonb :administrative
+      t.jsonb :description
+      t.jsonb :identification
+      t.jsonb :structural
+      t.jsonb :geographic
+      t.datetime :closed_at
```
justinlittman, line 16: Can you set this when closing?

mjgiarlo (commented):
@justinlittman I think I got it all?

app/models/repository_object.rb
justinlittman: Yeah, but in a high concurrency situation like this, AR uniqueness might not be effective, so need to handle PG uniqueness violations anyway. Easier to just have one code path for handling like https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina_object_store.rb#L174

Plus, AR unique requires an extra query.

justinlittman (approved):

PR #4766, opened 2024-03-29
Refactors logic for getting release tags for items and collections. F…

### LINKED ISSUE #4764: ReleaseTags.collection_tags should use ReleaseTag model objects when present
See `collection.administrative.releaseTags` in https://github.com/sul-dlss/dor-services-app/blob/main/app/services/release_tags.rb#L66

Getting the release tags for a collection should use the same approach as an item (see ReleaseTags.item_tags).

### LINKED ISSUE #4765: Self release tags should trump collection release tags
ReleaseTags.for_public_metadata creates a map of release targets to release value (boolean indicating whether the item should be released).

The logic should be:
For each release target ("to" in the release tag): Use the most recent release value (based on "date" in the release tag) for the item's release tags. Otherwise, use the most recent release value for the collection's collection release tags where "what"="collection".

See https://github.com/sul-dlss/dor-services-app/commit/5f2be17c6ab7410bd284ea2ce232297137e4df6b#diff-b04d1159cdb076ecce97b78ef87ff05e8c60d34d93938f460af4e5bf88d2c2a2 for when this was broken.

justinlittman:
…ixes logic for getting release tags for public metadata.

closes #4764
closes #4765
:

Fixes for release tags.

Unit

### DIFF
--- a/app/services/registration_csv_converter.rb
+++ b/app/services/registration_csv_converter.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'csv'
+
 # Convert CSV to JSON for registration
 class RegistrationCsvConverter
   include Dry::Monads[:result]
--- a/app/services/release_tags.rb
+++ b/app/services/release_tags.rb
@@ -44,31 +44,49 @@ class ReleaseTags
   attr_reader :cocina_object
 
   # Determine projects in which an item is released
-  # @return [Hash{String => Boolean}] all namespaces, keys are Project name Strings, values are Boolean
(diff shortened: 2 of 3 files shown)


app/services/registration_csv_converter.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'csv'
```
justinlittman, line 3: I don't know why this is necessary, but test failing without.

lwrubel (approved):
Approved, but noting the `require 'csv'` requirement is weird.

PR #4799, opened 2024-04-08
Add scopes to RepositoryObject for different object types

### LINKED ISSUE #4776: Add dros, admin_policies, collections scopes to RepositoryObject

jcoyne:

Fixes #4776 


jcoyne:
It seems more appropriate to use STI for this.

justinlittman:
I'm expecting direct use of the Active Record objects to be very, very limited. It that proves not to be the case, agree that should consider STI.

mjgiarlo (approved):

PR #4803, opened 2024-04-08
Destroy the RepositoryObject when calling CocinaObjectStore#destroy

### LINKED ISSUE #4796: Implement CocinaObjectStore destroy
In addition to deleting Dro / AdminPolicy / Collection, also delete RepositoryObject (and associated RepositoryObjectVersions).

Please test that a RepositoryObject with RepositoryObjectVersions can be deleted, as there are foreign keys in both directions.

jcoyne:

Fixes #4796 

### DIFF
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -142,7 +142,12 @@ class CocinaObjectStore
   def destroy(druid)
     cocina_object = CocinaObjectStore.find(druid)
 
-    ar_destroy(druid)
+    RepositoryObject.transaction do
+      # TODO: After migrating to RepositoryObjects, we can get rid of the nil check and use:
+      #   RepositoryObject.find_by!(external_identifier: druid).destroy
+      RepositoryObject.find_by(external_identifier: druid)&.destroy
+      ar_destroy(druid)
+    end
 
     Notifications::ObjectDeleted.publish(model: cocina_object, deleted_at: Time.zone.now)
   end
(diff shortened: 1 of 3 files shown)


app/services/cocina_object_store.rb
```diff
@@ -142,7 +142,10 @@ def version(druid)
   def destroy(druid)
     cocina_object = CocinaObjectStore.find(druid)
 
-    ar_destroy(druid)
+    RepositoryObject.transaction do
+      RepositoryObject.find_by!(external_identifier: druid).destroy
```
mjgiarlo, line 146: Does this imply that we won't be able to destroy a druid until it's been migrated into the new table?

justinlittman: Yes, it does imply that. It should only delete the RepositoryObject when it exists.

justinlittman (changes requested):

spec/services/cocina_object_store_spec.rb
```diff
@@ -187,16 +187,32 @@
...
     end
 
-    it 'destroys the object' do
-      described_class.destroy(dro.external_identifier)
-      expect(Dro).not_to exist(dro.external_identifier)
-      expect(Notifications::ObjectDeleted).to have_received(:publish)
+    context 'when a repository object does not exist' do
+      it 'destroys the object' do
+        described_class.destroy(dro.external_identifier)
+        expect(Dro).not_to exist(dro.external_identifier)
```
mjgiarlo, line 199: @jcoyne Should this and line 213 change to use the DB ID instead of the druid per https://github.com/sul-dlss/dor-services-app/pull/4802/files#diff-247bc7342b1377c6a8867011a39120dc287983499a26459b1cf1779782cc1678R198 ?

jcoyne: Yes, I got this when I rebased.

mjgiarlo (approved):

PR #4808, opened 2024-04-08
Implement CocinaObjectStore#exists? with RepositoryObject model

### LINKED ISSUE #4795: Implement CocinaObjectStore exists methods
For each of the CocinaObjectStore find methods (exists?, exists!), implement such that when enabled_features.repository_object is enabled, the RepositoryObject will be queried instead of the Dro / AdminPolicy / Collection.

jcoyne:

Fixes #4795 

### DIFF
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -113,6 +113,8 @@ class CocinaObjectStore
   end
 
   def exists?(druid, type: nil)
+    return RepositoryObject.exists?(external_identifier: druid) if Settings.enabled_features.repository_object
+
     ar_exists?(druid, type:)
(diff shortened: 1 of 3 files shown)


app/services/cocina_object_store.rb
```diff
@@ -122,14 +124,6 @@ def exists!(druid)
     raise CocinaObjectNotFoundError.new("Couldn't find object with 'external_identifier'=#{druid}", druid)
   end
 
-  # This is only public for migration use.
```
jcoyne, line 125: Moved this to private since we no longer have a migration use.

mjgiarlo (approved):

PR #4809, opened 2024-04-08
Update repository object source_id on head version update

### LINKED ISSUE #4790: Update RepositoryObject source_id.
Update RepositoryObject source_id when updated in **current** RepositoryObjectVersion’s administrative cocina.

Callback?

lwrubel:
Resolves #4790. 

Unit

### DIFF
--- a/app/models/repository_object_version.rb
+++ b/app/models/repository_object_version.rb
@@ -8,6 +8,7 @@ class RepositoryObjectVersion < ApplicationRecord
   scope :members_of_collection, ->(collection_druid) { where("structural -> 'isMemberOf' ? :druid", druid: collection_druid) }
 
   validates :version, :version_description, presence: true
+  after_save :update_object_source_id
 
   # @param [Cocina::Models::DRO, Cocina::Models::Collection, Cocina::Models::AdminPolicy] cocina_object a Cocina
   #   model instance, either a DRO, collection, or APO.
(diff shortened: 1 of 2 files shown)


app/models/repository_object_version.rb
```diff
@@ -8,4 +8,13 @@ class RepositoryObjectVersion < ApplicationRecord
   scope :members_of_collection, ->(collection_druid) { where("structural -> 'isMemberOf' ? :druid", druid: collection_druid) }
 
   validates :version, :version_description, presence: true
+  after_save :update_object_source_id
+
+  private
+
+  def update_object_source_id
+    return unless repository_object.head_version == self
```
mjgiarlo, line 16: Does this allow us to make updates to the last closed version, since head can either be the last closed or the currently open version? I wonder if we should only proceed if `self` is the `opened_version`?

app/models/repository_object_version.rb
```diff
@@ -8,4 +8,13 @@ class RepositoryObjectVersion < ApplicationRecord
...
+
+    repository_object.update!(source_id: identification['sourceId']) if identification&.fetch('sourceId', nil)
```
mjgiarlo, line 18: Should we also consider adding a guard clause above this that returns early if the source ID of the version already matches `repository_object.source_id`?

app/models/repository_object_version.rb
mjgiarlo: That's exactly what I had in mind, but explained with more clarity than I gave!

lwrubel: Thanks, made a change to add this.

app/models/repository_object_version.rb
lwrubel: Done.

mjgiarlo (approved):

PR #4810, opened 2024-04-08
Update the opened repository object version with values from Cocina instance

### LINKED ISSUE #4789: Add from_cocina, to_model_hash, update_cocina methods to RepositoryObjectVersion
See https://github.com/sul-dlss/dor-services-app/blob/main/app/models/dro.rb#L39-L64

Note that will need to handle DROs, Collections, and AdminPolicies.

Also note that `update_cocina` will always be an update so the existing upsert logic isn't necessary.

mjgiarlo:
# Why was this change made?

Fixes #4789

# How was this change tested?

CI

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -53,6 +53,12 @@ class RepositoryObject < ApplicationRecord
     end
   end
 
+  # @param [Cocina::Models::DRO, Cocina::Models::Collection, Cocina::Models::AdminPolicy] cocina_object a Cocina
+  #   model instance, either a DRO, collection, or APO.
+  def update_opened_version_from(cocina_object:)
+    opened_version.update!(**RepositoryObjectVersion.to_model_hash(cocina_object))
+  end
+
   def open_version!
     raise VersionAlreadyOpened, "Cannot open new version because one is already open: #{head_version.version}" if open?
 
--- a/app/models/repository_object_version.rb
+++ b/app/models/repository_object_version.rb
@@ -8,4 +8,25 @@ class RepositoryObjectVersion < ApplicationRecord
   scope :members_of_collection, ->(collection_druid) { where("structural -> 'isMemberOf' ? :druid", druid: collection_druid) }
 
   validates :version, :version_description, presence: true
+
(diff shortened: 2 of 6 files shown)


app/models/repository_object.rb
```diff
@@ -53,6 +53,12 @@ def current_scope_prefix
     end
   end
 
+  # @param [Cocina::Models::DRO, Cocina::Models::Collection, Cocina::Models::AdminPolicy] cocina_object a Cocina
+  #   model instance, either a DRO, collection, or APO.
+  def update_opened_version_from(cocina_object:)
```
mjgiarlo, line 58: I called an audible here on the name and opted for something a little more explanatory, namely that it's the opened version that will be updated.

app/models/repository_object_version.rb
```diff
@@ -8,4 +8,25 @@ class RepositoryObjectVersion < ApplicationRecord
...
+      if cocina_object.dro?
+        object_hash[:content_type] = object_hash.delete(:type)
+        object_hash[:geographic] ||= nil
+      elsif cocina_object.collection?
+        object_hash[:content_type] = object_hash.delete(:type)
+      elsif cocina_object.admin_policy?
+        object_hash.delete(:type)
+        object_hash[:description] ||= nil
+      end
+    end
```
mjgiarlo, line 30: I also thought that this method made more sense on the `RepositoryObjectVersion` class, which does not care about the external identifier/druid (line 18), but I don't think it matters too much since it's a class method (one that doesn't care about DB-persisted state).

db/migrate/20240408230311_change_repository_object_version_cocina_version_to_string.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeRepositoryObjectVersionCocinaVersionToString < ActiveRecord::Migration[7.1]
+  def change
+    change_column :repository_object_versions, :cocina_version, :string
```
mjgiarlo, line 3: This was initially set up as an int. I hadn't notice this before. Yay for testing!

justinlittman (approved):

PR #4821, opened 2024-04-09
Open RepositoryObject in VersionService.open

### LINKED ISSUE #4814: Open RepositoryObject in VersionService.open
If a RepositoryObject exists, it should be opened (new RepositoryObjectVersion created with incremented version, cocina copied from last closed version, provided description set). This should be in addition to existing work done in VersionService.open.

Mutating the RepositoryObject should be encapsulated in the CocinaObjectStore. (That is, RepositoryObjects should not be manipulated in VersionService code.)

Please leave a note indicating that the "if a RepositoryObject exists" should be removed. See https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina_object_store.rb#L140-L141

jcoyne:

Fixes #4814

### DIFF
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -70,6 +70,13 @@ class VersionService
       # This is for testing when we don't have the SDR container available
       new_object_version = ObjectVersion.increment_version(druid:, description:)
     end
+
+    # TODO: After migrating to RepositoryObjects, we can get rid of the nil check and use:
+    #   RepositoryObject.find_by!(external_identifier: druid).open_version!
+    RepositoryObject.find_by(external_identifier: druid)&.open_version!
+    # TODO: when we stop calling the UpdateObjectService, after we've migrated to RepostoryObjects, we may need to trigger indexing:
+    #       e.g.: Notifications::ObjectUpdated.publish(model: cocina_object_with_metadata)
+
     update_cocina_object = cocina_object
     update_cocina_object = UpdateObjectService.update(cocina_object.new(version: new_object_version.version)) if cocina_object.version != new_object_version.version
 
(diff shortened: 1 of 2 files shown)


spec/services/version_service_spec.rb
```diff
@@ -23,8 +23,16 @@
     let(:workflow_client) do
       instance_double(Dor::Workflow::Client, create_workflow_by_name: true)
     end
+    # Doing build(:repository_object) rather than create to support test where it doesn't exist.
+    # This case can be removed after we fully migrate to RepositoryObjects
+    let(:repository_object) { build(:repository_object, external_identifier: druid) }
 
     before do
+      # This check be removed after we fully migrate to RepositoryObjects
```
justinlittman, line 31: Can you prefix with TODO so we can find it easier later?

app/services/version_service.rb
```diff
@@ -70,6 +70,13 @@ def open(cocina_object:, description:, opening_user_name:, assume_accessioned:,
       # This is for testing when we don't have the SDR container available
       new_object_version = ObjectVersion.increment_version(druid:, description:)
     end
+
+    # TODO: After migrating to RepositoryObjects, we can get rid of the nil check and use:
+    #   RepositoryObject.find_by!(external_identifier: druid).open_version!
+    RepositoryObject.find_by(external_identifier: druid)&.open_version!
+    # TODO: when we stop calling the UpdateObjectService, after we've migrated to RepostoryObjects, we may need to trigger indexing:
```
justinlittman, line 77: I don't think we'll want to stop calling UpdateObjectService, as that service handles all of the activities to update an object, including publishing messages.

justinlittman (approved):

app/services/version_service.rb
jcoyne: The old way of incrementing a version was to persist with an update. Now, open_object! does the persistence. I don't think we want a duplicate update message. Just the reindex should be sufficient.

PR #4822, opened 2024-04-09
Add to_cocina, to_cocina_with_metadata, and to_h to RepositoryObjectVersion

### LINKED ISSUE #4788: Add to_cocina, to_h methods to RepositoryObjectVersion
Also, add to_cocina to RepositoryObject that delegates to current's to_cocina.

See https://github.com/sul-dlss/dor-services-app/blob/main/app/models/dro.rb#L17-L37

Note that will need to handle DROs, Collections, and AdminPolicies.

aaron-collier:

Closes #4788 

### DIFF
--- a/app/models/repository_object_version.rb
+++ b/app/models/repository_object_version.rb
@@ -31,7 +31,28 @@ class RepositoryObjectVersion < ApplicationRecord
     end
   end
 
-  private
+  def external_lock
+    # This should be opaque, but this makes troubeshooting easier.
+    # The external_identifier is included so that there is enough entropy such
+    # that the lock can't be used for an object it doesn't belong to as the
+    # lock column is just an integer sequence.
+    [repository_object.external_identifier, repository_object.lock.to_s].join('=')
+  end
+
+  def to_cocina_with_metadata
+    Cocina::Models.with_metadata(to_cocina, external_lock, created: created_at.utc, modified: updated_at.utc)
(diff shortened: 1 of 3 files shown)


app/models/repository_object_version.rb
```diff
@@ -31,6 +31,45 @@ def self.to_model_hash(cocina_object)
...
+    # lock column is just an integer sequence.
+    [repository_object.external_identifier, repository_object.lock.to_s].join('=')
+  end
+
+  def to_cocina_with_metadata
+    Cocina::Models.with_metadata(to_cocina, external_lock, created: created_at.utc, modified: updated_at.utc)
+  end
+
+  def to_cocina
+    case repository_object.object_type
```
justinlittman, line 47: Will Cocina::Models.build (https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models.rb#L83) work here?

justinlittman (changes requested):

app/models/repository_object_version.rb
aaron-collier: @justinlittman good catch, it does!

mjgiarlo (approved):

mjgiarlo:
@justinlittman Leaving for you to merge since you have a prior review. Consider squash-merging?

justinlittman (approved):

PR #4831, opened 2024-04-10
Add migrator to remove release tags from cocina objects

### LINKED ISSUE #4811: Write migrator to remove release tags field from cocina
This should use the existing migration framework: https://github.com/sul-dlss/dor-services-app/blob/main/bin/migrate-cocina

This will be executed before releaseTags is removed from cocina model.

jcoyne:

Fixes #4811 

### DIFF
--- a/spec/services/migrators/remove_release_tags_spec.rb
+++ b/spec/services/migrators/remove_release_tags_spec.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Migrators::RemoveReleaseTags do
+  subject(:migrator) { described_class.new(ar_cocina_object) }
+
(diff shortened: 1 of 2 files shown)


spec/services/migrators/remove_release_tags_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      let(:ar_cocina_object) { create(:ar_admin_policy) }
+
+      it { is_expected.to be false }
+    end
+  end
+
+  describe 'migrate' do
+    let(:ar_cocina_object) { create(:ar_dro, administrative: { hasAdminPolicy: 'druid:hy787xj5878', releaseTags: [] }) }
+
+    it 'updates label and title' do
```
mjgiarlo, line 35: Copypasta description?

justinlittman (approved):

PR #4834, opened 2024-04-10
Find objects persisted as RepositoryObjects

### LINKED ISSUE #4794: Find RepositoryObjects in CocinaObjectStore find methods
For each of the CocinaObjectStore find methods (find, find_by_source_id), implement such that:
* When enabled_features.repository_object_test is enabled and a RepositoryObject exists, the cocina from the RepositoryObject and the Dro / AdminPolicy / Collection active record models will be compared. If there is a discrepancy, Honeybadger will be alerted and the Dro / AdminPolicy / Collection will be returned.
* When a RepositoryObject exists, the cocina model from the RepositoryObject will be returned; otherwise, the cocina model from the Dro / AdminPolicy / Collection will be returned

jcoyne:
Fixes #4794

### DIFF
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -97,7 +97,9 @@ class CocinaObjectStore
   end
 
   def find_by_source_id(source_id)
-    ar_cocina_object = Dro.find_by_source_id(source_id) ||
+    # TODO: Nil check can be removed after migrating to RepositoryObject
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):

justinlittman (changes requested):
Per ticket, please also handle find_by_source_id method.

justinlittman (approved):

PR #4843, opened 2024-04-11
Add repository_object_test feature to CocinaObjectStore#find and #find_by_source_id

jcoyne:

Ref #4794


.solargraph.yml
```diff
@@ -0,0 +1,23 @@
+---
```
justinlittman, line 1: Intentional?

justinlittman (approved):
Approve pending confirmation that you intended to commit .solargraph.yml.

.solargraph.yml
jcoyne: Naw. I'll remove the file.

mjgiarlo (approved):

PR #4852, opened 2024-04-11
Add tests for full coverage of PublicCocinaService

jcoyne:

we were missing tests for published files and for a collection.

### DIFF
--- a/spec/services/publish/public_cocina_service_spec.rb
+++ b/spec/services/publish/public_cocina_service_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Publish::PublicCocinaService do
   subject(:json) { JSON.parse(create.to_json) }
 
   let(:create) { described_class.create(cocina_object) }
-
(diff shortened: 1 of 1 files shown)


spec/services/publish/public_cocina_service_spec.rb
```diff
@@ -3,149 +3,168 @@
...
 
 RSpec.describe Publish::PublicCocinaService do
-  subject(:json) { JSON.parse(create.to_json) }
+  subject(:create) { described_class.create(cocina_object) }
 
-  let(:create) { described_class.create(cocina_object) }
+  context 'with a Collection' do
+    let(:cocina_object) { build(:collection) }
 
-  let(:cocina_object) do
```
jcoyne, line 10: This is mostly an indenting change. I recommend reviewing by hiding whitespace differences.

mjgiarlo (approved):

PR #4854, opened 2024-04-11
Update repository object in UpdateObjectService

### LINKED ISSUE #4838: Update RepositoryObject is UpdateObjectService
When a RepositoryObject exists for the object, update it. This is in addition to updating the Dro/Collection/AdminPolicy via CocinaObjectStore.store.

There is NO feature flag for this operation.

jcoyne:

Fixes #4838

### DIFF
--- a/app/services/update_object_service.rb
+++ b/app/services/update_object_service.rb
@@ -35,6 +35,9 @@ class UpdateObjectService
 
     cocina_object_without_metadata = Cocina::Models.without_metadata(cocina_object)
 
+    # TODO: after migration to RepositoryObject we can remove these nil checks
+    RepositoryObject.find_by(external_identifier: cocina_object.externalIdentifier)&.head_version&.update!(RepositoryObjectVersion.to_model_hash(cocina_object_without_metadata))
+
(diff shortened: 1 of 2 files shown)


app/services/update_object_service.rb
```diff
@@ -35,6 +35,9 @@ def update
 
     cocina_object_without_metadata = Cocina::Models.without_metadata(cocina_object)
 
+    # TODO: after migration to RepositoryObject we can remove these nil checks
+    RepositoryObject.find_by(external_identifier: cocina_object.externalIdentifier)&.head_version&.update!(RepositoryObjectVersion.to_model_hash(cocina_object_without_metadata))
+
```
mjgiarlo, line 40: Do we always want to patch the updates into the head version? I.e., what if the head version is already closed? @jcoyne @justinlittman

mjgiarlo: Sounds good to me.

justinlittman: Should not automatically open versions.

justinlittman: Also, update service will be verifying that object is open.

justinlittman (approved):

PR #4855, opened 2024-04-11
Ignore last steps when determining if assembling.

### LINKED ISSUE #4851: Version can't be closed in assembly workflows
Currently the last step in assembly workflows is to close a version. However, versions can't be closed while an assembly workflow is running.

This is revealed by integration tests, e.g., preassembly_hfs_accessioning_spec.

To address this VersionService should check to make sure that all workflow steps are completed EXCEPT the last step.

justinlittman:
closes #4851

So that an assembly workflow can close a version as its last step.

Unit, integration


app/services/workflow_state_service.rb
```diff
@@ -79,4 +86,16 @@ def active_workflow?(workflow:)
...
   end
+
+  def active_workflow_except_step?(workflow:, process:)
+    workflow_response = workflow_client.workflow(pid: druid, workflow_name: workflow)
+
+    # Is there a workflow for the current version?
+    return false unless workflow_response.active_for?(version:)
+
+    incomplete_processes = workflow_response.incomplete_processes_for(version:)
+    return false if incomplete_processes.size == 1 && incomplete_processes.first.name == process
```
mjgiarlo, line 97: So this method only returns `true` if:

1. The specified workflow is active for the specified version/druid; **and either**
2. The number of incomplete processes is 0 or greater than 1; **or**
3. The first incomplete process is named the specified process

❓

justinlittman: Yes, but let me add some more code comments.

app/services/workflow_state_service.rb
```diff
@@ -73,10 +80,24 @@ def active_workflow?(workflow:)
...
+    workflow_response = workflow_client.workflow(pid: druid, workflow_name: workflow)
+
+    # Is there a workflow for the current version? If not, then it can't be active.
+    return false unless workflow_response.active_for?(version:)
+
+    incomplete_processes = workflow_response.incomplete_processes_for(version:)
+    # There is a workflow whose only incomplete process is the one that is excluded, so considering not active.
+    return false if incomplete_processes.size == 1 && incomplete_processes.first.name == process
+
+    # There is a workflow that has > 1 incomplete processes or the 1 incomplete process is not the excluded process.
```
mjgiarlo, line 100: Is it possible for us to get to this point if `incomplete_processes.size` is 0?

mjgiarlo (approved):

PR #4856, opened 2024-04-11
Remove stubbing from version service spec

jcoyne:

It's helpful to see what is actually persisted here


spec/requests/versions_spec.rb
```diff
@@ -73,7 +73,7 @@
     end
   end
 
-  describe '/versions' do
+  describe 'POST /v1/objects/:druid/versions' do
```
mjgiarlo, line 76: 🥇

mjgiarlo (approved):

PR #4859, opened 2024-04-11
Create RepositoryObjects in CreateObjectService

### LINKED ISSUE #4837: Create RepositoryObject in CreateObjectService
When the enabled_features.repository_object_create flag is set, create and persist a RepositoryObject in CreateObjectService. This is in addition to creating a Dro / Collection / AdminPolicy via CocinaObjectStore.store.

jcoyne:

Fixes #4837

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -52,6 +52,17 @@ class RepositoryObject < ApplicationRecord
       method_name.to_s.start_with?(current_scope_prefix) || super
     end
 
+    # @param [Cocina::Models::DRO, Cocina::Models::Collection, Cocina::Models::AdminPolicy] cocina_object a Cocina
+    #   model instance, either a DRO, collection, or APO.
+    def create_from(cocina_object:)
+      args = {
+        external_identifier: cocina_object.externalIdentifier,
+        object_type: cocina_object.class.name.demodulize.underscore
+      }
+      args[:source_id] = cocina_object.identification.sourceId if cocina_object.respond_to?(:identification)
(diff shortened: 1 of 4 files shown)


app/models/repository_object.rb
```diff
@@ -52,6 +52,15 @@ def respond_to_missing?(method_name, include_private = false)
       method_name.to_s.start_with?(current_scope_prefix) || super
     end
 
+    # @param [Cocina::Models::DRO, Cocina::Models::Collection, Cocina::Models::AdminPolicy] cocina_object a Cocina
+    #   model instance, either a DRO, collection, or APO.
+    def create_from(cocina_object:)
+      external_identifier = cocina_object.externalIdentifier
+      object_type = cocina_object.class.name.demodulize.underscore
```
mjgiarlo, line 59: `demodulize` is the method I couldn't remember when I wrote this: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/repository_object_migrator.rb#L65

app/models/repository_object.rb
```diff
@@ -52,6 +52,15 @@ def respond_to_missing?(method_name, include_private = false)
...
+      source_id = cocina_object.identification.sourceId
```
mjgiarlo, line 60: I think this will blow up if `cocina_object` is an APO.

jcoyne: In fact we have tests that confirm your suspicions. 🎉

mjgiarlo (approved):

PR #4877, opened 2024-04-16
Fix repository object test

justinlittman:
So that integration tests pass. See individual comments.

Unit, integration.

### DIFF
--- a/app/models/repository_object_version.rb
+++ b/app/models/repository_object_version.rb
@@ -51,6 +51,7 @@ class RepositoryObjectVersion < ApplicationRecord
   # @return [Hash] RepositoryObjectVersion instance as a hash
   def to_h
     {
+      cocinaVersion: cocina_version,
       type: content_type,
       externalIdentifier: repository_object.external_identifier,
       label:,
--- a/app/services/cocina_object_store.rb
+++ b/app/services/cocina_object_store.rb
@@ -114,7 +114,7 @@ class CocinaObjectStore
       return cocina unless Settings.enabled_features.repository_object_test
 
       legacy = (Dro.find_by_source_id(source_id) || Collection.find_by_source_id(source_id)).to_cocina_with_metadata
-      return cocina if legacy == cocina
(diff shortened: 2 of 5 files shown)


app/models/repository_object_version.rb
```diff
@@ -51,6 +51,7 @@ def to_cocina
   # @return [Hash] RepositoryObjectVersion instance as a hash
   def to_h
     {
+      cocinaVersion: cocina_version,
```
justinlittman, line 54: cocina version wasn't being copied to cocina from the RepositoryObjectVersion. This was causing equality tests to fail.

app/services/cocina_object_store.rb
```diff
@@ -114,7 +114,7 @@ def find_by_source_id(source_id)
       return cocina unless Settings.enabled_features.repository_object_test
 
       legacy = (Dro.find_by_source_id(source_id) || Collection.find_by_source_id(source_id)).to_cocina_with_metadata
-      return cocina if legacy == cocina
+      return cocina if Cocina::Models.without_metadata(legacy) == Cocina::Models.without_metadata(cocina)
```
justinlittman, line 117: Need to compare without_metadata since metadata contains update timestamps which are slightly different.

app/services/version_service.rb
```diff
@@ -79,7 +79,9 @@ def open(cocina_object:, description:, opening_user_name:, assume_accessioned:,
 
     # TODO: After migrating to RepositoryObjects, we can get rid of the nil check and use:
     #   RepositoryObject.find_by!(external_identifier: druid).open_version!
-    RepositoryObject.find_by(external_identifier: druid)&.open_version!
+    repo_obj = RepositoryObject.find_by(external_identifier: druid)
+    repo_obj = RepositoryObjectMigrator.migrate(external_identifier: druid) if repo_obj.nil? && Settings.enabled_features.repository_object_create
```
justinlittman, line 83: Perform opportunistic migration here rather than in update object service below so that object is correctly opened.

PR #4878, opened 2024-04-16
Adds transactions around RepositoryObject creation.

justinlittman:
To avoid race conditions that were causing RepositoryObject / legacy object comparisons to fail.

integration

### DIFF
--- a/app/services/update_object_service.rb
+++ b/app/services/update_object_service.rb
@@ -30,19 +30,22 @@ class UpdateObjectService
     Cocina::ObjectValidator.validate(cocina_object)
     notify_unless_open_version
 
+    cocina_object_without_metadata = Cocina::Models.without_metadata(cocina_object)
+
     # If this is a collection and the title has changed, then reindex the children.
     update_items = need_to_update_members?
 
     # Only update if already exists in PG (i.e., added by create or migration).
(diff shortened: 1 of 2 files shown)


app/services/update_object_service.rb
```diff
@@ -30,19 +30,22 @@ def update
...
-      RepositoryObjectMigrator.migrate(external_identifier: druid)
+    cocina_object_with_metadata = nil
+    RepositoryObject.transaction do
+      cocina_object_with_metadata = CocinaObjectStore.store(cocina_object, skip_lock:)
+
+      repo_object = RepositoryObject.find_by(external_identifier: druid)
+      if repo_object
+        repo_object.update_opened_version_from(cocina_object: cocina_object_without_metadata)
+      elsif Settings.enabled_features.repository_object_create
+        RepositoryObjectMigrator.migrate(external_identifier: druid)
```
jcoyne, line 47: Should we remove the existing transaction from the RepositoryObjectMigrator?

justinlittman: I don't think so, as the RepositoryObjectMigrator will be used in other non-transactional contexts.

jcoyne (approved):

PR #4888, opened 2024-04-17
Reindex locally instead of sending object update message.

### LINKED ISSUE #4885: Index after AdministrativeTags create/update/destroy
In here:
https://github.com/sul-dlss/dor-services-app/blob/e4b9d96d64ee0981745a5614f0dffc8cac45f49f/app/controllers/administrative_tags_controller.rb#L21-L66

Rather than sending `Notifications::ObjectUpdated`, call the indexer in DSA, but in a Synchronous manner.

This will allow Argo to stop calling the indexer:
https://github.com/sul-dlss/argo/blob/84743b4ceb85a55356f7f2323faa72983f06ae66/app/controllers/tags_controller.rb#L64
https://github.com/sul-dlss/argo/blob/84743b4ceb85a55356f7f2323faa72983f06ae66/app/jobs/import_tags_job.rb#L38

justinlittman:
closes #4885

Local indexing is better than remote.

Unit, integration test


app/services/indexer.rb
```diff
@@ -16,6 +16,15 @@ def self.reindex(cocina_object:)
...
   end
 
+  # @param [Cocina::Models::DROWithMetadata|CollectionWithMetadata|AdminPolicyWithMetadata]
+  def self.reindex_later(cocina_object:)
+    ReindexJob.perform_later(
+      model: cocina_object.to_h,
+      created: cocina_object.created,
+      modified: cocina_object.modified
+    )
+  end
```
mjgiarlo, line 26: It struck me as interesting that this method calls a job that calls back into this same service class. Can you say a bit about why you created this seam over having collaborators directly invoke `ReindexJob.perform_later`?

mjgiarlo (approved):
Looking over our [messaging documentation](https://github.com/sul-dlss/infrastructure-team/blob/main/sdr/messages.md), the only consumer of messages to the `sdr.objects.updated` topic is DIA, so this seems like a safe change.

I have a question about the service class design, but I don't consider it to be a blocker.

app/services/indexer.rb
justinlittman: I don't feel strongly about this choice, but I thought a future developer would be more likely to find the Indexer service, but might not be aware that there was a job available as well.

mjgiarlo: K. I don't think it's a big deal, either way. If we decide we feel differently about it in the future, it should be a quick refactor.

PR #4902, opened 2024-04-19
Update version description when opening and closing.

### LINKED ISSUE #4901: RepositoryObjectVersion description should be set
From VersionService.open and .close, the RepositoryObjectVersion description should be set.

justinlittman:
closes #4901

Version descriptions are supposed to be updated.

Unit

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -76,20 +76,27 @@ class RepositoryObject < ApplicationRecord
     opened_version.update!(**RepositoryObjectVersion.to_model_hash(cocina_object))
   end
 
-  def open_version!
+  def open_version!(description:)
     raise VersionAlreadyOpened, "Cannot open new version because one is already open: #{head_version.version}" if open?
 
     RepositoryObject.transaction do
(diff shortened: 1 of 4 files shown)


jcoyne (approved):

app/models/repository_object.rb
```diff
@@ -76,20 +76,27 @@ def update_opened_version_from(cocina_object:)
...
+  def close_version!(description: nil)
     raise VersionNotOpened, "Cannot close version because head version is closed: #{head_version.version}" if closed?
 
     RepositoryObject.transaction do
-      version_to_close = opened_version.tap { |object_version| object_version.closed_at = Time.current }
+      version_to_close = opened_version.tap do |object_version|
+        object_version.closed_at = Time.current
+        object_version.version_description = description if description
+      end
+      version_to_close.save!
```
mjgiarlo, line 99: @justinlittman Did you find you needed to add this? I don't think we needed it before, and we don't have an extra `#save!` up in `#open_version!`.

justinlittman: I was surprised as well for both of the reasons that you state. However, the tests verified that it was necessary here and not in open_version!

PR #4906, opened 2024-04-19
Return version content from RepositoryObject if it exists

### LINKED ISSUE #4903: Return version data from RepositoryObjectVersions
In Versions index endpoint (https://github.com/sul-dlss/dor-services-app/blob/main/app/controllers/versions_controller.rb#L8-L22), return version data from RepositoryObjectVersions when RepositoryObject is present, otherwise continue to return from ObjectVersions.

aaron-collier:

Fixes #4903 

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -6,9 +6,15 @@ class VersionsController < ApplicationController
   before_action :load_version, only: %i[current close_current openable status]
 
   def index
+    repository_object = RepositoryObject.find_by(external_identifier: params[:object_id])
     object_versions = ObjectVersion.where(druid: params[:object_id])
 
-    return render json: {} if object_versions.empty?
+    # return an empty JSON object if there are no versions and no repository object
+    return render json: {} if object_versions.empty? && repository_object.nil?
+
+    # add an entry with version id and description for each RepositoryObjectVersion
+
+    return render json: { versions: repository_object_version_content(repository_object.versions) } if repository_object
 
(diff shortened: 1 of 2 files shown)


app/controllers/versions_controller.rb
```diff
@@ -6,9 +6,15 @@ class VersionsController < ApplicationController
...
+    repository_object = RepositoryObject.find_by(external_identifier: params[:object_id])
     object_versions = ObjectVersion.where(druid: params[:object_id])
 
-    return render json: {} if object_versions.empty?
+    # return an empty JSON object if there are no versions and no repository object
+    return render json: {} if object_versions.empty? && repository_object.nil?
+
+    # add an entry with version id and description for each RepositoryObjectVersion
+
+    return render json: { versions: repository_object_version_content(repository_object.versions) } unless repository_object.nil?
```
justinlittman, line 17: I'd suggest `if repository_object` to avoid double negatives.

spec/requests/versions_inventory_spec.rb
```diff
@@ -5,21 +5,47 @@
...
+    let(:attrs) do
+      {
+        external_identifier: druid,
+        object_type: 'dro',
+        source_id: 'sul:dlss:testing'
+      }
+    end
+
+    before do
+      repository_object.save # we need at least one persisted version so we can run this test
```
justinlittman, line 40: Can you add one more version to the test please?

justinlittman (approved):

justinlittman:
I tested in stage.

PR #4910, opened 2024-04-19
Fix versioning tests and make repository object internally consistent

mjgiarlo:
# Why was this change made?

To fix the build and make a model more consistent.

# How was this change tested?

CI

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -80,11 +80,9 @@ class RepositoryObject < ApplicationRecord
     raise VersionAlreadyOpened, "Cannot open new version because one is already open: #{head_version.version}" if open?
 
     RepositoryObject.transaction do
-      version_to_open = last_closed_version.dup.tap do |object_version|
-        object_version.version += 1
-        object_version.version_description = description
-      end
-      update!(opened_version: version_to_open, head_version: version_to_open)
+      new_version = last_closed_version.dup
+      new_version.update!(version: new_version.version + 1, version_description: description)
+      update!(opened_version: new_version, head_version: new_version)
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/repository_object.rb
```diff
@@ -80,24 +80,18 @@ def open_version!(description:)
...
 
     RepositoryObject.transaction do
-      version_to_close = opened_version.tap do |object_version|
-        object_version.closed_at = Time.current
-        object_version.version_description = description if description
-      end
-      version_to_close.save!
-      update!(opened_version: nil, last_closed_version: version_to_close, head_version: version_to_close)
+      opened_version.update!(closed_at: Time.current, version_description: description || opened_version.version_description)
+      update!(opened_version: nil, last_closed_version: opened_version, head_version: opened_version)
```
mjgiarlo, line 94: Done in the name of consistency. Also perhaps the direct updates over the taps may be more readable?

spec/services/version_service_spec.rb
```diff
@@ -11,6 +11,7 @@
 
   before do
     allow(WorkflowStateService).to receive(:new).and_return(workflow_state_service)
+    allow(Indexer).to receive(:reindex_later)
```
mjgiarlo, line 14: This is what was breaking the tests.

spec/services/version_service_spec.rb
```diff
@@ -19,16 +20,9 @@
     let(:workflow_client) do
       instance_double(Dor::Workflow::Client, create_workflow_by_name: true)
     end
-    # Doing build(:repository_object) rather than create to support test where it doesn't exist.
-    # This case can be removed after we fully migrate to RepositoryObjects
-    let(:repository_object) { build(:repository_object, external_identifier: druid) }
+    let(:repository_object) { create(:repository_object, external_identifier: druid) }
```
mjgiarlo, line 23: Instead of defaulting to build, and patching in conditional behavior in the `before` block, default to create and override where needed.

aaron-collier (approved):

PR #4929, opened 2024-04-26
Inline indexing code.

justinlittman:
Reduce complexity. Indexing is now only performed in DSA.

Unit, integration tests


mjgiarlo (approved):
It's a heck of a lot of code, but it'll be easier to manage in DSA than off in a gem that's only used in DSA (now that DIA is being retired)!

Approve pending consideration of my only code review question: how might you feel about promoting `app/services/indexing/*` to `app/*`? So, moving `app/services/indexing/builders` to `app/builders`? Downside is maybe less clarity about relevance to indexing. Upside is promoting specific kinds of Ruby classes that hang together over making an already bloated `app/services` junk drawer into an even more bloated junk drawer?

PR #4934, opened 2024-04-29
Only use db container

lwrubel:
Is there a use case for bringing up a full set of services to run with dor-services-app in development? This PR shows what only bringing up a db container would involve.

If we do want to have the option of bring up a full set of containers, we should review and update them. 

Running locally.


justinlittman (approved):

jcoyne:
I think the full set of containers could be useful for doing an indexing integration test (now that indexing is in this repo), so I'd be `-1` on this change.

jcoyne:
You can always do `docker compose up db` if you only want the one.  I use an alias for this.

justinlittman:
The solr container configuration is all wrong, so would need to be replaced for an integration test.

lwrubel:
This would be an alternative approach: https://github.com/sul-dlss/dor-services-app/pull/4936 

However, I'm not sure how to test that the container configurations are correct in that PR. They come up, but I'm not sure beyond that.

PR #4935, opened 2024-04-29
[HOLD] Determine open status from RepositoryObject

justinlittman:
To move away from using workflow state to determine if object is open.

Unit, integration tests


justinlittman:
Hold until after dep updates on 5/6.

mjgiarlo (approved):

PR #4936, opened 2024-04-29
Updated docker compose

lwrubel:
Update container definitions for a future use case that could make use of them (e.g. integration testing of indexing). This would be an alternative approach to https://github.com/sul-dlss/dor-services-app/pull/4934 which removes all container definitions except for the db. 

However, I'm not sure how to test that the container configurations are correct. They come up, but I'm not sure beyond that. 

All containers come up locally.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,11 +1,10 @@
-version: '3.6'
-
 services:
   web:
     build: .
     volumes:
       - ./:/app
     working_dir: /app
+    platform: linux/amd64
     depends_on:
       - solr
       - suri
@@ -35,6 +34,7 @@ services:
     volumes:
       - ./:/app
     working_dir: /app
+    platform: linux/amd64
     command: bundle exec sidekiq
     depends_on:
       - redis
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -46,18 +46,19 @@ services:
...
   redis:
-    image: redis:4
+    image: redis
     ports:
       - 6379:6379
   solr:
-    image: solr:slim
+    image: solr:8.11.2
     volumes:
       - ./solr_conf/conf/:/myconfig
```
justinlittman, line 55: Is the solr_conf in DSA current?

lwrubel: @ndushay did quite a bit of updating of solr_config.xml in the past few months, including for docker. See https://github.com/sul-dlss/dor-services-app/pull/4656

mjgiarlo (approved):

mjgiarlo:
@justinlittman Sorry about the errant review request. Happy to revert if you wanted to take another look!

PR #4937, opened 2024-04-29
Creates initial UserVersion model validating a closed RepositoryObject

### LINKED ISSUE #4840: Create UserVersion active record model
Fields:
* version: integer, not null
* withdrawn: boolean, not null, default false
* belongs to a **closed** RepositoryObjectVersion. (That the RepositoryObjectVersion is closed should be validated.)

Unique index on version, repository_object_version_id.

aaron-collier:

Closes #4840 

### DIFF
--- a/app/models/user_version.rb
+++ b/app/models/user_version.rb
@@ -8,6 +8,6 @@ class UserVersion < ApplicationRecord
 
   def repository_object_version_is_closed
     # Validate that the repository object version is closed
-    errors.add(:repository_object_version, 'cannot open a user version on an open RepositoryObjectVersion') if repository_object_version.repository_object.open?
+    errors.add(:repository_object_version, 'cannot set a user version to an open RepositoryObjectVersion') if repository_object_version.repository_object.open?
   end
 end
--- a/app/models/user_version.rb
+++ b/app/models/user_version.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+# Models a user version of a specific repository object version.
+class UserVersion < ApplicationRecord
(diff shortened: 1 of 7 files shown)


app/models/user_version.rb
```diff
@@ -0,0 +1,13 @@
...
+
+# Models a user version of a specific repository object version.
+class UserVersion < ApplicationRecord
+  belongs_to :repository_object_version
+
+  validate :repository_object_version_is_closed
+
+  def repository_object_version_is_closed
+    # Validate that the repository object version is closed
+    errors.add(:repository_object_version, 'cannot open a user version on an open RepositoryObjectVersion') if repository_object_version.repository_object.open?
```
justinlittman, line 11: "cannot set a user version to an open RepositoryObjectVersion"

db/migrate/20240429201956_create_user_versions.rb
```diff
@@ -0,0 +1,13 @@
+class CreateUserVersions < ActiveRecord::Migration[7.1]
+  def change
+    create_table :user_versions do |t|
+      t.belongs_to :repository_object_version, index: true
```
justinlittman, line 4: Should this be a foreign key?

```
t.references :repository_object_version, null: false, foreign_key: true
```

justinlittman (changes requested):

aaron-collier:
@justinlittman applied changes. I added the FK as a follow up migration to avoid any commit/manual wonkiness. If you'd prefer it as a single migration, I can update. Thanks.

justinlittman (approved):

Issue #4940, opened 2024-04-30
Clean-up RepositoryObject conditional code


justinlittman:
That is code that handles the case where a RepositoryObject might not exist. Now that migration is complete, RepositoryObjects can be assumed to exist.

justinlittman:
Done for now.

PR #4945, opened 2024-04-30
Clean up RepositoryObject conditional code in CocinaObjectStore

justinlittman:
refs #4940 

Continue migration to RepositoryObjects.

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -11,7 +11,7 @@ class RepositoryObject < ApplicationRecord
   class VersionAlreadyOpened < StandardError; end
   class VersionNotOpened < StandardError; end
 
-  has_many :versions, class_name: 'RepositoryObjectVersion', dependent: :destroy, inverse_of: 'repository_object'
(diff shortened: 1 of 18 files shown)


jcoyne (approved):

app/controllers/versions_controller.rb
```diff
@@ -103,7 +103,7 @@ def load_version
   end
 
   def repository_object_version_content(repository_object_versions)
-    repository_object_versions.map do |repository_object_version|
+    repository_object_versions.order(version: :asc).map do |repository_object_version|
```
mjgiarlo, line 106: I wonder if we should consider doing this in the [model](https://github.com/sul-dlss/dor-services-app/blob/main/app/models/repository_object.rb#L14): 

```ruby
  has_many :versions, class_name: 'RepositoryObjectVersion', dependent: :destroy, inverse_of: 'repository_object', order: 'version ASC'
```

❓

mjgiarlo (commented):
Quick question for now

app/controllers/versions_controller.rb
justinlittman: Ended up with:
```
has_many :versions, -> { order(version: :asc) }, class_name: 'RepositoryObjectVersion', dependent: :destroy, inverse_of: 'repository_object'
```

PR #4948, opened 2024-04-30
Create UserVersionService

### LINKED ISSUE #4841: Create UserVersion service
Providing the following methods:
* create user version(druid, version, optional access) returns UserVersion version
* change withdrawn(druid, user version, withdrawn): Changes the withdrawn attribute for a UserVersion.
* move user version(druid, version, user version): Changes association of an existing UserVersion
* exists?(druid, user version)

Note that this service will encapsulate the UserVersion active record model.

aaron-collier:

Closes #4841 

### DIFF
--- a/app/services/user_version_service.rb
+++ b/app/services/user_version_service.rb
@@ -20,12 +20,16 @@ class UserVersionService
     new(druid:, version:).move(user_version, event_factory)
   end
 
+  def self.exist?(druid:, user_version:)
+    new(druid:, version: nil).exist?(user_version:)
+  end
+
   # @param [String] druid of the item
   # @param [Integer] version of the item
   def initialize(druid:, version:)
     @druid = druid
     @repository_object = RepositoryObject.find_by!(external_identifier: druid)
-    @repository_object_version = @repository_object.versions.find_by!(version:)
+    @repository_object_version = version.nil? ? @repository_object.head_version : @repository_object.versions.find_by!(version:)
   end
 
   # @param [Class] event_factory (EventFactory) the factory for creating events
@@ -47,6 +51,10 @@ class UserVersionService
     user_version.repository_object_version = repository_object_version
     user_version.save!
     event_factory.create(druid:, event_type: 'user_version_moved', data: { version: user_version.version.to_s })
+  end
+
+  def exist?(user_version:)
+    RepositoryObjectVersion.where(repository_object:, user_version:).exists?
   end
 
   attr_reader :druid, :repository_object, :repository_object_version
--- a/app/services/user_version_service.rb
+++ b/app/services/user_version_service.rb
@@ -5,26 +5,23 @@ class UserVersionService
   class UserVersioningError < StandardError; end
 
   # @param [RepositoryObjectVersion] repository_object_version of the item being acted upon
-  # @param [Class] event_factory (EventFactory) the factory for creating events
-  def self.create(druid:, version:, event_factory: EventFactory)
-    new(druid:, version:).create(event_factory:)
+  def self.create(druid:, version:)
+    new(druid:, version:).create
   end
 
   # @param [UserVersion] user_version to withdraw
-  # @param [Class] event_factory (EventFactory) the factory for creating events
-  def self.withdraw(user_version:, event_factory: EventFactory)
+  def self.withdraw(user_version:)
     user_version.withdrawn = true
     user_version.save!
     druid = user_version.repository_object_version.repository_object.external_identifier
-    event_factory.create(druid:, event_type: 'user_version_withdrawn', data: { version: user_version.version.to_s })
+    EventFactory.create(druid:, event_type: 'user_version_withdrawn', data: { version: user_version.version.to_s })
   end
 
   # @param [String] druid of the item
   # @param [Integer] version of the item
   # @param [UserVersion] user_version to move
-  # @param [Class] event_factory (EventFactory) the factory for creating events
-  def self.move(druid:, version:, user_version:, event_factory: EventFactory)
-    new(druid:, version:).move(user_version, event_factory)
+  def self.move(druid:, version:, user_version:)
+    new(druid:, version:).move(user_version)
   end
 
   # @param [String] druid of the item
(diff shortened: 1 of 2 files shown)


app/services/user_version_service.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+# Open and close versions
+class UserVersionService
+  class UserVersioningError < StandardError; end
+
+  # @param [RepositoryObjectVersion] repository_object_version of the item being acted upon
```
justinlittman, line 7: Params doc needs to be fixed.

app/services/user_version_service.rb
```diff
@@ -0,0 +1,77 @@
...
+  # @param [Class] event_factory (EventFactory) the factory for creating events
+  def self.create(druid:, version:, event_factory: EventFactory)
```
justinlittman, line 9: Can you change EventFactory to be consistent with new approach in https://github.com/sul-dlss/dor-services-app/pull/4956?

app/services/user_version_service.rb
```diff
@@ -0,0 +1,77 @@
...
+    new(druid:, version:).create(event_factory:)
+  end
+
+  # @param [UserVersion] user_version to withdraw
+  # @param [Class] event_factory (EventFactory) the factory for creating events
+  def self.withdraw(user_version:, event_factory: EventFactory)
```
justinlittman, line 15: This shouldn't take ActiveRecord models as an arg. It should take druid and user_version (integer) and lookup itself.

app/services/user_version_service.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  # @param [String] druid of the item
+  # @param [UserVersion] user_version to check
+  def self.exist?(druid:, user_version:)
+    new(druid:, version: nil).exist?(user_version:)
+  end
+
+  # @param [String] druid of the item
+  # @param [Integer] version of the item
+  def initialize(druid:, version:)
```
justinlittman, line 38: This initializer doesn't really work, as only some methods need version. It might just be cleaner to make all of the methods class methods.

app/services/user_version_service.rb
```diff
@@ -0,0 +1,77 @@
...
+    @repository_object_version = version.nil? ? @repository_object.head_version : @repository_object.versions.find_by!(version:)
+  rescue ActiveRecord::RecordNotFound
+    raise(UserVersioningError, "RepositoryObject not found for #{druid}")
+  end
+
+  # @param [Class] event_factory (EventFactory) the factory for creating events
+  # @return [UserVersion version] The version number of the new user version object
+  # @raise [VersionService::VersioningError] if the object hasn't been accessioned, or if a version is already opened
+  def create(event_factory:)
+    raise(UserVersioningError, 'RepositoryObjectVersion not closed') if repository_object_version.closed_at.nil?
```
justinlittman, line 50: `.closed?`

app/services/user_version_service.rb
```diff
@@ -0,0 +1,77 @@
...
+  # @param [String] druid of the item
+  # @param [Integer] version of the item
+  # @param [UserVersion] user_version to move
+  # @param [Class] event_factory (EventFactory) the factory for creating events
+  def self.move(druid:, version:, user_version:, event_factory: EventFactory)
+    new(druid:, version:).move(user_version, event_factory)
+  end
+
+  # @param [String] druid of the item
+  # @param [UserVersion] user_version to check
```
justinlittman, line 31: This should take a user_version (integer) not an active record object.

justinlittman (changes requested):

app/services/user_version_service.rb
aaron-collier: Good call. Refactoring.

app/services/user_version_service.rb
aaron-collier: Good catch, I had updated the other call and missed this one.

app/services/user_version_service.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+# Create, withdraw, and move UserVersions
+class UserVersionService
+  class UserVersioningError < StandardError; end
+
+  # @param [String] druid of the item
+  # @param [Integer] version of the item
+  # @return [UserVersion version] The version number of the new user version object
```
justinlittman, line 9: This should be returning an integer.

app/services/user_version_service.rb
```diff
@@ -0,0 +1,82 @@
...
+    next_user_version
+  end
+
+  # @param [String] druid of the item
+  # @param [integer] user_version version to withdraw
+  # @param [Boolean] withdraw true to withdraw the user version, false to unwithdraw
+  # @raise [UserVersionService::UserVersioningError] RepositoryObject not found for the druid
+  # @raise [UserVersionService::UserVersioningError] RepositoryObjectVersion not found for the version
+  def self.withdraw(druid:, user_version:, withdraw: true)
+    @druid = druid
```
justinlittman, line 32: I don't love this use of instance variables. Can't we just pass around as method args?

justinlittman: For example, `user_version_for(druid:, user_version:)`

app/services/user_version_service.rb
```diff
@@ -0,0 +1,82 @@
...
+  # @param [Integer] RepositoryObjectVersion version of the item to move to
+  # @param [integer] user_version version to move
+  # @raise [UserVersionService::UserVersioningError] RepositoryObject not found for the druid
+  # @raise [UserVersionService::UserVersioningError] RepositoryObjectVersion not found for the version
+  # @raise [UserVersionService::UserVersioningError] RepositoryObjectVersion not closed
+  def self.move(druid:, version:, user_version:)
+    @druid = druid
+    @version = version
+
+    raise(UserVersioningError, 'RepositoryObjectVersion not closed') unless repository_object_version.closed?
```
justinlittman, line 47: This using class methods like instance methods feels really un-idiomatic.

aaron-collier: Yeah, good point. I was trying to DRY things up, which I was able to do a bit cleaner based not his feedback (I think) :-)

justinlittman (approved):

PR #4951, opened 2024-05-01
Update UserVersion validation to check closed_at datetime

aaron-collier:

Update the validation to check `closed_at` on the RepositoryObjectVersion instead of the status of the parent RepositoryObject.


justinlittman (commented):
Can you add a `.closed?` method the RepositoryObjectVersion to make this a little cleaner?

aaron-collier:
@justinlittman Done. I went ahead and added an `open?` as well. If that is a reach, happy to remove it.

PR #4954, opened 2024-05-01
Retries creating workspace content dir in ShelvingService.

### LINKED ISSUE #4952: File exists @ dir_s_mkdir - ShelvingService
## Backtrace

    line 30 of [PROJECT_ROOT]/app/services/shelving_service.rb: shelve
    line 9 of [PROJECT_ROOT]/app/services/shelving_service.rb: shelve
    line 15 of [PROJECT_ROOT]/app/jobs/shelve_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/94972560)

justinlittman:
closes #4952

Less noisy integration tests.

Integration

### DIFF
--- a/app/services/shelving_service.rb
+++ b/app/services/shelving_service.rb
@@ -27,7 +27,17 @@ class ShelvingService
     preserve_diff = content_diff('preserve')
     shelve_diff = content_diff('shelve')
 
-    workspace_content_pathname = Pathname(workspace_druid.content_dir(true))
+    # In stage, we consistently get Errno::EEXIST.
+    # The theory is that this is a sync issue with the underlying filesystem.
+    # This addresses the issue by retrying the operation; upon retry, the directory should be found
+    # to exist and the operation should succeed properly.
+    workspace_content_pathname = nil
+    begin
+      workspace_content_pathname = Pathname(workspace_druid.content_dir(true))
+    rescue Errno::EEXIST
+      retry
+    end
+
     ShelvableFilesStager.stage(druid, preserve_diff, shelve_diff, workspace_content_pathname)
 
     # workspace_content_pathname = workspace_content_dir(shelve_diff, workspace_druid)


jcoyne (changes requested):
I believe this will make the problem worse.

Currently it fails because another process has already created the directory.   If you keep retrying, you will continue to find that the directory exists and still cannot be created.

justinlittman:
Here's the relevant code: https://github.com/sul-dlss/druid-tools/blob/main/lib/druid_tools/druid.rb#L88-L100

There are 2 checks to see if the directory exists prior to create (the first one should skip the create; the second one should raise). I'm guessing that there is probably a file system sync problem that is causing both of those checks to report that the directory doesn't exist. I'd expect that if retried the file system sync will resolve and the code will work as expected.

jcoyne:
@justinlittman If you have two processes running simultaneously and they each ask "Directory x created yet?" they will each get a "no". Then they will both try to create it.  One will succeed and the other will get this error.  A retry will just give the error again.

justinlittman:
On the second process's retry, it will discover that the directory exists (https://github.com/sul-dlss/druid-tools/blob/main/lib/druid_tools/druid.rb#L90) and skip the mkdir.

app/services/shelving_service.rb
```diff
@@ -27,7 +27,14 @@ def shelve
     preserve_diff = content_diff('preserve')
     shelve_diff = content_diff('shelve')
 
-    workspace_content_pathname = Pathname(workspace_druid.content_dir(true))
+    # This addresses a race condition.
```
jcoyne, line 30: can you be more specific about the race condition and why the retry resolves it in this comment?

jcoyne (changes requested):

app/services/shelving_service.rb
jcoyne: Should the retry be upstream in the gem? Don't we want it fixed for all users?

justinlittman: I'm reluctant to fix in the gem, as the fix might have unintended consequences on other file systems.

Adding more detail to the comment.

jcoyne (approved):

PR #4957, opened 2024-05-02
Make tests less flaky, less noisy, and faster

mjgiarlo:
# Why was this change made?

Developer happiness and efficiency: while working on #4956, I found flaky tests, noisy tests, and some slow ones. I decided to try improving all of these. This commit includes those changes.

# How was this change tested?

- [x] CI
- [x] ran `rspec` 10-15 times in a row with random seeds and saw zero failures and zero noisy test output

### DIFF
--- a/config/settings/test.yml
+++ b/config/settings/test.yml
@@ -1,4 +1,4 @@
 catalog:
   folio:
     max_lookup_tries: 3
-    sleep_seconds: 1
+    sleep_seconds: 0
--- a/spec/services/cleanup_service_spec.rb
+++ b/spec/services/cleanup_service_spec.rb
@@ -80,7 +80,7 @@ RSpec.describe CleanupService do
       end
 
       it 'backups, cleans up content, and delete workflows' do
-        described_class.stop_accessioning(druid)
+        expect { described_class.stop_accessioning(druid) }.to output.to_stdout
         expect(described_class).to have_received(:backup_content_by_druid).once.with(druid)
         expect(described_class).to have_received(:cleanup_by_druid).once.with(druid)
         expect(described_class).to have_received(:delete_accessioning_workflows).once.with(druid, version)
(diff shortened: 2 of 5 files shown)


config/settings/test.yml
```diff
@@ -1,4 +1,4 @@
 catalog:
   folio:
     max_lookup_tries: 3
-    sleep_seconds: 1
+    sleep_seconds: 0
```
mjgiarlo, line 4: I don't think we need tests to sit around waiting. The passage of time here is not important; what is important is that the retries happen, which they do even after this change.

spec/services/cleanup_service_spec.rb
```diff
@@ -80,15 +80,15 @@
       end
 
       it 'backups, cleans up content, and delete workflows' do
-        described_class.stop_accessioning(druid)
+        expect { described_class.stop_accessioning(druid) }.to output.to_stdout
```
mjgiarlo, line 83: I didn't think it was helpful to see `puts` statements interrupting test output. The `puts` statements in the cleanup service are there primarily for users running that service via rake task, not for folks running tests.

spec/services/member_service_spec.rb
```diff
@@ -37,12 +37,20 @@
...
+        allow(workflow_client).to receive(:active_lifecycle).with(
+          druid: dro1.external_identifier,
+          milestone_name: 'opened',
+          version: 1
+        ).and_return(true)
+        allow(workflow_client).to receive(:active_lifecycle).with(
+          druid: dro2.external_identifier,
+          milestone_name: 'opened',
+          version: 1
+        ).and_return(false)
```
mjgiarlo, line 49: This test was flaky until I made this change. Before the change, this was order-dependent and now it isn't, so it no longer flaps.

justinlittman (approved):

spec/services/member_service_spec.rb
```diff
@@ -37,12 +37,20 @@
...
       end
 
       it 'returns members' do
         expect(members).to eq [dro2.external_identifier]
-        expect(workflow_client).to have_received(:active_lifecycle).with(druid: dro1.external_identifier, milestone_name: 'opened', version: dro1.version.to_s)
```
mjgiarlo, line 45: I don't believe the stubs here (and on lines 59-60) are intended as spies. They are there to make sure the class operates as intended. What we *really* care about here is the effect it has on filtering `members`.

PR #4963, opened 2024-05-02
change how accessioning? is determined

peetucket:

Ref sul-dlss/common-accessioning/pull/1180 

note: that  PR does not yet have the logic that will be added to end-accession that will actually kick off ocrWF

Proposal to change how we determine if an object is in accessioning, which is used to determine if it can be opened. This makes the requirement that there are no active steps in accessionWF instead of looking at the lifecycle.  This allows us to skip checking the `end-accession` step, which allows us to start ocrWF and open a version, even if that last step is being executed.  Otherwise we are likely to end up in a race condition, where `end-accession` triggers `ocrWF`, which tries to open the version and is blocked, because `end-accession` may not have yet been marked as complete.  

CI

### DIFF
--- a/app/services/workflow_state_service.rb
+++ b/app/services/workflow_state_service.rb
@@ -47,14 +47,15 @@ class WorkflowStateService
   # As such, they may not represent the current best practice for determining workflow state
   # and will probably be subject to further refactoring or removal.
 
-  # Checks if the active (latest) version has any incomplete workflow steps and there is an accessionWF (known by the presence of a submitted milestone).
-  # If so we don't want to start another acceession workflow.  This is also true if a preservationAuditWF has returned an error, so that no further
-  # accessioning can take place until that is resolved.
+  # Checks if the active (latest) version has any incomplete workflow steps in accessionWF (other than end-accession).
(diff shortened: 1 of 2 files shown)


app/services/workflow_state_service.rb
```diff
@@ -47,14 +47,15 @@ def open?
...
+  # If there are other active steps in accessionWF, we don't want to start another accession workflow.
+  # Note that this is also true if a preservationAuditWF or preservationIngestWF has returned an error,
+  # because it will not complete the `sdr-ingest-received` step, which will cause this to return true as well.
   # @return [Boolean] true if object is currently being accessioned or is failing an audit
   def accessioning?
-    return true if workflow_client.active_lifecycle(druid:, milestone_name: 'submitted', version: version.to_s)
-
-    false
+    active_workflow_except_step?(workflow: 'accessionWF', process: 'end-accession')
   end
```
peetucket, line 59: now similar to the `assembling?` check above: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/workflow_state_service.rb#L25-L36

justinlittman (approved):

PR #4971, opened 2024-05-06
Update user version on close

### LINKED ISSUE #4947: Add user version updating to DSA VersionService.close and version close endpoint
Version.close and version close endpoint should accept a new optional parameter for updating the user version.

The options should be:
* Nothing (default)
* Create new user version
* Update last user version to this object version. (If update last user version to this object version and there is no last user version create a new user version).

Please add the parameter to the openapi.

aaron-collier:

Closes #4947 

### DIFF
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -38,10 +38,12 @@ class VersionService
   # @param [String] description describes the version change
   # @param [String] user_name add username to the events datastream
   # @param [Boolean] start_accession (true) set to true if you want accessioning to start, false otherwise
-  def self.close(druid:, version:, description: nil, user_name: nil, start_accession: true)
+  # @param [Symbol] user_version (:none) one of :none, :new, :update
+  def self.close(druid:, version:, description: nil, user_name: nil, start_accession: true, user_version: :none)
     new(druid:, version:).close(description:,
                                 user_name:,
-                                start_accession:)
+                                start_accession:,
+                                user_version:)
   end
 
   # @param [String] druid of the item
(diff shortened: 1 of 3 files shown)


jcoyne:
@aaron-collier can you squash these commits?

app/services/version_service.rb
```diff
@@ -121,21 +123,39 @@ def can_open?(assume_accessioned: false)
...
+    repository_object = RepositoryObject.find_by!(external_identifier: druid)
+    current_closed_version = repository_object&.last_closed_version
+
     # Default to creating accessionWF when calling close_version
     workflow_client.close_version(druid:,
                                   version: version.to_s,
                                   create_accession_wf: start_accession)
 
-    RepositoryObject.find_by!(external_identifier: druid).close_version!(description:)
+    repository_object&.close_version!(description:)
```
jcoyne, line 148: I don't think we need this safe-nav here.

jcoyne (changes requested):

app/services/version_service.rb
```diff
@@ -121,21 +123,39 @@ def can_open?(assume_accessioned: false)
...
+
+    raise ArgumentError, "user_version must be one of #{user_version_options.join(', ')}" unless user_version_options.include?(user_version)
+
     ObjectVersion.update_current_version(druid:, description:) if description
 
     ensure_closeable!
 
+    # Get the currently closed version before we close the head version
+    repository_object = RepositoryObject.find_by!(external_identifier: druid)
+    current_closed_version = repository_object&.last_closed_version
```
jcoyne, line 141: I don't think we need this safe-nav either, do we?

jcoyne (changes requested):

app/services/version_service.rb
aaron-collier: Probably not. I'll update. (same with below)

aaron-collier:
@jcoyne squashed :-)

jcoyne (approved):

PR #4979, opened 2024-05-09
Memoized the value based on the input value to this method

jcoyne:


justinlittman (approved):
Haven't seen this technique before, but makes sense.

PR #4983, opened 2024-05-10
allow the start accession DSA endpoint to accept workflow context

peetucket:

Part of sul-dlss/dor-services-client#427 -- allows us to send workflow context to the workflow service via json in the body of the `accession` request.  The workflow client and workflow server are already setup to receive it and store it.

New specs and QA

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -76,7 +76,9 @@ class ObjectsController < ApplicationController
   # The versioning params are included below for reference.
   #  :descriptions [String] (required) description of version change
   #  :opening_user_name [String] (optional) opening sunetid to add to the events datastream
+  # You can also pass information that will be used to start the workflow:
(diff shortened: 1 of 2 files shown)


app/controllers/objects_controller.rb
```diff
@@ -95,7 +97,7 @@ def accession
     end
 
     # initialize workflow
-    workflow_client.create_workflow_by_name(@cocina_object.externalIdentifier, workflow, version: updated_cocina_object.version.to_s)
+    workflow_client.create_workflow_by_name(@cocina_object.externalIdentifier, workflow, version: updated_cocina_object.version.to_s, context: workflow_context)
     head :created
```
peetucket, line 101: context comes into DSA and then gets passed on to the workflow client, where it will be sent to workflow-server-rails (this change is already done: https://github.com/sul-dlss/workflow-server-rails/pull/745 and https://github.com/sul-dlss/dor-workflow-client/pull/299)

PR #4988, opened 2024-05-14
add report for auditing contents of technical metadata database

jmartin-sul:

to run the audit described in https://github.com/sul-dlss/technical-metadata-service/issues/515 (will close that issue once the report has actually been run)

on QA.  once the related puppet changes were applied and the supporting shared_configs updated, i `scp`ed the report code over.  then tested, fixed, refactored for usability, tested, fixed, repeated, until this felt like it'd be likely to run through on prod (or produce helpful logging if it didn't).

i did one full run on QA with this code (448 druids with techMD inconsistencies were found), and also ran both the test report and the single druid method.

example test report stdout (same stdout as full report would be, just limited to 20 druids audited):
```
$ bin/rails r -e p 'AuditTechnicalMetadataFileList.test_report'
druid:bk509qs2277: found technical-metadata-service database; inconsistencies with v11 cocina: {"missing_filenames"=>["asawa.jp2", "maps-africa.jp2", "moore.jp2", "hay.jp2"]}
druid:ww089jm4506: not found in technical-metadata-service database
druid:wy216td0310: found technical-metadata-service database; inconsistencies with v1 cocina: {"missing_filenames"=>["example.jp2"]}
druid:xb970cg0663: found technical-metadata-service database; inconsistencies with v1 cocina: {"missing_filenames"=>["example.jp2"]}
druid:gv185zf2539: not found in technical-metadata-service database
```

progress logging worked, as did the other info and debug logging.

example running on a single druid:
```
$ bin/rails r -e p 'AuditTechnicalMetadataFileList.audit_one_druid(druid: "druid:ww089jm4506")'
druid:ww089jm4506: not found in technical-metadata-service database
```

### DIFF
--- a/app/reports/audit_technical_metadata_file_list.rb
+++ b/app/reports/audit_technical_metadata_file_list.rb
@@ -1,0 +1,187 @@
+# frozen_string_literal: true
+
+# @see https://github.com/sul-dlss/technical-metadata-service/issues/510
+# @see https://github.com/sul-dlss/technical-metadata-service/issues/515
+# @see https://github.com/sul-dlss/technical-metadata-service/issues/517
+# @see https://github.com/sul-dlss/common-accessioning/pull/1154
+#
+# Iterate over all closed DROs, and use the technical metadata service's
+# audit endpoint to determine whether it
+# * is aware of all the files it should be aware of.
+# * has info on extraneous files.
+# * has info for a file with the same name but different content (based on the MD5 of the file).
+#
+# This is intended to audit the full contents of SDR, and the techMD problem being audited was
+# present for a while, so this report may produce a lot of output. It is suggested that output be
+# piped through gzip to save disk space, e.g.
+#
+#   bin/rails r -e production 'AuditTechnicalMetadataFileList.report' | gzip > techmd_audit.`date -Iseconds`.gz
+#
+# Probably best to do that in a screen session since it might take a while.
+#
+# For a small test run, you can do something like
+#   bin/rails r 'AuditTechnicalMetadataFileList.test_report'
+# To follow up on one druid
+#   bin/rails r 'AuditTechnicalMetadataFileList.audit_one_druid(druid: "druid:bc123df4567")'
+#
+# Error and info output other than an individual druid audit result (e.g. totally unexpected
+# responses and exceptions, progress info) will go to a log file named for the class.
+#
+# @note Since we shouldn't be creating new instance of this problem, we should be able to retire
+# this report once extant instances are remediated.
+#
+# @note If you remove this report, consider removing technical_metadata settings section and
+# techMD tokens for DSA from vault if nothing else has come to need that info (and related env
+# var setting from puppet).
+class AuditTechnicalMetadataFileList
+  FILE_INFOS_JSONPATH = '$.contains[*].structural.contains'
+  FILES_FROM_RESOURCES_SQL = Arel.sql("jsonb_path_query_array(structural, '#{FILE_INFOS_JSONPATH}')")
+  EXPECTED_RESPONSE_CODES = [200, 404].freeze
+
+  def self.test_report
+    report(limit: 20)
+  end
+
+  def self.report(...)
+    new(...).report
+  end
+
+  def self.audit_one_druid(druid:)
+    new.audit_one_druid(druid:)
+  end
+
+  def initialize(limit: nil)
+    @limit = limit
+  end
+
+  attr_reader :limit
+
+  def report
+    logger.info("=== auditing technical-metadata-service for inconsistencies with current cocina file information (limit: #{limit})")
+
+    # We iterate over the druid/version pairs from the Dro list and feed them to VersionService as soon as we get them,
+    # to minimize the chance we try a stale latest version, since latest is what VersionService.open? needs to operate on.
+    # We could get more clever/efficient/transactional, but that'd require copypasta'ing VersionService and/or RepositoryObject
+    # internal code, or refactoring VersionService and/or RepositoryObject a bit.  Which didn't seem worth the effort for a
+    # report that should become obsolete in not too long.
+    # Using in_batches means we can pluck only the info we need to use, without having to instantiate an ActiveRecord obj for
+    # each Dro in each batch.
+    num_processed = 0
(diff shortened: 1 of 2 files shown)


jmartin-sul:
interestingly, the QA audit turned up some unexpected files in techMD DB, and some instances where file names matched for a druid but md5 differed.  i was only expecting missing files, but i wasn't sure enough of my understanding of the original problem for that to be a strong expectation.  either way, glad i did had the techMD endpoint return all 3 categories of problem.

app/reports/audit_technical_metadata_file_list.rb
```diff
@@ -0,0 +1,185 @@
...
+  def audit_one_druid(druid:)
+    Dro.where(external_identifier: druid).pick(:external_identifier, :version, FILES_FROM_RESOURCES_SQL).tap do |dro_row|
+      self.class.process_dro_row(dro_row, techmd_connection, logger)
+    end
+  end
+
+  class << self
+    # @param [Array] an array of arrays.  each element in the top level list is an array containing
+    #   druid, version, and cocina-models file hashes grouped by resource (plucked from Dro)
+    def process_dro_row(dro_row, techmd_connection, logger)
```
jmartin-sul, line 91: perhaps this should've been a service class.  but i think the constructor would take these 3 params, which would only save one param among the other class methods (`dro_row` to `.druid_version_files`).  so i'm not sure much concision would be gained from a service class.  and this feels like an ok amount of reusability (`.process_drow_row`) and boxing of complexity into private helper methods.  so i'm inclined to leave this factored as-is, since i'm not sure it's worth spending more time cleaning this up (though happy to take suggestions if i'm missing something obvious).

config/settings.yml
```diff
@@ -74,6 +74,10 @@ purl_fetcher:
   url: 'https://purl-fetcher.example.edu'
   token: 'fake-token'
 
+technical_metadata:
```
justinlittman, line 77: Maybe a note why this config is here?

app/reports/audit_technical_metadata_file_list.rb
```diff
@@ -0,0 +1,185 @@
...
+      max: 3,
+      methods: %i[delete get head options put post]
+    }
+  end
+
+  def techmd_connection
+    @techmd_connection ||= Faraday.new(Settings.technical_metadata.url) do |builder|
+      builder.request :retry, retry_options
+      builder.use Faraday::Request::UrlEncoded
+      # builder.use Faraday::Response::RaiseError # raise exceptions on 40x, 50x responses
```
justinlittman, line 165: Cruft

app/reports/audit_technical_metadata_file_list.rb
```diff
@@ -0,0 +1,185 @@
...
+  private
+
+  def logger
+    @logger ||= Logger.new(Rails.root.join('log', "#{self.class.name}.log"))
+  end
+
+  def retry_options
+    {
+      max: 3,
+      methods: %i[delete get head options put post]
```
justinlittman, line 157: Just `get`?

justinlittman (approved):
Approved with comments.

config/settings.yml
jmartin-sul: good call, will do

app/reports/audit_technical_metadata_file_list.rb
jmartin-sul: yes, good catch

app/reports/audit_technical_metadata_file_list.rb
jmartin-sul: in this case `post` is the only type of request we make, and is actually retriable, because it's just a read with no side-effects.  should've been a `get` semantically, but the file list we're sending could overrun the traditional max size for URL params in a `GET`, hence the `POST` usage.

i could make the list just `post`.  was on the fence about that, felt a little weird.  either way, seems like a comment would be good since this is an unusual config?

jmartin-sul:
added a comment to `settings.yml`, got rid of the Faraday config cruft in `#techmd_connection`, added a comment to the confusing `methods` list in `#retry_options`.  squashed that into the existing commit, but here's the diff from before i pushed for easy review:

```diff
% git diff origin/audit-techmd HEAD
diff --git a/app/reports/audit_technical_metadata_file_list.rb b/app/reports/audit_technical_metadata_file_list.rb
index aa94cc72..483620d5 100644
--- a/app/reports/audit_technical_metadata_file_list.rb
+++ b/app/reports/audit_technical_metadata_file_list.rb
@@ -154,6 +154,9 @@ class AuditTechnicalMetadataFileList
   def retry_options
     {
       max: 3,
+      # The default retriable/idempotent HTTP methods, plus POST. As of May 2024, POST is the only
+      # HTTP method this class actually uses (for a read-only operation, but the file list parameter
+      # might exceed the traditional max for total URL query param size, hence POST instead of GET).
       methods: %i[delete get head options put post]
     }
   end
@@ -162,7 +165,6 @@ class AuditTechnicalMetadataFileList
     @techmd_connection ||= Faraday.new(Settings.technical_metadata.url) do |builder|
       builder.request :retry, retry_options
       builder.use Faraday::Request::UrlEncoded
-      # builder.use Faraday::Response::RaiseError # raise exceptions on 40x, 50x responses
       builder.adapter Faraday.default_adapter
       builder.headers[:user_agent] = "dor-services-app #{Socket.gethostname}"
       builder.headers['Authorization'] = "Bearer #{Settings.technical_metadata.token}"
diff --git a/config/settings.yml b/config/settings.yml
index e2593ca8..a3fc94aa 100644
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -74,6 +74,9 @@ purl_fetcher:
   url: 'https://purl-fetcher.example.edu'
   token: 'fake-token'
 
+# As of may 2024, this is just here to support auditing contents of
+# technical-metadata-service. If the report that uses it (AuditTechnicalMetadataFileList)
+# is removed, see whether this section can be removed.
 technical_metadata:
   url: 'https://techmd-svc.example.edu'
   token: 'secret-token'
```

merging since approved and i think i addressed the comments, but happy to revisit the `retry_options` thing or whatever else in a follow up PR if you'd like.

PR #4991, opened 2024-05-15
Diff cocina from purl to determine what to shelve.

### LINKED ISSUE #4982: Use cocina structural diff to determine files to send to purl-fetcher
Currently, the files to send to purl-fetcher are determined based on the existing content diffing strategy: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/shelving_service.rb#L50

For this ticket, implement a new strategy that compares the current cocina structural against the cocina structural retrieved from purl (https://purl.stanford.edu/<druid>.json). For any file which is (1) shelved and (2) not in purl cocina or in purl cocina but has different checksums, send to purl-fetcher.

justinlittman:
closes #4982

Less janky.

Unit, integration tests.

### DIFF
--- a/app/services/digital_stacks_differ.rb
+++ b/app/services/digital_stacks_differ.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+# Determines which files should be shelved by comparing the current
+# cocina object against PURL's current cocina object.
+# A file should be shelved when it included in PURL's current cocina object
+# or is included, but has changed.
+class DigitalStacksDiffer
+  class Error < StandardError; end
+
+  def self.call(...)
+    new(...).call
+  end
+
+  def initialize(cocina_object:)
+    @cocina_object = cocina_object
+  end
+
+  # return [Array<String>] filenames of files that should be shelved
+  def call
+    cocina_object_file_map.reject do |filename, sha1|
+      purl_cocina_object_file_map[filename] == sha1
+    end.keys
+  end
+
+  private
+
+  attr_reader :cocina_object
+
+  def cocina_object_file_map
+    @cocina_object_file_map ||= file_map_for(cocina_object)
+  end
+
+  def purl_cocina_object_file_map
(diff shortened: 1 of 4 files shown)


app/services/digital_stacks_differ.rb
```diff
@@ -0,0 +1,78 @@
...
+    @cocina_object_file_map ||= file_map_for(cocina_object)
+  end
+
+  def purl_cocina_object_file_map
+    @purl_cocina_object_file_map ||= file_map_for(purl_cocina_object)
+  end
+
+  def purl_cocina_object
+    @purl_cocina_object ||= begin
+      response = connection.get("/#{bare_druid}")
```
jcoyne, line 39: Do you need to add `.json` as a suffix here?  Won't it give HTML by default?

app/services/digital_stacks_differ.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def self.call(...)
+    new(...).call
+  end
+
+  def initialize(cocina_object:)
+    @cocina_object = cocina_object
+  end
+
+  # return [Array<String>] filenames of files that should be shelved
```
jcoyne, line 18: What about items that should be removed from stacks?

jcoyne (changes requested):

app/services/digital_stacks_differ.rb
justinlittman: Stacks needs to figure that out.

app/services/digital_stacks_differ.rb
justinlittman: I'd expect so. Fixing.

app/services/digital_stacks_differ.rb
jcoyne: Currently we don't have a stacks API, it's just a file system.  I would agree with you, but only after https://github.com/sul-dlss/purl-fetcher/issues/718 is done.

jcoyne (approved):

PR #4994, opened 2024-05-16
Adds user version index endpoint.

### LINKED ISSUE #4993: Create user version index endpoint
The endpoint should return a list of user versions for a repository object. For each user version, it should also provide its repository object version. For example:
```
{
  user_versions: [
    { user_version: 1, version: 1},
    { user_version: 2, version: 4}
  ]
}
```

For reference, see https://github.com/sul-dlss/dor-services-app/blob/main/app/controllers/versions_controller.rb#L8

justinlittman:
closes #4993

So that user versions can be displayed in argo.

Unit

### DIFF
--- a/app/controllers/user_versions_controller.rb
+++ b/app/controllers/user_versions_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# Controller for user versions
+class UserVersionsController < ApplicationController
+  def index
+    repository_object = RepositoryObject.find_by!(external_identifier: params[:object_id])
+
+    render json: { user_versions: user_versions_content(repository_object.user_versions) }
+  end
+
+  private
+
+  def user_versions_content(user_versions)
+    user_versions.map do |user_version|
+      {
+        userVersion: user_version.version,
+        version: user_version.repository_object_version.version
+      }
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -78,6 +78,8 @@ Rails.application.routes.draw do
         end
       end
 
+      resources :user_versions, only: [:index]
+
       resources :release_tags, only: [:create, :index]
     end
   end
--- a/openapi.yml
+++ b/openapi.yml
@@ -1102,6 +1102,26 @@ paths:
           schema:
             type: string
           example: "some_sunetid"
+  "/v1/objects/{object_id}/user_versions":
+    get:
+      tags:
+        - versions
+      summary: The user version log for this object
+      operationId: "user_versions#index"
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/UserVersionLog"
+      parameters:
+        - name: object_id
+          in: path
+          description: ID of object
+          required: true
+          schema:
+            $ref: "#/components/schemas/Druid"        
   /v1/objects:
     post:
       tags:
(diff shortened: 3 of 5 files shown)


config/routes.rb
```diff
@@ -78,6 +78,8 @@
         end
       end
 
+      resources :user_versions, only: [:index]
```
jcoyne, line 81: should this be resource (singular)? there's never more than one, right?

app/controllers/user_versions_controller.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+# Controller for user versions
+class UserVersionsController < ApplicationController
+  before_action :check_cocina_object_exists, only: %i[index]
```
jcoyne, line 5: Is this necessary? Why?

jcoyne (changes requested):

app/controllers/user_versions_controller.rb
justinlittman: It's not. Removed.

openapi.yml
```diff
@@ -1102,6 +1102,26 @@ paths:
...
+        - versions
+      summary: The user version log for this object
+      operationId: "user_versions#index"
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/UserVersionLog"
```
mjgiarlo, line 1117: Should we add the 404 response here as well?

openapi.yml
```diff
@@ -3274,6 +3294,24 @@ components:
...
+    UserVersion:
+      type: object
+      description: "Information about a user version."
+      properties:
+        version:
+          type: integer
+          description: the number of the object version
+        userVersion:
+          type: string
+          description: the number of the user version
```
mjgiarlo, line 3306: I'm curious why one version if a string and one is an int if both are numbers.

mjgiarlo (approved):
Approve pending consideration of comment

app/controllers/user_versions_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Controller for user versions
+class UserVersionsController < ApplicationController
+  rescue_from(ActiveRecord::RecordNotFound) do |e|
```
jcoyne, line 5: Isn't this already done by `ActionDispatch::ExceptionWrapper.rescue_responses`?

jcoyne (changes requested):

openapi.yml
justinlittman: Copypasta. Fixed.

app/controllers/user_versions_controller.rb
justinlittman: When I comment out and run the test:
```
Failures:

  1) User versions when not found returns a 404
     Failure/Error: repository_object = RepositoryObject.find_by!(external_identifier: params[:object_id])

     ActiveRecord::RecordNotFound:
       Couldn't find RepositoryObject with [WHERE "repository_objects"."external_identifier" = $1]
     # ./app/controllers/user_versions_controller.rb:10:in `index'
     # ./spec/requests/user_versions_inventory_spec.rb:26:in `block (3 levels) in <top (required)>'
```

jcoyne: You have configured the test suite to not allow ActionDispatch to handle the response.  It will be handled in prod or you can switch this setting: https://github.com/sul-dlss/dor-services-app/blob/main/config/environments/test.rb#L34

justinlittman: You learn something new every day. Removing.

jcoyne (approved):

Issue #4990, opened 2024-05-14
Report request: items where one or more files is missing file size information


andrewjbtw:
Required so that we can implement: https://github.com/sul-dlss/cocina-models/issues/708

We need file sizes in Cocina to keep the embed viewer happy. Some items, probably all pre-dating the Fedora migration, may have files where file size was never recorded.

**Ideal output:**
druid,list,of,files,without,size

**Acceptable output:**
druid (only)

jcoyne:
There were only four items:
```
dor_services@dor-services-app-prod-b:~/dor_services/current$ bin/rails r -e production "FilesWithNoSize.report" > files_with_no_size_attribute.csv
dor_services@dor-services-app-prod-b:~/dor_services/current$ wc -l files_with_no_size_attribute.csv
5 files_with_no_size_attribute.csv
dor_services@dor-services-app-prod-b:~/dor_services/current$ cat files_with_no_size_attribute.csv
item_druid
druid:pb386cr1662
druid:np008tj3137
druid:hg262qg8763
druid:yx310tq1633
```

lwrubel:
@andrewjbtw what would you advise about these four druids?

andrewjbtw:
Two of these items have problems related to being partially but not actually accessioned:
- druid:hg262qg8763
- druid:pb386cr1662
Both are from the same project. I've contacted the content owner about one of them and wasn't aware of the other one before. I can try to work out what to do. The issue is that the files are listed in the Cocina because accessioning must have been started, but accessioning was never completed.

These two I think are fine now and came up in the report because they were still in accessioning:
- druid:np008tj3137
- druid:yx310tq1633

In the big picture, I don't think we can require file size in Cocina (for https://github.com/sul-dlss/cocina-models/issues/708) unless we change accessioning so that file size is generated differently. In assemblyWF, I don't think file size is computed until `exif-collect`. That's probably because assemblyWF doesn't have the full set of files for an object until after `jp2-create` has run. This must be why we never required file size in the first place.

PR #5001, opened 2024-05-17
Create a report for files with no size

jcoyne:
Ref #4990


justinlittman (approved):

app/reports/files_with_no_size.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Invoke via:
+# bin/rails r -e production "FilesWithNoSize.report"
+class FilesWithNoSize
```
justinlittman, line 5: Maybe a comment indicating that this is missing the size attribute, not size is 0.

PR #5022, opened 2024-05-20
Switch model used by migrators from legacy classes to RepositoryObject

### LINKED ISSUE #5015: Update migrators framework to use RepositoryObjects.
It is currently using Dro / AdminPolicy / Collection.

If possible, update all of the existing migrators. If onerous, they can be deleted as long as the exemplar is retained.

mjgiarlo:
# Why was this change made?

Fixes #5015

# How was this change tested?

CI


app/services/migrate_release_tags.rb
```diff
@@ -5,19 +5,17 @@
 # bin/rails r -e production "MigrateReleaseTags.run"
 class MigrateReleaseTags
```
justinlittman, line 6: This can just be deleted.

justinlittman (approved):
Approve pending comment.

app/services/migrate_release_tags.rb
mjgiarlo: Excellent, will do.

PR #5024, opened 2024-05-21
Use version XML from RepositoryObject model in PreservationMetadataExtractor

### LINKED ISSUE #5020: Update PreservationMetadataExtractor to use RepositoryObjectVersions
instead of ObjectVersions, which are deprecated.

mjgiarlo:
# Why was this change made?

Fixes #5020

# How was this change tested?

CI

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -108,6 +108,19 @@ class RepositoryObject < ApplicationRecord
     head_version == last_closed_version
   end
 
+  # @return [String] xml representation of version metadata
+  def version_xml
+    Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
+      xml.versionMetadata({ objectId: external_identifier }) do
+        versions.each do |object_version|
+          xml.version({ versionId: object_version.version }.compact) do
+            xml.description(object_version.version_description)
+          end
+        end
+      end
+    end.to_xml
+  end
+
   private
 
   def open_first_version
--- a/app/services/preservation_metadata_extractor.rb
+++ b/app/services/preservation_metadata_extractor.rb
@@ -38,7 +38,7 @@ class PreservationMetadataExtractor
   end
 
   def version_xml
-    ObjectVersion.version_xml(cocina_object.externalIdentifier)
+    RepositoryObject.find_by!(external_identifier: cocina_object.externalIdentifier).version_xml
   end
 
   def content_xml
--- a/spec/factories/repository_object_versions.rb
+++ b/spec/factories/repository_object_versions.rb
@@ -7,7 +7,7 @@ FactoryBot.define do
       is_member_of { [] }
       source_id { "sul:#{SecureRandom.uuid}" }
     end
-    sequence(:version)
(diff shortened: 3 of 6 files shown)


app/models/repository_object.rb
```diff
@@ -108,6 +108,19 @@ def closed?
...
+    Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
+      xml.versionMetadata({ objectId: external_identifier }) do
+        versions.each do |object_version|
+          xml.version({ versionId: object_version.version }.compact) do
+            xml.description(object_version.version_description)
+          end
+        end
+      end
+    end.to_xml
+  end
```
mjgiarlo, line 122: Ported from `ObjectVersion.version_xml`

app/services/preservation_metadata_extractor.rb
```diff
@@ -38,7 +38,7 @@ def generate_xml
   end
 
   def version_xml
-    ObjectVersion.version_xml(cocina_object.externalIdentifier)
+    RepositoryObject.find_by!(external_identifier: cocina_object.externalIdentifier).version_xml
```
mjgiarlo, line 41: Stop using the deprecated `ObjectVersion` model.

spec/factories/repository_object_versions.rb
```diff
@@ -7,7 +7,7 @@
       is_member_of { [] }
       source_id { "sul:#{SecureRandom.uuid}" }
     end
-    sequence(:version)
+    version { 1 }
```
mjgiarlo, line 10: I don't think we want or need to use a sequence here. We only need version numbers to be unique *within* a given `RepositoryObject` association. Instead provide a sensible default and expect tests to override this value when a higher version is needed. Without this change, I saw unexpected behavior in `spec/services/preservation_metadata_extractor_spec.rb`, namely an object with two versions: one version `1` and another version `n` which could range from a single-digit to a double-digit one.

spec/factories/repository_objects.rb
```diff
@@ -25,7 +25,7 @@
   trait :with_repository_object_version do
     transient do
       repository_object_version { nil }
-      sequence(:version)
+      version { 1 }
```
mjgiarlo, line 28: I don't think we want or need to use a sequence here. We only need version numbers to be unique *within* a given `RepositoryObject` association. Instead provide a sensible default and expect tests to override this value when a higher version is needed. Without this change, I saw unexpected behavior in `spec/services/preservation_metadata_extractor_spec.rb`, namely an object with two versions: one version `1` and another version `n` which could range from a single-digit to a double-digit one.

spec/factories/repository_object_versions.rb
```diff
@@ -87,8 +87,7 @@
     end
   end
 
-  trait :dro_repository_object_version do # rubocop:disable Lint/EmptyBlock
-  end
+  trait :dro_repository_object_version
```
mjgiarlo, line 90: We don't need the empty block.

justinlittman (approved):

PR #5033, opened 2024-05-21
Use RepositoryObject for locking.

### LINKED ISSUE #5030: Change locks to use RepositoryObject

justinlittman:
closes #5030

Legacy models are going away.

Unit


app/models/repository_object.rb
```diff
@@ -96,8 +98,9 @@ def close_version!(description: nil)
     raise VersionNotOpened, "Cannot close version because head version is closed: #{head_version.version}" if closed?
 
     RepositoryObject.transaction do
-      opened_version.update!(closed_at: Time.current, version_description: description || opened_version.version_description)
-      update!(opened_version: nil, last_closed_version: opened_version, head_version: opened_version)
+      closing_version = opened_version
+      update!(opened_version: nil, last_closed_version: closing_version, head_version: closing_version)
+      closing_version.update!(closed_at: Time.current, version_description: description || closing_version.version_description)
```
mjgiarlo, line 103: Can you say more about this change?

app/models/repository_object_version.rb
```diff
@@ -85,6 +78,7 @@ def update_object_source_id
     return unless repository_object.opened_version == self && source_id
     return if repository_object.source_id == source_id
 
-    repository_object.update!(source_id:)
+    # Reloading to avoid ActiveRecord::StaleObjectError
```
mjgiarlo, line 81: This is presumably because the lock changed (for some reason)?

app/services/version_service.rb
```diff
@@ -80,12 +80,13 @@ def open(cocina_object:, description:, assume_accessioned:, opening_user_name: n
...
     update_cocina_object = cocina_object
-    RepositoryObject.find_by!(external_identifier: druid).open_version!(description:)
+    repository_object = RepositoryObject.find_by!(external_identifier: druid)
+    repository_object.open_version!(description:)
     # TODO: when we stop calling the UpdateObjectService, after we've migrated to RepostoryObjects, we may need to trigger indexing:
     #       e.g.: Notifications::ObjectUpdated.publish(model: cocina_object_with_metadata)
 
     # Skipping open check since not yet opened.
-    update_cocina_object = UpdateObjectService.update(cocina_object.new(version: new_object_version.version), skip_open_check: true) if cocina_object.version != new_object_version.version
+    update_cocina_object = UpdateObjectService.update(cocina_object.new(version: new_object_version.version, lock: repository_object.reload.external_lock), skip_open_check: true) if cocina_object.version != new_object_version.version
```
mjgiarlo, line 89: With the various `reload`s and at least one spot where we need a manual `check_lock!`, I'm wondering whether it might be worth a story-time at some point. I can see either of these bits being difficult to remember as we continue to tweak and extend DSA. Maybe it's just me? cc: @sul-dlss/infrastructure-team

mjgiarlo (commented):
I can't say I grok how DSA is using locking, at least not to the extend that all the reloads and such make sense to me. Maybe worth some team clumping to push this along?

app/models/repository_object.rb
justinlittman: This makes sure that updates are performed in an order that does not result in a stale error. (Updating the head version causes the lock on the RepositoryObject to be incremented.)

app/models/repository_object_version.rb
justinlittman: Yes, this is called after save so the repository object has changed.

app/services/version_service.rb
justinlittman: Note that L89 is going away once shortly, so that added complexity is only temporary.

This is confusing. I'd suggest that for long term maintainability, well-placed code comments might be helpful. Let me add a few more.

app/models/repository_object.rb
mjgiarlo: OK, makes sense. Comment-worthy?

justinlittman:
@mjgiarlo I added comments to try to make this clearer to our future selves.

mjgiarlo (approved):
Appreciate the replies & new comments! Let's roll with this. Feel free to merge once the merge conflict's gone.

PR #5035, opened 2024-05-21
Raise when version mismatch with Preservation.

### LINKED ISSUE #5034: Raise when version mismatch with Preservation
Instead of the current approach of removing any extra ObjectVersions.

justinlittman:
closes #5034

ObjectVersions are going to be removed, but still want to make version check with pres.

Unit

### DIFF
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 # Open and close versions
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):
LGTM, but would be good to know @andrewjbtw is on board with this change before we ship it.

justinlittman:
@andrewjbtw This change just makes sure that we continue to check that DSA versions are in sync with Pres.

PR #5042, opened 2024-05-24
allow the workspace call to return the location created

peetucket:

ocrWF will be making a call to create the relevant `/dor/workspace` druid tree directory. It will not need the source path and symlink, it just wants to the workspace folder.   So this:

- removes the HB alert since we this will be the first consumer to not be sending in the source
- allows DSA to return the directory location created, so it can be used by the caller
- allows client to request metadata and/or content subdirectories also be created (but only if no source is passed in, since if source is passed in, the subdirectories should already exist at the source, and the created directory is actually just a symlink there)

There is a corresponding [dor-services-client PR](https://github.com/sul-dlss/dor-services-client/pull/436) to grab the location and return to the consumer.

Updated existing spec
Created a new spec
Localhost using updated dor-services-client: https://github.com/sul-dlss/dor-services-client/pull/436

### DIFF
--- a/app/controllers/workspaces_controller.rb
+++ b/app/controllers/workspaces_controller.rb
@@ -11,7 +11,7 @@ class WorkspacesController < ApplicationController
   # POST /v1/objects/:druid/workspace
   def create
     result = WorkspaceService.create(params[:object_id], params[:source])
-    head :created, location: result
+    render status: :created, json: { path: result }
   end
 
   def destroy
--- a/app/controllers/workspaces_controller.rb
+++ b/app/controllers/workspaces_controller.rb
@@ -10,7 +10,7 @@ class WorkspacesController < ApplicationController
 
   # POST /v1/objects/:druid/workspace
   def create
-    result = WorkspaceService.create(params[:object_id], params[:source])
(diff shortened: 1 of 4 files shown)


app/controllers/workspaces_controller.rb
```diff
@@ -10,8 +10,8 @@ class WorkspacesController < ApplicationController
 
   # POST /v1/objects/:druid/workspace
   def create
-    WorkspaceService.create(params[:object_id], params[:source])
-    head :created
+    result = WorkspaceService.create(params[:object_id], params[:source])
+    head :created, location: result
```
jcoyne, line 14: This seems like abuse of the location header. This response header is supposed to contain a URL.

jcoyne (changes requested):

app/controllers/workspaces_controller.rb
peetucket: Moved to body.

peetucket:
Added the ability to request the content and/or metadata subfolders be created as well.  This will be useful for the OCR work too (where we want the content subfolder if it doesn't exist).

PR #5049, opened 2024-05-29
Stop asking workflow service to close version.

### LINKED ISSUE #5047: Stop closing versioningWF
At https://github.com/sul-dlss/dor-services-app/blob/main/app/services/version_service.rb#L132

Instead, create an accessioningWF.

justinlittman:
closes #5047

The versioningWF now only has one step and that step is completed, so no need to close.

Unit, stage integration

### DIFF
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -128,12 +128,8 @@ class VersionService
 
     repository_object = RepositoryObject.find_by!(external_identifier: druid)
 
-    # Default to creating accessionWF when calling close_version
-    workflow_client.close_version(druid:,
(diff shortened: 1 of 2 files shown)


edsu (approved):
So it looks like the workflow_client was accidentally being used to close the version?

justinlittman:
No, it was intentional. In the past, the way to close a version was to tell the workflow service to end all of the steps in the versioningWF. After doing that it also kicked off the accessioningWF.

Now, we no longer need to end all of the steps in the versioningWF and DSA is responsible for kicking off accessioningWF.

PR #5050, opened 2024-05-30
Add report to find repository objects with files shelved but not publ…

### LINKED ISSUE #5045: Create a report for shelve: yes, publish: no
record the APO to identify an ETDs, which are expected to have this state
and the content-type to identify webarchive-binary which also have this state.

aaron-collier:
Closes #5045 

Adds a report for all objects that have files where `shelve = true` and `publish = false`.

### DIFF
--- a/app/reports/shelved_not_published.rb
+++ b/app/reports/shelved_not_published.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'json'
+
+# Generates a report of repository objects with files that are
+# shelved but not published
+
+# bin/rails r -e production "ShelvedNotPublished.report"
+#
+class ShelvedNotPublished
+  SQL = <<~SQL.squish.freeze
+    SELECT ro.external_identifier,
+      jsonb_path_query(rov.administrative, '$.hasAdminPolicy') AS admin_policy,
+      rov.content_type,
+      jsonb_path_query_array(rov.structural, '$.contains[*].structural.contains[*].administrative[*].shelve') AS shelve,
+      jsonb_path_query_array(rov.structural, '$.contains[*].structural.contains[*].administrative[*].publish') AS publish
+      FROM repository_objects AS ro, repository_object_versions AS rov
+      WHERE ro.head_version_id = rov.id
+      AND ro.object_type = 'dro'
+  SQL
+
+  def self.report
+    puts 'druid,admin_policy,type'
+    ActiveRecord::Base.connection.execute(SQL).each do |row|
+      next unless row['shelve'].include?('true') && row['publish'].include?('false')
+
+      JSON.parse(row['shelve']).each_with_index do |val, key|
+        next unless val == true
+        next unless JSON.parse(row['publish'])[key] == false
+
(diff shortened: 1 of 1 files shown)


justinlittman (approved):

app/reports/shelved_not_published.rb
```diff
@@ -0,0 +1,35 @@
...
+      rov.content_type,
+      jsonb_path_query_array(rov.structural, '$.contains[*].structural.contains[*].administrative[*].shelve') AS shelve,
+      jsonb_path_query_array(rov.structural, '$.contains[*].structural.contains[*].administrative[*].publish') AS publish
+      FROM repository_objects AS ro, repository_object_versions AS rov
+      WHERE ro.head_version_id = rov.id
+      AND ro.object_type = 'dro'
+  SQL
+
+  def self.report
+    puts 'druid,admin_policy,type,shelve,publish'
```
mjgiarlo, line 23: The header suggests there are five columns and line 31 looks like it's only spitting out three columns. Did I get that right?

app/reports/shelved_not_published.rb
```diff
@@ -0,0 +1,35 @@
...
+    ActiveRecord::Base.connection.execute(SQL).each do |row|
+      next unless row['shelve'].include?('true') && row['publish'].include?('false')
+
+      JSON.parse(row['shelve']).each_with_index do |val, key|
+        next unless val == true
+        next unless JSON.parse(row['publish'])[key] == false
```
mjgiarlo, line 29: `key` here is the index injected by `each_with_index`, yeah? (No changes needed, just confirming.)

app/reports/shelved_not_published.rb
```diff
@@ -0,0 +1,35 @@
...
+      WHERE ro.head_version_id = rov.id
+      AND ro.object_type = 'dro'
+  SQL
+
+  def self.report
+    puts 'druid,admin_policy,type,shelve,publish'
+    ActiveRecord::Base.connection.execute(SQL).each do |row|
+      next unless row['shelve'].include?('true') && row['publish'].include?('false')
+
+      JSON.parse(row['shelve']).each_with_index do |val, key|
```
mjgiarlo, line 27: I assume that we always expect the count of `row['shelve']` to match the count of `row['publish']`, but it's prolly not worth checking that given Cocina wouldn't allow anything else? Confirm/deny?

mjgiarlo (approved):
No blockers, just some minor questions. Approving

app/reports/shelved_not_published.rb
aaron-collier: Cruft, good catch, fixing.

app/reports/shelved_not_published.rb
aaron-collier: Yes, because it's an array of the values from the object files which always have `shelve`, `publish`, and `sdrPreserve` (which isn't getting queried here).

mjgiarlo: 🚀

PR #5056, opened 2024-05-30
Update docker base image to debian base

aaron-collier:

Update to a debian based docker image in order to support recent version(s) of openapi_parser and use a base closer to our deployed environments.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,6 +45,8 @@ services:
       SETTINGS__SOLR__URL: http://solr:8983/solr/dorservices
       SETTINGS__SURI__URL: http://suri:3000
       SETTINGS__WORKFLOW_URL: http://workflow:3000
+      # We don't actually use this anywhere but the rails server needs it in production env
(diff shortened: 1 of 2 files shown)


jcoyne (approved):

docker-compose.yml
```diff
@@ -45,6 +45,8 @@ services:
       SETTINGS__SOLR__URL: http://solr:8983/solr/dorservices
       SETTINGS__SURI__URL: http://suri:3000
       SETTINGS__WORKFLOW_URL: http://workflow:3000
+      # We don't actually use this anywhere but the rails server needs it in production env
+      SECRET_KEY_BASE: e0221b0233dbe4914fae9405c1e179eb1db71379fd999c51265123a6dde45c8281235307dda0b8c06633b702a05679391a950a483f8c624642eb3c3211d4241d
```
jcoyne, line 49: @aaron-collier can we set https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt#L86 SECRET_KEY_BASE_DUMMY=1 instead?

PR #5066, opened 2024-06-03
Update validate-cocina to use RepositoryObject

### LINKED ISSUE #5065: Update validate-cocina to use RepositoryObjects
Now that Dro/Collection/AdminPolicy no longer exist, validat-cocina doesn't work until updated to use RepositoryObject

aaron-collier:

Fixes #5065 by switching from the removed `Dro`, `Collection`, and `AdminPolicy` models to `RepositoryObject`

Ran this branch on sdr-infra against stage

### DIFF
--- a/bin/validate-cocina
+++ b/bin/validate-cocina
@@ -66,8 +66,8 @@ def validate(clazz, sample_size, processes)
       Cocina::Models::Mapping::ToMods::Description.transform(cocina_obj.description, cocina_obj.externalIdentifier)
       [obj.external_identifier, nil]
     rescue StandardError => e
-      collection = if obj.is_a?(Dro)
-                     obj.structural['isMemberOf'].join(' ')
+      collection = if obj.object_type == 'dro'
+                     obj.head_version.structural['isMemberOf'].join(' ')
                    else
                      ''
                    end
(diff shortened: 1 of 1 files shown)


bin/validate-cocina
```diff
@@ -76,8 +76,8 @@ def validate(clazz, sample_size, processes)
   end.flatten(1)
 end
 
-results = validate(Dro, options[:sample],
-                   options[:processes]) + validate(Collection, options[:sample], options[:processes]) + validate(AdminPolicy, options[:sample], options[:processes])
+results = validate(RepositoryObject.dros, options[:sample],
```
jmartin-sul, line 79: with this change, i wonder if we might want to rename the `validate` method's first param from `clazz` to something like `ar_relation`, since the type of the parameter we're passing in seems to be changing from a class to an active record relation?  and then if that's right, `validate` has a few uses of `clazz.name` that might need to be updated to e.g. `ar_relation.class.name`?

justinlittman, line 79: Is there a need for separate runs by type anymore? Can't there just be a single run of all RepositoryObjects?

jmartin-sul (commented):
a question about a change to the first parameter that's passed to `validate`

bin/validate-cocina
aaron-collier: Sure - I started down that path and thought it might be clearer to keep not change the shape, but happy to work through that.

aaron-collier: @jmartin-sul it seemed better to just remove the param :-)

aaron-collier: @justinlittman - Updated.

jmartin-sul (approved):

bin/validate-cocina
jmartin-sul: makes sense given the change justin suggested above!

approved, feel free to merge if you want, or to wait from feedback from @justinlittman if you'd prefer.

PR #5068, opened 2024-06-04
Restore the remove release tags migratator but for repository objects

aaron-collier:

Restores the remove release tags migrator so that we can officially remove them from cocina-models.

Ran the migration in stage and then validated the cocina against the https://github.com/sul-dlss/cocina-models/pull/721 branch of cocina-models

### DIFF
--- a/app/services/migrators/remove_release_tags.rb
+++ b/app/services/migrators/remove_release_tags.rb
@@ -10,7 +10,6 @@ module Migrators
     def migrate
       repository_object.versions.select { |version| version.administrative&.key?('releaseTags') }.each do |version|
         version.administrative.delete('releaseTags')
-        version.save!
       end
     end
   end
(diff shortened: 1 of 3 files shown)


app/services/migrators/remove_release_tags.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Migrators
+  # Migrator that will be used to remove release tags.
+  class RemoveReleaseTags < Base
+    def migrate?
+      repository_object.head_version.administrative.key?('releaseTags')
```
justinlittman, line 7: You're going to need to migrate every version, not just the head version.

aaron-collier: Ah, OK - Good to know. I'll update

justinlittman (approved):

PR #5071, opened 2024-06-05
Replace Cocina::Models::ReleaseTag with Dor::ReleaseTag

### LINKED ISSUE #5070: Replace Cocina::Models::ReleaseTag with local ReleaseTag model
This is a step towards removing ReleastTags from cocina-models. A mirror of: https://github.com/sul-dlss/dor-services-client/pull/439

aaron-collier:

Fixes #5070 

### DIFF
--- a/app/models/dor/release_tag.rb
+++ b/app/models/dor/release_tag.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Dor
+  # Include the Dry types module
+  module Types
+    include Dry.Types()
+  end
+
+  # A tag that indicates the item or collection should be released.
+  # This duplicates the model in dor-services-client:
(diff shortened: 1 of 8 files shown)


app/models/dor/release_tag.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Dor
+  # Include the Dry types module
+  module Types
+    include Dry.Types()
+  end
+
+  # A tag that indicates the item or collection should be released.
```
justinlittman, line 9: Can you put a comment that this is similar to class in DSC?

justinlittman (approved):

app/models/dor/release_tag.rb
aaron-collier: Doh - I meant to do that and got caught up in testing and getting this pushed up. Fixed. Will merge when green.

PR #5076, opened 2024-06-07
Only shelve WAS files via ShelveJob

### LINKED ISSUE #5072: Replace existing shelving service with WARC-specific shelving service
With the new publish-shelve API, the only files that need to be shelved are WARC files to the web archive stacks. The existing logic for this (ShelvingService) is for general shelving and involves a complicated comparison with metadata from preservation.

For the current state of shelving, see https://github.com/sul-dlss/dor-services-app/blob/main/app/jobs/shelve_job.rb.

The existing shelving service should be replaced with a WARC-specific shelving service that for every file that is both in the cocina structural and in the workspace, copies the file to the web archive stacks.

For additional context, see #5046.

Please mark the PR as HOLD, as it is blocked by enabling the new publish-shelve API in production.

lwrubel:
Resolves #5072 for shelving web archiving files only. No longer held since we are ready to use purl-fetcher for shelving all other files in production. 

Unit and integration tests on stage. Re-tested since re-based.


.rubocop_todo.yml
```diff
@@ -67,6 +68,7 @@ RSpec/Be:
 RSpec/BeforeAfterAll:
   Exclude:
     - 'spec/services/digital_stacks_service_spec.rb'
```
justinlittman, line 70: Looks like there are TODOs for files that you deleted.

Is it possible to not add any new TODOs? Either disable the rules if will never enforce or add inline disables.

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+class WasShelvingService
+  class ConfigurationError < RuntimeError; end
+  class WasShelvingError < StandardError; end
+
+  def self.shelve(cocina_object)
+    new(cocina_object).shelve
+  end
+
+  def initialize(cocina_object)
+    raise ConfigurationError, 'Missing configuration Settings.stacks.local_workspace_root' if Settings.stacks.local_workspace_root.nil?
```
justinlittman, line 13: I would just raise StandardError here; no reason to create a special error class.

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+    raise WasShelvingService::WasShelvingError, 'Missing structural' if cocina_object.structural.nil?
```
justinlittman, line 14: Why would this happen? (Also, I would put this in the shelve method instead of the initializer.)

justinlittman, line 14: Why would this happen? (Also, I would put this in the shelve method instead of the initializer.)

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    @cocina_object = cocina_object
+    @druid = cocina_object.externalIdentifier
+  end
+
+  attr_reader :cocina_object, :druid
+
+  def shelve
+    # determine destination web archiving stacks location
```
justinlittman, line 23: I'd suggest moving these into methods to make this method clearer.

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def was_stacks_location
+    collection_druid = cocina_object.structural&.isMemberOf&.first
+
+    raise WasShelvingService::WasShelvingError, 'Web archive object missing collection' unless collection_druid
+
+    "/web-archiving-stacks/data/collections/#{collection_druid.delete_prefix('druid:')}"
+  end
+
+  def files_for(cocina_object)
```
justinlittman, line 47: cocina_object is an instance variable so no need to pass in.

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+      copy_file(workspace_pathname, stacks_pathname) if File.exist?(workspace_pathname)
+    end
+  end
+
+  def was_stacks_location
+    collection_druid = cocina_object.structural&.isMemberOf&.first
+
+    raise WasShelvingService::WasShelvingError, 'Web archive object missing collection' unless collection_druid
+
+    "/web-archiving-stacks/data/collections/#{collection_druid.delete_prefix('druid:')}"
```
justinlittman, line 44: Should `/web-archiving-stacks` be hardcoded?

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  def copy_file(workspace_pathname, stacks_pathname)
+    stacks_pathname.parent.mkpath unless stacks_pathname.parent.exist?
+
+    Rails.logger.debug("[Was Shelve] Copying #{workspace_pathname} to #{stacks_pathname}")
+    FileUtils.cp workspace_pathname.to_s, stacks_pathname.to_s
+    # Change permissions
+    FileUtils.chmod 'u=rw,go=r', stacks_pathname.to_s
```
justinlittman, line 63: Why do we do this? (Can you add reason to L62?)

.rubocop_todo.yml
lwrubel: Thanks for catching those deleted files. Disabled rules.

app/services/was_shelving_service.rb
lwrubel: Done

app/services/was_shelving_service.rb
lwrubel: This was a possible situation that was being handled when shelving other files, but since there is really only one pathway for web archive files to get accessioned, seems extremely unlikely. I've taken it out.

app/services/was_shelving_service.rb
lwrubel: Agree that is an improvement. Moved the methods.

app/services/was_shelving_service.rb
lwrubel: Duh, thanks.

app/services/was_shelving_service.rb
lwrubel: I've added a setting for it.

app/services/was_shelving_service.rb
lwrubel: I did some more testing, and I think that may not be a concern with web archiving stacks. Seems like the permissions were already as needed. I'll check again when testing on stage just to be sure.

app/services/was_shelving_service.rb
lwrubel: Also updated tests.

app/services/was_shelving_service.rb
lwrubel: Checked on stage and permissions are being set to 644. No need to change them.

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+    # get the list of shelvable files from cocina and shelve those that are available in the workspace
+    files_for.each do |file|
+      workspace_pathname = workspace_content_dir.join(file)
+      was_stacks_pathname = was_stacks_dir.join(file)
+      copy_file(workspace_pathname, was_stacks_pathname) if File.exist?(workspace_pathname)
+    end
+  end
+
+  def was_stacks_dir
+    # determine destination web archiving stacks directory
```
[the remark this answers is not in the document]
lwrubel: Done

app/services/was_shelving_service.rb
```diff
@@ -0,0 +1,65 @@
...
+    "#{Settings.stacks.web_archiving_stacks}/#{collection_druid.delete_prefix('druid:')}"
+  end
+
+  def workspace_content_dir
+    # determine current workspace location of object's content file
+    workspace_druid = DruidTools::Druid.new(druid, Settings.stacks.local_workspace_root)
+    Pathname(workspace_druid.content_dir(true))
+  end
+
+  def files_for
```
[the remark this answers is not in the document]
lwrubel: Ah, I see. Thanks for the suggestion.

lwrubel:
Be sure to squash when merging.

justinlittman (approved):

PR #5080, opened 2024-06-12
[HOLD] add endpoint to check if text extraction workflows are running

peetucket:

Part of sul-dlss/argo#4470

We need the ability to know if an object is in the middle of a text extraction WF (e.g. ocrWF) so we can disable various things (like editing or starting a new ocrWF in Argo).

HOLD for further changes in dor-services-client and Argo

see also https://github.com/sul-dlss/argo/pull/4460 and https://github.com/sul-dlss/argo/issues/4458 for related work in assemblyWF

Updated tests

### DIFF
--- a/app/services/workflow_state_service.rb
+++ b/app/services/workflow_state_service.rb
@@ -31,6 +31,14 @@ class WorkflowStateService
       active_workflow?(workflow: 'gisAssemblyWF')
   end
 
+  # Checks if the latest version has any text extraction workflows with incomplete steps.
+  # @return [Boolean] true if object is currently having text extracted
+  def text_extracting?
+    # Omitting the last step for these workflows since the last step is closing the version.
+    # NOTE: we can add captioning workflows here if/when they are created
+    active_workflow_except_step?(workflow: 'ocrWF', process: 'end-ocr')
+  end
+
   # The following methods were extracted from VersionService.
   # As such, they may not represent the current best practice for determining workflow state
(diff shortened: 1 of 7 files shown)


app/services/workflow_state_service.rb
```diff
@@ -31,6 +31,14 @@ def assembling?
       active_workflow?(workflow: 'gisAssemblyWF')
   end
 
+  # Checks if the latest version has any text extraction workflows with incomplete steps.
+  # @return [Boolean] true if object is currently having text extracted
+  def text_extracting?
+    # Omitting the last step for these workflows since the last step is closing the version.
+    # NOTE: we can add captioning workflows here if/when they are created
+    active_workflow_except_step?(workflow: 'ocrWF', process: 'end-ocr')
```
dnoneill, line 39: Shouldn't this be active_workflow?, not active_workflow_except_step?

peetucket: The reason it is this way is the same reason as for "assembling?" above --- there are other parts of the code that ensure the object is closeable (e.g. https://github.com/sul-dlss/dor-services-app/blob/main/app/services/version_service.rb#L192-L196) and I think we want to make the object not closeable if it is in the middle of text extracting too.  But when we do that, end-ocr tries to close the object and is blocked, because it itself is what is doing the closing.

peetucket:
Closing in favor of simpler approach.

PR #5102, opened 2024-06-24
Set resourceType to an empty string for Datacite if it's nil

aaron-collier:

This avoids including the option `resourceType` in the datacite package if it is nil.

### DIFF
--- a/app/services/cocina/to_datacite/form.rb
+++ b/app/services/cocina/to_datacite/form.rb
@@ -19,10 +19,10 @@ module Cocina
       def type_attributes
         return unless resource_type_general || resource_type
 
-        {
-          resourceTypeGeneral: resource_type_general,
-          resourceType: resource_type
-        }
+        {}.tap do |types_attributes|
+          types_attributes[:resourceTypeGeneral] = resource_type_general
+          types_attributes[:resourceType] = resource_type || ''
+        end
       end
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):

mjgiarlo:
@aaron-collier Approved. Might be worth adding a test or it might not depending on how the tests are currently written and how much work it'd add to this PR.

app/services/cocina/to_datacite/form.rb
```diff
@@ -19,10 +19,10 @@ def initialize(cocina_desc)
...
         return unless resource_type_general || resource_type
 
-        {
-          resourceTypeGeneral: resource_type_general,
-          resourceType: resource_type
-        }
+        {}.tap do |types_attributes|
+          types_attributes[:resourceTypeGeneral] = resource_type_general
+          types_attributes[:resourceType] = resource_type || ''
+        end
```
mjgiarlo, line 25: Adding the `|| ''` means we don't need the `tap` jazz now, I think?

```suggestion
        {
          resourceTypeGeneral: resource_type_general,
          resourceType: resource_type || ''
        end
```

aaron-collier: Probably, I can test that.

PR #5105, opened 2024-06-25
Adds create and update endpoints for user versions.

### LINKED ISSUE #5098: Create user version create and update endpoints
This should allow:
* Add a new user version.
* Moving a user version to a different object version.
* Withdrawing and unwithdrawing

justinlittman:
closes #5098

Endpoints are going to be necessary to manage user versions from Argo.

Unit

### DIFF
--- a/spec/requests/create_user_version_spec.rb
+++ b/spec/requests/create_user_version_spec.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Create user version' do
+  let(:repository_object) { repository_object_version.repository_object }
+
+  context 'when the user version can be created' do
+    let(:repository_object_version) { create(:repository_object_version, :with_repository_object, closed_at: Time.zone.now) }
+
+    it 'returns created' do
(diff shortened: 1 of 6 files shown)


spec/requests/create_user_version_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  context 'when the user version cannot be created' do
+    let(:repository_object_version) { create(:repository_object_version, :with_repository_object) }
+
+    it 'returns unprocessable' do
+      post "/v1/objects/#{repository_object.external_identifier}/user_versions",
+           headers: { 'Authorization' => "Bearer #{jwt}", 'Content-Type' => 'application/json' },
+           params: { version: repository_object_version.version }.to_json
+
+      expect(response).to have_http_status(:unprocessable_entity)
+      # .parsed_body won't work here because of content-type
```
jcoyne, line 32: What about content type?

justinlittman: Comment updated to provide more complete explanation.

jcoyne (approved):

PR #5135, opened 2024-07-24
Check that shelving was successful.

### LINKED ISSUE #5117: Add a basic check that the "shelve" step ran correctly
**Background**

When we shelve files during accessionWF, the result should be that the /stacks filesystem ends up with exactly the files listed as "publish=yes, shelve=yes" in the public Cocina. If there are no shelved files, the filesystem for that druid (or that druid version) should be empty.

We do not currently have a check that shelving has succeeded. Over the years, we've found numerous cases where the filesystem does not match the Cocina. In testing new features, we've also missed cases where shelving did not do what it was supposed to do. The shelve step rarely reports anything other than success - even if files did not get shelved (or unshelved) as expected.

**The request**

We should add a basic check in accessioning to verify that the Stacks filesystem is in sync with the Cocina. As a sanity check, this could be as basic as checking that the filenames and paths line up properly between Stacks and Cocina. To be thorough, we could check checksums, but this could cause significant slowdowns for large deposits.

In the coming versioned world, the check should verify that the correct files are present for each version of the druid. The "withdraw" use case means that there can be changes to the files of a previous version. Withdrawal means that a version will be taken down, so files unique to that version should be removed.  

**How this will help**
Having this basic check will likely save us a lot of time in the long run. For testing and feature development, it would lessen the burden on testers to watch the Stacks filesystem closely. Currently, a person has to be in the Stacks directory for a specific druid and eyeball the before and after states in order to verify shelving success.

This check should also prevent situations like the one we had where we found over 1 million files on Stacks that should not have been there.

justinlittman:
closes #5117

So that @andrewjbtw can sleep well at night.

Unit

### DIFF
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -24,6 +24,7 @@ module Publish
       return publish_delete_on_success unless discoverable?
 
       publish_shelve
+      check_stacks if public_cocina.dro?
       republish_virtual_object_constituents!
       release_tags_on_success
     end
@@ -76,6 +77,12 @@ module Publish
       end
       PurlFetcher::Client::Publish.publish(cocina: public_cocina, file_uploads: filepath_map, version: public_version,
                                            must_version: must_version?, version_date:)
+    end
+
+    def check_stacks
+      missing_filepaths = DigitalStacksDiffer.call(cocina_object: public_cocina)
+
+      raise "Files are missing from stacks: #{missing_filepaths}" if missing_filepaths.present?
     end
 
     def filepaths_to_shelve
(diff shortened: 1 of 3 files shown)


justinlittman:
HOLD pending https://github.com/sul-dlss/purl-fetcher/issues/876

app/services/publish/metadata_transfer_service.rb
```diff
@@ -24,6 +24,7 @@ def publish
       return publish_delete_on_success unless discoverable?
 
       publish_shelve
+      check_shelf if public_cocina.dro?
```
mjgiarlo, line 27: Minor naming tidbit: should we consider renaming `shelf`/`shelves` on lines 27, 82, and 85 to `stacks`? `shelf` is logical, considering we're shelving, but using this word in this context is new to me.

justinlittman: Agree. Fixing.

PR #5137, opened 2024-07-24
Adds additional to user version log. Adds validation so that cannot withdraw head.

### LINKED ISSUE #5136: Add user version status endpoint and validate user version withdraw changes
For a particular user version, the status should provide:
* Is the version withdrawn?
* Can the version be withdrawn?
* Can the version be restored?
* Is this the head user version?

Validation rules:
* The latest user version cannot be withdrawn. (This also includes when there is only a single user version.)
* Only a withdrawn user version can be restored.
* Only a non-withdrawn user version can be withdrawn.

The validation rules should also be applied when updating a User Version.

justinlittman:
closes #5136

To support withdrawing from argo.

Unit

### DIFF
--- a/app/models/user_version.rb
+++ b/app/models/user_version.rb
@@ -7,6 +7,7 @@ class UserVersion < ApplicationRecord
 
   validate :repository_object_version_is_closed
   validate :repository_object_version_has_cocina
+  validate :can_withdraw
 
   def repository_object_version_is_closed
     # Validate that the repository object version is closed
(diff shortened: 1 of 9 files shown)


justinlittman:
Note that I added the additional info about the user version to an existing endpoint rather than adding new.

app/models/user_version.rb
```diff
@@ -7,28 +7,53 @@ class UserVersion < ApplicationRecord
...
 
   def repository_object_version_has_cocina
-    # Validate that the repository object version has cocina (lecacy versions may not)
+    # Validate that the repository object version has cocina (legacy versions may not)
     errors.add(:repository_object_version, 'cannot set a user version to an RepositoryObjectVersion without cocina') unless repository_object_version.has_cocina?
   end
 
+  def can_withdraw
+    # Validate that the user version can be withdrawn or restored
+    errors.add(:repository_object_version, 'head version cannot be withdrawn') if withdrawn && (head? || version.nil?)
```
mjgiarlo, line 24: Naïvely, could/should the condition here use the logic in `#withdrawable?` ?

mjgiarlo: If you think that's comment-worthy, I wouldn't mind seeing a bit more here. If not, ship it.

mjgiarlo (approved):

PR #5141, opened 2024-07-25
Support new user version mode for update if exists.

### LINKED ISSUE #5140: Create user version mode for update if exists
This would perform an update only if there is an existing user version. This is unlike the existing update mode which will create a new user version if one doesn't exist.

This should be made the default mode.

justinlittman:
closes #5140

So that users can modify objects with user versions (i.e., created in H2) from Argo. Otherwise, the user version doesn't move to the latest object version.

Unit, stage

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -91,6 +91,7 @@ class VersionsController < ApplicationController
       :user_name,
       :user_versions
     ).to_h.symbolize_keys
+    new_params[:user_version_mode] = new_params.delete(:user_versions).to_sym if new_params.key?(:user_versions)
     boolean_param(new_params, :start_accession)
   end
 
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 # Open and close versions
+# rubocop:disable Metrics/ClassLength
 class VersionService
   class VersioningError < StandardError; end
 
   class CocinaObjectNotFoundError < VersioningError; end
+
+  DEFAULT_USER_VERSION_MODE = :update_if_existing
 
   # @param [String] druid of the item
   # @param [Integer] version of the item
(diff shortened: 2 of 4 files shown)


app/controllers/versions_controller.rb
```diff
@@ -91,6 +91,7 @@ def close_params
       :user_name,
       :user_versions
     ).to_h.symbolize_keys
+    new_params[:user_version_mode] = new_params.delete(:user_versions).to_sym if new_params.key?(:user_versions)
```
justinlittman, line 94: Currently, `user_versions` param is a string. Changed to `user_version_mode` and symbol, which I think is clearer.

app/services/version_service.rb
```diff
@@ -115,14 +118,14 @@ def can_open?(assume_accessioned: false)
...
   # @param [Boolean] :start_accession set to true if you want accessioning to start (default), false otherwise
-  # @param [String] :user_versions create, update, or do nothing with user_versions on close
+  # @param [Symbol] :user_version_mode :create, :update, :update_if_existing (default), or :none (do nothing) with user_versions on close
   # @raise [VersionService::VersioningError] if the object hasn't been opened for versioning, or if accessionWF has
   #   already been instantiated or the current version is missing a description
   # @raise [ArgumentError] if user_versions is not one of none, new, update
-  def close(description:, user_name:, start_accession: true, user_versions: 'none')
-    user_version_options = %w[none new update]
+  def close(description:, user_name:, start_accession: true, user_version_mode: DEFAULT_USER_VERSION_MODE)
+    user_version_mode_options = %i[none new update update_if_existing]
```
mjgiarlo, line 126: Line 44 suggests the four modes are `:create, :update, :update_if_existing, :none` which are different from the four modes here. Intended?

app/services/version_service.rb
```diff
@@ -195,15 +198,27 @@ def ensure_closeable!
...
+  def update_user_version(user_version_mode:, repository_object:)
+    case user_version_mode
+    when :new
+      create_user_version(repository_object)
+    when :update
+      no_user_versions?(repository_object) ? create_user_version(repository_object) : move_user_version(repository_object)
+    when :update_if_existing
+      move_user_version(repository_object) unless no_user_versions?(repository_object)
+    end
+  end
```
mjgiarlo, line 210: So there's no special handling for `:none`, right? It just falls through.

mjgiarlo (approved):
Approve pending consideration of comments

app/services/version_service.rb
justinlittman: No. Fixed.

app/services/version_service.rb
justinlittman: Correct. Adding comment.

app/services/version_service.rb
mjgiarlo: @justinlittman I'm still seeing `:create` on line 44 and `:new` on lines 121, 126, etc.

justinlittman: Grrr ... fixing in separate PR.

mjgiarlo: No prob! I'll review and ship it as soon as it's ready

Issue #5181, opened 2024-10-01
Identify and remediate published items that are missing public cocina in the purl-fetcher db


andrewjbtw:
This comes out of the discussion [in this Slack thread](https://stanfordlib.slack.com/archives/C056A6GDE7Q/p1727725743240559) about an item that was failing publish because of missing public Cocina JSON in the purl-fetcher DB. There are probably more items with a similar problem.

More specifically, items that meet these criteria are likely to fail `publish` until remediated:

1. Item is closed (Accessioned for in accessioning)
2. Item has been previously published, but prior to the versioning work
3. Item has rights that are not dark (i.e. there should be a purl)
4. Item does not have public cocina in the purl-fetcher DB (but may have public cocina as a static file on purl)

The publish problem seems to be that the publish step tries to diff the cocina for the currently accessioning version with the cocina for the previously-published version. But if there's no cocina in the purl-fetcher DB, the diff can't proceed.

We think it's likely that items that lack public cocina in purl-fetcher ran into a migration problem when we were populating the purl-fetcher db during the versioning work.

**Additional background**

There are also "Opened" items that don't have public cocina in the purl-fetcher DB even though they have purls. We should treat those differently and not include them in this remediation. Those items will fail a standalone publish (triggered from Argo) but if they are closed, they should successfully be accessioned. 

We had to treat "Opened" items differently when migrating to the new versioning model for reasons that I'm finding difficult to summarize concisely in this issue. I may file a separate issue about them but need to gather more information. Ideally, we can solve the issue by closing them but they may contain unfinished in-progress changes.

lwrubel:
I haven't yet figured out how to determine if the item has been previously published since I'm bumping into some problems with querying the workflow service. I'm also trying to figure out in the meantime the best way to know if that previous publishing happened prior to the versioning work.

But in the meantime, here is a report on the druids that don't have public json in the purl-fetcher DB. It includes:
* closed?
* publishable? Meaning it has a last closed version and there's cocina for that last closed version--it has been closed at least once since we moved to the new version model.
* not dark?

[publish_status_report.csv](https://github.com/user-attachments/files/17232870/publish_status_report.csv)

lwrubel:
@andrewjbtw I've updated the report to include each druid's version 1 "published" milestone date. It looks like all of these were published at least once before migration (the latest published date is 2023-07-06). 

[publish_status_report.csv](https://github.com/user-attachments/files/17250249/publish_status_report.csv)

[...]

PR #5182, opened 2024-10-03
Report about druids lacking a public json record in purl-fetcher's db

lwrubel:
Gathering info for #5181.

### DIFF
--- a/app/reports/publish_status.rb
+++ b/app/reports/publish_status.rb
@@ -15,11 +15,12 @@ class PublishStatus
       csv << %w[druid closed publishable not_dark first_published]
 
       druids.each do |druid|
-        object = RepositoryObject.where(external_identifier: druid).first
+        object = RepositoryObject.find_by(external_identifier: druid)
         latest = object.head_version
         not_dark = latest.access.fetch('view') != 'dark'
 
         milestones = workflow_client.milestones(druid:)
+        # to determine if this was published before versioning changes took effect, get the date of the first published milestone
         published = milestones.filter { |milestone| milestone[:version] == '1' && milestone[:milestone] == 'published' }
         date = published.first[:at] if published.any?
 
--- a/app/reports/publish_status.rb
+++ b/app/reports/publish_status.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# Exports publish information about druids lacking public cocina.
+#   https://github.com/sul-dlss/dor-services-app/issues/5181
+# Input is a CSV of druids generated on purl-fetcher with the query:
+#   Purl.status('public').where.missing(:public_json).pluck(:druid)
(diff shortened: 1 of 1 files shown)


app/reports/publish_status.rb
```diff
@@ -0,0 +1,35 @@
...
+
+# Exports publish information about druids lacking public cocina.
+#   https://github.com/sul-dlss/dor-services-app/issues/5181
+# Input is a CSV of druids generated on purl-fetcher with the query:
+#   Purl.status('public').where.missing(:public_json).pluck(:druid)
+# Invoke via:
+# bin/rails r -e production "PublishStatus.report('druids.csv')"
+class PublishStatus
+  def self.report(druids_file)
+    druids = File.readlines(druids_file).map(&:strip)
```
mjgiarlo, line 11: Is the `strip` bit there to remove trailing newlines? If so, you could also consider this hard-to-ferret-out argument to `File::readlines` (because it's defined in `IO::readlines`):

```suggestion
    druids = File.readlines(druids_file, chomp: true)
```

But this is mostly FYI, and I don't expect you to make a change and run the report again.

app/reports/publish_status.rb
```diff
@@ -0,0 +1,35 @@
...
+    output_file = 'tmp/publish_status_report.csv'
+
+    CSV.open(output_file, 'w') do |csv|
+      csv << %w[druid closed publishable not_dark first_published]
+
+      druids.each do |druid|
+        object = RepositoryObject.where(external_identifier: druid).first
```
mjgiarlo, line 18:
```suggestion
        object = RepositoryObject.find_by(external_identifier: druid)
```

app/reports/publish_status.rb
```diff
@@ -0,0 +1,35 @@
...
+        latest = object.head_version
+        not_dark = latest.access.fetch('view') != 'dark'
+
+        milestones = workflow_client.milestones(druid:)
+        published = milestones.filter { |milestone| milestone[:version] == '1' && milestone[:milestone] == 'published' }
```
mjgiarlo, line 23: Maybe this is obvious, but can you drop a comment about the milestone version matching `1` here, like, why we care about only the first version in this report.

app/reports/publish_status.rb
lwrubel: Thanks, wasn't aware of the `chomp` argument, so good to know!

app/reports/publish_status.rb
lwrubel: True, I should totally be using `find_by` instead. Just realized this the other day.

lwrubel:
Thanks for the comments, @mjgiarlo. I went ahead and made the tweaks so I'd have them if I ever use this report as the basis for another in the future.

PR #5183, opened 2024-10-03
remove resetworkspace job - use cleanup job for the work

### LINKED ISSUE #5110: Combine reset workspace and cleanup job logic
We currently have two different jobs called ResetWorkspace and Cleanup that do similar things, calling out to ResetWorkspaceService and CleanupService.  These jobs are invoked by DSA, which in turn starts then via calls from accessionWF:reset-workspace and accessionWF:end-accession.

It may make sense to have a single job that combines the logic of these two cleanup steps, and then (1) invokes this single job from accessionWF:reset-workspace and (2) removes the invocation of the job from accessionWF:end-accession, so that there are no jobs being called from the last step in accessionWF.  

Reasoning?  See the logic here: https://github.com/sul-dlss/common-accessioning/issues/1289#issuecomment-2197469689

Would also possibly require changes to dor-services-client if any endpoints change in DSA.

See also sul-dlss/common-accessioning#1322

Per Captioning work cycle planning meeting 2024-09-25:  This warrants a quick check-in with the wider infra team about whether to proceed, and much (maybe most) of the work will be coordinating testing.  Also, we think that maybe the retries (in DSA job?) will keep the race condition from bothering Andrew with failed workflow steps, but not totally sure about that?

peetucket:

Fixes #5110

Get rid of `ResetWorkspaceJob`, the `reset` endpoint that called it and the service that did the work.  The `reset` call is removed in dsa-client here: https://github.com/sul-dlss/dor-services-client/pull/472

This is because the work done by `ResetWorkspaceJob` is completely duplicated by the existing `CleanupJob`, which will now be called by the accessionWF:reset-workspace robot (see https://github.com/sul-dlss/common-accessioning/pull/1380).  This will invoke the `CleanupJob` instead, which will NOT be invoked by end-accession anymore.  This is the key change to remove the race condition.

If you closely examine what `CleanupService` does and what `ResetWorkspaceService` does, you will note that 
1. They both remove the export bag in the same way.
2. ResetWorkspaceServce copies the druid tree folder (e.g. `/dor/workspace/ab/123/bc/1234/ab123bc1234` -> `/dor/workspace/ab/123/bc/1234/ab123bc1234_v1`, and then the CleanupService just immediately removes everything from the end point of the druid tree anyway (e.g. `/dor/workspace/ab/123/bc/1234`). Since they run one after the other (even now), there is no point in making that copy, just to immediately delete it.

Note that nobody else is using the `reset` endpoint (see see https://github.com/search?q=org%3Asul-dlss%20workspace.reset&type=code) so once we stop calling it in the `ResetWorkspaceJob`, it can safely go away.

We need to coordinate changes in [dsa-client](https://github.com/sul-dlss/dor-services-client/pull/472) and [common-accessioning](https://github.com/sul-dlss/common-accessioning/pull/1380).

Specs
Integration tests

Test strategy on QA:
- deploy `5110-cleanup-jobs` branch of DSA
- deploy `remove-cleanup-from-end-accession` branch of common-accessioning (which includes pointing to the `combine-reset-and-cleanup` branch of dor-services-client gem)
- run integration tests on QA


app/controllers/workspaces_controller.rb
```diff
@@ -16,16 +16,7 @@ def create
...
-  end
-
-  # Once an object has been transferred to preservation, reset the workspace by
-  # renaming the druid-tree to a versioned directory and removing the export directory
-  def reset
-    result = BackgroundJobResult.create
-    version = CocinaObjectStore.version(params[:object_id])
+    EventFactory.create(druid: params[:object_id], event_type: 'reset_request_received', data: { background_job_result_id: result.id })
     ResetWorkspaceJob.set(queue: params['lane-id']).perform_later(druid: params[:object_id], version:, background_job_result: result, workflow: params[:workflow])
     head :created, location: result
```
peetucket, line 21: 1. the `reset` endpoint is going away
2. the `reset` method in dsa-client will send the `delete` verb to DSA, which will call the `destroy` method
3. the destroy method will now start the `ResetWorkspaceJob` -- which will do both a cleanup and a reset (see below)

app/jobs/reset_workspace_job.rb
```diff
@@ -15,15 +15,22 @@ def perform(druid:, version:, background_job_result:, workflow:)
 
     begin
       ResetWorkspaceService.reset(druid:, version:)
+      CleanupService.cleanup_by_druid druid
+
+      EventFactory.create(druid:,
+                          event_type: 'reset-workspace',
+                          data: { status: 'success' })
     rescue ResetWorkspaceService::DirectoryAlreadyExists
```
peetucket, line 23: this job now also does the cleanup (the CleanupJob is gone)

app/controllers/workspaces_controller.rb
```diff
@@ -21,15 +21,6 @@ def destroy
...
 
-  # Once an object has been transferred to preservation, reset the workspace by
-  # renaming the druid-tree to a versioned directory and removing the export directory
-  def reset
-    result = BackgroundJobResult.create
-    version = CocinaObjectStore.version(params[:object_id])
-    ResetWorkspaceJob.set(queue: params['lane-id']).perform_later(druid: params[:object_id], version:, background_job_result: result, workflow: params[:workflow])
-    head :created, location: result
-  end
-
```
peetucket, line 32: this endpoint is going away, all of the work will be done by the existing cleanup endpoint

jmartin-sul (approved):
traced through the DruidTools and PruneServices stuff myself just to double-check, and the reasoning sounds right to me re: the deletion and the essential non-effect of the backup copy that's deleted right after creation.

PR #5186, opened 2024-10-08
remove workflow param from cleanup

peetucket:

Part of https://github.com/sul-dlss/common-accessioning/issues/1387

DisseminationWF is gone (see #1385), so remove it's reference.  We also don't need to bother even asking for the workflow anymore for this endpoint since it's always going to be accessionWF.  

This changes needs to be coordinated with https://github.com/sul-dlss/common-accessioning/pull/1386 and https://github.com/sul-dlss/dor-services-client/pull/475

HOLD for #1385

Process:
1. Merge https://github.com/sul-dlss/dor-services-client/pull/475 and then cut a new release
2. Update https://github.com/sul-dlss/common-accessioning/pull/1386 to use new dor-services-client release, merge it
3. Merge https://github.com/sul-dlss/dor-services-app/pull/5186
4. Deploy all

Spec

### DIFF
--- a/app/jobs/cleanup_job.rb
+++ b/app/jobs/cleanup_job.rb
@@ -6,11 +6,12 @@ class CleanupJob < ApplicationJob
 
   # @param [String] druid the identifier of the object to be cleaned up
   # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
-  # @param [String] workflow Which workflow should this be reported to?
-  def perform(druid:, background_job_result:, workflow:)
+  def perform(druid:, background_job_result:)
     background_job_result.processing!
 
-    workflow_process = workflow == 'accessionWF' ? 'reset-workspace' : 'cleanup'
+    # this job is only called by accessionWF:reset-workspace, so it should complete it when done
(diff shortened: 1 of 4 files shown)


app/jobs/cleanup_job.rb
```diff
@@ -6,11 +6,12 @@ class CleanupJob < ApplicationJob
 
   # @param [String] druid the identifier of the object to be cleaned up
   # @param [BackgroundJobResult] background_job_result identifier of a background job result to store status info
-  # @param [String] workflow Which workflow should this be reported to?
-  def perform(druid:, background_job_result:, workflow:)
+  def perform(druid:, background_job_result:)
     background_job_result.processing!
 
-    workflow_process = workflow == 'accessionWF' ? 'reset-workspace' : 'cleanup'
```
peetucket, line 13: we used to need to pass in the workflow because it could be either `accessionWF` or `disseminationWF`, each of which had a different step to mark as completed

that is no longer true, only accessionWF is making this call

jmartin-sul (approved):

jmartin-sul (approved):

PR #5195, opened 2024-10-23
Support discarding open object versions.

### LINKED ISSUE #5194: Support discarding open versions (discard draft)
Currently, this is supported by RepositoryObject. This should be exposed via API and also include additional discard operations link cleaning up the workspace.

justinlittman:
closes #5194

This functionality is needed for H3.

Unit

### DIFF
--- a/app/models/repository_object.rb
+++ b/app/models/repository_object.rb
@@ -111,15 +111,21 @@ class RepositoryObject < ApplicationRecord # rubocop:disable Metrics/ClassLength
     end
   end
 
+  def check_discard_open_version!
+    raise VersionNotDiscardable, 'Cannot discard version because head version is closed' if closed?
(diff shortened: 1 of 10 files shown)


app/models/repository_object.rb
```diff
@@ -111,15 +111,21 @@ def close_version!(description: nil)
     end
   end
 
+  def check_discard_open_version!
```
lwrubel, line 114: Why is the check turned into a bang method? Just curious, not a blocker.

lwrubel (approved):

lwrubel:
Approved although I am wondering if there any way to further test this in a deployed env?

justinlittman:
I've tested it locally; it will get exercised by H3 eventually.

app/models/repository_object.rb
justinlittman: Because public and it raises, it is a general convention.

PR #5196, opened 2024-10-23
Changes strategy when staging files from preservation for shelving.

### LINKED ISSUE #5193: publish/shelve is sometimes overwriting "good" files on stacks with non-working files
I've discovered that under certain conditions "publish" will overwrite files that are already on Stacks with the wrong content. Specifically, files are being shelved that are not deposited content, but rather error messages that say:

>400 Bad Request: version parameter must be positive integer

Example: the viewer shows an error on this druid: https://purl.stanford.edu/fs305xf2702

This item should have:

- 1 PDF
- 1 XML
- ! JP2

Download the files and you will find that all 3 files are 59 bytes and all contain the error message above. I've verified that these files are the same files on Stacks, so this is not a download issue - the wrong files have actually been shelved.

I've also verified that "real" files were previously present for this druid. I happened to run a report back in July 2024 listing all files on Stacks and their sizes. At that time, the three files had the sizes and modification dates you would expect from the correct files:

```
{"permissions":"644", "size":"2481", "fullpath":"/stacks/fs/305/xf/2702/91100086.xml", "file_modificaton":"2015-07-10 18:10:32.000000000 -0700"}
{"permissions":"644", "size":"35153", "fullpath":"/stacks/fs/305/xf/2702/91100086.pdf", "file_modificaton":"2015-07-10 18:10:32.000000000 -0700"}
{"permissions":"644", "size":"1006278", "fullpath":"/stacks/fs/305/xf/2702/91100086_0001.jp2", "file_modificaton":"2015-07-10 18:10:32.000000000 -0700"}
```

The current files on Stacks for that druid now all have modification dates of today, when I republished:

```
total 12
-rw-rw-r-- 1 stacks stacks 59 Oct 22 04:36 91100086_0001.jp2
-rw-rw-r-- 1 stacks stacks 59 Oct 22 04:36 91100086.pdf
-rw-rw-r-- 1 stacks stacks 59 Oct 22 04:36 91100086.xml
```

That means they must have been written today, overwriting the files that were originally there.

**Possible cause**

I think what could be happening is this:

- an item is on version 1
- Argo sets off a publish job on the item
- even though files are on stacks, the metadata in purl-fetcher may not indicate that, so the item is considered to not have files on stacks
- there's an attempt to shelve from preservation
- shelve from preservation assumes that you want the files from the **last** preserved version, so the request for files is for version (1 - 1 = 0)
- preservation responds with an error message that the version number must be a positive integer
- the error message is written to **files** that have the same name as the files that were requested from preservation
- the new files containing the error message are shelved

I have only seen this issue on items that are on version 1 when republished. When the version is > 1, I've found some files shelved from preservation.

**Level of impact**

So far, I think the affected items are only within two collections that are not highly-used (one is citation-only). Because of previous problems where republishing and/or updating druids has led to errors at the publish step, I've been making an effort to verify that every accessioned object can be republished cleanly. All the publish errors since about 6 PM last night have been in a small number of collections/APOs that I was republishing in their own batches.

There have been other publish errors over the past couple of months, and I have had to shelve files to fix many of them, but I have not previously seen this overwriting issue. The risk of overwriting has likely been there the whole time, it just happens I had not yet reached a set of items that meet all the conditions that cause overwriting.

justinlittman:
closes #5193

So that errors are correctly handled when getting files from preservation so that garbage is not shelved.

Unit

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ gem 'mais_orcid_client'
 gem 'marc'
 gem 'marc-vocab', '~> 0.3.0' # for indexing
 gem 'moab-versioning', '~> 6.0', require: 'moab/stanford'
-gem 'preservation-client', '~> 6.0'
+gem 'preservation-client', '~> 7.0'
 gem 'purl_fetcher-client', '~> 2.1'
 gem 'stanford-mods' # for indexing
 gem 'sul_orcid_client', '~> 0.3'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -363,7 +363,7 @@ GEM
     patience_diff (1.2.0)
       optimist (~> 3.0)
     pg (1.5.8)
-    preservation-client (6.2.0)
+    preservation-client (7.0.1)
       activesupport (>= 4.2, < 8)
       faraday (~> 2.0)
       moab-versioning (>= 5.0.0, < 7)
@@ -618,7 +618,7 @@ DEPENDENCIES
   okcomputer
   parallel
   pg
-  preservation-client (~> 6.0)
+  preservation-client (~> 7.0)
   propshaft
   puma (~> 6.0)
   purl_fetcher-client (~> 2.1)
--- a/app/services/shelvable_files_stager.rb
+++ b/app/services/shelvable_files_stager.rb
@@ -27,7 +27,7 @@ class ShelvableFilesStager
       file_pathname = workspace_content_pathname.join(filepath)
       next if file_pathname.exist?
 
-      copy_file_from_preservation(file_pathname:, filepath:)
+      retrieve_from_preservation(file_pathname:, filepath:)
     end
   end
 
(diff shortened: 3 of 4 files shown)


peetucket:
Any reasonable way of testing this in stage?  I guess at a minimum would be nice to deploy there and run some integration tests that exercise accessioning.

andrewjbtw:
I think there are bascially two cases to test:

1. Open an object, update metadata to shelve previously preserve-only file(s), close object.
2. Accessioned object, do not open or update, but press "republish" from Argo

Case 1 should be testable without much trouble. It's the original motivating use case/workflow for shelve from preservation.

Case 2 is trickier because we have to trigger the shelve without modifying the object. I think we'd need to do one or both:

- delete files from Stacks for that object that are marked as publish/shelve = yes in Cocina, so the shelve logic will determine that the files need to be reshelved
- and/or do something to the data in purl-fetcher to make the shelve logic lead to a shelve from preservation

For a "healthy" object, there would be no shelve from preservation during republish because the files would already be present.

jmartin-sul (approved):

jmartin-sul (approved):
code changes look good to me.  my read of @andrewjbtw's ticket is that we aren't 100% sure of the cause, so given that and the quiet and subtle nature of the bug, i'd personally vote for reproducing it using `main`, then confirming that this branch fixes it.  seems like we at least have a really solid hypothesis about the cause with a concrete plan for testing.

andrewjbtw:
Pasting in my testing results from Slack:

 I tested:

- version 1 republish (after deleting files from Stacks to cause a reshelve)
- version >1 republish
- shelve/unshelve via open version, update structure, close version
- small number of files reshelve
- up to 700 files re-shelve

The big issue is that we’re still writing bad files to the workspace if a file can’t be retrieved from preservation. This item has 2 JP2s that aren’t in preservation: https://argo-stage.stanford.edu/view/druid:yg564qg3921

When I republished, I first got an error for unable to find the file in the workspace: https://app.honeybadger.io/projects/50568/faults/113812706

But a few minutes later, I noticed that somewhere along the way, two 0-byte files got written as JP2s and then shelved. I know the file size check should mitigate this, but I don’t understand why we’re writing files that don’t exist.

andrewjbtw:
After retesting the fix for the zero-byte file issue, I think this is good to go.

PR #5218, opened 2024-10-31
Rails 7.2 upgrade

jmartin-sul:

up to date dependencies good, out of date dependencies bad

needs some commit squashing/cleanup before merge.

- [x] unit
- [x] integration


config/initializers/new_framework_defaults_7_2.rb
aaron-collier, line 1: Would you be OK deleting this file? We've tried not to keep them in the other PRs.

config/environments/production.rb
```diff
@@ -30,19 +32,34 @@
   # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
   # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
 
+  # Assume all access to the app is happening through a SSL-terminating reverse proxy.
+  # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
+  # config.assume_ssl = true
+
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
mjgiarlo, line 40: We saw problems when we turned this on in the workflow server, so I think we have been leaving it commented out.

config/environments/production.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
+require 'active_support/core_ext/integer/time'
```
mjgiarlo, line 3: Do you know why we needed to add this `require` to all envs?

config/environments/production.rb
```diff
@@ -30,19 +32,34 @@
...
 
+  # "info" includes generic and useful information about system operation, but avoids logging too much
+  # information to avoid inadvertent exposure of personally identifiable information (PII). If you
+  # want to log everything, set the level to "debug".
+  config.log_level = ENV.fetch('RAILS_LOG_LEVEL', 'info')
+
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
 
+  config.action_dispatch.show_exceptions = :none
```
mjgiarlo, line 61: Can you say more about where this new config comes from?

config/initializers/new_framework_defaults_7_2.rb
mjgiarlo: Yeah, and when you use `config.load_defaults 7.2` in `config/application.rb`, it has the same effect as keeping this file and uncommenting everything, so it shoudl be safe to delete as long as tests are passing.

config/environments/production.rb
jmartin-sul: this didn't seem to cause any trouble with the integration test i ran, but i had seen that in other PRs (was my first suspect when all the integration tests were failing for me friday), so i'm happy to be consistent with everything else if it's not useful/possibly problematic.

config/environments/production.rb
jmartin-sul: not sure, i believe it was just part of the code that `bin/rails app:update` spit out (and i think the test suite errored out when i left it out, though i'm not 100% sure about that memory).  in retrospect though, i think `app:update` also removed the same `require` from [`config/application.rb`](https://github.com/sul-dlss/dor-services-app/pull/5218/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9L19) at first, and i may have left it out (and these new lines in) when fixing unit tests, then restored the `require` in `application.rb` when looking at the diff and committing after getting things working.  i'd meant to try getting rid of the env specific `require`s when i realized later that we might've centralized to DRY things up, but forgot and didn't get back to it.  will try now...

config/environments/production.rb
jmartin-sul: i think the `app:update` command defaulted to `:rescuable` (or maybe `:all` since that's the default?), and i saw some discussion about it here: https://github.com/sul-dlss/sul_pub/pull/1741#pullrequestreview-2405987510

but [the guide says](https://guides.rubyonrails.org/configuring.html#config-action-dispatch-show-exceptions) that this is about showing error pages, which it seems we wouldn't want to do for an API app.  though maybe it's a moot point given that all the requests ask for JSON and so wouldn't get error pages anyway?

happy to remove if you'd prefer.

jmartin-sul: (in which case i'd probably want to re-run integration tests, out of an abundance of caution)

config/environments/production.rb
jmartin-sul: the test suite passed with all of them gone just now, including the one that this PR didn't remove 🤷

jmartin-sul: > we might've centralized to DRY things up

and also to obey the docs, apparently:
> Prefer requiring dep in config/application.rb if it used in all envs
> And require base active_support path before requiring any parts of AS. This has been recommended in the Rails docs for a long while but now, as of Rails 7.1, if you do not obey the docs, the code breaks.

https://github.com/sul-dlss/dor-services-app/commit/c0d2263b4f714ae6a1129ba1d6ed0259210858ce from https://github.com/sul-dlss/dor-services-app/pull/4620/commits

jmartin-sul: looks like the require was added in a prior rails update: https://github.com/sul-dlss/dor-services-app/commit/87d7f8627c1903b56b24e169442d693b2843f30e#diff-d3c4b3f41072daa416f1920511e9b2e26caea8c5cec0a14cb9508589a4dafa47 in https://github.com/sul-dlss/dor-services-app/pull/4055

i'm going to try removing them all and running integration tests.  maybe never necessary in the first place, maybe no longer necessary, but either way, seems like maybe something unnecessary can be removed.

config/initializers/new_framework_defaults_7_2.rb
jmartin-sul: ah, good to know, thanks!

jmartin-sul: removed, passed specs and integration tests

config/environments/production.rb
jmartin-sul: commented out, passed specs and integration tests

config/environments/production.rb
jmartin-sul: removed them all, passed specs and integration tests

config/environments/production.rb
jmartin-sul: ended up leaving this as-is, but happy to revisit if you'd prefer.

jmartin-sul:
@aaron-collier and @mjgiarlo this is ready for another look when you get a chance.  i think i addressed all the suggestions/questions (with the only thing not resulting in a change being the `config.action_dispatch.show_exceptions` one, but i'm happy to revisit that too).  specs and integration tests both pass with these most recent changes.

config/environments/production.rb
mjgiarlo: @jmartin-sul Excellent, thank you!

config/environments/production.rb
mjgiarlo: Nope, sounds good 🙂

mjgiarlo (approved):

PR #5247, opened 2025-01-09
Report with counts of related resource types

lwrubel:
Example report demonstrating counting values at a path. 

Run on QA


lwrubel:
@arcadiafalcone here are the reports:

[related_resource_types_qa.csv](https://github.com/user-attachments/files/18377847/related_resources_types_qa.csv)
[related_resource_types_stage.csv](https://github.com/user-attachments/files/18377851/related_resource_types_stage.csv)
[related_resource_types_prod.csv](https://github.com/user-attachments/files/18377853/related_resource_types_prod.csv)

app/reports/related_resource_types.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# Invoke via:
+# bin/rails r -e production "RelatedResourceTypes.report" > related_resource_types.csv
+class RelatedResourceTypes
+  # path within description
+  JSON_PATH = '$.relatedResource[*].type'
```
mjgiarlo, line 7: I'm not sure we've been consistent with this, but I wonder if we should make this a ["strict"](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-SQLJSON-STRICT-AND-LAX-MODES) JSON path

```suggestion
  JSON_PATH = 'strict $.relatedResource[*].type'
```

I haven't internalized when we need to make this distinction and when we don't, but I do vaguely recall bogus results in reports early on before we discovered that `lax` was the default for JSON path expressions in Postgres.

mjgiarlo: Since it's an example, we may want to toss it in there so we don't get bitten with weird lax data bugs in the future. It may or may not be valuable to re-run the changed report in an env like QA to verify the prior results.

lwrubel: Agree this is a good approach and updated the report to use strict with a pointer to the docs.

lwrubel:
@arcadiafalcone I re-ran the reports after accounting for nested relatedResources. There are some slight differences in the numbers. 

[related_resource_types_qa.csv](https://github.com/user-attachments/files/18381470/related_resource_types_qa.csv)
[related_resource_types_stage.csv](https://github.com/user-attachments/files/18381471/related_resource_types_stage.csv)
[related_resource_types_prod.csv](https://github.com/user-attachments/files/18381504/related_resource_types_prod.csv)

mjgiarlo (approved):

Issue #5231, opened 2024-11-20
Indexing is not keeping up with accessioning


andrewjbtw:
The indexing statuses in Argo have been falling behind lately and I'm guessing the problem is the indexer itself is falling behind or failing to index as items complete accessioning.

As I write this issue, the Argo workflow grid shows 11 items waiting at the start of the accessionWF, but most of these items have actually been accessioned, some almost a day ago. The same problem applies to the 6 items listed as "started" at end-accession.
<img width="1297" alt="Screenshot 2024-11-19 at 5 23 10 PM" src="https://github.com/user-attachments/assets/2d8db35b-734b-4d67-a3d5-e3b62c1af1fc">

The show page has the same problem, where the status line shows "In accessioning" but the workflow display shows all the steps are complete
<img width="945" alt="Screenshot 2024-11-19 at 5 25 13 PM" src="https://github.com/user-attachments/assets/0f371df0-a719-4fa8-9886-2d8b5f4107e3">

Manually reindexing corrects the issue so it seems like the indexing status is getting dropped somewhere while accessioning runs. Without manual reindex, these items will have incorrect statuses until the rolling indexer comes along.

jcoyne:
@justinlittman log should be in `/var/log/apache/error_log`

justinlittman:
@andrewjbtw Are you able to replicate this in any environment or is it sporadic?

andrewjbtw:
I've tried to reproduce this on stage but haven't been able to do it. I can almost certainly make it happen on prod if I can come up with a set of items to update for real.

andrewjbtw:
Most often, I see this issue in ones and twos every few days. For example, there's [an item that was accessioned yesterday](https://argo.stanford.edu/view/druid:rq089xm3950) that is still indexed as "in accessioning."

justinlittman:
This seems to indicate that https://github.com/sul-dlss/argo/issues/3817 has not been resolved.

andrewjbtw:
Possibly, but it was more or less resolved for a while. That earlier issue was when dor-indexing-app was a separate thing.

mjgiarlo:
@justinlittman I see you've been doing some work related to this issue. OK if I assign you and move to in progress? (This issue is showing up on the prod board as needing to be triaged.)

justinlittman:
Logging is shows that there are indexing requests after the indexing request that created the document that is in Solr.

I have 2 remaining theories:
[...]

PR #5249, opened 2025-01-15
Adds tracing id to troubleshoot indexing.

justinlittman:
refs #5231

To allow better debugging of indexing.

Stage, unit

### DIFF
--- a/app/services/indexer.rb
+++ b/app/services/indexer.rb
@@ -3,11 +3,12 @@
 # Indexes a Cocina object.
 class Indexer
   # @param [Cocina::Models::DROWithMetadata|CollectionWithMetadata|AdminPolicyWithMetadata]
-  def self.reindex(cocina_object:)
+  def self.reindex(cocina_object:, trace_id: nil)
     return unless Settings.solr.enabled
 
     solr_doc = Indexing::Builders::DocumentBuilder.for(
-      model: cocina_object
+      model: cocina_object,
+      trace_id: trace_id || trace_id_for(druid: cocina_object.externalIdentifier)
     ).to_solr
     solr.add(solr_doc)
     solr.commit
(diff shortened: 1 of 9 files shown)


app/services/indexer.rb
```diff
@@ -21,16 +22,23 @@ def self.delete(druid:)
...
   def self.solr
     RSolr.connect(timeout: 120, open_timeout: 120, url: Settings.solr.url)
   end
   private_class_method :solr
+
+  def self.trace_id_for(druid:, source:)
+    SecureRandom.uuid.tap do |trace_id|
+      Rails.logger.info("Reindexing #{druid} from #{source} with trace_id=#{trace_id}")
+    end
+  end
```
mjgiarlo, line 43: Is the long-story-short explanation of all these changes that we want a way to relate log entries and honeybadger exceptions to solr documents? Those are the three spots where these IDs wind up, if I'm reading this right.

app/services/indexer.rb
```diff
@@ -3,11 +3,12 @@
...
   # @param [Cocina::Models::DROWithMetadata|CollectionWithMetadata|AdminPolicyWithMetadata]
-  def self.reindex(cocina_object:)
+  def self.reindex(cocina_object:, trace_id:)
     return unless Settings.solr.enabled
 
     solr_doc = Indexing::Builders::DocumentBuilder.for(
-      model: cocina_object
+      model: cocina_object,
+      trace_id:
     ).to_solr
```
mjgiarlo, line 12: I'm wondering if we need to add this arg to `Indexer.reindex` and `Indexer.reindex_later` since both of these methods already have:

1. A Cocina object, which should always have the druid in question (?), and
2. `Kernel#caller_locations`, which should get us the path back to the caller

app/services/indexer.rb
justinlittman: That's correct. We have Solr docs that are out of date; we don't know how they got written. Is there a race condition? Is stale data being indexed? Are Solr commits being dropped? Something else?

app/services/indexer.rb
mjgiarlo: OK, then would ☝🏻 work for DSA-internal use? If so, then we could cut down the changes in this PR and incorporate the remote trace stuff when we work on that.

justinlittman: @mjgiarlo I've reduced the footprint using your suggested technique. I still left the method signature hooks in place so that an existing trace_id can be passed in.

mjgiarlo (approved):

app/services/indexer.rb
mjgiarlo: Thank you!

PR #5250, opened 2025-01-15
Add more context to HB when editing via QuickMARC

mjgiarlo:
# Why was this change made?

Errors from the QuickMARC API were previously being eaten. This gives us more actionable exceptions and alerts.

If there's a QuickMARC error, don't bother retrying subsequent Folio API interactions. Instead, log a *single* HB alert with a useful title and sufficient context for a developer to figure out what is going on.

# How was this change tested?

CI, tested with a prod object

### DIFF
--- a/app/services/catalog/folio_writer.rb
+++ b/app/services/catalog/folio_writer.rb
@@ -19,6 +19,12 @@ module Catalog
     def save
       return if catalog_record_ids.empty? && previous_catalog_record_ids.empty?
 
+      Honeybadger.context(
+        druid: cocina_object.externalIdentifier,
+        catalog_record_ids:,
+        previous_catalog_record_ids:
+      )
+
       # remove 856 for previous catkeys
       previous_catalog_record_ids.each { |previous_id| delete_previous_ids(catalog_record_id: previous_id, ignore_not_found: true) }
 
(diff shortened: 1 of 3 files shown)


app/services/catalog/folio_writer.rb
```diff
@@ -64,24 +74,29 @@ def update_current_ids(catalog_record_id:)
...
 
+    # @param [Hash,NilClass] response quickmarc response e.g.,
+    #   {"message"=>"Record is too long to be a valid MARC binary record, it's length would be 100106 which is more than 99999 bytes", "type"=>"-4", "code"=>"INTERNAL_SERVER_ERROR"}
+    def validate_quickmarc_response!(response)
+      return if response.nil?
+
+      Honeybadger.context(quickmarc_response: response)
+
+      raise response['message']
+    end
```
mjgiarlo, line 85: AFAICT, we should only get a `nil` response if the [PUT request to the QuickMARC API](https://github.com/sul-dlss/folio_client/blob/9e521089a49774543117ecea38cc89473eae9acb/lib/folio_client/records_editor.rb#L35) succeeds. Why? The [API docs](https://github.com/folio-org/mod-quick-marc/blob/2bb8a994f08b02786775232fbd19910ba82c58cb/src/main/resources/swagger.api/records-editor-async.yaml#L12-L41) list five possible responses:

1. A successful 202 response with no body (parsed to `nil` by the Folio client)
2. A failed 400 response with a body
3. A failed 409 response with a body
4. A failed 422 response with a body
5. A failed 500 response with a body

justinlittman (approved):

PR #5263, opened 2025-02-13
Return version description with version status.

justinlittman:
refs https://github.com/sul-dlss/hungry-hungry-hippo/issues/704

So that it would be available for use in H3.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -87,8 +87,6 @@ GEM
     attr_extras (7.1.0)
     base64 (0.2.0)
     bcrypt_pbkdf (1.1.1)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -87,8 +87,6 @@ GEM
     attr_extras (7.1.0)
     base64 (0.2.0)
     bcrypt_pbkdf (1.1.1)
-    bcrypt_pbkdf (1.1.1-arm64-darwin)
-    bcrypt_pbkdf (1.1.1-x86_64-darwin)
```
mjgiarlo, line 91: 😆 here we go again

mjgiarlo (approved):

PR #5264, opened 2025-02-14
Fix docker compose build

mjgiarlo:
# Why was this change made?

For some reason, the Sidekiq Pro env variable was no longer being automagically injected. No idea why or what changed. This takes a more explicit approach.

# How was this change tested?

Successfully ran `docker compose build`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ ENV BUNDLER_WITHOUT="development test"
 
 # For Sidekiq Pro
 ARG BUNDLE_GEMS__CONTRIBSYS__COM
-ENV BUNDLE_GEMS__CONTRIBSYS__COM $BUNDLE_GEMS__CONTRIBSYS__COM
 
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get -y install --no-install-recommends \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -5,7 +5,6 @@ ENV BUNDLER_WITHOUT="development test"
 
 # For Sidekiq Pro
 ARG BUNDLE_GEMS__CONTRIBSYS__COM
-ENV BUNDLE_GEMS__CONTRIBSYS__COM $BUNDLE_GEMS__CONTRIBSYS__COM
```
mjgiarlo, line 8: AFAIK, we only need this value at build time, not run time, and we probably don't want a public image with a secret in its env 🙈

compose.yaml
```diff
@@ -1,6 +1,9 @@
 services:
   web:
-    build: .
+    build:
+      context: .
+      args:
+        BUNDLE_GEMS__CONTRIBSYS__COM: "${BUNDLE_GEMS__CONTRIBSYS__COM}"
```
mjgiarlo, line 6: No idea why we needed this change, though.

justinlittman (approved):

PR #5267, opened 2025-02-18
Digital serials report

lwrubel:
Gather data on possible digital serials druids.

### DIFF
--- a/app/reports/digital_serials.rb
+++ b/app/reports/digital_serials.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# Invoke via:
+# bin/rails r -e production "DigitalSerials.report" > digital_serials.csv
+class DigitalSerials
+  # NOTE: Prefer strict JSON querying over lax when using the `.**` operator, per
+  # https://www.postgresql.org/docs/14/functions-json.html#STRICT-AND-LAX-MODES
+  # relatedResource may be within a relatedResource, so we need to use .**
+  SQL = <<~SQL.squish.freeze
+    SELECT ro.external_identifier as druid,
+      jsonb_path_query(rov.identification, '$.catalogLinks[*] ? (@.catalog == "folio").catalogRecordId') ->> 0 as catalog_record_id,
+      jsonb_path_query(rov.identification, '$.catalogLinks[*] ? (@.catalog == "folio").refresh') ->> 0 as refresh
+      FROM repository_objects AS ro, repository_object_versions AS rov WHERE
+        jsonb_path_exists(rov.identification, '$.catalogLinks[*] ? (@.catalog == "folio")')
+        AND (jsonb_path_exists(rov.description, 'strict $.title.**.type ? (@ like_regex "part name|part number")')
+        OR jsonb_path_exists(rov.description, 'strict $.note.**.type ? (@ like_regex "date\/sequential designation")'))
+        AND ro.head_version_id = rov.id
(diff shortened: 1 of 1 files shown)


app/reports/digital_serials.rb
```diff
@@ -0,0 +1,43 @@
...
+# bin/rails r -e production "DigitalSerials.report" > digital_serials.csv
+class DigitalSerials
+  # NOTE: Prefer strict JSON querying over lax when using the `.**` operator, per
+  # https://www.postgresql.org/docs/14/functions-json.html#STRICT-AND-LAX-MODES
+  SQL = <<~SQL.squish.freeze
+    SELECT ro.external_identifier as druid,
+      jsonb_path_query(rov.structural, '$.isMemberOf') ->> 0 as collection_id,
+      jsonb_path_query(rov.identification, '$.catalogLinks[*] ? (@.catalog == "folio").catalogRecordId') ->> 0 as catalog_record_id,
+      jsonb_path_query(rov.identification, '$.catalogLinks[*] ? (@.catalog == "folio").refresh') ->> 0 as refresh
+      FROM repository_objects AS ro, repository_object_versions AS rov WHERE
```
edsu, line 13: This might just be me, but I'm used to seeing WHERE on a new line indented along with SELECT and FROM.

app/reports/digital_serials.rb
```diff
@@ -0,0 +1,43 @@
...
+        jsonb_path_exists(rov.identification, '$.catalogLinks[*] ? (@.catalog == "folio")')
+        AND (jsonb_path_exists(rov.description, 'strict $.title.**.type ? (@ like_regex "part name|part number")') OR
+        jsonb_path_exists(rov.description, 'strict $.note.**.type ? (@ like_regex "date\/sequential designation")'))
+        AND jsonb_path_exists(rov.structural, '$.isMemberOf[*] ? (@ != "druid:yh583fk3400")')
```
edsu, line 17: It looks like this is to ignore Google Books?

justinlittman (approved):

app/reports/digital_serials.rb
```diff
@@ -0,0 +1,43 @@
...
+    sql_result_rows = ActiveRecord::Base.connection.execute(sql_query).to_a
+
+    sql_result_rows.map do |row|
+      collection_druid = row['collection_id']
+      collection_name = RepositoryObject.collections.find_by(external_identifier: collection_druid)&.head_version&.label
+
+      [
+        row['druid'],
+        collection_druid,
+        "\"#{collection_name}\"",
```
edsu, line 37: Hopefully there are no quotes in collection names!

app/reports/digital_serials.rb
lwrubel: I meant to add a note to this report like I have in the other. Will do that.

app/reports/digital_serials.rb
lwrubel: We have a lot of variability in reports it seems. Thanks for the suggestion to make queries more readable.

app/reports/digital_serials.rb
lwrubel: Will double check and take a look at simplifying this.

lwrubel: Seeing that collection names had this treatment in other reports, I figured it was something to cautious about. There are collection names with quotes, believe it or not: https://argo.stanford.edu/view/druid:bc677qs7597

PR #5278, opened 2025-03-17
Faster version statuses

justinlittman:
So that H3 is faster.

Unit, stage, qa

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -147,7 +147,7 @@ class VersionsController < ApplicationController # rubocop:disable Metrics/Class
     {
       versionId: version,
       open: version_service.open?,
-      openable: version_service.can_open?,
+      openable: version_service.can_open?(check_preservation: false),
       assembling: workflow_state_service.assembling?,
       accessioning: workflow_state_service.accessioning?,
       closeable: version_service.can_close?,
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -139,8 +139,8 @@ class VersionsController < ApplicationController # rubocop:disable Metrics/Class
   end
 
   def status_for(druid:, version:)
-    version_service = VersionService.new(druid:, version:)
     workflow_state_service = WorkflowStateService.new(druid:, version:)
+    version_service = VersionService.new(druid:, version:, workflow_state_service:)
     repository_object = RepositoryObject.find_by!(external_identifier: druid)
     version_description = repository_object.versions.select(:version_description).find_by!(version:).version_description
 
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -67,7 +67,7 @@ class VersionService
     raise ArgumentError, 'description is required to open a new version' if description.blank?
 
     ensure_openable!(assume_accessioned:)
-    repository_object = RepositoryObject.find_by!(external_identifier: cocina_object.externalIdentifier)
+    repository_object = RepositoryObject.includes(:head_version).find_by!(external_identifier: cocina_object.externalIdentifier)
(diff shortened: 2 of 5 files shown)


app/controllers/versions_controller.rb
```diff
@@ -139,8 +139,8 @@ def load_repository_object_version
   end
 
   def status_for(druid:, version:)
-    version_service = VersionService.new(druid:, version:)
     workflow_state_service = WorkflowStateService.new(druid:, version:)
+    version_service = VersionService.new(druid:, version:, workflow_state_service:)
```
justinlittman, line 143: Passing in the existing workflow state service to avoid making duplicates calls to WF.

app/services/version_service.rb
```diff
@@ -65,7 +67,7 @@ def open(cocina_object:, description:, assume_accessioned:, opening_user_name: n
     raise ArgumentError, 'description is required to open a new version' if description.blank?
 
     ensure_openable!(assume_accessioned:)
-    repository_object = RepositoryObject.find_by!(external_identifier: cocina_object.externalIdentifier)
+    repository_object = RepositoryObject.includes(:head_version).find_by!(external_identifier: cocina_object.externalIdentifier)
```
justinlittman, line 70: Avoid extra db calls.

app/services/version_service.rb
```diff
@@ -86,25 +88,30 @@ def open(cocina_object:, description:, assume_accessioned:, opening_user_name: n
   # @raise [CocinaObjectNotFoundError] if the object is not found
   # @raise [VersioningError] if the version does not match the head version
   def open?
-    repo_obj = RepositoryObject.find_by(external_identifier: druid)
+    @open ||=  begin
```
justinlittman, line 91: Memoization to avoid re-running code.

app/services/version_service.rb
```diff
@@ -86,25 +88,30 @@ def open(cocina_object:, description:, assume_accessioned:, opening_user_name: n
...
+  # @param [Boolean] check_preservation If true, checks Preservation for the current version.
   # @return [Boolean] true if a new version can be opened.
   # @raise [Preservation::Client::Error] if bad response from preservation catalog.
-  def can_open?(assume_accessioned: false)
-    ensure_openable!(assume_accessioned:)
-    retrieve_version_from_preservation if Settings.version_service.sync_with_preservation
-    true
-  rescue VersionService::VersioningError
-    false
+  def can_open?(assume_accessioned: false, check_preservation: true)
```
justinlittman, line 107: Yeah, let's not hit preservation if not necessary.

app/services/workflow_state_service.rb
```diff
@@ -74,23 +72,31 @@ def workflow_client
 
   # @return [Boolean] true if there is a workflow for the current version and it has incomplete steps.
   def active_workflow?(workflow:)
-    workflow_response = workflow_client.workflow(pid: druid, workflow_name: workflow)
+    workflow_response = workflow_for(workflow_name: workflow)
```
justinlittman, line 75: Guess what? All of the workflows can be retrieved in 1 call to WF instead of 1 call per WF.

mjgiarlo (approved):
Approve pending successful round of integration tests

app/services/version_service.rb
aaron-collier: Just to make sure I'm understanding, you're making checking preservation optional, but not turning off checking preservation anywhere in this PR. Is that correct and expected?

justinlittman: Oh, you're right. That got dropped in a rebase. Fixing.

aaron-collier (changes requested):
HOLD while running integration tests

PR #5279, opened 2025-03-19
Prevent indexing the same repository object concurrently to avoid rac…

### LINKED ISSUE #5231: Indexing is not keeping up with accessioning

justinlittman:
…es to commit solr record.

closes #5231

### DIFF
--- a/app/services/redis_lock.rb
+++ b/app/services/redis_lock.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+# A simple Redis-based lock implementation.
+# This is useful for ensuring that only one worker is processing a given job at a time.
+# Based on https://github.com/sul-dlss/preservation_catalog/blob/main/app/jobs/concerns/unique_job.rb
+class RedisLock
+  # @param [String] key the key to lock
+  # @param [Integer] lock_timeout the number of seconds before the lock expires
+  # @return [Boolean] true if the lock can be acquired
+  def self.lock(key:, lock_timeout:)
+    now = Time.now.to_i
+    new_expiry_time = now + lock_timeout + 1
+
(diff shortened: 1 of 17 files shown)


app/services/redis_lock.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+# A simple Redis-based lock implementation.
+# This is useful for ensuring that only one worker is processing a given job at a time.
+# Based on https://github.com/sul-dlss/preservation_catalog/blob/main/app/jobs/concerns/unique_job.rb
+class RedisLock
+  # @param [String] key the key to lock
+  # @param [Integer] lock_timeout the number of seconds before the lock expires
+  # @return [Boolean] true if the lock can be acquired
+  def self.lock(key:, lock_timeout:)
```
justinlittman, line 10: There's really no good way to test this class.

aaron-collier (approved):

PR #5326, opened 2025-04-30
Make reports output to STDOUT consistently per cocina-models documentation

mjgiarlo:


app/reports/digital_serials.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Invoke via:
-# bin/rails r -e production "DigitalSerials.report" > digital_serials.csv
+# bin/rails r -e production "DigitalSerials.report"
```
mjgiarlo, line 4: All of DSA's reports output to STDOUT and most do not suggest a filename to which to redirect output. I removed the few examples that specify a filename so as not to suggest that we need to prescribe the output filename (suggesting it's special for some reason, maybe because something else depends on it being there). The DSA README points to the cocina-models README for report documentation which already says that output goes to STDOUT.

lwrubel (approved):

lwrubel:
Thanks for improving the consistency in the reports!

PR #5327, opened 2025-04-30
Make FOLIO update honeybadger context more interesting

mjgiarlo:
Let the interesting exception bubble up to Honeybadger instead of eating it.

### DIFF
--- a/app/services/catalog/folio_writer.rb
+++ b/app/services/catalog/folio_writer.rb
@@ -108,9 +108,9 @@ module Catalog
         retry
       end
 
-      Honeybadger.notify('Error updating Folio record', error_message: e.message)
+      Honeybadger.context(try_count: @try_count)
 
-      raise StandardError, 'FOLIO update not completed.'
+      raise e
     end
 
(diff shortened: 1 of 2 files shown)


app/services/catalog/folio_writer.rb
```diff
@@ -108,9 +108,9 @@ def retry_lookup
         retry
       end
 
-      Honeybadger.notify('Error updating Folio record', error_message: e.message)
```
mjgiarlo, line 111: Before, we were eating what was raised (on line 102 as exception `e`) after raising interesting exceptions elsewhere in this code. Instead of eating that useful exception, let it bubble up. Why? 

# Before

Two HB notifications are logged at the same time, neither of which gives any clues as to the underlying error:

1. `FOLIO update not completed.`
2. `Error updating Folio record`

# After

Only one HB notification is logged and gives us a clue about what's going on:

Error message: `No matching PURL found in instance record after update.`

aaron-collier (approved):

PR #5334, opened 2025-05-02
Digital serials report with part label field

### LINKED ISSUE #5331: Add field to DigitalSerialsTitleParts report
Add a field to https://github.com/sul-dlss/dor-services-app/blob/main/app/reports/digital_serials_title_parts.rb that shows what the new partLabel will be, based on the concatenation logic for the migrator:

> Concatenate the title "part name" and "part number" values in their current order. There may be multiple part name or part number fields.
> Concatenate with a comma between each part.

This logic can then be used in the migrator, once reviewed. 

Run the report in prod using the two druids files in https://github.com/sul-dlss/dor-services-app/issues/5321.

lwrubel:
Resolves #5331 to include a part label field that shows what the label would look like with migration. There are some questions about this, so keeping this in draft until I can get answers to those.

### DIFF
--- a/app/reports/digital_serials_title_parts.rb
+++ b/app/reports/digital_serials_title_parts.rb
@@ -4,6 +4,7 @@
 # bin/rails r -e production "DigitalSerialsTitleParts.report(filename: '/opt/app/deploy/dor-services-app/druid_list.txt')"
 class DigitalSerialsTitleParts
   # Report on title part names and part numbers for a provided set of druids.
+  # Also generates a part label for each druid if one does not already exist in the catalogLinks.
   # The provided file contains a list of druids, one per line, without the 'druid:' prefix.
   # filename should be the full path.  There won't be shell expansion, so e.g. "~" for home dir won't work.
   def self.report(...)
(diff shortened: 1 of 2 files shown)


app/reports/digital_serials_title_parts.rb
```diff
@@ -19,26 +20,37 @@ def initialize(filename:)
   def report
     raise "Input file missing: #{filename}" unless File.exist?(filename)
 
-    puts 'druid,catalogRecordId,part_names,part_numbers'
-    File.foreach(filename) do |raw_line|
+    puts 'druid,catalogRecordId,part_names,part_numbers,part_label'
+    File.foreach(filename) do |raw_line| # rubocop:disable Metrics/BlockLength
```
mjgiarlo, line 24: I wonder if we should add `app/reports/**/*.rb` to the exclude list for this rule in `.rubocop.yml`. I'm guessing we're not going to care much about block length for any reports.

app/reports/digital_serials_title_parts.rb
```diff
@@ -19,26 +20,37 @@ def initialize(filename:)
...
       druid = "druid:#{raw_line.chomp}"
```
mjgiarlo, line 25: Totally trivial, admittedly, but:

```suggestion
    File.foreach(filename, chomp: true) do |bare_druid| # rubocop:disable Metrics/BlockLength
      druid = "druid:#{bare_druid}"
```

app/reports/digital_serials_title_parts.rb
```diff
@@ -19,26 +20,37 @@ def initialize(filename:)
...
+      structured_part_names = []
+      parallel_part_names = []
+      structured_part_numbers = []
+      parallel_part_numbers = []
 
-      structured_part_numbers = cocina_object.description.title.map(&:structuredValue).flatten.select { |part| part.type == 'part number' }.map(&:value)
-      parallel_part_numbers = cocina_object.description.title.map(&:parallelValue).flatten.map(&:structuredValue).flatten.select { |part| part.type == 'part number' }.map(&:value)
+      if @title.structuredValue.any?
+        structured_part_names = @title.structuredValue.select { |part| part.type == 'part name' }&.map(&:value)
+        structured_part_numbers = @title.structuredValue.select { |part| part.type == 'part number' }.map(&:value)
```
mjgiarlo, line 37: Looks like the part names use safe navigation and the part numbers don't. Just making sure that's intended!

app/reports/digital_serials_title_parts.rb
```diff
@@ -50,4 +62,45 @@ def report
...
+
+    part_parts = []
+    structured_values.each do |structured_value|
+      structured_value.each do |part|
+        part_parts << part if part_types.include?(part.type)
+      end
+    end
+
+    part_parts.each_with_index.map do |part, index|
+      # if the part is not the first one, check the previous part type to determine the delimiter
```
mjgiarlo, line 95: 🙏🏻 🎉

app/reports/digital_serials_title_parts.rb
```diff
@@ -50,4 +62,45 @@ def report
...
+    end
+
+    part_parts = []
+    structured_values.each do |structured_value|
+      structured_value.each do |part|
+        part_parts << part if part_types.include?(part.type)
+      end
+    end
+
+    part_parts.each_with_index.map do |part, index|
```
mjgiarlo, line 94: Trivial:

```suggestion
    part_parts.map.with_index do |part, index|
```

mjgiarlo (approved):
Approving because all of my comments and questions are trivial 😄

app/reports/digital_serials_title_parts.rb
lwrubel: Thanks for the suggestion, adding this.

PR #5338, opened 2025-05-07
Faster missing druids

justinlittman:
Parallelizes reindexing rake task and makes indexing logic configurable.

So that SDR can recover faster from Solr indexing issues.

### DIFF
--- a/lib/tasks/missing_druids.rake
+++ b/lib/tasks/missing_druids.rake
@@ -6,9 +6,9 @@ namespace :missing_druids do
     results = SolrService.query('id:*', fl: 'id', rows: 10_000_000, wt: 'csv')
     druids_from_solr = results.pluck('id')
 
-    druids_from_db = RepositoryObject.order(external_identifier: :asc).pluck(:external_identifier)
+    druids_from_db = RepositoryObject.order(updated_at: :desc).pluck(:external_identifier)
 
-    missing_druids = druids_from_db - druids_from_solr.sort
+    missing_druids = druids_from_db - druids_from_solr
     File.open('missing_druids.txt', 'w') do |file|
       missing_druids.map { |druid| file.write("#{druid}\n") }
     end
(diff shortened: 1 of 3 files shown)


lib/tasks/missing_druids.rake
```diff
@@ -22,12 +22,35 @@ namespace :missing_druids do
...
+      batch_start_time = Time.zone.now
+      solr_docs = Parallel.filter_map(batch, in_processes: Settings.rolling_indexer.num_parallel_processes) do |druid|
+        cocina_object = CocinaObjectStore.find(druid)
+        # This returns a Solr doc hash
+        Indexing::Builders::DocumentBuilder.for(
+          model: cocina_object,
+          trace_id: Indexer.trace_id_for(druid:)
+        ).to_solr
+      rescue CocinaObjectStore::CocinaObjectNotFoundError
+        # Return `nil`, which is compacted, so the Solr add isn't grumpy
```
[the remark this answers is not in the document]
justinlittman: `Parallel.filter_map` takes care of it.

lib/tasks/missing_druids.rake
```diff
@@ -22,12 +22,35 @@ namespace :missing_druids do
...
+  # To parallelize: SETTINGS__ROLLING_INDEXER__NUM_PARALLEL_PROCESSES=4 bin/rake missing_druids:index_unindexed_objects
   task index_unindexed_objects: :environment do
-    File.readlines('missing_druids.txt').each do |line|
-      druid = line.chomp
-      puts "Indexing #{druid}"
-      cocina_object = CocinaObjectStore.find(druid.chomp)
-      Indexer.reindex(cocina_object:)
+    solr_conn = RSolr.connect(timeout: 120, open_timeout: 120, url: Settings.solr.url)
+
+    druids = File.readlines('missing_druids.txt').map(&:chomp)
```
mjgiarlo, line 30: Trivial change, but: 

```suggestion
    druids = File.readlines('missing_druids.txt', chomp: true)
```

lib/tasks/missing_druids.rake
```diff
@@ -22,12 +22,35 @@ namespace :missing_druids do
...
+
+    batches = druids.each_slice(Settings.rolling_indexer.batch_size)
+    batches.each_with_index do |batch, index|
```
mjgiarlo, line 33: Since we seem to want the human-readable index (`index + 1`), we can tell `#each_with_index` the starting value here and get rid of the addition below.

```suggestion
    batches.each_with_index(1) do |batch, index|
```

PR #5341, opened 2025-05-08
Use versions association for autosave

lwrubel:
To get migrated cocina to be saved. 


app/services/migrators/move_digital_serials.rb
```diff
@@ -14,10 +15,13 @@ def migrate?
     end
 
     def migrate
-      versions = [repository_object.head_version]
+      # This migration is meant to run in commit mode, which uses autosave on a repository object's versions.
+      # Since only the versions association has autosave at this time,
+      # we can't use the head_version or last_closed_version associations here
```
mjgiarlo, line 20: 🎉

mjgiarlo (approved):

PR #5344, opened 2025-05-12
Fix digital serials migrator to mutate versions

lwrubel:

Mutate the repository object versions, not an array of versions

On QA with a couple of druids. Will do broader testing once merged.

### DIFF
--- a/app/services/migrators/move_digital_serials.rb
+++ b/app/services/migrators/move_digital_serials.rb
@@ -14,18 +14,15 @@ module Migrators
       DRUIDS.include?(repository_object.external_identifier)
     end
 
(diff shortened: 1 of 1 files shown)


app/services/migrators/move_digital_serials.rb
```diff
@@ -14,18 +14,15 @@ def migrate?
       DRUIDS.include?(repository_object.external_identifier)
     end
 
-    def migrate # rubocop:disable Metrics/AbcSize
```
mjgiarlo, line 17: Bonus!

mjgiarlo (approved):

PR #5357, opened 2025-05-27
Map to Datacite related resource

justinlittman:
Datacite now supports related resource. H3 supports additional related resource types.

Unit


arcadiafalcone:
Looks good to me.

PR #5366, opened 2025-06-04
Switch how we pass event data from dsa-client to be specific list instea of json

peetucket:

Follow on from https://github.com/sul-dlss/dor-services-app/pull/5364
Fixes https://github.com/sul-dlss/hungry-hungry-hippo/issues/1411

Ref https://github.com/sul-dlss/hungry-hungry-hippo/issues/1411 and https://github.com/sul-dlss/hungry-hungry-hippo/issues/1371

Goes with https://github.com/sul-dlss/dor-services-client/pull/527 and https://github.com/sul-dlss/dor-services-client/pull/528

- switches from using event data as a general hash and instead uses specific params
- updates the open api to reflect this
- adds the ability to send the who param when registering objects and pass this into the event
- refactor the `UpdateObjectService.update` method to use cocina object has a param for consistency

CI and deploying to QA

### DIFF
--- a/app/services/create_object_service.rb
+++ b/app/services/create_object_service.rb
@@ -11,6 +11,7 @@
 class CreateObjectService
   # @param [Cocina::Models::RequestDRO,Cocina::Models::RequestCollection,Cocina::Models::RequestAdminPolicy] cocina_object # rubocop:disable Layout/LineLength
   # @param [boolean] assign_doi
+  # @param [string] who the sunetid of the user performing the update
   # @param [#call] id_minter assigns identifiers. You can provide your own minter if you want to use a specific druid
   # for an item.
   # @return [Cocina::Models::DROWithMetadata,Cocina::Models::CollectionWithMetadata,Cocina::Models::AdminPolicyWithMetadata] # rubocop:disable Layout/LineLength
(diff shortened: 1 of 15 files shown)


spec/requests/update_metadata_spec.rb
```diff
@@ -188,20 +188,40 @@
...
+        { who: 'test_user', description: 'updating stuff' }
+      end
+
+      it 'returns the updated object' do
+        patch("/v1/objects/#{druid}?event_data=#{event_data.to_json}",
+              params: data,
+              headers:)
+        expect(response).to have_http_status(:ok)
+        expect(response.body).to include('Not a title')
+      end
```
peetucket, line 224: this test fails with the previous code, and passes now

justinlittman (commented):
Oh, I really don't like passing JSON in query parameters. Is there a reason to not just pass `who` and `description` as query parameters?

I'd suggest that this is idiomatic and would allow description/validation in our JSON schema.

peetucket:
> Oh, I really don't like passing JSON in query parameters. Is there a reason to not just pass `who` and `description` as query parameters?
> 
> I'd suggest that this is idiomatic and would allow description/validation in our JSON schema.

That's fine, except if we ever wanted to pass other event fields, we'd need to add to the params list.   And touch two places besides the consumer (the dsa client here:https://github.com/sul-dlss/dor-services-client/blob/main/lib/dor/services/client/mutate.rb#L47 and in DSA too).  The knowledege of the fields we were passing in would be also be potentially be spread across these two locations

justinlittman:
That seems fine and is consistent with our documented and validated API.

app/services/create_object_service.rb
```diff
@@ -42,7 +43,7 @@ def create(cocina_request_object, assign_doi: false) # rubocop:disable Metrics/A
 
     add_project_tag(druid, cocina_request_object)
 
-    EventFactory.create(druid:, event_type: 'registration', data: cocina_object.to_h)
+    EventFactory.create(druid:, event_type: 'registration', data: { who:, request: cocina_object.to_h })
```
peetucket, line 46: this mimics how we record events for an update

justinlittman (approved):

PR #5411, opened 2025-06-26
Adds additional workflow methods to workflow service.

### LINKED ISSUE #5410: Create workflow service
To encapsulate all invocations of dor-workflow-client.

justinlittman:
closes #5410

Encapsulate the workflow client.

Unit


peetucket:
As part of https://github.com/sul-dlss/technical-metadata-service/issues/617, I was updating the README and noticed we were missing the "show" route in `WorkflowProcessController`, which will get a step status for the given workflow, e.g. equivalent of this call: https://github.com/sul-dlss/dor-workflow-client/blob/main/lib/dor/workflow/client/workflow_routes.rb#L70-L78

I still think we will want this, even with this PR work, correct?

peetucket (approved):

PR #5412, opened 2025-06-26
add workflow-status route

peetucket:

Add the route which allows to fetch the status of a single workflow step, ie. lets us call this: https://github.com/sul-dlss/dor-workflow-client/blob/main/lib/dor/workflow/client/workflow_routes.rb#L70-L78

Specs


peetucket:
closed in favor of https://github.com/sul-dlss/dor-services-client/pull/547

PR #5444, opened 2025-07-16
Port WorkflowProcessService to use local workflows.

### LINKED ISSUE #5443: Port WorkflowProcessService

justinlittman:
closes #5443

Merge workflow server

QA


app/services/workflow_process_service.rb
```diff
@@ -57,4 +69,69 @@ def update_error(error_msg:, error_text: nil)
   def workflow_client
     @workflow_client ||= WorkflowClientFactory.build
   end
+
+  def update_status(status:, elapsed: 0, lifecycle: nil, note: nil, # rubocop:disable Metrics/ParameterLists
```
justinlittman, line 73: This is the crux of this PR.

app/services/workflow_template_loader.rb
```diff
@@ -3,7 +3,8 @@
 ##
 # Loading workflow templates
 class WorkflowTemplateLoader
-  WORKFLOWS_DIR = 'config/workflows'
+  WORKFLOWS_DIR = Rails.root.join('config/workflows').freeze
```
justinlittman, line 6: This fixed a bug whereby the templates could not be found from the rolling indexer, since it has a different execution context.

lib/tasks/rabbitmq.rake
```diff
@@ -18,13 +18,6 @@ namespace :rabbitmq do
     queue = channel.queue('dsa.create-event', durable: true)
     queue.bind(exchange, routing_key: '#')
 
-    # These messages look like this: { druid: step.druid, action: 'workflow updated' }
```
justinlittman, line 21: No longer need to read from this queue.

mjgiarlo (approved):
Heavy lift, good work! LGTM

app/services/workflow_process_service.rb
```diff
@@ -57,4 +69,69 @@ def update_error(error_msg:, error_text: nil)
...
+
+    if last_accession_step?(step)
+      # https://github.com/sul-dlss/argo/issues/3817
+      # Theory is that many commits to solr are not being executed in the correct order, resulting in
+      # older data being indexed last.  This is an attempt to force a delay when indexing the very
+      # last step of the accessionWF.
+      sleep 1
+
+      # In theory, notifications should be sent for every step.
+      # However, currently consumers only care about the end-accession step.
```
mjgiarlo: @justinlittman Do you know when/where this came from? It looks like workflow-server-rails sent notifications for every step. Did we dial that back after workflow functionality was ported to DSA? I ask because I believe pre-assembly is a consumer that also cares about workflow error messages being sent. cc: @aaron-collier

justinlittman: At some point I dialed it back on the apparently mistaken assumption that all consumers only cared about the end-accession step.

PR #5448, opened 2025-07-21
Restore indexing job.

justinlittman:
This is needed when the feature flag is not enabled.

### DIFF
--- a/app/jobs/reindex_by_druid_job.rb
+++ b/app/jobs/reindex_by_druid_job.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+# Reindexes an object given a druid.
+# This worker will connect to "dor.indexing-by-druid" queue. This queue is populated when
+# the workflow service makes new RabbitMQ messages on the sdr.workflow exchange.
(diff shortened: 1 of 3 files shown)


app/jobs/reindex_by_druid_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Reindexes an object given a druid.
+# This worker will connect to "dor.indexing-by-druid" queue. This queue is populated when
+# the workflow service makes new RabbitMQ messages on the sdr.workflow exchange.
+# @see https://github.com/sul-dlss/dor_indexing_app/blob/8546e8aabb76d506fe3f5f5ea43ae499a442d75f/lib/tasks/rabbitmq.rake#L20-L21
```
jmartin-sul, line 6: not for this PR since it's just restoring a mis-deleted file to fix a prod bug, but this line could use some freshening

jmartin-sul (approved):
one nitpick for a follow-on

app/jobs/reindex_by_druid_job.rb
jmartin-sul: https://github.com/sul-dlss/dor-services-app/pull/5449

PR #5470, opened 2025-07-29
Reduce the number of db queries to retrieve batches of version statuses.

justinlittman:
Faster rendering of H3 dashboard.

Unit, stage


app/models/repository_object.rb
```diff
@@ -158,11 +214,11 @@ def reopen!
   end
 
   def open?
-    head_version == opened_version
+    head_version_id == opened_version_id
```
justinlittman, line 217: This avoids lazy loading when not necessary.

app/models/version_context.rb
```diff
@@ -2,6 +2,9 @@
 
 # Models optional context that is associated with a druid/version pair for any workflow
 class VersionContext < WorkflowApplicationRecord
+  has_many :workflow_steps, foreign_key: %i[druid version], primary_key: %i[druid version], dependent: :destroy,
```
justinlittman, line 5: Switching VersionContext to use plain-old Rails associations.

app/services/cocina_object_store.rb
```diff
@@ -102,7 +102,7 @@ def exists!(druid)
   end
 
   def version(druid)
-    RepositoryObject.find_by!(external_identifier: druid).head_version.version
+    RepositoryObject.joins(:head_version).select(:version).find_by!(external_identifier: druid).version
```
justinlittman, line 105: More efficient query.

app/services/version_service.rb
```diff
@@ -49,10 +46,12 @@ def self.discard(...)
   # @param [String] druid of the item
   # @param [Integer] version of the item
   # @param [WorkflowStateService] workflow_state_service
-  def initialize(druid:, version:, workflow_state_service: nil)
+  # @param [RepositoryObject] repository_object optional object to check against, otherwise it will be fetched
+  def initialize(druid:, version:, workflow_state_service: nil, repository_object: nil)
```
justinlittman, line 50: Added passing in repository_object to minimize the number of times it has to be retrieved.

app/services/workflow_service.rb
```diff
@@ -106,7 +105,7 @@ def workflow(workflow_name:)
       steps = WorkflowStep.where(
         druid:,
         workflow: workflow_name
-      ).order(:workflow, created_at: :asc)
+      ).left_outer_joins(:version_context).order(:workflow, created_at: :asc)
```
justinlittman, line 108: This avoids lazy loading.

app/services/version_batch_status_service.rb
```diff
@@ -0,0 +1,135 @@
...
+    end
+  end
+
+  def self.call(...)
+    new(...).call
+  end
+
+  # raises CocinaObjectStore::CocinaObjectNotFoundError if the druid is not found
+  def self.call_single(druid:)
+    new(druids: [druid]).call[druid].tap do |status|
```
mjgiarlo, line 26: Nothing wrong with this, but it does look kinda funky. I did a double-take anyway. Could consider a tweak like:


```suggestion
    new(druids: [druid]).call.fetch(druid, nil).tap do |status|
```

app/services/version_batch_status_service.rb
```diff
@@ -0,0 +1,135 @@
...
+                                               gisDeliveryWF ocrWF speechToTextWF gisAssemblyWF])
+                           .incomplete
+                           .where.not(workflow: 'assemblyWF', process: 'accessioning-initiate')
+                           .where.not(workflow: 'wasCrawlPreassemblyWF', process: 'end-was-crawl-preassembly')
+                           .where.not(workflow: 'wasSeedPreassemblyWF', process: 'end-was-seed-preassembly')
+                           .where.not(workflow: 'gisDeliveryWF', process: 'start-accession-workflow')
+                           .where.not(workflow: 'ocrWF', process: 'end-ocr')
+                           .where.not(workflow: 'speechToTextWF', process: 'end-stt')
+                           .select(:druid).distinct.pluck(:druid)
+  end
```
mjgiarlo, line 103: I wondered whether these three query chains could be moved into the `WorkflowStep` model as named scopes, but if this is the only collaborator that cares about these queries, it probably doesn't make a difference.

app/services/version_batch_status_service.rb
```diff
@@ -0,0 +1,135 @@
...
+    workflow_state_service = workflow_state_service_for(druid)
+    version_service = VersionService.new(druid:, version:, workflow_state_service:, repository_object:)
+
+    {
+      versionId: version,
+      open: version_service.open?,
+      # checks workflow state service accessioned?, accessioning?
+      openable: version_service.can_open?(check_preservation: false),
+      assembling: workflow_state_service.assembling?,
+      accessioning: accessioning_druids.include?(druid),
```
mjgiarlo, line 120: Could this be:


```suggestion
      accessioning: workflow_state_service.accessioning?,
```

spec/support/db_query_matchers_config.rb
```diff
@@ -0,0 +1,18 @@
...
+  # http://edgeguides.rubyonrails.org/active_support_instrumentation.html#sql-active-record
+  config.on_query_counted do |payload|
+    # do something arbitrary with the query
+  end
+
+  config.log_backtrace = true
+  config.backtrace_filter = proc do |backtrace|
+    backtrace.select { |line| line.start_with?(Rails.root.to_s) }.take(100)
+  end
+end
```
mjgiarlo, line 18: How/where is this used? (It looks useful.)

app/services/version_batch_status_service.rb
justinlittman: Yeah, these will be used only here.

spec/support/db_query_matchers_config.rb
justinlittman: For example, https://github.com/sul-dlss/dor-services-app/pull/5470/files#diff-5ee10b04a34b5fdcf8e4d3e6c7681ed1fa8059377cbc9eb2a23716b1cd9835eeR78

PR #5473, opened 2025-07-30
Use direct db queries to get workflow state

justinlittman:
So that there is only one approach to getting workflow state.

Unit

### DIFF
--- a/spec/requests/versions_spec.rb
+++ b/spec/requests/versions_spec.rb
@@ -279,11 +279,6 @@ RSpec.describe 'Operations regarding object versions' do
       end
 
       context 'when the object is not found' do
-        # before do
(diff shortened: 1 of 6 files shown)


spec/requests/versions_spec.rb
```diff
@@ -279,11 +279,6 @@
       end
 
       context 'when the object is not found' do
-        # before do
-        #   allow(VersionBatchStatusService).to receive(:new).and_raise(CocinaObjectStore::CocinaObjectNotFoundError,
-        #                                                               'Object not found')
-        # end
```
mjgiarlo, line 285: Cruft?

mjgiarlo (approved):

PR #5475, opened 2025-07-31
Cache templates

justinlittman:
Performance optimization.

Unit


app/services/workflow/template_loader.rb
```diff
@@ -22,6 +15,8 @@ def self.load_as_xml(workflow_name)
     # @param [String] workflow_name name/id of workflow, e.g., accessionWF
     def initialize(workflow_name)
       @workflow_name = workflow_name
+      @@cache ||= {} # rubocop:disable Style/ClassVars
```
mjgiarlo, line 18: Given the recent situations we've seen with odd thread behavior and the fact that class variables and threading do not get along, I wonder if we should cache these values using a different approach, e.g., https://guides.rubyonrails.org/caching_with_rails.html#low-level-caching-using-rails-cache.

mjgiarlo: I was just about to ask about that. Good enough for now.

PR #5483, opened 2025-08-04
Adds spec for when next step service. Moves next step service to workflow module.

justinlittman:
https://stanfordlib.slack.com/archives/C09M7P91R/p1754085036484509


app/services/workflow/next_step_service.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+module Workflow
```
justinlittman, line 3: Ooops, I missed adding this to the workflow module.

spec/services/workflow/next_step_service_spec.rb
```diff
@@ -96,5 +96,20 @@
         expect(next_steps).to eq []
       end
     end
+
+    context 'when an error' do
```
justinlittman, line 100: A new spec.

PR #5497, opened 2025-09-03
[HOLD] Include the request cocina file set IDs when create objects

mjgiarlo:
Connects to sul-dlss/purl#1454

By skipping the cocina ID generator, the create object service generates bogus FS IDs when the request cocina already has a non-URI value set. This is what pre-assembly does, in fact, expecting for its IDs to be URI-ified. This does not happen currently.


mjgiarlo:
@andrewjbtw We worked ☝🏻 up for the versioning work cycle when the Access devs pointed out the bug. I haven't been tracking that work cycle since, so this is hanging out for now. I suspect we'd want to test this change before taking it forward, but I defer to the versioning WC team whether this is a change we're ready for.

mjgiarlo:
@andrewjbtw 💬 
> @mjgiarlo I was going to ping you about this. Based on discussion with the Access team, they think this can go in but I do want to take a quick look on stage.
> 
> Am I correct that the only way to see the effect is to run Preassembly?

I am fuzzy on whether there are other scenarios that trigger this affect, but yes, pre-assembly is the one place we were confident it was happening.

Feel free to ping either me or the FR on slack whenever you have time for testing (branch to deploy: `cocina-id-generator/purl#1454`)

mjgiarlo:
I believe @andrewjbtw tested this and we concluded it didn't have the intended effect? Should we add this to the Argo WC? I'm fuzzy on where we left this, sorry.

PR #5501, opened 2025-09-09
Handle related arXiv and pmid resources

### LINKED ISSUE #5499: Datacite DOI registration is failing
We appear to be getting an HTTP 422 error from DataCite's API when trying to register a DOI for this item:

https://argo.stanford.edu/view/druid:cz537wr8540

The body of the error looks like:

```json
{
  "errors": [
    {
      "source": "metadata", 
      "title": "Is invalid", 
      "uid": "10.25740/cz537wr8540"
    }
  ]
}
```

The Honeybadger alert is here:

https://app.honeybadger.io/projects/50568/faults/122916314

From the stacktrace it appears that DSA is using the `Datacite::Client.update` method to register a DOI. 

The client appears to be doing an HTTP PUT to `https://api.datacite.org/dois/10.25740/cz537wr8540`:

https://github.com/sul-dlss/datacite-ruby/blob/58785fb6ae026dae12a0173b29c77d36fe760b07/lib/datacite/client.rb#L42-L51

This DOI doesn't exist yet, so it kinda makes sense why it wouldn't be able to update the Datacite resource? It seems like a calling `Datacite::Client.register` would be more appropriate, but it doesn't look like we are actually using that in H3 or DSA (or anywhere in Github):

https://github.com/sul-dlss/datacite-ruby/blob/58785fb6ae026dae12a0173b29c77d36fe760b07/lib/datacite/client.rb#L33-L40

I sent an email to support@datacite.org to see if the API changed recently, since @amyehodge believes that this is the way we've always registered DOIs.

Or perhaps something wrong with the JSON payload that is being posted? The JSON that was posted was copy and pasted out of the HB alert, prettified, and attached here:

[datacite.json](https://github.com/user-attachments/files/22151761/datacite.json)

edsu:

We weren't adding related resources fully if the Cocina related resource identifier type was `arxiv` or `pmid`.

Fixes #5499

Unit tests, and sending some live requests to the test API at api.test.datacite.org.


app/services/cocina/to_datacite/related_resource.rb
```diff
@@ -109,32 +93,17 @@ def related_identifier_blank?
       end
 
       def related_item_title
-        @related_item_title ||= preferred_citation || other_title || related_item_doi || related_item_identifier_url
-      end
-
```
edsu, line 114: I removed these comments since they weren't uniformly applied as examples, and seemed redundant.

spec/services/cocina/to_datacite/related_resource_spec.rb
```diff
@@ -157,4 +163,72 @@
       )
     end
   end
```
edsu, line 165: These are new tests to ensure that arXiv and pmid related resources get identifiers assigned correctly.

justinlittman (approved):

app/services/cocina/to_datacite/related_resource.rb
```diff
@@ -47,43 +47,41 @@ def initialize(related_resource)
...
+            relationType: relation_type,
+            relatedItemIdentifier: id,
+            relatedItemIdentifierType: type
+          }
+        else
+          {
+            relatedItemType: 'Other',
+            titles: titles,
+            relationType: relation_type
+          }
```
edsu, line 66: I thought it made sense to not add `nil` values if they aren't available, even though it seems like DataCite API doesn't care.

app/services/cocina/to_datacite/related_resource.rb
```diff
@@ -47,43 +47,41 @@ def initialize(related_resource)
...
       # @return [Hash,nil] Array of DataCite relatedIdentifier attributes, conforming to the expectations of HTTP PUT
-      # request to DataCite or nil if blank
+      # request to DataCite or nil if blank or the identifier lacks a URI or Type
       #  see https://support.datacite.org/reference/dois-2#put_dois-id
       def related_identifier_attributes
         return if related_identifier_blank?
 
+        id, type = unpack_related_uri_and_type
+        return unless id && type
+
```
edsu, line 78: It doesn't make sense to have a related identifier without a URL or type. So this will return nil unless those can be determined.

app/services/cocina/to_datacite/related_resource.rb
```diff
@@ -109,32 +107,17 @@ def related_identifier_blank?
...
       end
 
-      # example cocina relatedResource:
-      #   {
-      #     title: [
-      #       {
-      #         value: 'A paper'
-      #       }
-      #     ]
-      #   }
```
edsu, line 137: I removed these examples because they weren't applied uniformly, and seemed redundant.

app/services/cocina/to_datacite/related_resource.rb
```diff
@@ -147,21 +130,35 @@ def related_item_doi
...
+        if related_item_doi
+          [related_item_doi, 'DOI']
+        elsif related_item_arxiv
+          [related_item_arxiv, 'arXiv']
+        elsif related_item_pmid
+          [related_item_pmid, 'PMID']
+        elsif related_item_identifier_url
+          [related_item_identifier_url, 'URL']
+        end
+      end
```
edsu, line 161: This is a new method for determining the URL and URL Type and having the logic in one place.

spec/services/cocina/to_datacite/related_resource_spec.rb
```diff
@@ -157,4 +157,72 @@
...
+        relatedItemIdentifierType: 'PMID'
+      )
+      expect(related_identifier_attributes).to eq(
+        resourceTypeGeneral: 'Other',
+        relationType: 'IsReferencedBy',
+        relatedIdentifier: 'https://pubmed.ncbi.nlm.nih.gov/31060017/',
+        relatedIdentifierType: 'PMID'
+      )
+    end
+  end
```
edsu, line 227: These tests ensure that arXiv and pmid are handled correctly.

PR #5524, opened 2025-10-01
Add find_all endpoint to ObjectController

aaron-collier:

Part of https://github.com/sul-dlss/argo/issues/4824

This will be called by argo initially to find all APO collections for registration without making individual SOLR queries.

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -84,6 +84,15 @@ class ObjectsController < ApplicationController
     render json: Cocina::Models.without_metadata(cocina_object)
   end
 
+  def find_all
+    druids = params.require([:externalIdentifiers]).first
+    cocina_objects = druids.map do |druid|
+      cocina_object = CocinaObjectStore.find(druid)
+      Cocina::Models.without_metadata(cocina_object)
+    end
+    render json: { items: cocina_objects }
+  end
+
   # Initialize specified workflow (assemblyWF by default), and also version if needed
   # called by pre-assembly and goobi kick off accessioning for a new or existing object
   #
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -84,11 +84,15 @@ class ObjectsController < ApplicationController
     render json: Cocina::Models.without_metadata(cocina_object)
   end
 
+  # Find all objects by their external identifiers
+  # Ignores any identifiers that are not found
   def find_all
     druids = params.require([:externalIdentifiers]).first
-    cocina_objects = druids.map do |druid|
-      cocina_object = CocinaObjectStore.find(druid)
-      Cocina::Models.without_metadata(cocina_object)
+    cocina_objects = []
+    druids.each do |druid|
+      cocina_objects << Cocina::Models.without_metadata(CocinaObjectStore.find(druid))
+    rescue CocinaObjectStore::CocinaObjectNotFoundError
+      next
     end
     render json: cocina_objects
   end
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -88,7 +88,7 @@ class ObjectsController < ApplicationController
   # Ignores any identifiers that are not found
   def find_all
     druids = params.require([:externalIdentifiers]).first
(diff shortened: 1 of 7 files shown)


app/controllers/objects_controller.rb
```diff
@@ -84,6 +84,19 @@ def find
     render json: Cocina::Models.without_metadata(cocina_object)
   end
 
+  # Find all objects by their external identifiers
+  # Ignores any identifiers that are not found
+  def find_all
+    druids = params.require([:externalIdentifiers]).first
+    cocina_objects = []
+    druids.each do |druid|
```
justinlittman, line 92: Can you make this so that it makes a single query to db instead of a query for each item? This will require adding a new method to the CocinaObjectStore.

aaron-collier: Done.

app/services/cocina_object_store.rb
```diff
@@ -32,6 +32,15 @@ def self.find(druid)
     new.find(druid)
   end
 
+  # Retrieves a list of Cocina objects from the datastore.
+  # @param [Array<String>] druids
+  # @return [Array<Cocina::Models::DROWithMetadata,
+  #                Cocina::Models::CollectionWithMetadata,
+  #                Cocina::Models::AdminPolicyWithMetadata>] cocina_objects
+  def self.find_all(druids)
+    new.find_all(druids)
```
aaron-collier, line 41: I'm unclear why this is showing as uncovered as it's tested below.

app/controllers/objects_controller.rb
```diff
@@ -84,6 +84,14 @@ def find
     render json: Cocina::Models.without_metadata(cocina_object)
   end
 
+  # Find all objects by their external identifiers
+  # Ignores any identifiers that are not found
+  def find_all
+    druids = params.require([:externalIdentifiers]).first
+    cocina_objects = CocinaObjectStore.find_all(druids).map { |cocina_object| Cocina::Models.without_metadata(cocina_object) }
```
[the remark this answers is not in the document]
aaron-collier: Yeah, maybe - I was going to ask about that as I'm working through the client.

app/controllers/objects_controller.rb
```diff
@@ -84,6 +84,14 @@ def find
...
+    render json: cocina_objects
+  end
```
mjgiarlo, line 93: I wonder about using `index` for the name of this route instead of `find_all` since `index` is the built-in resourceful route name for "get me all of the things"

app/controllers/objects_controller.rb
mjgiarlo, line 91: It looks like the object store find_all is mapping to Cocinas without metadata and then we're mapping through them again here to strip it away. I wonder if `CocinaObjectStore#find_all` should put the objects in whatever state the controller wants them in, so the controller can call the method without mucking about.

app/services/cocina_object_store.rb
```diff
@@ -85,6 +94,14 @@ def find(druid, version: :head)
     raise CocinaObjectNotFoundError.new("Couldn't find object with 'external_identifier'=#{druid}", druid)
   end
 
+  # @param [Array] druids to find
+  # @return [Array<Cocina::Models::DROWithMetadata,Cocina::Models::CollectionWithMetadata,Cocina::Models::AdminPolicyWithMetadata>] for the head version #rubocop:disable Layout/LineLength
+  def find_all(druids)
+    RepositoryObject.where(external_identifier: druids).map do |repo_object|
+      repo_object.head_version.to_cocina_with_metadata
+    end
+  end
```
mjgiarlo, line 103: Since this method doesn't require or track state, how about this moves up into the body of the class method?

app/controllers/objects_controller.rb
aaron-collier: @mjgiarlo that's fixed - Now mapping to with metadata

app/controllers/objects_controller.rb
aaron-collier: I think that would take a bit of rework that may be outside of the scope of this PR (https://github.com/sul-dlss/dor-services-app/blob/main/openapi.yml#L1130) as we already post to index...

PR #5527, opened 2025-10-02
Fix batch reindexing

justinlittman:
Bugs found running in stage.


app/jobs/batch_reindex_job.rb
```diff
@@ -92,7 +94,7 @@ def parent_collections_for(repository_object:)
   end
 
   def parent_collections_release_tags_for(repository_object:)
-    collection_druids_from(repository_object:).flat_map do |collection_druid|
+    collection_druids_from(repository_object:).index_with do |collection_druid|
```
mjgiarlo, line 97: Nice.

mjgiarlo (approved):

PR #5538, opened 2025-10-08
Add option for reverting metadata from last closed version to cleanup service

### LINKED ISSUE #5506: "cleanup" rake task should include option to revert cocina to last accessioned version
When an item gets stuck in accessioning (assemblyWF or accessionWF) and the error is fatal*, we need to:

- stop the accessionWF (and possibly remove assemblyWF for that version too)
- clean up the filesystem locations for that item
- set the item status back to Registered or Opened as appropriate

We have a rake task that takes care of those actions but it does not modify the item's cocina to revert it back to the last accessioned version. This can be a problem if the fatal error happens to be that the item's cocina is invalid. An item can end up with invalid cocina because there are cases where the descriptive cocina validation fails to catch a problem at upload.

These types of cocina validation problems are typically caught during accessioning - publish will fail on invalid cocina - and/or during reindexing.

Recently, a bulk descriptive upload caused 26 items to fail in accessioning due to validation problems. I was able to use the cleanup rake task to bring these back to a state where the cocina could be updated via Argo again, but reindexing began to alert repeatedly on these items while they were invalid. 

I ended up manually restoring the previous version cocina for each item one by one so that HB would stop getting error alerts from reindexing. For each druid, I 

- went to the previous version in Argo
- downloaded that version's descriptive CSV
- went to the current version
- uploaded the previous version's description

Ideally, it would not be necessary to restore the cocina manually. I could have ignored the HB alerts but when an item can't be reindexed, the display is misleading and the item doesn't appear in the expected facets. It's also not always possible for a depositor to return to the item and fix the metadata with a new upload in a short period of time. Restoring the last-accessioned (and valid) version of the descriptive cocina is the only way to stop the alerts.

*Examples of fatal errors:

- checksum mismatch on a file
- cocina descriptive is invalid

mjgiarlo:
# Why was this change made?

Fixes #5506

# How was this change tested?

- [x] CI
- [x] QA


mjgiarlo:
Closes #5506

PR #5551, opened 2025-10-10
Add indexable route to verify descriptive metadata is indexable

aaron-collier:

Parts of argo/4871

### DIFF
--- a/app/controllers/objects_controller.rb
+++ b/app/controllers/objects_controller.rb
@@ -151,16 +151,13 @@ class ObjectsController < ApplicationController
   end
 
   def indexable
-    cocina_object = Cocina::Models.with_metadata(Cocina::Models.build(params.except(:action,
-                                                                                    :controller,
-                                                                                    :id,
-                                                                                    :event_description,
-                                                                                    :user_name).to_unsafe_h),
-                                                 'temp-lock',
-                                                 created: Time.zone.now,
-                                                 modified: Time.zone.now)
+    cocina_object = Cocina::Models.without_metadata(Cocina::Models.build(params.except(:action,
+                                                                                       :controller,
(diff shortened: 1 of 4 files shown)


app/controllers/objects_controller.rb
```diff
@@ -150,6 +150,21 @@ def reindex
...
 
+  def indexable
+    cocina_object = Cocina::Models.with_metadata(Cocina::Models.build(params.except(:action,
+                                                                                    :controller,
+                                                                                    :id,
+                                                                                    :event_description,
+                                                                                    :user_name).to_unsafe_h),
+                                                 'temp-lock',
+                                                 created: Time.zone.now,
+                                                 modified: Time.zone.now)
```
mjgiarlo: Question for my own edification: does it add value to the indexable endpoint to validate the metadata (lock, created, modified)? Or are we passing a Cocina model with metadata so that something doesn't break?

aaron-collier: Yeah - there was an error when I was doing the full solr document, let me see if that changes for only using descriptive and I can maybe remove it now.

openapi.yml
```diff
@@ -194,6 +194,41 @@ paths:
...
+            # Note that additionalProperties needs to be set to true to allow id parameter.
+            schema:
+              oneOf:
+                - $ref: "#/components/schemas/DRO"
+                - $ref: "#/components/schemas/Collection"
+                - $ref: "#/components/schemas/AdminPolicy"
+      responses:
+        "204":
+          description: OK
+        "412":
```
justinlittman: 422 not 412

justinlittman (approved):
Approve pending changing 412 to 422

PR #5566, opened 2025-10-17
Decouple from Solr where DB usage is sufficient

### LINKED ISSUE #4459: Decouple from Solr

mjgiarlo:
# Why was this change made?

Fixes #4459
Supersedes #4578

# How was this change tested?

- [x] CI
- [x] QA (integration tests)


app/models/repository_object.rb
```diff
@@ -11,13 +11,27 @@
...
+  end
+
   class VersionAlreadyOpened < StandardError; end
   class VersionNotOpened < StandardError; end
   class VersionNotDiscardable < StandardError; end
 
-  has_many :versions, lambda {
-    order(version: :asc)
-  }, class_name: 'RepositoryObjectVersion', dependent: :destroy, inverse_of: 'repository_object', autosave: true
+  has_many :versions, -> { order(version: :asc) }, class_name: 'RepositoryObjectVersion',
```
mjgiarlo: Going with stabby lambdas.

app/services/virtual_object_service.rb
```diff
@@ -6,36 +6,32 @@ def self.constituents(...)
     new(...).constituents
   end
 
-  # @param [Cocina::Models::DRO] cocina_dro the cocina model for the virtual object
+  # @param [Cocina::Models::DRO] cocina_item the cocina model for the virtual object
```
mjgiarlo: Per naming conventions in cocina models README

spec/services/published_relationships_filter_spec.rb
mjgiarlo: Dead code

app/models/repository_object.rb
```diff
@@ -38,6 +52,27 @@ class VersionNotDiscardable < StandardError; end
   scope :collections, -> { where(object_type: 'collection') }
   scope :admin_policies, -> { where(object_type: 'admin_policy') }
   scope :closed, -> { where('last_closed_version_id = head_version_id') }
+  scope :currently_has_constituents, ->(constituent_druids) {
```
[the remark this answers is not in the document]
mjgiarlo: Done.

PR #5587, opened 2025-10-22
Do not treat dark objects as released to SearchWorks

justinlittman:
So that 856 fields are not added to MARC records. See https://github.com/sul-dlss/dor-services-app/blob/main/app/services/catalog/folio_writer.rb#L65

Unit


mjgiarlo (approved):
Approve pending consideration of suggestion

PR #5609, opened 2025-10-29
Adds initiating releaseWF to accessionWF and to publish job.

### LINKED ISSUE #5471: Do not run releaseWF on items/collections that are dark
[ This issue should be paired with https://github.com/sul-dlss/dor-services-app/issues/5472 We may actually need to cover that issue first since we will need a mechanism for removing items from SW when rights are changed to dark. ]

Currently, the releaseWF is applied to objects that are dark and can't be released. This can put items in a bad state in three systems:

- Argo (because the item is indexed as "released" even though it isn't public)
- FOLIO (because the item may be included in an 856 that says it's available online even though it is not)
- SW (because the display expects a purl link and/or embed view and there isn't one for dark items)

The simplest approach to this problem would be to not run the releaseWF at all on items/collections that are dark.

Currently, when the releaseWF runs on items that are dark, it correctly skips the `release-publish` step since a dark item is not published and has no purl. 

> Image

But the other parts of the release process are not skipped, which is what leads to the bad state. Argo still indexes the object as "released" and the FOLIO 856 still gets created as if the item is public. That 856 then leads SW to show purl links and/or a non-working viewer.

**Additional context**

I think releaseWF has been running on items that are dark because it has to run in order to perform "un-release"/withdraw actions. When an item is unreleased, we need to update any linked MARC record to remove the 856. So we have to solve that issue in order to solve this one. 

One option would be to make sure that all dark objects have been withdrawn before we prevent the releaseWF from running on dark items.

justinlittman:
closes #5471

Normalize the release process by always using the releaseWF.

Unit, QA

### DIFF
--- a/app/jobs/robots/dor_repo/release/release_members.rb
+++ b/app/jobs/robots/dor_repo/release/release_members.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Robots
+  module DorRepo
+    module Release
+      # Release the member of a collection
+      class ReleaseMembers < Robots::Robot
+        def initialize
+          super('releaseWF', 'release-members')
+        end
+
+        def perform_work
+          return unless cocina_object.collection?
+          return unless add_wf_to_members?
+
+          published_members.each do |member|
+            create_release_workflow(druid: member.external_identifier, version: member.version)
+          end
+        end
+
+        private
+
+        def published_members
+          # .currently_members_of_collection returns a sparse RepositoryObject.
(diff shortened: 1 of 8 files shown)


app/jobs/robots/dor_repo/release/release_members.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module Robots
```
justinlittman: This was moved from common accessioning.

PR #5661, opened 2026-01-15
Set up MARC feature flag for refreshing metadata

### LINKED ISSUE #5665: Add feature flag for using MARC mapping for Cocina
Set up a feature flag for using a `Cocina::Models::FromMarc` mapping instead of `FromMods`. This will allow testing as the MARC mapping is progressively built out.

lwrubel:
Resolves #5665. Sets up a feature flag in order to use a FromMarc cocina mapping instead of MODS.  

Also changes the logic for generating a label when registering, using cocina (via Cocina Display) instead of MODS. 

Registered an item via folio HRID, with no label, and it pulled in the title and assigned a label.  (see https://argo-qa.stanford.edu/view/druid:qr308ht1128)

### DIFF
--- a/app/services/catalog/marc_service.rb
+++ b/app/services/catalog/marc_service.rb
@@ -8,13 +8,15 @@ module Catalog
     class CatalogRecordNotFoundError < MarcServiceError; end
     class TransformError < MarcServiceError; end
 
-    def self.mods(barcode: nil, folio_instance_hrid: nil)
-      new(barcode:, folio_instance_hrid:).mods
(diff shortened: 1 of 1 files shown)


spec/services/create_object_service_spec.rb
```diff
@@ -9,8 +9,6 @@
 
   describe '#create' do
     let(:druid) { 'druid:bc123df4567' }
-    let(:catalog_links) { [] }
```
lwrubel: Unused.

PR #5699, opened 2026-02-12
Suppress INFO-level logging in reindex job runs

mjgiarlo:
# Why was this change made?

Setting the log_level sidekiq option in the job does not suppress AJ logging.

# How was this change tested?

CI, stage


justinlittman (approved):
If you could please add some comments to the actual jobs that are affected, this is good to go.

PR #5761, opened 2026-02-23
Add latest-versions-only option

### LINKED ISSUE #5717: Add option to bin/validate-cocina script to validate certain versions
It will be helpful to be able to limit validation to run on versions that are the head version or last_closed_version (when object is open). This supports the use case when `description` validation changes are implemented and validation needs to be checked for those versions only. 

This issue is for adding an option to the script for latest versions validating. Not sure what the best wording for this is, so add to the script docs / README to explain what it is checking. 

See **https://github.com/sul-dlss/dor-services-app/wiki/Migrating-Cocina** for context.

lwrubel:
Fixes #5717. To be accompanied by a cocina-models README PR. Note this also removes checking the `ToMods` mapping for the description. `ToMods` will be moving to purl-fetcher and removed from Cocina::Models in the future. 

Tested on QA with various versions made invalid.

### DIFF
--- a/bin/validate-cocina
+++ b/bin/validate-cocina
@@ -53,10 +53,9 @@ def tty_progress_bar(count)
   )
 end
 
-def versions_to_validate(repository_object, latest_versions_only)
+def versions_to_validate(repository_object)
   versions = repository_object.versions.where.not(cocina_version: nil)
-
-  return versions unless latest_versions_only
+  return versions unless options[:latest_versions_only]
 
   # Only include head_version and last_closed_version (if object is open)
   latest_versions = []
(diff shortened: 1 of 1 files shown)


bin/validate-cocina
```diff
@@ -50,7 +53,22 @@ def tty_progress_bar(count)
   )
 end
 
-def validate(sample_size, processes)
+def versions_to_validate(repository_object, latest_versions_only)
```
mjgiarlo: Does `latest_versions_only` need to be passed in, or can this method directly access `options[:latest_versions_only]`?

PR #5840, opened 2026-03-03
publish messages to the rabbit queue when an error occurs

peetucket:

See sul-dlss/pre-assembly#1648

As per Mike's analysis, this may only be happening when an error occurs during accessioning, since we do not appear to be publishing messages to RabbitMQ when an error in a workflow step occurs.  Since pre-assembly counts on these messages, it will never know.

We already have a rabbitmq exchage setup so that "error" messages to the `preassembly.accession_complete` queue, so everything else should be setup to process these messages.  See https://sul-rabbit-stage.stanford.edu/#/exchanges/workflow/sdr.workflow

~~Not yet. Hold for testing.  Plan:~~

1.  Deploy to stage, run an integration test with pre-assembly that succeeds.  Verify status is still updated in pre-assembly (it works now, it should continue to work).
2. Run an integration test with pre-assembly that fails during accessioning.  Verify status is now updated in pre-assembly to error (it likely does not work now, and now it should)

Test was successful, see https://github.com/sul-dlss/pre-assembly/issues/1648#issuecomment-3999594747

### DIFF
--- a/app/services/workflow/next_step_service.rb
+++ b/app/services/workflow/next_step_service.rb
@@ -18,17 +18,15 @@ module Workflow
     # @return [ActiveRecord::Relation] a list of WorkflowSteps that have been enqueued
     def enqueue_next_steps(step:)
       next_steps = find_next(step:)
-      if last_accession_step?(step)
-        # https://github.com/sul-dlss/argo/issues/3817
-        # Theory is that many commits to solr are not being executed in the correct order, resulting in
-        # older data being indexed last.  This is an attempt to force a delay when indexing the very
-        # last step of the accessionWF.
-        sleep 1
 
-        # In theory, notifications should be sent for every step.
-        # However, currently consumers only care about the end-accession step.
-        Notifications::WorkflowStepUpdated.publish(step:)
-      end
+      # https://github.com/sul-dlss/argo/issues/3817
+      # Theory is that many commits to solr are not being executed in the correct order, resulting in
+      # older data being indexed last.  This is an attempt to force a delay when indexing the very
+      # last step of the accessionWF.
+      sleep 1 if last_accession_step?(step)
+
+      # send message to RabbitMQ for accessioning complete/error steps, pre-assembly/H3 consume one or both to update UI
+      Notifications::WorkflowStepUpdated.publish(step:) if step.status == 'error' || last_accession_step?(step)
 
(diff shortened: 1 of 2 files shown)


app/services/workflow/next_step_service.rb
```diff
@@ -18,17 +18,15 @@ def initialize
...
-      end
+
+      # https://github.com/sul-dlss/argo/issues/3817
+      # Theory is that many commits to solr are not being executed in the correct order, resulting in
+      # older data being indexed last.  This is an attempt to force a delay when indexing the very
+      # last step of the accessionWF.
+      sleep 1 if last_accession_step?(step)
+
+      # send message to RabbitMQ for accessioning complete or error occurred, pre-assembly consumes these to update UI
+      Notifications::WorkflowStepUpdated.publish(step:) if step.status == 'error' || last_accession_step?(step)
```
peetucket: this is the real change, it basically does the same as above but also does the notification for an error

spec/services/workflow/next_step_service_spec.rb
```diff
@@ -106,10 +100,25 @@
                active_version: true)
       end
 
-      it 'does not enqueue any steps' do
+      it 'does not enqueue any steps but publishes a message' do
         expect { next_steps }.not_to(change { step.reload.status })
+        expect(Notifications::WorkflowStepUpdated).to have_received(:publish).with(step:)
```
peetucket: this new expectation would have failed before since no messages are currently sent with a workflow error

mjgiarlo (approved):

app/services/workflow/next_step_service.rb
```diff
@@ -18,17 +18,15 @@ def initialize
...
-        Notifications::WorkflowStepUpdated.publish(step:)
-      end
+
+      # https://github.com/sul-dlss/argo/issues/3817
+      # Theory is that many commits to solr are not being executed in the correct order, resulting in
+      # older data being indexed last.  This is an attempt to force a delay when indexing the very
+      # last step of the accessionWF.
+      sleep 1 if last_accession_step?(step)
+
+      # send message to RabbitMQ for accessioning complete or error occurred, pre-assembly consumes these to update UI
```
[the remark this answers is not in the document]
peetucket: updated code comment

PR #5855, opened 2026-03-06
Catch other errors with migrate script

### LINKED ISSUE #5851: better error handling in cocina migrator
unexpected individual object migration failures should be caught and logged, instead of blowing up the script with a stack trace. that could be a `rescue` after the one in `Migrators::MigrationRunner.migrate_repository_object` (but also fine if it's a `rescue` in the `Migrators::MigrationRunner.migrate_druid_list` ActiveRecord query, or whatever the implementer chooses): https://github.com/sul-dlss/dor-services-app/blob/f08d2e4294fc1f5170f94f9c0b66bfeba13152f6/app/services/migrators/migration_runner.rb#L73-L75

lwrubel:
Fixes #5851. 

sdr-infra

### DIFF
--- a/app/services/migrators/migration_runner.rb
+++ b/app/services/migrators/migration_runner.rb
@@ -69,12 +69,8 @@ module Migrators
       end
       Rails.logger.info("#{obj.external_identifier} successfully migrated")
       { obj:, status: 'SUCCESS' }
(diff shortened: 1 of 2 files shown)


lwrubel:
I also added a test, so maybe you could look at this again, @mjgiarlo?

mjgiarlo (approved):
🎉

PR #5898, opened 2026-03-16
Return invalid cocina for user versions

### LINKED ISSUE #5850: Return invalid JSON when requesting a user version that has invalid cocina.
Similar to how handled in https://github.com/sul-dlss/dor-services-app/pull/5738, we need to return the invalid JSON when requesting a user version that doesn't validate. This requires an accompanying DSC method. 

Argo will need to be able to use this to support https://github.com/sul-dlss/argo/issues/5022.

mjgiarlo:
# Why was this change made?

Fixes #5850

Follows the same approach used for system versions.

# How was this change tested?

CI

### DIFF
--- a/app/controllers/user_versions_controller.rb
+++ b/app/controllers/user_versions_controller.rb
@@ -4,6 +4,7 @@
 class UserVersionsController < ApplicationController
   before_action :find_repository_object, except: %i[create]
   before_action :find_user_version, only: %i[show update solr]
+  before_action :validate_from_openapi, except: %i[show solr]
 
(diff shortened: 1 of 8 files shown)


app/controllers/user_versions_controller.rb
```diff
@@ -4,6 +4,7 @@
 class UserVersionsController < ApplicationController
   before_action :find_repository_object, except: %i[create]
   before_action :find_user_version, only: %i[show update solr]
+  before_action :validate_from_openapi, except: %i[show solr]
```
mjgiarlo: This is what we have in the versions controller.

PR #5904, opened 2026-03-19
Add a general report for getting details about a druid list

jcoyne:
```
./bin/find-item-details.rb -f bad_druids.csv > info_about_bad_druids.csv
```

So we can get details about a list of druids. Many times it is faster to do calculations on an extracted set of data.  Once the set is winnowed down, we need to know details about the members of the set.  This script gets those details.

run on sdr-infra


jcoyne:
Add readme docs too.

PR #5909, opened 2026-03-19
Change version validation in migration and expand RemoveParallelContributor 

lwrubel:
Checks validity of each version when running migrator. HOWEVER, this existing logic will not migrate an object once ANY version is found to be invalid. 

Also updates the RemoveParallelContributor migrator to cover all existing locations of `parallelContributor`, based on a dryrun migration on prod. 


lwrubel:
Closing since we have completed this migration with a refactored migration runner.

PR #5937, opened 2026-04-01
Fix bug in release tags controller that always returns public tags

mjgiarlo:
# Why was this change made?

Fixes sul-dlss/argo#5094

# How was this change tested?

CI, QA

### DIFF
--- a/spec/requests/release_tags_spec.rb
+++ b/spec/requests/release_tags_spec.rb
@@ -3,6 +3,17 @@
 require 'rails_helper'
 
 RSpec.describe 'Operations on release tags' do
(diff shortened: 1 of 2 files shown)


spec/requests/release_tags_spec.rb
```diff
@@ -3,6 +3,17 @@
 require 'rails_helper'
 
 RSpec.describe 'Operations on release tags' do
+  let(:auth_headers) { { 'Authorization' => "Bearer #{jwt}" } }
```
mjgiarlo: More RSpec reorganization.

PR #5938, opened 2026-04-02
Refactor migration.

### LINKED ISSUE #5934: Separate RepositoryObject migration from Cocina migration
Currently, the MigrationRunner handles both RepositoryObject migration (updates multiple versions, does not involve versioning, persisted by saving RepositoryObject) and Cocina migration (updates head version, involves versioning, persisted by UpdateObjectService.update). This has made the logic murky, especially around updating multiple versions.

For this ticket, Cocina migration should be separated from RepositoryObject migration.

Notes for Cocina migration:
* migrate? is required because will determine whether versioning actions should be performed.

Notes for RepositoryObject migration:
* Migrator.must_be_valid_after_migrate = indicates if every version that is currently_valid must be valid after migrate
* For a RepositoryObject, for each repository_object_version:
  1. Skip if no cocina
  2. Determine if the repository_object_version is currently valid.
  3. Invoke migrator.migrate(repository_object_version:, is_head_version:, is_last_closed_version:, is_opened_version:, currently_valid:)
  4. If last closed version or opened version or (currently valid and must be valid after migrate) fail if not valid. Make sure to report which version.
  5. Save RepositoryObject in transaction.

justinlittman:
closes #5934

Existing migration is muddled about how to persist, especially when multiple versions are migrated.

Unit, QA


app/services/migrators/base.rb
```diff
@@ -19,62 +19,75 @@ module Migrators
   #   version_description - description for the version
   # See app/services/migrators/exemplar.rb for an example.
   class Base
-    # Return an array of druids to be migrated, or nil if all druids should be migrated.
-    def self.druids
```
justinlittman: Druids can no longer be specified in the migrator; they are provided when invoking the CLI.

app/services/migrators/base.rb
```diff
@@ -19,62 +19,75 @@ module Migrators
...
-    # association, so that if the parent repository_object is saved by the migration runner framework (which
-    # invokes this #migrate method), the updated version(s) touched by the migrator will also be saved; but the
-    # #migrate method SHOULD NOT explicitly save any ActiveRecord objects, so that dry run and validation behavior
-    # can be coordinated by the migration runner.
-    # @see Migrators::Exemplar#head_rov
-    # @todo A future enhancement could be to have the constructor take an optional mode, so that future migrator classes
-    # can look at the mode flag and explicitly save as needed, to remove the current autosave requirement.
-    def migrate
+    # @return [String] the migration strategy to use for this migrator. See STRATEGIES for valid strategies.
+    def self.migration_strategy
```
aaron-collier: Not a blocker, but I wonder if we should validate `migration_strategy` against `STRATEGIES` in case a migrator is added with in invalid strategy. I may also just be missing it in the larger diff if this works (perhaps through a raise).

aaron-collier (approved):
Approved with a couple of minor questions for here or possible follow up.

PR #5950, opened 2026-04-10
Adds ability to load a cocina object without validation. Skip validation when indexing.

justinlittman:
Validation is very slow, especially for some objects. In some read-only contexts, validation can be skipped for increased performance.

Unit


jcoyne (approved):
👍 Very good idea. I was wondering if we should make validation opt-in, rather than opt out since it's so expensive.

PR #5980, opened 2026-05-05
Remove trailing punctuation from contributor role.

### LINKED ISSUE #5955: MARC mapping: remove trailing punctuation from role code
When parsing `..contributor.role.code`, drop any trailing punctuation before checking the code against the list of valid codes.

MARC fields:
100/110/700/710 $e and $4
111/711 $j and $4

Example of record with punctuation: druid:vh974xq8713/a1985876

justinlittman:
closes #5955

Thanks copilot.

Per @arcadiafalcone 

Unit

### DIFF
--- a/app/services/cocina/from_marc/contributor.rb
+++ b/app/services/cocina/from_marc/contributor.rb
@@ -71,10 +71,16 @@ module Cocina
 
       def build_roles(field, code: 'e') # rubocop:disable Metrics/AbcSize
         expanded = field.subfields.select { |sf| sf.code == '4' }
-                                  .map { |role| { value: MARC_RELATORS[role.value] }.compact_blank }
+                                  .map do |role|
+                                    { value: MARC_RELATORS[normalize_role_code(role.value)] }.compact_blank
+                                  end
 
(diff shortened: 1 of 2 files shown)


lwrubel:
We have some "discontinued" roles in the `MARC_RELATORS` values that end in punctuation. (https://github.com/sul-dlss/dor-services-app/blob/c7f4251e3a33127d4e53c8d3fec797e12a76f9d3/app/services/cocina/from_marc/marc_relators.rb#L13). I'm not sure what "discontinued" means-- I guess it's no longer used in MARC records? If we did get any from a MARC record, the mapping would fail. But maybe that error would be useful to us in that case, to be aware of the old data.

lwrubel (approved):

justinlittman:
@arcadiafalcone ?

justinlittman:
I was wondering the same thing, but unsaved changes doesn't apply to these buttons.

arcadiafalcone:
I think that's fine for now - we can revisit if there are a lot of discontinued codes failing mapping.

PR #6005, opened 2026-05-18
Prepare for next cocina-models release

mjgiarlo:
Connects to sul-dlss/cocina-models#1027

### DIFF
--- a/spec/requests/indexable_spec.rb
+++ b/spec/requests/indexable_spec.rb
@@ -12,7 +12,6 @@ RSpec.describe 'Indexable' do
   let(:apo_druid) { apo.external_identifier }
   let(:label) { 'This is my label' }
   let(:title) { 'This is my title' }
-  let(:who) { 'test_user' }
 
   let(:view) { 'world' }
   let(:download) { 'world' }
@@ -76,7 +75,6 @@ RSpec.describe 'Indexable' do
   let(:data) do
     <<~JSON
       {
-        "user_name": "#{who}",
         "cocinaVersion": "#{Cocina::Models::VERSION}",
         "externalIdentifier": "#{druid}",
         "type":"#{content_type}",
--- a/spec/requests/update_metadata_spec.rb
+++ b/spec/requests/update_metadata_spec.rb
@@ -51,7 +51,6 @@ RSpec.describe 'Update object' do
   let(:data) do
     <<~JSON
       {
-        "user_name": "#{who}",
         "cocinaVersion": "#{Cocina::Models::VERSION}",
         "externalIdentifier": "#{druid}",
         "type":"#{content_type}",
@@ -108,9 +107,10 @@ RSpec.describe 'Update object' do
       end
 
       it 'updates the object' do
-        patch("/v1/objects/#{druid}",
+        patch("/v1/objects/#{druid}?user_name=#{who}",
               params: data,
               headers:)
+
         expect(response).to have_http_status(:ok)
         expect(response.body).to equal_cocina_model(Cocina::Models.build(JSON.parse(data).except('user_name')))
         expect(response.headers['Last-Modified']).to end_with 'GMT'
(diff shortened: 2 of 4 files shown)


spec/requests/show_user_version_solr_spec.rb
```diff
@@ -31,11 +31,13 @@
     end
 
     context 'when validate param is set to true (default value)' do
-      it 'raises a Cocina validation error' do
-        expect do
-          get "/v1/objects/#{object.external_identifier}/user_versions/1/solr?validate=true",
-              headers: { 'Authorization' => "Bearer #{jwt}" }
-        end.to raise_error(Cocina::Models::ValidationError)
+      it 'returns a 200' do
```
justinlittman: Maybe I'm misunderstanding, but doesn't this defeat the purpose of `validate=true1`?

spec/requests/show_version_solr_spec.rb
```diff
@@ -30,11 +30,13 @@
     end
 
     context 'when validate param is set to true (default value)' do
-      it 'raises a Cocina validation error' do
-        expect do
-          get "/v1/objects/#{object.external_identifier}/versions/1/solr?validate=true",
-              headers: { 'Authorization' => "Bearer #{jwt}" }
-        end.to raise_error(Cocina::Models::ValidationError)
+      it 'returns a 200' do
```
justinlittman: Same.

spec/requests/update_metadata_spec.rb
```diff
@@ -13,7 +13,6 @@
   let(:modified) { DateTime.now }
   let(:label) { 'This is my label' }
   let(:title) { 'This is my title' }
-  let(:who) { 'test_user' }
```
justinlittman, line 16: Why this change?

spec/requests/show_user_version_solr_spec.rb
mjgiarlo: Yes, it does. We intentionally ignore `validate=true` only in the Solr endpoints so that we can always render the Argo show page. It is set up like that right now, too.

spec/requests/update_metadata_spec.rb
mjgiarlo: Cruft from earlier work, good catch. I'll add this back.

spec/requests/show_user_version_solr_spec.rb
justinlittman: I'm a bit confused. If we ignore the parameter, why is it part of the API?

mjgiarlo: @justinlittman patched.

justinlittman (commented):
Can you say more about what is going on in the changes to the specs?

spec/requests/indexable_spec.rb
```diff
@@ -12,7 +12,6 @@
   let(:apo_druid) { apo.external_identifier }
   let(:label) { 'This is my label' }
   let(:title) { 'This is my title' }
-  let(:who) { 'test_user' }
```
mjgiarlo, line 15: This is never tested below, so take it out (noise).

spec/requests/indexable_spec.rb
```diff
@@ -72,11 +71,10 @@
       }
     }
   end
-  let(:content_type) { Cocina::Models::ObjectType.book }
+  let(:content_type) { Cocina::Models::ObjectType.object }
```
justinlittman: Why the switch from book to object? (Curious)

mjgiarlo: I was seeing HTTP 400 errors with book. I'll look into why.

mjgiarlo: @justinlittman I figured out what had been going wrong and fixed things up.

spec/requests/update_metadata_spec.rb
```diff
@@ -51,7 +51,6 @@
   let(:data) do
     <<~JSON
       {
-        "user_name": "#{who}",
```
mjgiarlo, line 54: This param does not go into the body, because it's not part of the Cocina JSON. It should be a query parameter.

justinlittman (approved):

spec/requests/update_metadata_spec.rb
mjgiarlo: In practice, as far as I know, ☝🏻 is what's already happening in production code. I believe including the username in the HTTP body was only done in the test here, which now triggers an openapi exception because `user_name` is an unevaluated property.

PR #6032, opened 2026-05-26
Update the definition for assembling_druids

thatbudakguy:
PR #5695 added the new gisDerivativeWF, but didn't update the
definition of assembling_druids to take it into account.

### DIFF
--- a/app/services/workflow/state_batch_service.rb
+++ b/app/services/workflow/state_batch_service.rb
@@ -47,7 +47,7 @@ module Workflow
         .where.not(workflow: 'assemblyWF', process: 'accessioning-initiate')
         .where.not(workflow: 'wasCrawlPreassemblyWF', process: 'end-was-crawl-preassembly')
         .where.not(workflow: 'wasSeedPreassemblyWF', process: 'end-was-seed-preassembly')
-        .where.not(workflow: 'gisAssemblyWF', process: 'start-accession-workflow')
+        .where.not(workflow: 'gisAssemblyWF', process: 'start-derivative-workflow')
         .where.not(workflow: 'ocrWF', process: 'end-ocr')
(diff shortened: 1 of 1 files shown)


app/services/workflow/state_batch_service.rb
```diff
@@ -47,7 +47,7 @@ def assembling_druids
         .where.not(workflow: 'assemblyWF', process: 'accessioning-initiate')
         .where.not(workflow: 'wasCrawlPreassemblyWF', process: 'end-was-crawl-preassembly')
         .where.not(workflow: 'wasSeedPreassemblyWF', process: 'end-was-seed-preassembly')
-        .where.not(workflow: 'gisAssemblyWF', process: 'start-accession-workflow')
```
peetucket, line 50: i think instead of deleting this line, we may want to leave it, but replace 'gisAssemblyWF' with 'gisDerivativeWF', so: `.where.not(workflow: 'gisDerivativeWF', process: 'start-accession-workflow')`  

like, do we want to consider that step in either workflow as potentially in the middle of "assembling"?

thatbudakguy: I don't 100% understand what this method is supposed to return, but it seemed like it was "steps in a workflow that aren't really work, but are actually just invoking a subsequent workflow".

If that's true, then both `start-derivative-workflow` in gisAssemblyWF and `start-accession-workflow` in gisDerivativeWF would qualify, I think

thatbudakguy: but if it's just "the step in a workflow that invokes accessionWF" then it would be only the latter, as you suggest

peetucket: actually, i think my comment is unwarranted, let's just go with your change for now - i think it's sufficient

peetucket (approved):

PR #6100, opened 2026-06-11
skip releaseWF based on a workflow context value being set

### LINKED ISSUE #6096: Add support for a skip release context flag
As discussed in #6063, in order to support skipping the releaseWF with a mass FOLIO refresh: 

1. Add support for setting context via DSA close version endpoint. (This endpoint will start an accessionWF.)
2. Add support for setting a skip release context flag to the MigrationRunner.
3. Have ReleaseInitiate robot skip initiating the releaseWF based on the presence of the skip release context flag. If the `release-initiate` is skipped, add a note that the step was skipped as requested. @andrewjbtw if there is specific wording you'd like to see here, just add to the ticket.

peetucket:

Fixes #6096 

1. initiate release step will do nothing if workflow context is present telling it to be skipped (thus, no releaseWF)
2. migrators can now set workflow context to pass to the accessioning process
3. version closing service now accepts context as a parameter and hands it off to the workflow service to be set

New specs

### DIFF
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -108,20 +108,32 @@ class VersionsController < ApplicationController
     boolean_param(new_params, :assume_accessioned)
   end
 
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
   def close_params
     new_params = params.permit(
       :description,
       :start_accession,
       :user_name,
       :user_versions,
-      :'lane-id'
+      :'lane-id',
+      context: {}
     ).to_h.symbolize_keys
     new_params[:user_version_mode] = new_params.delete(:user_versions).to_sym if new_params.key?(:user_versions)
+
     if new_params.key?(:'lane-id')
       new_params[:accession_args] = { lane_id: LaneSupport.lane_for(new_params.delete(:'lane-id')) }
     end
+
+    if new_params.key?(:context)
+      new_params[:accession_args] ||= {}
+      new_params[:accession_args][:context] = new_params.delete(:context)
+    end
+
     boolean_param(new_params, :start_accession)
   end
+  # rubocop:enable Metrics/MethodLength
+  # rubocop:enable Metrics/AbcSize
 
   def load_version
     @version = CocinaObjectStore.version(params[:object_id])
--- a/app/controllers/versions_controller.rb
+++ b/app/controllers/versions_controller.rb
@@ -108,9 +108,7 @@ class VersionsController < ApplicationController
     boolean_param(new_params, :assume_accessioned)
   end
 
-  # rubocop:disable Metrics/MethodLength
-  # rubocop:disable Metrics/AbcSize
-  def close_params
+  def close_params # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
     new_params = params.permit(
       :description,
       :start_accession,
@@ -132,8 +130,6 @@ class VersionsController < ApplicationController
 
     boolean_param(new_params, :start_accession)
   end
-  # rubocop:enable Metrics/MethodLength
-  # rubocop:enable Metrics/AbcSize
 
   def load_version
     @version = CocinaObjectStore.version(params[:object_id])
--- a/app/jobs/robots/dor_repo/accession/release_initiate.rb
+++ b/app/jobs/robots/dor_repo/accession/release_initiate.rb
@@ -15,8 +15,19 @@ module Robots
                                               note: 'Admin policy objects are not released')
           end
 
+          if skipped?
+            return LyberCore::ReturnState.new(status: :skipped,
+                                              note: 'releaseWF was skipped because workflow context indicated this')
+          end
+
           Workflow::Service.create(druid:, workflow_name: 'releaseWF',
                                    version: cocina_object.version)
+        end
+
+        def skipped?
+          # checks if user has indicated that releaseWF should be skipped (sent as workflow_context)
+          # default to false if not present
+          workflow.context['skipReleaseWF'] || false
         end
       end
     end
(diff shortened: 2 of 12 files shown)


mjgiarlo (approved):

app/controllers/versions_controller.rb
```diff
@@ -108,20 +108,32 @@ def create_params
...
+
+    if new_params.key?(:context)
+      new_params[:accession_args] ||= {}
+      new_params[:accession_args][:context] = new_params.delete(:context)
+    end
+
     boolean_param(new_params, :start_accession)
   end
+  # rubocop:enable Metrics/MethodLength
+  # rubocop:enable Metrics/AbcSize
```
peetucket: we could refactor this method to get rid of these complaints, but I'm not sure it'll be an improvement

app/jobs/robots/dor_repo/accession/release_initiate.rb
```diff
@@ -15,9 +15,20 @@ def perform_work
...
+
           Workflow::Service.create(druid:, workflow_name: 'releaseWF',
                                    version: cocina_object.version)
         end
+
+        def skipped?
+          # checks if user has indicated that releaseWF should be skipped (sent as workflow_context)
+          # default to false if not present
+          ActiveModel::Type::Boolean.new.cast(workflow.context['skipReleaseWF'])
+        end
```
peetucket, line 31: workflow context can come as a string, let's be sure "true" comes out to `true`

blank/nil values will be cast to false, leaving it as the default when there is no workflow context

app/controllers/versions_controller.rb
```diff
@@ -108,20 +108,32 @@ def create_params
...
+
     if new_params.key?(:'lane-id')
       new_params[:accession_args] = { lane_id: LaneSupport.lane_for(new_params.delete(:'lane-id')) }
     end
+
+    if new_params.key?(:context)
+      new_params[:accession_args] ||= {}
+      new_params[:accession_args][:context] = new_params.delete(:context)
+    end
+
```
peetucket, line 130: this ensures we pass the "context" into the accession params when it is passed to the close method

app/jobs/robots/dor_repo/accession/release_initiate.rb
```diff
@@ -15,9 +15,20 @@ def perform_work
                                               note: 'Admin policy objects are not released')
           end
 
+          if skipped?
+            return LyberCore::ReturnState.new(status: :skipped,
+                                              note: 'releaseWF was skipped because workflow context indicated this')
+          end
```
peetucket, line 21: skip this step (which skips the creation of releaseWF) with a note if the workflow context says we should

app/services/migrators/migration_runner.rb
```diff
@@ -217,8 +217,12 @@ def migrated_model_hash_for_update(migrated_model_hash)
...
       VersionService.close(druid:,
                            version: repository_object.head_version.version,
-                           accession_args: { lane_id: 'low' }, # Always send to low queue.
+                           accession_args: { lane_id: 'low', context: }, # Always send to low queue.
                            description: nil) # Use the existing version description
     end
+
+    def context
+      migrator_class.workflow_context || {}
+    end
```
peetucket, line 226: pulls in the context that may have been set in the migrator class and then passes it on to the object close service method

app/controllers/versions_controller.rb
```diff
@@ -108,20 +108,32 @@ def create_params
     boolean_param(new_params, :assume_accessioned)
   end
 
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
   def close_params
```
mjgiarlo: If you make the following tweak:

```suggestion
  def close_params # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```

You can dump lines 111, 112, 135, and 136. (Rubocop will apply the exclusions only to the method block.)

openapi.yml
```diff
@@ -900,6 +900,15 @@ paths:
           in: query
           schema:
             $ref: "#/components/schemas/LaneId"
+        - name: context
+          in: query
+          description: Optional workflow context to pass to accessionWF.
+          required: false
+          style: deepObject
+          explode: true
```
mjgiarlo: Interesting. What do these two parameters do? They're new to me and I don't see them used elsewhere in the spec.

mjgiarlo (approved):
LGTM. Non-blocking comments/questions only.

app/controllers/versions_controller.rb
peetucket: will adjust

openapi.yml
peetucket: I had the same question, since Claude made the openAPI spec changes when I was coding with it. Here was it's explanation when I asked about the changes:


Those OpenAPI directives tell API clients how to encode the nested `context` object as query parameters:

- **`style: deepObject`** – Serializes the object using bracket notation rather than as a flat JSON string.
- **`explode: true`** – Expands nested properties into separate query parameter key-value pairs.

Together they mean a client sending:
```json
{ "skipReleaseWF": true, "ocrLanguages": ["Russian"] }
```

Should encode it as query string:
```
?context[skipReleaseWF]=true&context[ocrLanguages][]=Russian
```

Instead of (without deepObject):
```
?context={"skipReleaseWF":true,"ocrLanguages":["Russian"]}
```

The OpenAPI spec is documenting to clients how to pass nested context objects via query parameters using this bracket notation.

peetucket: Having said that, I do not recall doing this in my previous work on workflow context, though that may be because I never had to modify an openapi spec when doing that.

mjgiarlo: Huh, neat. If it works, great! How about dropping a comment here that explains ☝🏻 ❓

peetucket: comments added

mjgiarlo: :shipit:

peetucket:
integration tests passed, so no regressions.  merging

Issue #6150, opened 2026-06-19
Rerun report on form source codes


arcadiafalcone:
Rerun the report in #6009 to take account of remediation so far.

peetucket:
Report re-run.  Here: https://drive.google.com/file/d/1Zx0NIiQxkXrukvKK0oPznG_V7hPtpk-m/view
Has 11589 rows.

arcadiafalcone:
These have been remediated with errors. Please add `marccarrier` and `lcmpt` to the list of valid codes and rerun.

justinlittman:
[invalid_form_source_codes.csv](https://github.com/user-attachments/files/29258865/invalid_form_source_codes.csv)

arcadiafalcone:
These have been remediated. Please rerun the report to confirm.

peetucket:
Updated report attached.

Now has 152 rows, the same invalid code `marcmuscomp`, all appear to be the same APO.

[invalid_form_source_codes.csv](https://github.com/user-attachments/files/29355000/invalid_form_source_codes.csv)

arcadiafalcone:
My bad, I missed a step. They should be fixed now.

peetucket:
Re-ran report - all clear (no rows left).

PR #6154, opened 2026-06-23
Add 'marccarrier' and 'lcmpt' to valid codes for invalid form source …

justinlittman:
…codes report

Per https://github.com/sul-dlss/dor-services-app/issues/6150


mjgiarlo (approved):
Worth retaining the alpha sort order or nah?

Issue #6135, opened 2026-06-18
Allow creating an object from FOLIO instance record instead of MARC source storage record


lwrubel:
Either HOLD the PR for this until the accompanying Argo front-end work (to be ticketed) is completed or add a feature flag. 

Here's one approach to creating Cocina from an instance record, but feel free to come up with a different, better way:

- [x] Add a param to [RefreshDescription.run](https://github.com/sul-dlss/dor-services-app/blob/e93f538e410bb21adb657304202e7cd8aa7ec1d2/app/services/refresh_description_from_catalog.rb#L13) for `use_marc_only`, defaulting to `true`. The default is set to `true` so that when the refresh description button is used in Argo or in migrators, refreshing will fail if no MARC record exists (existing behavior). This param should be set to `false` when called from [CreateObjectServce#sync_from_catalog](https://github.com/sul-dlss/dor-services-app/blob/1aaa41dd3eeef81302f0f0e9b8c2df1762bdda42/app/services/create_object_service.rb#L94) to allow using a fallback instance record. 
- [x] The `use_marc_only` value would need to get passed to `Catalog::MarcService.marc` so that `FolioReader`, which [currently gets the MARC from FOLIO](https://github.com/sul-dlss/dor-services-app/blob/707386a9e2bd45c5c237af7709e566c701a2cb88/app/services/catalog/folio_reader.rb#L22) could instead get an instance record.
- [x] Get an instance record using `FolioClient.fetch_instance_info`. This could be a new service?
- [x] Use the existing SW mapping from FOLIO instance to MARC to create "synthetic" MARC. https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/folio/marc_record_instance_mapper.rb#L14-L77. Note we don't need the holdings record or an 856 here. FolioReader can still apply all its other MARC record modifications before returning the record.

PR #6157, opened 2026-06-24
Refactor MARC fetch logic and error handling for catalog

mjgiarlo:
# Why was this change made?

Connects to #6135

This commit sets up work to create a fallback mechanism that will generate MARC from instance records. It does so by adding a new seam (the create_marc_if_missing arg) and getting some preliminary refactoring out of the way given that fetching MARC will soon be a little more complicated.

The prior design was:
* `Catalog::MarcService` - this class did nothing more than hold some error classes and wrap `Catalog::FolioReader`, which was only called in this one place. That suggested to me that it could be worth looking into pulling some of the Folio reader logic into the MARC service to give it more responsibility.
* `Catalog::FolioReader` - this class had arguably too many responsibilities: optionally fetching the HRID given a barcode; fetching the MARC hash to normalize abstracts;  building the new MARC record (w/ changes to 001 and 003); and updating the MARC cache.

The new design tries to balance the responsibilities:
* `Catalog::MarcService` - this class optionally fetches the HRID given a barcode and orchestrates creating the MARC record hash by collaborating with two new classes, to be described next; it then updates the MARC cache
* `Catalog::SourceStorageFetcher` - this class's job is to fetch the MARC hash from Folio source storage and normalize abstracts
* `Catalog::ControlFieldsTransformer` - this class's job is to build the new MARC record w/ changes to the 001 and 003 fields
* `Catalog::Errors` - a new module to hold the various catalog errors

## Includes

- Remove FolioReader in favor of SourceStorageFetcher and ControlFieldsTransformer
- Introduce Catalog::Errors for unified error handling across services
- Update MarcService to use new fetcher/transformer and error classes
- Refactor controllers and services to rescue new error types
- Update specs and initializers to use new MARC fetch flow and error classes
- Ensure barcode/HRID lookups pass create_marc_if_missing param explicitly
- Add and test SourceStorageFetcher for MARC normalization and error propagation
- Improve documentation and param handling for service classes and methods
- Remove legacy error classes and update all references to new error structure

# How was this change tested?

CI


app/services/catalog/control_fields_transformer.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Catalog
+  # Fetch a MARC record from Folio's JSON API to fetch marc json given a HRID
```
justinlittman: This description doesn't seem right.

app/services/catalog/marc_service.rb
```diff
@@ -3,45 +3,52 @@
...
     def marc
+      marc_record ||= begin
+        marc_hash = SourceStorageFetcher.fetch(folio_instance_hrid:)
+        ControlFieldsTransformer.transform(marc_hash:, folio_instance_hrid:)
+      rescue StandardError
+        raise Errors::RecordNotFoundError,
+              "Catalog record not found. HRID: #{folio_instance_hrid} | Barcode: #{barcode}"
+        # raise e unless create_marc_if_missing
+
+        # marc_record_from_catalog_instance_record
```
lwrubel, line 31: These are placeholders for the next steps in the work?

app/services/catalog/marc_service.rb
```diff
@@ -3,45 +3,52 @@
...
       @barcode = barcode
-      @folio_instance_hrid = folio_instance_hrid
-    end
-
-    # @return [MARC::Record] MARC record
-    # @raise CatalogResponseError
-    # @raise CatalogRecordNotFoundError
-    def marc_record
-      @marc_record ||= marc_record_from_folio
+      @folio_instance_hrid = folio_instance_hrid || FolioClient.fetch_hrid(barcode:)
```
justinlittman, line 16: What happens if it can't find by barcode?

app/services/catalog/marc_service.rb
```diff
@@ -3,45 +3,52 @@
...
-    # @raise CatalogRecordNotFoundError
+    # @raise [Error]
     def marc
+      marc_record ||= begin
+        marc_hash = SourceStorageFetcher.fetch(folio_instance_hrid:)
+        ControlFieldsTransformer.transform(marc_hash:, folio_instance_hrid:)
+      rescue StandardError
+        raise Errors::RecordNotFoundError,
+              "Catalog record not found. HRID: #{folio_instance_hrid} | Barcode: #{barcode}"
+        # raise e unless create_marc_if_missing
```
justinlittman, line 29: Cruft?

app/services/catalog/marc_service.rb
```diff
@@ -3,45 +3,52 @@
...
 
     # @return [Hash] MARC record as a hash
-    # @raise CatalogResponseError
-    # @raise CatalogRecordNotFoundError
+    # @raise [Error]
     def marc
+      marc_record ||= begin
+        marc_hash = SourceStorageFetcher.fetch(folio_instance_hrid:)
+        ControlFieldsTransformer.transform(marc_hash:, folio_instance_hrid:)
+      rescue StandardError
```
justinlittman: Why does StandardError -> RecordNotFoundError?

lwrubel:
Thanks, refactoring this makes a lot of sense. The responsibilities of MarcService vs FolioReader were indeed unclear.

lwrubel (approved):
Recommend integration testing with [spec/features/accessioning/item_creation_with_folio_hrid_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test/blob/64581d967ed1a370806df01597bbe61977032c63/spec/features/accessioning/item_creation_with_folio_hrid_spec.rb#L8) just to be safe.

app/services/catalog/marc_service.rb
```diff
@@ -3,45 +3,52 @@
...
-      FolioReader.to_marc(folio_instance_hrid:, barcode:)
-    rescue FolioClient::ResourceNotFound
-      raise CatalogRecordNotFoundError, "Catalog record not found. HRID: #{folio_instance_hrid} | Barcode: #{barcode}"
-    rescue FolioClient::Error => e
-      raise CatalogResponseError, "Error getting record from catalog: #{e.message}"
+    attr_reader :barcode, :create_marc_if_missing, :folio_instance_hrid
+
+    def update_marc_cache!(marc_record:)
+      # Cache the MARC, so that we can use it for creating the Argo index without having to fetch it again.
+      MarcCacheEntry.upsert( # rubocop:disable Rails/SkipsModelValidations
```
justinlittman, line 47: Why do we need this? And by this, I mean the MarcCache.

app/services/catalog/source_storage_fetcher.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Catalog
+  # Fetch a MARC record from Folio's JSON API to fetch marc json given a HRID
+  class SourceStorageFetcher
```
justinlittman, line 5: Can this be called by MarcHashFetcher? I don't know what source storage is.

app/services/catalog/source_storage_fetcher.rb
```diff
@@ -0,0 +1,48 @@
...
+    # @param folio_instance_hrid [String] Folio instance HRID
+    def initialize(folio_instance_hrid:)
+      @folio_instance_hrid = folio_instance_hrid
+    end
+
+    # @return [Hash] MARC record as a hash
+    # @raise [Catalog::MarcService::Error]
+    def fetch # rubocop:disable Metrics/AbcSize
+      FolioClient.fetch_marc_hash(instance_hrid: folio_instance_hrid).tap do |record_hash|
+        # Only normalize if abstracts present
```
justinlittman, line 20: What is the difference between this normalization and the normalization done by the ControlFieldsTransformer? (That is, why is normalization happening in 2 places?)

app/services/refresh_description_from_catalog.rb
```diff
@@ -6,23 +6,27 @@ class RefreshDescriptionFromCatalog
...
+    new(...).run
   end
 
-  def initialize(cocina_object:, druid:, use_barcode:)
+  # @param cocina_object [Cocina::Models::DRO, Cocina::Models::Collection, Cocina::Models::APO, Cocina::Models::Agreement,
+  #   Cocina::Models::RequestDRO, Cocina::Models::RequestCollection, Cocina::Models::RequestAPO,
+  #   Cocina::Models::RequestAgreement] cocina object to refresh
+  # @param druid [String] the druid identifier
+  # @param use_barcode [Boolean] whether to use barcode as an identifier (default: false)
+  # @param create_marc_if_missing [Boolean] whether to create a MARC record if missing in the catalog (default: false)
```
justinlittman, line 19: Can you say more about what it means to create a MARC record? In what context is it created?

app/services/catalog/marc_service.rb
mjgiarlo: @lwrubel Yep!

app/services/catalog/source_storage_fetcher.rb
lwrubel: Changing the class name sounds good to me. I had suggested that "source storage" terminology so good to know it could be clearer. It also could help to change the comment to something like "Fetch a MARC record hash from FOLIO Source Storage, given an HRID". We're going to be fetching FOLIO instance records in the case a source storage MARC record is not available. That will use a different FOLIO API and why the distinction between source storage and instance is relevant.

app/services/catalog/control_fields_transformer.rb
mjgiarlo: Good catch.

app/services/catalog/marc_service.rb
mjgiarlo: Then the fetch below in `#marc` raises an exception which is re-raised with an error message saying the record could not be found with the user-provided HRID and barcode.

app/services/catalog/marc_service.rb
mjgiarlo: I reworked this to be less generic, now rescuing `FolioClient::ResourceNotFound, FolioClient::MultipleResourcesFound`

app/services/catalog/marc_service.rb
mjgiarlo: No, this is what's coming next. I left it in to show why I restructured things a bit.

app/services/catalog/marc_service.rb
mjgiarlo: @jcoyne added it a few months ago, and I didn't want to include major behavior changes in this refactoring.

app/services/catalog/source_storage_fetcher.rb
mjgiarlo: I've fixed up the comment above. How about we leave this class name unchanged for now and I will consider a rename when I add its sibling fetcher (from the instance record)?

app/services/catalog/source_storage_fetcher.rb
mjgiarlo: Nothing, really. How about I reconsider normalization in the follow-up PR?

app/services/refresh_description_from_catalog.rb
mjgiarlo: This will be clearer in the next PR.

app/services/catalog/source_storage_fetcher.rb
justinlittman: Works for me.

app/services/catalog/source_storage_fetcher.rb
justinlittman: Sure.

justinlittman (approved):

Issue #6156, opened 2026-06-23
"publish" step is not reliable for large amounts of data (2+ TB)


andrewjbtw:
The `publish` step has multiple problems handling large data deposits (where large is defined as 2+ TB):

- it takes a very long time, in part because all data is copied twice
    - data is copied to `/access-transfers` then copied again to `/stacks`
- it often fails: 4 out of the last 5 deposits of 2+ TB hit an error at publish
    - the length of time the step takes may make it more likely to fail
- when the publish step fails
    - retries do not pick up from where the step failed; instead, all data is recopied
    - data is left partially copied to the `/access-transfers` mount and must be deleted manually before retrying
      - this creates a risk that retries will fill up all disk space on the mount if someone fails to clear space in advance
    - a partially copied file may be left on Stacks which then needs to be deleted manually before a retry can succeed

We recently took in 5 self-deposits of 2+ TB each. Of the 5, 4 failed at least once in the publish step and I had to intervene to make sure the deposits succeeded. In two cases, I ended up copying the files to `/stacks` myself because each retry of publish took 10-12 hours each time to fail and I didn't want to spend days trying to get the deposits to work through the retry process.

**Walk through of a `publish` failure**
To the best of my knowledge, this is what has been happening when the step fails and I need to retry it:

1. User starts deposit
2. Files are copied to accessioning pipeline under `/dor/workspace`
3. DSA copies files to `/access-transfer`
4. purl-fetcher copies files from `/access-transfer` and starts copying them to `/stacks`
5. There's an error. Frustratingly, the main error I've seen is a generic 500 https://app.honeybadger.io/projects/50568/faults/126350160
6. Files that have not been copied to `/stacks` yet are left on `/access-transfer`: this could mean up to 2 TB are left on `access-transfer`
7. If a file being copied to `/stacks` was interrupted during the copying, a truncated copy of the file is left on `/stacks`
8. I ssh into the purl-fetcher server and look at the situation
9. If there are multiple TB in `/access-transfer`, I remove the files to avoid the risk of filling up `/access-transfer` on retry
10. I identify any truncated file on `/stacks` and delete it
11. I then rerun the `publish` step from Argo
12. The entire set of files is copied again to `/access-transfer` as if publishing never ran before
13. In 10-12 hours, I find out if the retry succeeded.

This is not a sustainable way of supporting large data deposits. We need to work out a more workable way of handling this step and its possible failures.

justinlittman:
[...]

PR #6159, opened 2026-06-24
Cleanup if publish/shelve fails.

justinlittman:
refs https://github.com/sul-dlss/dor-services-app/issues/6156

Avoid filling up the staging area.

Unit

### DIFF
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -58,14 +58,28 @@ module Publish
       # It's fine. The object is already deleted.
     end
 
-    def publish_shelve
+    def publish_shelve # rubocop:disable Metrics/AbcSize
       if filepaths_to_shelve.present?
         ShelvableFilesStager.stage(cocina_object:, filepaths: filepaths_to_shelve, workspace_content_pathname:)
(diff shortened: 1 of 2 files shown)


app/services/publish/metadata_transfer_service.rb
```diff
@@ -58,14 +58,28 @@ def publish_delete_on_success
...
         TransferStager.copy(druid:, filepath_map: filepath_uuid_map, workspace_content_pathname:)
       end
+
       PurlFetcher::Client::Publish.publish(cocina: public_cocina, file_uploads: filepath_uuid_map,
                                            version: public_version,
                                            must_version: must_version?, version_date:)
+    rescue StandardError
+      delete_staged_files
+      raise
+    end
```
mjgiarlo, line 73: I think this could also be:

```suggestion
    ensure
      delete_staged_files
    end
```

but no big diff

mjgiarlo (approved):

app/services/publish/metadata_transfer_service.rb
justinlittman: My change: delete_stages_files only if there is an error.
Your change: always delete_staged_files

Is that a correct reading? If so, then the intended behavior is my change.

mjgiarlo: Oof, no, you're right. Nevermind!

PR #6160, opened 2026-06-24
When indexing, get SW format from cocina display, not MARC.

justinlittman:
Reduced complexity.

Unit


mjgiarlo (approved):
LGTM. Ideally, @jcoyne could sign off on this

PR #6161, opened 2026-06-25
Refactor MigrationRunner for clarity and performance.

justinlittman:
I went to make some performance improvements in the migrator (reducing repetitive validations and building of cocina objects), but found the code handling both cocina updates and commits confusing. This separates them so that logic is clearer.

Unit

### DIFF
--- a/app/services/migrators/migration_runner.rb
+++ b/app/services/migrators/migration_runner.rb
@@ -2,7 +2,7 @@
 
 module Migrators
   # Performs migration on a RepositoryObject given a migrator class.
-  class MigrationRunner # rubocop:disable Metrics/ClassLength
+  class MigrationRunner
     # see README for explanation of modes
     MODES = %i[dryrun migrate].freeze
     DEFAULT_MODE = :dryrun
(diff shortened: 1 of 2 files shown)


app/services/migrators/migration_runner.rb
```diff
@@ -35,198 +35,284 @@ def initialize(migrator_class:, repository_object:, mode:)
...
-      @opened_version ||= repository_object.opened_version&.version
-    end
+        invalid_results = validate_migrated_model_hashes(migrated_model_hashes:, original_model_hash_valid_map:)
+        return invalid_results if invalid_results.present?
 
-    # @return [Integer, nil] the last closed version number if the object has been closed, nil otherwise
-    def last_closed_version
-      @last_closed_version ||= repository_object.last_closed_version&.version
-    end
+        was_already_open = open_version! if migrator_class.version? # Note that open_version! is dryrun-aware.
```
lwrubel: I'm getting a bit stuck on calling the result of `open_version!` `was_already_open`. If the version was already open, then `open_version!` returns `false`. If was already closed, then `open_version!` returns `true`. What I am I missing?

app/services/migrators/migration_runner.rb
```diff
@@ -35,198 +35,284 @@ def initialize(migrator_class:, repository_object:, mode:)
...
+        was_already_open = open_version!
 
-    def close_version!
-      VersionService.close(druid:,
-                           version: repository_object.head_version.version,
-                           accession_args: { lane_id: 'low', context: }, # Always send to low queue.
-                           description: nil) # Use the existing version description
-    end
+        cocina_update!(migrated_cocina_object:) unless dryrun?
+        close_version! if was_already_open # Not closing objects that were already open.
```
lwrubel: `close_version!` is dryrun-aware by way of `open_version!` being dryrun-aware?

app/services/migrators/migration_runner.rb
```diff
@@ -35,198 +35,284 @@ def initialize(migrator_class:, repository_object:, mode:)
...
-        # This allows for migration errors to occur before actually opening the version.
-        if migrator_class.version? && repository_object.closed?
-          new_version = repository_object.last_closed_version.version + 1
-          model_hash = to_model_hash(repository_object.last_closed_version, version: new_version)
-          original_model_hash_map[new_version] = model_hash.freeze
-          @last_closed_version = head_version
-          @opened_version = new_version
-          @head_version = new_version
+    def call
+      runner_class = migrator_class.cocina_update? ? CocinaUpdateRunner : CommitRunner
```
mjgiarlo, line 39: Naïve question, as I haven't mucked about w/ the new migrator design much: is it worth considering abandoning the `migration_strategy` abstraction entirely, and replacing it with direct specification of `runner_class` within each migrator class?

app/services/migrators/migration_runner.rb
```diff
@@ -35,198 +35,284 @@ def initialize(migrator_class:, repository_object:, mode:)
...
+
+    private
+
+    attr_reader :migrator_class, :repository_object, :mode
+
+    # Base class for migration runners.
+    class BaseRunner
+      def initialize(runner)
+        @runner = runner
+      end
```
mjgiarlo: This isn't complicated, but I still found myself wondering how surprised I was by instantiating a runner by passing it a runner.

app/services/migrators/migration_runner.rb
mjgiarlo: I'm a bit stuck on this too. `#open_version!` seems to be doing two or three different things all by itself: returning a boolean (predicate method), raising an error if not openable, and opening the version. I wonder if this is worth separating into twoish methods so each can do a smaller job (thinking about command-query separation) and be named in a way that is more intention-revealing.

app/services/migrators/migration_runner.rb
```diff
@@ -35,198 +35,284 @@ def initialize(migrator_class:, repository_object:, mode:)
...
+        @head_version = new_version
       end
-    rescue StandardError => e
-      Rails.logger.info("#{druid} failed to migrate#{' (dry run)' if dryrun?}: #{e.message} -- #{e.backtrace}")
-      [Result.new(status: 'ERROR', exception: e.message, **result_id_attrs)]
     end
 
-    private
+    # Runner that migrates a RepositoryObject by committing the migrated model hashes to the database.
+    # Note that this migrates all versions (though it is up to the migrator to decide which actually to change.)
```
mjgiarlo, line 147: 👏🏻

app/services/migrators/migration_runner.rb
justinlittman: `was_already_open` was misnamed. But the larger point is true -- this needs a refactor. Fixing.

app/services/migrators/migration_runner.rb
justinlittman: Yes, but this is going to be changed.

app/services/migrators/migration_runner.rb
justinlittman: You could take that approach, but not sure the juice is worth the squeeze.

app/services/migrators/migration_runner.rb
justinlittman: Gone.

mjgiarlo (approved):
👏🏻

PR #6165, opened 2026-06-25
Synthesize MARC record from Folio instance

### LINKED ISSUE #6135: Allow creating an object from FOLIO instance record instead of MARC source storage record

mjgiarlo:
# Why was this change made?

Fixes #6135

## Includes:

- Introduce `Catalog::InstanceMarcBuilder` to build MARC records from
  Folio instance hashes when no MARC is found in Folio
- Update `Catalog::MarcService` to use `InstanceMarcBuilder` as fallback
  only when `create_marc_if_missing` is true
- Change `CreateObjectService` to pass `create_marc_if_missing: true` when
  refreshing description from catalog
- Add comprehensive spec for `InstanceMarcBuilder` covering MARC field
  generation from instance data

## Also includes set up commit:

- Replace SourceStorageFetcher and ControlFieldsTransformer with
  AbstractNormalizer and ControlFieldNormalizer for improved separation of
  concerns and testability
- Move abstract field normalization logic to AbstractNormalizer, ensuring
  escaped dollar signs in 520 fields are replaced with literal '$'
- Move control field normalization to ControlFieldNormalizer, updating 001
  and 003 fields to match FOLIO instance HRID and source
- Update MarcService to use FolioClient directly and compose normalization
  steps, simplifying the MARC record retrieval pipeline
- Remove obsolete SourceStorageFetcher and ControlFieldsTransformer classes
  and their specs
- Update specs to reflect new normalization pipeline and dependency
  injection

# How was this change tested?

CI


app/services/catalog/instance_marc_builder.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+module Catalog
+  # TODO
```
lwrubel: cruft?

justinlittman: Cruft.

justinlittman (approved):
Approve pending green and consideration of comment.

app/services/catalog/marc_service.rb
```diff
@@ -18,17 +18,19 @@ def initialize(barcode: nil, folio_instance_hrid: nil, create_marc_if_missing: f
...
       rescue FolioClient::ResourceNotFound, FolioClient::MultipleResourcesFound
-        raise Errors::RecordNotFoundError,
-              "Catalog record not found for HRID '#{folio_instance_hrid}' or barcode '#{barcode}'"
-        # raise e unless create_marc_if_missing
+        error_message = "Catalog record not found for HRID '#{folio_instance_hrid}' or barcode '#{barcode}'"
+        raise Errors::RecordNotFoundError, error_message unless create_marc_if_missing
 
-        # marc_record_from_catalog_instance_record
+        FolioClient.fetch_instance_info(hrid: folio_instance_hrid)
+                   .then { |instance_hash| InstanceMarcBuilder.build(instance_hash:) }
```
lwrubel, line 33: If `create_marc_if_missing` is `true`, and the HRID is truly not found (e.g. a typo), what happens--it will still raise a `FolioClient::ResourceNotFound` error (which is what we would want I think)?

app/services/catalog/instance_marc_builder.rb
mjgiarlo: It was a placeholder for me to actually write a top-level class comment. Which I've now done thanks to your keen eye! Thanks.

mjgiarlo: Got it.

app/services/catalog/marc_service.rb
mjgiarlo: Good q. I'll write a test to check/confirm that.

mjgiarlo: Yes, you're exactly right. Line 32 will raise `FolioClient::ResourceNotFound`.

I'm thinking I won't add a test, since I'd be stubbing the raise behavior (to avoid making an HTTP API call) and then verifying the behavior I just put in place.

PR #6169, opened 2026-06-26
Update purl-fetcher client.

justinlittman:
refs #6156

So that publish retries.

### DIFF
--- a/app/services/publish/metadata_transfer_service.rb
+++ b/app/services/publish/metadata_transfer_service.rb
@@ -64,6 +64,7 @@ module Publish
         TransferStager.copy(druid:, filepath_map: filepath_uuid_map, workspace_content_pathname:)
       end
 
+      # This is going to retry several times if publish fails. This will avoid a re-copy.
       PurlFetcher::Client::Publish.publish(cocina: public_cocina, file_uploads: filepath_uuid_map,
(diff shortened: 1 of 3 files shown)


app/services/publish/metadata_transfer_service.rb
```diff
@@ -64,6 +64,7 @@ def publish_shelve # rubocop:disable Metrics/AbcSize
         TransferStager.copy(druid:, filepath_map: filepath_uuid_map, workspace_content_pathname:)
       end
 
+      # This is going to retry several times if publish fails. This will avoid a re-copy.
       PurlFetcher::Client::Publish.publish(cocina: public_cocina, file_uploads: filepath_uuid_map,
```
peetucket, line 68: i guess the retry is already implemented in the job triggered by this call?  so the comment is an FYI

peetucket (approved):
any value in deploying DSA with this change and then running integration tests?  or are you reasonably confident in no issues with the upgrade?

app/services/publish/metadata_transfer_service.rb
justinlittman: Exactly.

justinlittman:
Deployed to stage and ran some integration tests.

PR #6176, opened 2026-06-29
Dispatch reindex jobs more efficiently.

### LINKED ISSUE #6175: Avoid putting all druids in an array
to avoid memory spike.

https://github.com/sul-dlss/dor-services-app/blob/main/lib/tasks/indexer.rake#L6-L11

justinlittman:
closes #6175

To avoid loading all druids into a single array.

QA

### DIFF
--- a/lib/tasks/indexer.rake
+++ b/lib/tasks/indexer.rake
@@ -3,12 +3,9 @@
 namespace :indexer do
   desc 'Reindex all objects in jobs'
   task :reindex_jobs, %i[batch_size] => :environment do |_task, args|
-    druids = RepositoryObject.pluck(:external_identifier)
-
-    batches = druids.each_slice(args.batch_size&.to_i || 100)
-
-    batches.each do |batch|
-      BatchReindexJob.perform_later(batch)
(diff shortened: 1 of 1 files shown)


lib/tasks/indexer.rake
```diff
@@ -3,12 +3,9 @@
...
   desc 'Reindex all objects in jobs'
   task :reindex_jobs, %i[batch_size] => :environment do |_task, args|
-    druids = RepositoryObject.pluck(:external_identifier)
-
-    batches = druids.each_slice(args.batch_size&.to_i || 100)
-
-    batches.each do |batch|
-      BatchReindexJob.perform_later(batch)
+    RepositoryObject.find_in_batches(batch_size: args.batch_size&.to_i || 100) do |repository_objects|
+      druids = repository_objects.pluck(:external_identifier)
```
peetucket: does this only select the `external_identifier` column when querying each object?  maybe an additional saving to not select all columns when not necessary

justinlittman: Good catch @peetucket. Fixed.

peetucket (approved):

PR #6188, opened 2026-07-01
[HOLD] Skip incomplete releaseWF steps when opening a new version

### LINKED ISSUE #6172: Incomplete releaseWF steps are cluttering the 24 hour workflow notice
The HB warning for steps that have been running for 24 hours has become overwhelmed by incomplete releaseWF steps. https://app.honeybadger.io/projects/50568/faults/122433514

This makes it almost useless because it's very difficult to identify accessioning workflow steps that need more immediate attention. I've spent some time resolving these 24-hour only to see more warnings come right back within a day or two. I don't like ignoring this warning but it's become noise that I don't check like I used to.

I think the problem for releaseWF comes down to this:

- an item can be released at any time, regardless of whether it's opened or accessioned (release publishes the last-closed cocina version)
- an active releaseWF does not (and should not) block opening the item to create a new version
- the FOLIO steps are single-threaded and items can be queued for a long time
- each version of the item has workflows associated with that version, so there's could be a version 5 releaseWF step and a version 6 releaseWF step

That means that:

- if an item is queued in a releaseWF step
- and the item is accessioned in a new version before that step finishes
- then the item gets a new version for the releaseWF
- and the item's releaseWF for the previous version never finishes
- causing the 24 hour notice to include that unfinished step

You can see this in the [workflow history for bg324db4339](https://argo.stanford.edu/items/druid:bg324db4339/workflows/releaseWF?raw=true), the first item currently listed in the HB notice:

- version 5 of the releaseWF `update-marc` step is queued
- version 6 of the releaseWF `update-marc` step is completed
- version 6 of the item must have been created before version 5 managed to finish releaseWF

The item is fully functional for future version accessioning, but the queued past version step will cause the HB notice until someone updates the workflow DB. There may be 100s of items like this. I recently resolved nearly 2000 items like this via the rails console and a loop.

**Options for resolving this problem**

I prefer the third option listed below (mark steps skipped) but included the others in case those turn out to be more implement-able.

- We could have the 24-hour notice ignore past versions

I'm not sure if that's a good idea. The system currently blocks opening a new version if a past version step is incomplete in other workflows. I think past version incomplete steps in assemblyWF, accessionWF, and preservationIngestWF will block opening a new version. I actually prefer this because incomplete steps in those workflows should be investigated, even if in a past version.

- We could have the 24-hour notice ignore releaseWF for past versions

I'm still not completely comfortable with this but it probably has minimal impact.

- We could automatically "skip" the steps in releaseWF if they are incomplete when a new version of the item is created

This might be the cleanest approach in terms of workflow DB data. But it could have a side effect of leaving an "Opened" item un-released, caught between the past version releaseWF not finishing and the current version releaseWF not existing yet.

peetucket:
alternate approach in progress

## Summary
- Fixes #6172 (option 3 from the discussion): a releaseWF step for a previous version can be left incomplete forever once a new version is opened, since an active releaseWF doesn't block versioning — this trips the 24-hour stuck workflow Honeybadger notice indefinitely.
- `VersionService#open` now skips any incomplete `releaseWF` steps for the version being opened-from (and any earlier stragglers) right after `ensure_openable!` passes and before the new version is created.
- Adds `Workflow::Service#skip_incomplete`, scoped to incomplete steps up through a given version.

## Test plan

Specs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/services/version_service.rb
+++ b/app/services/version_service.rb
@@ -73,6 +73,8 @@ class VersionService
 
     ensure_openable!(assume_accessioned:, check_preservation_version: !from_version)
 
+    skip_incomplete_release_steps!
+
     from_repository_object_version = from_version ? repository_object.versions.find_by!(version: from_version) : nil
 
     repository_object.open_version!(description:, from_version: from_repository_object_version)
@@ -251,6 +253,15 @@ class VersionService
 
   def workflow_state_service
     @workflow_state_service ||= Workflow::StateService.new(druid:, version:)
+  end
+
+  # An active releaseWF doesn't block opening a new version, so a releaseWF step from a previous
+  # version can be left incomplete forever once that version is no longer the active one. Skip any
+  # such steps so they stop showing up in stuck workflow monitoring.
+  def skip_incomplete_release_steps!
+    Workflow::Service.skip_incomplete(druid:, workflow_name: 'releaseWF', through_version: version,
+                                      note: 'Automatically skipped because a new object version was opened ' \
+                                            'while this releaseWF step was still incomplete.')
   end
 
   def ensure_closeable!
--- a/app/services/workflow/service.rb
+++ b/app/services/workflow/service.rb
@@ -59,6 +59,10 @@ module Workflow
       new(druid:).skip_all(workflow_name:, note:)
     end
 
+    def self.skip_incomplete(druid:, workflow_name:, through_version:, note: nil)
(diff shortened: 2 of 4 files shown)


justinlittman (commented):
Would a cleaner option to add releaseWF to the list of workflows that when running count as "accessioning"?

PR #6196, opened 2026-07-02
Allow version to be taken into account when running robots/updating workflow

### LINKED ISSUE #6172: Incomplete releaseWF steps are cluttering the 24 hour workflow notice

peetucket:

Fixes https://github.com/sul-dlss/dor-services-app/issues/6172

Alternative to #6188 

Coordinate with [lyber-core PR](https://github.com/sul-dlss/lyber-core/pull/297) and [dor-services-client PR](https://github.com/sul-dlss/dor-services-client/pull/630), see release plan below for timing.  This should be last.

Note: we allow version: nil as the default so that old long running jobs will not break during rollout.

## Release plan: ##

- [x] Merge [lyber-core PR](https://github.com/sul-dlss/lyber-core/pull/300) and release new full point version of lyber-core gem.
- [x] Go to PRs above and bump to new release of lyber-core
- [x] Pause stage accessioning queues (pause button):
- https://robot-console-stage.stanford.edu/queues 
- [x] Quiet stage workers and wait until all running jobs at bottom of page are done (quiet all workers button):
- https://robot-console-stage.stanford.edu/busy
- [x] Deploy all apps above to stage which should restart quieted workers.
- [x] Unpause all queues in stage
- https://robot-console-stage.stanford.edu/queues 
- [x]  Run accessioning integration tests in stage
- [x]  Unhold, rebase and merge all PRs below and then quiet/pause workers in qa, stage and prod.  Note that you will likely need to fix merge conflicts with Gemfiles since lyber-core was pinned in DSA, gis-robot-suite, and preservation-robots.  Common-accessioning and was-robot-suite were previously already pinning lyber-core, so should not need to do anything with those to resolve conflicts. 
- Consumer apps:
  - #6196 (this PR)
  - https://github.com/sul-dlss/common-accessioning/pull/1639
  - https://github.com/sul-dlss/gis-robot-suite/pull/1148
  - https://github.com/sul-dlss/preservation_robots/pull/692
  - https://github.com/sul-dlss/was_robot_suite/pull/880
- https://robot-console-stage.stanford.edu/queues (pause button for all queues) 
- https://robot-console-qa.stanford.edu/queues (pause button for each queue)
- https://robot-console-prod.stanford.edu/queues (pause button for each queue)
- https://robot-console-stage.stanford.edu/busy (quiet all workers button)
- https://robot-console-qa.stanford.edu/busy (quiet all workers button)
- https://robot-console-prod.stanford.edu/busy (quiet all workers button)
- [x]  Deploy all apps that had PRs above to stage, qa and prod, which should unquiet all workers.
- [x]  Unpause all queues in prod and qa 
- https://robot-console-stage.stanford.edu/queues (unpause button for all queues) 
- https://robot-console-qa.stanford.edu/queues (unpause button for all queues) 
- https://robot-console-prod.stanford.edu/queues (unpause button for all queues)
 
**Failure/Rollback:**
-  [x] Revert PRs above
-  [x] Make new PRs that pin lyber-core to previous full point release ( < 9) in all those apps, merge those PRs, re-deploy them all

- [x] Specs
- [x] Integration tests

### DIFF
--- a/app/controllers/workflow_processes_controller.rb
+++ b/app/controllers/workflow_processes_controller.rb
@@ -5,11 +5,12 @@ class WorkflowProcessesController < WorkflowApplicationController
   def update # rubocop:disable Metrics/AbcSize
     if status == 'error'
       Workflow::ProcessService.update_error(druid:, workflow_name:, process: workflow_process,
-                                            error_msg: params[:error_msg], error_text: params[:error_text])
+                                            error_msg: params[:error_msg], error_text: params[:error_text],
+                                            version:)
     else
       Workflow::ProcessService.update(druid:, workflow_name:, process: workflow_process, status:,
                                       elapsed: params[:elapsed] || 0, lifecycle: params[:lifecycle],
-                                      note: params[:note], current_status: params[:current_status])
+                                      note: params[:note], current_status: params[:current_status], version:)
     end
 
     head :no_content
@@ -31,5 +32,9 @@ class WorkflowProcessesController < WorkflowApplicationController
 
   def status
     params[:status]
+  end
+
+  def version
+    params[:version]
   end
 end
--- a/app/jobs/robots/robot.rb
+++ b/app/jobs/robots/robot.rb
@@ -14,17 +14,18 @@ module Robots
     private
 
     def workflow
-      @workflow ||= RobotWorkflow.new(workflow_name: workflow_name, process: process, druid:)
+      @workflow ||= RobotWorkflow.new(workflow_name: workflow_name, process: process, druid:, version:)
     end
 
     # This encapsulates the workflow operations that lyber-core does.
     # This implementation uses local services instead of the dor-services-client.
     # It should have the same interface as LyberCore::Workflow.
     class RobotWorkflow
-      def initialize(workflow_name:, process:, druid:)
+      def initialize(workflow_name:, process:, druid:, version:)
         @workflow_name = workflow_name
         @process = process
         @druid = druid
+        @version = version
       end
 
       def object_workflow
@@ -46,24 +47,31 @@ module Robots
       end
 
       def start!(note)
-        Workflow::ProcessService.update(druid:, workflow_name:, process:, status: 'started', note:, elapsed: 1.0)
+        Workflow::ProcessService.update(druid:, workflow_name:, process:, status: 'started', note:, elapsed: 1.0,
+                                        version:)
       end
 
       def complete!(status, elapsed, note)
-        Workflow::ProcessService.update(druid:, workflow_name:, process:, status:, note:, elapsed:)
+        Workflow::ProcessService.update(druid:, workflow_name:, process:, status:, note:, elapsed:, version:)
       end
 
       def retrying!
-        Workflow::ProcessService.update(druid:, workflow_name:, process:, status: 'retrying', note: nil, elapsed: 1.0)
+        Workflow::ProcessService.update(druid:, workflow_name:, process:, status: 'retrying', note: nil,
+                                        elapsed: 1.0, version:)
       end
 
       def error!(error_msg, error_text)
-        Workflow::ProcessService.update_error(druid:, workflow_name:, process:, error_msg:, error_text:)
+        Workflow::ProcessService.update_error(druid:, workflow_name:, process:, error_msg:, error_text:, version:)
+      end
+
+      def skip!(note)
+        Workflow::ProcessService.update(druid:, workflow_name:, process:, status: 'skipped', note:, elapsed: 0,
+                                        version:)
       end
 
       delegate :context, :status, :lane_id, :active_version?, to: :process_response
 
-      attr_reader :workflow_name, :process, :druid
+      attr_reader :workflow_name, :process, :druid, :version
     end
   end
 end
--- a/app/services/workflow/process_service.rb
+++ b/app/services/workflow/process_service.rb
@@ -4,22 +4,24 @@ module Workflow
   # Service for interacting with workflows processes (steps).
   class ProcessService
     def self.update(druid:, workflow_name:, process:, status:, elapsed: 0, lifecycle: nil, note: nil, # rubocop:disable Metrics/ParameterLists
-                    current_status: nil, enqueue_next_steps: true)
-      new(druid:, workflow_name:, process:).update(status:, elapsed:, lifecycle:, note:, current_status:,
-                                                   enqueue_next_steps:)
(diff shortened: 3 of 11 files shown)


app/controllers/workflow_processes_controller.rb
```diff
@@ -5,11 +5,12 @@ class WorkflowProcessesController < WorkflowApplicationController
   def update # rubocop:disable Metrics/AbcSize
     if status == 'error'
       Workflow::ProcessService.update_error(druid:, workflow_name:, process: workflow_process,
-                                            error_msg: params[:error_msg], error_text: params[:error_text])
+                                            error_msg: params[:error_msg], error_text: params[:error_text],
```
justinlittman, line 8: This probably needs a change to openapi.

justinlittman: Also, changes to the controller should be tested in a request spec.

justinlittman (changes requested):

app/services/workflow/process_service.rb
```diff
@@ -71,11 +73,12 @@ def update_status(status:, elapsed: 0, lifecycle: nil, note: nil, # rubocop:disa
...
     end
 
-    def find_step_for_process
+    def find_step_for_process # rubocop:disable Metrics/AbcSize
       query = WorkflowStep.where(druid:,
                                  workflow: workflow_name,
                                  process: process)
-                          .order(version: :desc)
+      query = query.where(version:) if version.present?
+      query = query.order(version: :desc)
```
justinlittman, line 87: While we're at it, should probably refactor this a bit. Perhaps replace L77-87 with:
```
query = WorkflowStep.where(druid:,
                                 workflow: workflow_name,
                                 process: process)
if version.present?
   query.find_by(version:)
else
   query.find_by(active_version: true)
```

peetucket: Making this change and an additional refactor.

app/controllers/workflow_processes_controller.rb
peetucket: 1. open api updated
2. this request spec was already updated: https://github.com/sul-dlss/dor-services-app/pull/6196/changes#diff-f23b1a766b52ce3dc418315db74c5db44ab52814f85b25720a78d616e1816b51    is there another one you think I missed?

app/jobs/robots/robot.rb
```diff
@@ -46,22 +46,29 @@ def process_response
...
-        Workflow::ProcessService.update_error(druid:, workflow_name:, process:, error_msg:, error_text:)
+        Workflow::ProcessService.update_error(druid:, workflow_name:, process:, error_msg:, error_text:, version:)
       end
 
-      delegate :context, :status, :lane_id, to: :process_response
+      def skip!(note)
+        Workflow::ProcessService.update(druid:, workflow_name:, process:, status: 'skipped', note:, elapsed: 0,
+                                        version:)
+      end
+
```
peetucket, line 71: skip method to match what we will have in lyber-core to perform our automated skipping if a previous version workflow step is not done yet when a new one starts (to implement item 2 in https://github.com/sul-dlss/dor-services-app/issues/6172#issuecomment-4859940091)

app/jobs/robots/robot.rb
```diff
@@ -14,17 +14,18 @@ def object_client
...
+      @workflow ||= RobotWorkflow.new(workflow_name: workflow_name, process: process, druid:, version:)
     end
 
     # This encapsulates the workflow operations that lyber-core does.
     # This implementation uses local services instead of the dor-services-client.
     # It should have the same interface as LyberCore::Workflow.
     class RobotWorkflow
-      def initialize(workflow_name:, process:, druid:)
+      def initialize(workflow_name:, process:, druid:, version:)
         @workflow_name = workflow_name
```
peetucket, line 25: this matches the new interface in lyber-core `LyberCore::Workflow` class, where we expect a version to be passed in: https://github.com/sul-dlss/lyber-core/pull/297/changes#diff-c6c673a8ace4e93ba0950d77584936b3ecb9fbb83b0ad555320c52a9cb9e9280L6

justinlittman (approved):

PR #6199, opened 2026-07-02
Bump folio_client

### LINKED ISSUE #6198: Better handling of FOLIO MARC validation errors
SDR contacts FOLIO during the `releaseWF` whenever an item has an HRID attached. At some point in this process, if FOLIO's QuickMARC API deems a MARC record invalid, the `update-marc` step stops with an error and reports back a FOLIO validation message.

Example: 
- https://argo.stanford.edu/view/druid:bd798vs7660
- [HB error](https://app.honeybadger.io/projects/50568/faults/119318456/01KWHZVRJEDXXRDTEJEEPHCGMC?page=0)
>FolioClient::ValidationError: There was a validation problem with the request: {"message":"Wrong value '?', on position 18. Allowed only: [\\,  , ??, a, c, i, n, u]","type":"-1","parameters":[{"key":"Descriptive cataloging form"}]}

So far as we know, these types of errors block SDR from updating the MARC record via QuickMARC. This means that for any item where the 856 needs to be update - to release or un-release - the MARC record must be updated to meet the QuickMARC validation rules.

There's one other problem with this error: it appears even in situations where SDR does not need to update the MARC record.

Example: 

- the [druid listed above](https://argo.stanford.edu/view/druid:bd798vs7660) is not released and has never been released
- that means there is no need to update the MARC record
- this means the FOLIO validation problem should not block the step

That raises questions:

- is SDR trying to update MARC even when there is no change to the MARC?
- when does the FOLIO validation message appear? Is it only when an update is attempted or is it a response to a query earlier in the process?

lwrubel:
Fixes #6198 

Needs to be tested in a deployed env

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -236,7 +236,7 @@ GEM
     faraday-retry (2.4.0)
       faraday (~> 2.0)
     fiber-storage (1.0.1)
-    folio_client (1.1.0)
+    folio_client (1.2.0)
       activesupport (>= 4.2)
       deprecation
       dry-monads


lwrubel:
folio_client will be bumped via dependency updates

Issue #6204, opened 2026-07-06
Fix pages build and deploy


lwrubel:
Github Pages site is not building: https://sul-dlss.github.io/dor-services-app

There are two errors/warnings in Github Actions when building the docs. See:
See: https://github.com/sul-dlss/dor-services-app/actions/runs/28793073550

One is an issue with the version of Node specified and the other has to do with a syntax issue in openapi.yml. See:
* https://validator.swagger.io/validator/debug?url=https://raw.githubusercontent.com/sul-dlss/dor-services-app/main/openapi.yml

PR #6205, opened 2026-07-06
Fix openapi errors.

justinlittman:
refs #6204

```
{"messages":["attribute paths.'/v1/objects/{object_id}/user_versions/{id}'(patch).operationId is repeated","Duplicate field $ref in `https://raw.githubusercontent.com/sul-dlss/dor-services-app/main/openapi.yml`"]}
```

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1000,8 +1000,9 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "https://raw.githubusercontent.com/sul-dlss/cocina-models/refs/tags/v0.113.0/schema.json#/$defs/DRO"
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -1212,50 +1214,6 @@ paths:
                 withdrawn:
                   description: whether the version is withdrawn
                   type: boolean
-    put:
```
lwrubel, line 1215: Why is this removed? Because we use patch instead of put?

lwrubel (approved):

openapi.yml
justinlittman: Yes.

PR #6215, opened 2026-07-07
Adds nested related resource migrator.

justinlittman:
refs https://github.com/sul-dlss/cocina-models/issues/884

Allow new validator.

QA

### DIFF
--- a/app/services/migrators/remove_nested_related_resource.rb
+++ b/app/services/migrators/remove_nested_related_resource.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Migrators
+  # Removes a nested related resource property from description in all objects and versions.
+  # See parent class and Migrators::MigrationRunner for more information.
+  class RemoveNestedRelatedResource < Base
+    def migrate
+      Array(model_hash.dig('description', 'relatedResource')).each do |related_resource_hash|
+        nested_related_resource = related_resource_hash['relatedResource'].present?
+
+        # If the nested relatedResource is empty, delete the property.
+        next related_resource_hash.delete('relatedResource') unless nested_related_resource
+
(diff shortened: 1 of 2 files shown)


app/services/migrators/remove_nested_related_resource.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Migrators
+  # Removes a nested related resource property from description in all objects and versions.
+  # See parent class and Migrators::MigrationRunner for more information.
+  class RemoveNestedRelatedResource < Base
+    def migrate
+      Array(model_hash.dig('description', 'relatedResource')).each do |related_resource_hash|
+        raise 'Nested relatedResource found' if related_resource_hash['relatedResource'].present?
```
lwrubel: This is helpful for dryrun. There was remediation in the last WC which involved usage of the field. I think it's possible that some older versions may still have a nested `relatedResource` since the head version was remediated, but it would be expected that it would get wiped out in migration. Here's an example that will get encountered: https://argo.stanford.edu/items/druid:pg723bw3438/version/9.json

justinlittman: How do we want to handle this? One possibility is to raise for head_version or last_closed_version, but leave invalid for earlier versions?

lwrubel: Let's discuss. Leaving invalid older versions is likely best, but I want to also ask about deleting the field that is now invalid.

lwrubel: As confirmed in discussion earlier, raise for head_version or last_closed_version only.

lwrubel (approved):

PR #6218, opened 2026-07-07
Filter invalid MARC dates when mapping event dates from MARC.

### LINKED ISSUE #6217: Drop invalid encoded dates when mapping from MARC 008/07-14
The 008 field in MARC contains dates encoded in the MARC format at characters 07-10 and 11-14. When any of these dates are determined to be invalid according to MARC date encoding rules, they should not be included in the output Cocina.

See also https://github.com/sul-dlss/dor-services-app/issues/6146 for the same issue with 008/00-05.

Examples:

HRID a3918041
`008 260625m1992????pl            000 0 pol u`
008/06 = m, so mapped as multiple dates ( see https://docs.google.com/spreadsheets/d/1hO8GULPIez5BSUb9DhHyu-B6Yqfk2_3CXZ8DtWPK2Oc/edit?gid=1934624845#gid=1934624845&range=A25)
date1: 1992
date2: ???? (invalid)

Current output:

Image

Desired output:
```
{ event: [ {
  type: "publication",
  date: [ {
    value: "1992",
    type: "publication",
    encoding: { code: "marc" }
  } ]
} ] }
```

--

HRID a2676614
`008 890328d198901  dcuab    tb  f000 0 eng d`
008/06 = d, so mapped as date range (see https://docs.google.com/spreadsheets/d/1hO8GULPIez5BSUb9DhHyu-B6Yqfk2_3CXZ8DtWPK2Oc/edit?gid=1934624845#gid=1934624845&range=A29)
date1: 1989
date2: 01 (invalid)

Current output:

Image

Desired output:
```
{ event: [ {
  type: "publication",
  date: [ {
    structuredValue: [ { value: "1989", type: "start } ],
    type: "publication",
    encoding: { code: "marc" }
  } ]
} ] }
```

justinlittman:
closes #6217

Per @arcadiafalcone 

Unit

### DIFF
--- a/app/services/cocina/from_marc/event_date.rb
+++ b/app/services/cocina/from_marc/event_date.rb
@@ -35,11 +35,12 @@ module Cocina
       end
 
       # @return [Array<Hash>] an array of event date hashes
-      def build
-        return if no_date?
+      def build # rubocop:disable Metrics/CyclomaticComplexity
+        return if no_date? || invalid_marc_date?(start_date) || invalid_marc_date?(end_date)
         return structured_open_range if open_range?
         return simple_range if simple_range?
         return structured_range if structured_range?
+
         return single_date if start_date
 
         nil
@@ -117,6 +118,12 @@ module Cocina
         return if date.blank? || date == BLANK_DATE
 
         date
+      end
+
+      def invalid_marc_date?(date)
+        return false unless date
+
+        !Cocina::Models::Validators::MarcDateValidator.validate(date)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


justinlittman:
@arcadiafalcone Do the specs look correct?

arcadiafalcone:
Yes. For the multiple dates, either the first or the second value could be invalid and dropped. It's also possible for all 8 characters to be part of a single date, but there weren't any instances of that in the error report.

peetucket (approved):

PR #6220, opened 2026-07-07
add active attribute to response

peetucket:

Part of https://github.com/sul-dlss/dor-services-app/issues/6172

We need to expose the active_version attribute in the response.  This is a pre-req to exposing this in the DSA client response.

Once merged, will deploy to stage and then update dsa-client and lyber-core as needed.  And then new PRs and new gems released.

Specs

### DIFF
--- a/spec/jobs/robots/robot_spec.rb
+++ b/spec/jobs/robots/robot_spec.rb
@@ -42,8 +42,11 @@ RSpec.describe Robots::Robot do
       described_class.new(workflow_name: 'testWF', process: 'test-step', druid: 'druid:gv054hp4128')
     end
 
-    let(:workflow_response) { instance_double(Dor::Services::Response::Workflow, process_for_recent_version: process_response) }
-    let(:process_response) { instance_double(Dor::Services::Response::Process, name: 'test-step', status: 'waiting', lane_id: 'low', context: { foo: 'bar' }) }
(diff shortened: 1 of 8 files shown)


spec/jobs/robots/robot_spec.rb
```diff
@@ -42,8 +42,11 @@ def perform_work; end
       described_class.new(workflow_name: 'testWF', process: 'test-step', druid: 'druid:gv054hp4128')
     end
 
-    let(:workflow_response) { instance_double(Dor::Services::Response::Workflow, process_for_recent_version: process_response) }
-    let(:process_response) { instance_double(Dor::Services::Response::Process, name: 'test-step', status: 'waiting', lane_id: 'low', context: { foo: 'bar' }) }
+    let(:workflow_response) { instance_double(Workflow::WorkflowResponse, process_for_recent_version: process_response) }
+    let(:process_response) do
+      instance_double(Workflow::ProcessResponse, name: 'test-step', status: 'waiting', lane_id: 'low',
+                                                 context: { foo: 'bar' }, active_version?: true)
+    end
```
peetucket, line 49: I think we want to instance double the classes in DSA and not in dsa-client...else we have a circular dependency causing test failures (the new methods to test don't exist in dsa-client yet because they can't be released until DSA exposes them in the response)

justinlittman (approved):

PR #6234, opened 2026-07-10
Protect against empty fields when mapping from MARC.

### LINKED ISSUE #6232: Drop empty fields when mapping MARC
If a MARC field has a subfield with no value, ignore it in the mapping.

2026-07-09 17:00:46 UTC Failed Dor::Services::Client::UnprocessableContentError Cocina validation error (Blank value in description: note6) for druid:vj213rj1935
14177439 - 500 $a has no value - do not map and do not report error

2026-07-09 15:51:54 UTC Failed Dor::Services::Client::UnprocessableContentError Cocina validation error (Blank value in description: subject2) for druid:tg738bw7717
a14185562 - 050 $a has no value - do not map and do not report error

### LINKED ISSUE #6233: Ignore 099 $b when mapping MARC
Ignore 099 $b when present and do not raise error message.

2026-07-10 01:59:46 UTC Failed Dor::Services::Client::UnprocessableContentError Cocina validation error (When validating Description: null is not of type "string" at /access/physicalLocation/0/value, Unevaluated properties are not allowed ('type', 'value' were unexpected) at /access/physicalLocation/0) for druid:fk742rz2402
a14418927 
099   b| SC1607
no cocina mapping

justinlittman:
closes #6232
closes #6233

Produce valid cocina.

Unit

### DIFF
--- a/app/services/cocina/from_marc/access.rb
+++ b/app/services/cocina/from_marc/access.rb
@@ -25,7 +25,7 @@ module Cocina
 
       def physical_location
         field = marc['099']
-        return unless field
+        return unless field && field['a'].present?
 
         [{
           value: field['a'],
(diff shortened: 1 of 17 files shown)


justinlittman:
Just to say out loud -- Claude did this work. I've reviewed the code, but as you can see it is very much in the weeds. However, since no tests were changed (only added), it gives me additional confidence.

lwrubel (approved):
Agree there are a lot of details changed in this, but I'm OK with it considering the test coverage and that we'll run another dry-run on the data.

PR #6241, opened 2026-07-10
remove label from migrators

### LINKED ISSUE #6211: Don't edit the top-level label in migrators
Change migrators to not update the top-level label:

- [ ] Don't edit the label field in `Migrators::Exemplar`:
https://github.com/sul-dlss/dor-services-app/blob/802f8e7525af18e788e2767559cfa801ff901a90/app/services/migrators/exemplar.rb
- [ ] `Migrators::CheckCocinaUpdate` will need a different fake edit to make to all objects:
https://github.com/sul-dlss/dor-services-app/blob/802f8e7525af18e788e2767559cfa801ff901a90/app/services/migrators/check_cocina_update.rb#L6-L10
- [ ] Update any specs if relevant.

peetucket:

Fixes #6211 

Specs

### DIFF
--- a/app/services/migrators/check_cocina_update.rb
+++ b/app/services/migrators/check_cocina_update.rb
@@ -4,8 +4,15 @@ module Migrators
   # Fake migrator that can be used to check that objects can be updated.
   class CheckCocinaUpdate < Base
     def migrate
-      # This changes every object.
-      model_hash['description']['title'].first['value'] = 'Test'
+      # This changes every object by adding a title to the description,
+      # and creating a description first if it doesn't exist.
+      description = model_hash['description'] || {}
+      titles = description['title'] || []
+
+      titles << { 'value' => 'Test' }
+      description['title'] = titles
+      model_hash['description'] = description
+
       model_hash
     end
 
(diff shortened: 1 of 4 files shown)


app/services/migrators/check_cocina_update.rb
```diff
@@ -5,7 +5,7 @@ module Migrators
   class CheckCocinaUpdate < Base
     def migrate
       # This changes every object.
-      model_hash['label'] = 'Test'
+      model_hash['description']['title'].first['value'] = 'Test'
```
justinlittman: Not every object has `model_hash['description']['title'].first['value']`

justinlittman (changes requested):

app/services/migrators/check_cocina_update.rb
peetucket: Do you have suggestions for a field that every object type is guaranteed to have?  I've only come up with, none of which seem a great fit to change.  Couldn't we just add a description block to things like APOs even if they don't have one? 

- cocinaVersion
- type
- externalIdentifier
- version

justinlittman: Yes, one option would be to append an additional title or note to the description.

peetucket: Done

justinlittman (approved):

PR #6250, opened 2026-07-14
Make adjustments for control of DB connections.

justinlittman:


lwrubel:
Can you describe the reason for adjusting the DB connections in the CreateEventJob and/or add a comment why it's set to 1 process?

I see the CreateEventJob config is mentioned in the README, but why do we want only 1 process?

peetucket (approved):

justinlittman:
Sure. Comment added @lwrubel

PR #6263, opened 2026-07-16
Bump cocina-models to 0.129.0

lwrubel:
Keep cocina-models in sync and also handle new validation that a catalog link or title must be present. 

### DIFF
--- a/app/services/refresh_description_from_catalog.rb
+++ b/app/services/refresh_description_from_catalog.rb
@@ -27,7 +27,7 @@ class RefreshDescriptionFromCatalog
   # @return [Dry::Monads::Results] Returns Failure if description unchanged (e.g., no refreshable identifiers),
   #   otherwise Success (Result with description_props)
   # @raise [Catalog::MarcService::Error]
-  def run
+  def run # rubocop:disable Metrics/AbcSize
     # Admin policies don't have identification.
     return Failure() if cocina_object.admin_policy?
     return Failure() unless refreshable?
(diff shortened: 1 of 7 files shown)


app/services/registration_csv_converter.rb
```diff
@@ -71,11 +72,18 @@ def model_params(row)
         barcode: row['barcode']
       }.compact
     }
-
     model_params[:structural] = structural(row)
     model_params[:access] = access(row)
+    title = row['title']
+    model_params[:description] = description(row) if title.present?
```
lwrubel, line 78: A title or folio hrid must be present for a RequestDro to be valid. This allows adding a title to the description and adds the folio_id to identification (it didn't look like this was handled at all previously).

app/services/refresh_description_from_catalog.rb
```diff
@@ -40,6 +40,7 @@ def run
     description_props = Cocina::FromMarc::Description.props(marc:, druid:)
 
     return Failure() if description_props.nil?
+    return Failure() if description_props[:title].blank?
```
lwrubel, line 43: There is a catalog link and `RefreshDescriptionFromCatalog::Result` would return success even if no title is mapped in. This closes that gap and returns Failure instead. 

Arguably, we could let this still return Success since there is a description and let downstream validation in `UpdateObjectService` raise.

spec/requests/metadata_refresh_spec.rb
```diff
@@ -167,9 +167,9 @@
     context 'when Cocina validation error' do
       let(:result) { Success(RefreshDescriptionFromCatalog::Result.new(description_props)) }
       let(:description_props) do
-        # Missing title
+        # Invalid title type
         {
-          purl: "https://purl.stanford.edu/#{druid.delete_prefix('druid:')}"
+          title: [{ value: 'Paying for College', type: 'not a type' }]
```
lwrubel, line 172: Missing a title wouldn't raise anymore since there is a folio catalogLink.

justinlittman (approved):

Issue #6278, opened 2026-07-21
releaseWF publish hits error if item is Opened or In accessioning


andrewjbtw:
The releaseWF publish step hits an error if an item is released while in an Opened or in an accessioning state where the accessionWF `publish` step has not run yet. This leads to repeated retries:
> [[Robots::DorRepo::Release::ReleasePublish::PublishNotCompleteError: Robots::DorRepo::Release::ReleasePublish::PublishNotCompleteError]](https://robot-console-stage.stanford.edu/retries)

The error message suggests that the item has not been published but I think the problem is that the item was previously published in the previous version. The workflow is probably looking for a completed "publish" step in the head version, but there isn't one because the head version number has been incremented since the previous publish.

**Expected behavior**

The releaseWF `release-publish` step should publish the cocina for the last accessioned version and not depend on the head version to be Opened or to complete accessioning. I think that means the check to determine if the item has been published should take into account whether the item was published in the previous version if no publish can be found for the head version.

Issue #6246, opened 2026-07-13
Robots::DorRepo::Release::ReleasePublish::PublishNotCompleteError


justinlittman:
## Backtrace

    line 21 of [PROJECT_ROOT]/app/jobs/robots/dor_repo/release/release_publish.rb: Robots::DorRepo::Release::ReleasePublish#perform_work

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50568/faults/124943476)

justinlittman:
This is happening because `ReleasePublish` is checking the most recent version of the object instead of the version of the object that the workflow is for.

In one such case, the `ReleasePublish` is for version 2, but there is an opened version 3. It should be checking if version 2 is published, not version 3.

PR #6287, opened 2026-07-21
any previously published version is ok when running releaseWF

peetucket:

Note: Alternative to #6289 in case we need it later (to deal with #6278)

A previous fix for https://github.com/sul-dlss/dor-services-app/issues/6246 causes an issue when you open a new version of an object, and then run releaseWF on it.  The fix in #6289 checks if the last closed version has a publish lifecycle.  There may edge cases where there was a previously published version other than in the last closed version.  This alternative implementation would allow release to continue in those situations.

This code adds a new Lifecycle service helper method to find the latest version of a particular lifecycle.  It then uses this in the release step.

Question: should we constraint this further so that the previous version must be n-1 the workflow version?  This might be too tight a constraint.

Note: Codex helped identify the underlying cause and suggested some fixes, which I iterated on and narrowed down in scope.  I wrote most of the code using some of it's suggestions.  It wrote most of the specs.

Specs, but could be worth deploying to stage, and seeing if the current stage object succeeds on retries: https://argo-stage.stanford.edu/view/druid:bb975xt8207

### DIFF
--- a/app/jobs/robots/dor_repo/release/release_publish.rb
+++ b/app/jobs/robots/dor_repo/release/release_publish.rb
@@ -40,10 +40,23 @@ module Robots
         end
 
         def published_version_available?
-          last_closed_version = RepositoryObject.find_by!(external_identifier: druid).last_closed_version_version
-          return false unless last_closed_version
+          workflow_version = version.to_i
 
-          Workflow::LifecycleService.milestone?(druid:, milestone_name: 'published', version: last_closed_version)
+          # The release workflow version itself has been published.
+          return true if version_published?(version: workflow_version)
+
+          # The release workflow may have started for a newer open version while an earlier version
+          # was already published. See https://github.com/sul-dlss/dor-services-app/issues/6278.
+          latest_version = latest_published_version
+          latest_version.present? && latest_version < workflow_version
+        end
+
+        def version_published?(version:)
+          Workflow::LifecycleService.milestone?(druid:, milestone_name: 'published', version:)
+        end
+
+        def latest_published_version
+          Workflow::LifecycleService.latest_milestone_version(druid:, milestone_name: 'published')
         end
       end
     end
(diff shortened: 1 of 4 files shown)


app/jobs/robots/dor_repo/release/release_publish.rb
```diff
@@ -39,9 +39,25 @@ def dark?
           Cocina::Support.dark?(cocina_object)
         end
 
-        def published?
+        def published_version_available?
```
justinlittman, line 42: Should this just check the last closed version of the object?

peetucket: can releaseWF run on v1 of an object? i.e. publish and releaseWF are both v1, which is ok, but there is no last closed version...

justinlittman: releaseWF can run on v1 of an object if has been closed and accessioned (and therefore published). If v1 is still open, then it cannot be released. (I think.)

peetucket: alternate (simpler) implementation that uses this approach: https://github.com/sul-dlss/dor-services-app/pull/6289   

just deployed that to stage and we will what happens with https://argo-stage.stanford.edu/view/druid:bb975xt8207 when it retries

andrewjbtw: An "Opened" v1 is the equivalent of "Registered" and releaseWF should not run. There shouldn't be any data in the pipeline to publish in that state so I would expect an error if releaseWF did run.

peetucket: That would be fine - there would be no "latest published version" in that case, and the code in this PR would throw the same "PublishNotComplete" exception and fail to run ... which would be the correct thing to do here.

PR #6289, opened 2026-07-21
check if there is a last closed version published

### LINKED ISSUE #6278: releaseWF publish hits error if item is Opened or In accessioning

peetucket:

Fixes #6278 - alternate to #6287 which is simpler in scope.  Instead of checking the workflow version as having been published (which will fail if we have a releaseWF opened in a new version for an object that was published in a previous version), we now check to see if there any closed version that has been published.

This is hopefully sufficient?  The only way this will fail is if try to release something for an opened v1 (even if published has occurred in that open version) OR for some reason there is a closed v1 that was never published, but we are on a later opened version (even if published but not yet closed).  Not sure if those scenarios will ever happen.

Specs, but could be worth deploying to stage, and seeing if the current stage object succeeds on retries: https://argo-stage.stanford.edu/view/druid:bb975xt8207

### DIFF
--- a/spec/jobs/robots/dor_repo/release/release_publish_spec.rb
+++ b/spec/jobs/robots/dor_repo/release/release_publish_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Robots::DorRepo::Release::ReleasePublish, type: :robot do
   subject(:perform) { test_perform(robot, druid, version: 4) }
 
   let(:druid) { 'bb222cc3333' }
-  let!(:repository_object) { create(:repository_object, :closed, external_identifier: druid) } # rubocop:disable RSpec/LetSetup
+  let!(:repository_object) { create(:repository_object, :closed, external_identifier: druid) }
   let(:cocina_object) { instance_double(Cocina::Models::DRO, dro?: true, access: dro_access, admin_policy?: false) }
   let(:dro_access) { instance_double(Cocina::Models::DROAccess, view: 'world') }
   let(:robot) { described_class.new }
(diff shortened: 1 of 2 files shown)


spec/jobs/robots/dor_repo/release/release_publish_spec.rb
```diff
@@ -77,14 +96,4 @@
...
-
-  context 'when an unpublished item' do
-    before do
-      allow(Workflow::LifecycleService).to receive(:milestone?).and_return(false)
-    end
-
-    it 'raises' do
-      expect { perform }.to raise_error(Robots::DorRepo::Release::ReleasePublish::PublishNotCompleteError)
-    end
-  end
```
peetucket, line 89: moved above into a new context block

justinlittman (approved):

PR #6295, opened 2026-07-24
allow user to pass in a csv file with druids to validate-cocina script

peetucket:

Update `validate-cocina` script to allow user to pass in a specific list of druids to validate against live data.  Helpful if you need to re-run validation against recently remediate objects.  You can pass in the CSV file generated by the `validate-data` script in cocina-models.

If merged, I can update the README in cocina-models.

On sdr-infra

### DIFF
--- a/bin/validate-cocina
+++ b/bin/validate-cocina
@@ -7,6 +7,7 @@
 # For example, gem 'cocina-models', github: 'sul-dlss/cocina-models', branch: 'test_me'
 
 require_relative '../config/environment'
+require 'csv'
 require 'optparse'
 require 'tty-progressbar'
 require 'tmpdir'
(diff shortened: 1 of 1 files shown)


bin/validate-cocina
```diff
@@ -86,13 +90,23 @@ def write_results_to_temp_file(results, temp_dir, process_index)
   temp_file
 end
 
+def druids_from_csv(csv_file)
+  CSV.foreach(csv_file).filter_map do |row|
+    druid = row.first&.strip&.delete_prefix("\uFEFF")
```
justinlittman: What is this for?

peetucket: could be simplified to `row.first` (basically assuming druid is the first column), the delete_prefix was some codex suggested defensive cleanup for start of line characters that might be in a UTF-8, but don't think it's needed... the next line lets us skip the header if it exists (assuming the header is "druid").... could also skip this and just assume no header (which is what the output of the validate-data script produces)

justinlittman: The `delete_prefix` seems like slop to me, but I'm good with the rest of it.

justinlittman (approved):
Approved pending consideration of comment.

bin/validate-cocina
peetucket: concur, i will remove

PR #6296, opened 2026-07-24
Prepare for cocina-models schema changes for label and AdminPolicy

lwrubel:
Changes that will be required in DSA when the next set of cocina-models changes to the schema are released 
(removing label and making description required for AdminPolicy).

Changes to reports are all to remove references to the AdminPolicy label field.
 
OK to merge before the cocina-models release that removes label is deployed. The only further change needed with the cocina-models release is removing the blank label insertion from app/service/serializer.rb

CI, To be tested with integration tests.

### DIFF
--- a/app/reports/date_encoding_values.rb
+++ b/app/reports/date_encoding_values.rb
@@ -49,7 +49,11 @@ class DateEncodingValues
           collection_name = Cocina::Models::Builders::TitleBuilder.build(collection_head_version.to_cocina.description.title)
         end
         apo_druid = rows.first['apo']
-        apo_name = RepositoryObject.admin_policies.find_by(external_identifier: apo_druid)&.head_version&.label
+        apo_head_version = RepositoryObject.admin_policies.find_by(external_identifier: apo_druid)&.head_version
+        if apo_head_version&.has_cocina?
+          apo_description = apo_head_version.to_cocina.description
+          apo_name = Cocina::Models::Builders::TitleBuilder.build(apo_description.title) if apo_description
+        end
         title = (rows.first['structured_title'] || rows.first['title'])&.delete("\n")
 
         [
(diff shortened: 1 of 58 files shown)


app/services/cocina/serializer.rb
```diff
@@ -9,7 +9,9 @@ def serialize?(argument)
     end
 
     def serialize(cocina_item)
-      super(Cocina::Models.without_metadata(cocina_item).to_h)
+      cocina_object = super(Cocina::Models.without_metadata(cocina_item).to_h)
+      # Remove when label fully gone from cocina-models schema
+      cocina_object.merge({ label: '' })
```
lwrubel, line 14: In order to merge this PR before the cocina-models release that removes labels, temporarily populating a label field when serializing.

app/reports/date_encoding_values.rb
```diff
@@ -49,7 +49,11 @@ def self.rows(sql)
           collection_name = Cocina::Models::Builders::TitleBuilder.build(collection_head_version.to_cocina.description.title)
         end
         apo_druid = rows.first['apo']
-        apo_name = RepositoryObject.admin_policies.find_by(external_identifier: apo_druid)&.head_version&.label
+        apo_head_version = RepositoryObject.admin_policies.find_by(external_identifier: apo_druid)&.head_version
+        if apo_head_version&.has_cocina?
+          apo_description = apo_head_version.to_cocina.description
+          apo_name = Cocina::Models::Builders::TitleBuilder.build(apo_description.title) if apo_description
```
justinlittman: cocina_display instead of TitleBuilder? (Here and below)

spec/factories/repository_object_versions.rb
```diff
@@ -12,7 +12,6 @@
     lock { 2 }
     cocina_version { Cocina::Models::VERSION }
     content_type { Cocina::Models::ObjectType.book }
-    label { 'Test DRO' }
```
justinlittman, line 15: Should there be an ability to provide a title (like label)?

justinlittman (approved):

app/reports/date_encoding_values.rb
lwrubel: Updated across all reports.

spec/factories/repository_object_versions.rb
lwrubel: Thanks for the idea, added that.

